PR #1757, opened 2014-09-29
Mhotar/cz inputs

mhotar:
It fixes problem with special characters used in INPUT names generated for Czech language.
INPUT names are different than those used for stick names to avoid special characters in these strings, because Companion can not import/export these characters yet.


kilrah:
Hmm that branch is still there, is something still needed here?

Issue #1849, opened 2014-11-16
LCD screen stays blank after power on


projectkk2glider:
Report from forums, http://www.rcgroups.com/forums/showpost.php?p=29953267&postcount=879

> I have a first generation Taranis. Never had a problem. Today, several time while powering it up I would hear the welcome message, but the screen would stay blank. and nearly as I can tell, the radio would not transmit. Usually turning the power off and on again solved it. Running OpenTx 2.0.12 which so far (2 months) worked fine. Anybody has had issues like this before? The radio is 14 months old. 

Then another user:

> About 4 hours after I received my Taranis I experienced this failure. My supplier gave me an RMA number immediately, I returned the system and I was sent a replacement which has been good for two months now. 

And another:

> Mine which is 16 months old has just started doing the same. It seems to be when the temperature is quite low.

And another:

> Mine did that also in the beginning with the stock FW. Then once or twice recently since updating to 2.0.12. Temperature doesn't seem to be a factor.Turning it off then on again always corrected it. I thought perhaps the FW was not loading properly. Mine is the ver.B, about a year old.

mgx0:
Hello after 4 years :) It's getting colder outside, is anybody having this problem again? I experienced it in temperatures below 15 degrees, repeatedly. I'm on the currently latest stable release of OpenTX, Taranis X9D (No Plus). Maybe something needs replacing, but is there anything I can check please?

I have had a pause in flying for 2 years, so I don't really know whether this problem would occur earlier or not :(

raphaelcoeffic:
@mgx0 could it be that it's outside the specs of the LCD? -15° is quite cold for the poor LCD....

mgx0:
@raphaelcoeffic nope that 15 was above zero 😊 i remember when this issue has been opened, it was around the same temperature.  And also turning off and back on the radio helped. And display never stopped working once started correctly until next shutdown. Same symptoms

Should I open new Issue? How do I find out whether it's an issue with my TX or general? My Taranis is one from the earliest X9D non-plus

kilrah:
In any case it has nothing to do with the software...

mgx0:
@kilrah could you prove your statement?

i can prove mine: this whole thread started with the same problem and has been fixed in the software. If I roll back to openTX 2.0.14, I have no problem, I just have found out

bsongis:
@mgx0 given that on your X9D 2.3 is KO would you tell me if :
- 2.0 is OK / KO
- 2.1 is OK / KO
[...]

PR #1877, opened 2014-11-26
Fixes #1875 and #1849: implemented LCD initialization for low temperatures

### LINKED ISSUE #1875: How to integrate low temperature LCD support
It has been determined in issue #1849 that the lcdInit() which works for low temperatures is (from commit de793502fa4241b0d2502cd539e28ae8725384d5):

``` c
void lcdInit()
{
  LCD_BL_Config();
  LCD_Hardware_Init();
#if defined(REVPLUS)
  initLcdSpi() ;
#endif

  // LCD_RST_HIGH();
  // Delay(20);

  LCD_RST_LOW();
  Delay(2);     // only 3us needed according to data-sheet

  LCD_RST_HIGH();
  Delay(1300);     // 150ms needed according to data-sheet, reserve for low temperature

  LCD_Init();
  AspiCmd(0xAF);    //dc2=1, IC into exit SLEEP MODE, dc3=1 gray=ON, dc4=1 Green Enhanc mode disabled
  Delay(20);      //neded for start of internal DC-DC converter
}
```

The main changes from current `master` are:
- removal of Aspi command that issues LCD controller reset. This reset was redundant, it was right after reset via RST pin. ( Even remaining reset via RST pin seems redundant, as LCD controler does reset by itself after power is applied.)
- the delay after the reset was increased from 83ms to 1300ms (this value was determined after extensive testing, see issue #1849 )

Ali in all the new code (that works at low temperatures) is about **one second slower** than before. This poses a new problem in case of unexpected shutdown (watchdog resest) where we want radio to be operational as soon as possible.

We need to decide how to integrate new code into opentx. Choices are:
1. integrate as is and we live with longer lcd initialization
2. in case of watchdog reset we use a shorter delay. This could quite possibly work, since it has been observed in tests that quick power-off/power-on sequence properly initializes display even with the current code in `master`. 
3. we could modify delay depending on temperature
4. we could modify the whole initializaiton sequence, reset LCD controller as soon as possible, then do other stuff (like reading EEPROM) and then finish initializing LCD.

Choices 3 and 4 are not my favorites, too many changes, my vote is for choice 2. @bsongis @kilrah @MikeBland what is your opinion?

projectkk2glider:
Closes #1875 and #1849

Initialization skipped in case of watchdog reset or boot-loader exit.

Only tested on Taranis, somebody else needs to check on Taranis Plus!

### DIFF
--- a/radio/src/targets/taranis/lcd_driver.cpp
+++ b/radio/src/targets/taranis/lcd_driver.cpp
@@ -16,8 +16,10 @@
 
 #if defined(REVPLUS)
   #define CONTRAST_OFS 160
+  #define RESET_WAIT_DELAY_MS      300        //wait time after LCD reset before first command
 #else
   #define CONTRAST_OFS 5
+  #define RESET_WAIT_DELAY_MS     1300        //wait time after LCD reset before first command
 #endif
 
 #if defined(BOOT)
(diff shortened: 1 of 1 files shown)


bob9:
I'm having this problem the taranis plus I got last week after upgrading to the latest firmware. 

I turn on taranis and intermittently the screen stays blue and the audio starts talking. The taranis seems to work fine just with no screen. 

I'm concerned I have a faulty taranis plus.. Should I be sending it back?

tomkfly:
Have you loaded the correct Firmware?  Also this is the wrong place to ask your type of question, you should ask on one of the Taranis forums. This facility is for reporting Firmware bugs or requesting new functions.

Issue #1935, opened 2014-12-15
BUG Mavlink Serial Speed - Max speed actually supported is 57600 - General param 115200 Not Functionnal 


macfly1202:
Hi, 
On 9XR with EXT=MAVLINK, serial speed max supported is 57600.
If we choose to use 115200 in General Param / Baudrate. .it's not fonctionnel on radio screen "Inv mav type"

After take a look on master branch / radio/src/serial.h and cpp, it seem max serial speed is 76800. 

I've  tested it with mission planner and Ctrl+F output. Max Serial supported speed is 57200. I know this output on computer work because my Ardustaion Mega (Antenne Tracker ) use it at 115200 to output Xplane Mavlink data for dev.

It could be great to have this working for connexion with ULTIMATE LRS for example (from flipflap on RC Group - More info there  http://www.itluxembourg.lu/site/ ) 
Serial Output is 115200 in this project (and mav'baud is 19200) and unfortunatly, it's not working

bsongis:
@adriano65 This one is for you I guess!

adriano65:
@bsongis  You guess right :-)
I added 115200 serial speed to 9xr PCB but i get
#warning "baud rate achieved is higher than allowed"
I am investigating about F_CPU and friends to check if that speed
is out of Atmega64-128 capabilities (mmmmh...) or simply out of tolerance..
Please have patience

bsongis:
We will have to release 2.0.14 soon. Winter is here and my mates in the club need it for the LCD corrections from @projectkk2glider 

Would you only remove the 115200bauds choice on 9X/9XR for 2.0.14 
115200bauds support will come in 2.1 as it seems more difficult (unless you succeed of course)

adriano65:
Hi Bertrand,
i submit a pull request with corrections.

I test them and seems to work: i hope macfly1202 will help us too.

Anyway, if you prefer i can remove 115200 speed,
that wasn't managed at all in previous version.

Nice day

macfly1202:
Hi, Thanks all for this. It's not a problem for me to wait. It was just a bug submission. I bought a Taranis for Christmas, and for sure, I will not use often this 9XR now. I could test it with pleasure.

adriano65:
Thank you macfly :-)
Connection is not 'perfectly' stable because serial and internal structures
are reinitializated every n seconds...
I haven't touched this part of code on 9XR, because it seemed working and i am not
the original coder (it was Pienk)
[...]

PR #1971, opened 2015-01-04
Correction to "compilation error on 9XR with MAVLINK extension"

adriano65:
It compiles ok, but I have't tested on a 9XR radio.
Modifications are minimal and it should work


bsongis:
Thanks!

bsongis:
@adriano65 Would you take care of #1935 as well? I guess the guy who opened the issue will be able to do some tests on both this one and #1935

PR #1974, opened 2015-01-06
Issue #1935 - added Mavlink 115200 Baud speed - to test

adriano65:
I used a trick to avoid compiler warning about 115200 speed.
Tested on my APM2.5 and openlrsng settled @ 115200 and works.

Note: 115200 to 115400 difference is 1.8% -> should be compatible with RS232C standards.


bsongis:
You were quick, thanks a lot!

Issue #1399, opened 2014-06-29
[OpenTX] [minor] Option to hide inv parameter effect in the channel monitor


davxlw:
I like to set-up my models with the mixers weight sign reflecting the direction of the surfaces: "plus" to indicate up-travel and "minus" for the reverse:
![mixers](https://cloud.githubusercontent.com/assets/7569433/3422836/5720b81c-ff6a-11e3-947c-19ac9f26e91a.jpg)

In this 4 servos wing example, I have to reverse every other servo with the inv setting so that they turn in the right direction:
![servos_inv](https://cloud.githubusercontent.com/assets/7569433/3422798/99267e58-ff66-11e3-809b-13375e1fd493.jpg)

And now the channel monitor:
![cm_inv](https://cloud.githubusercontent.com/assets/7569433/3422800/daa13e18-ff66-11e3-9745-5efe055058fa.jpg)

The channel monitor shows the correct servos output, as they are mounted in a symmetrical way in the wing, but I would much prefer see the actual surfaces direction:
![cm_surfaces](https://cloud.githubusercontent.com/assets/7569433/3422807/5100ba70-ff67-11e3-9715-40e12e2f5f37.jpg)

It could be a build option, maybe non official...but I would appreciate it !!

PR #1977, opened 2015-01-06
Mixers monitor

### LINKED ISSUE #1945: [OpenTX] [Companion] [enhancement] MIXERS MONITOR
In the mind of #1399, it could be handy to have MIXERS MONITOR to see clearly what's going on, without being distracted by the servos settings:

channel_monitor
                          **short [ENT]**
mixers_monitor

It could be accessed from the CHANNEL MONITOR (by the way, CHANNELS or maybe SERVOS would be more appropriate) with a short [ENT] press, and remember it's last state (channels or mixers).

davxlw:
Fixes  #1945.

In the CHANNELS MONITOR view, pressing [ENT] would alternate between channels and mixers view.

The build option from  #1399 could be removed.


davxlw:
I understand that this new screen could confuse people...
However, I would prefer not to have to keep track of my branch every time the code evolves.
So, I suggest to replace the "CHANNELS_MONITOR_INV_HIDE" buid option by "MIXERS_MONITOR", I will try to do it myself if you wish.

bsongis:
Exactly with a Makefile option everybody will be happy. Careful, won't you break things for those who use this CHANNELS_MONITOR_INV_HIDE option and don't need this new screen? Or perhaps you are the only one using this option? I can't remember who added it, not me for sure.

davxlw:
As you can see above with #1399, I was the original applicant and yes, probably the only one using that.
I don't see any problem in replacing this option, as the mixers monitor provides this functionality without changing the normal behavior of the channels monitor. It's only better IMHO.

At worst, I would just have to add it again :wink:

MicheleVilla:
Is possible to add this option in "Companion compilations options"?
it is possible that someone else might appreciate this option

davxlw:
Ok, sorry for the mess with the commits...

When we'll agree with what to do, I'll close this pull request and make a new one.

bsongis:
Yes it is possible to add another option on the compilation server.

MicheleVilla:
With option "Change view  from Mixer monitor  to Channel monitor" by ENT Key

davxlw:
I've replaced "CHANNELS_MONITOR_INV_HIDE" with "MIXERS_MONITOR" and added that to Companion firmware build options as "mixersmon".

 It should work but it needs to be reviewed ;)

davxlw:
No comments ?

bsongis:
I still didn't read it, I will do, I promised the BT driver to FrSky first

davxlw:
No problem.

bsongis:
Would you remove the unneeded files from the review?

davxlw:
Do you talk about the .ts files ?
If so, I've planned to remove them.

projectkk2glider:
I think you should just change the line endings on files that contain huge amount of changes like companion/src/translations/companion_it.ts

davxlw:
I have already removed the .ts files...

But now that you're talking about line endings, I think the repo is a little messy in that respect, there's some inconsistency between windows and linux type. I'm on linux so it shouldn't be a problem.

bsongis:
Right the line endings will be fixed, but not now or all pull requests will be invalid! Once done I will add a check-lines-ends job to jenkins at each commit.

bsongis:
Would you add the same feature to the 'next' branch?

davxlw:
Ok, will do it this week end.

PR #1978, opened 2015-01-06
Little lcd_putsCenter correction

davxlw:
Still doesn't take the different characters sizes into account...


projectkk2glider:
Do we need this in `next`?

PR #1988, opened 2015-01-09
Port ATmega2561 PWM_BACKLIGHT fix from master branch

### LINKED ISSUE #1970: Build fails with PCB=9X2561 and PWM_BACKLIGHT=YES
It looks like the PWM_BACKLIGHT code didn't get updated with ATmega2561-specific defines and code, so there are compile errors related to the different timer registers.

Happy to fix this myself and submit a pull request, but I could use some help to determine which timer I should use. The current mega64/128 code uses timer 2. The [2561 migration app note] (http://www.atmel.com/Images/doc2581.pdf) suggests that its timer0 is equivalent to timer2 on the 64/128, but I'm concerned that timer0 might be used elsewhere. Someone familiar with the code will be able to answer this much faster and more reliably than me.

kh90909:
Fixes #1970 for the next branch.


bsongis:
Thanks!

PR #1962, opened 2014-12-29
Projectkk2glider/simu audio sdl

projectkk2glider:
This is just a proof of concept for simulator audio using SDL library. Please test and comment. It is working for me on Linux Mint 17.1 in all three simulators (simu, Companion and Simulator). 

This pull is not ready for merge!!!

PR #2026, opened 2015-01-18
Projectkk2glider/simulator audio

### LINKED ISSUE #457: Audio not playing in companion9x simulator
_From [mickwp...@gmail.com](https://code.google.com/u/111814648479692963065/) on September 16, 2013 07:59:48_

What steps will reproduce the problem? Companion9x does not play audio or sound track on PC

Should be able to preview/simulate the audio custom function or audio custom switch using the companion9x simulator

Version Companion9x  1.47
OS: windows 7 64bit
Firmware opentx0r2751 (Taranis)

_Original issue: http://code.google.com/p/companion9x/issues/detail?id=257_

projectkk2glider:
Re #457: case insensitive f_stat() and f_open() for SIMU
Re #457: corrections for Windows build

Closes #457 

This pull replaces #1962

I tested this on Linux Mint 17.1 64bit and Windows XP 32bit (running inside virtualbox). On both systems simulator audio is working good. 

TODO: we have to add some kind of gain setting to Companion so user can adjust audio volume independently of the radio settings (mainly we need positive gain, because the audio is usually too quiet).


bsongis:
I think there is some inconsitency between those 2 variables:
DISKIO_SIMU = NO
SIMUAUDIO = NO

projectkk2glider:
Do you have any recommendation?

bsongis:
I would have recommanded SIMU_AUDIO SIMU_DISKIO because we already have TRACE_SDCARD TRACE_AUDIO etc.

projectkk2glider:
Looks like a rebase is needed also.

projectkk2glider:
Rebase done, also squashed all commits into one.

bsongis:
Yes, that's a nice one! Thanks for the boy scout work you did here and there (some come from th9x). I guess we have to split simpgmspace in the future! And also perhaps make the mixer run in his own thread, the simulation will be better.

bsongis:
This one has to be porte to 'next', right?

PR #2035, opened 2015-01-21
Fixes #1937: reverse polarity on external CPPM output (Taranis)

### LINKED ISSUE #1937: Taranis  and Taranis Plus   Ext. Modul setting +/- CPPM wrong
Taranis  and Taranis Plus   both openTx V2.013

Setting of ext Modul  PPM Protokolll
22,5ms 300us   + or -  for rising Edge and falling Edge of CPPM startimpuls ist wrong
 Plus    + must be rising edge  of CPPM-Signal, but its falling
 Minus  - must be falling edge of CPPM-Signal, but its rising 

------------------------
Setting  of Master / Trainer Slave PPM Protokoll is ok
Plus     + is rising Edge
Minus  - is falling Edge

Have both measerd today with a Digital Oszi

HRenz

projectkk2glider:
Removed support for internal module PPM mode (Taranis)
Hw specific code moved to drivers

I have tested this on my Taranis radio and oscilloscope. I measured both external module and trainer port signals before and after this modification. My tests show that it now works correct in every mode. 

**But I won't let this be merged until somebody else also tests this on their radio**.

Another thing: There are already upset/confused users on forums. They are mainly annoyed by the fact that such important thing as PPM polarity has changed in 2.0.15 (fortunately in reality it hasn't changed yet) and they were not aware of that. They say that they were not informed about this at the upgrade time.  We have a note about this in the firmware changelog, but I gues that is not enough for everybody. Do we have any solution for this, because this pull will **really change** PPM polarity?

### DIFF
--- a/radio/releasenotes.txt
+++ b/radio/releasenotes.txt
@@ -1,3 +1,11 @@
+<h2>Version 2.0.16 / todo</h2>
+
+[Taranis]
+<ul>
+<li>Fixed reverse polarity on external PPM output. <b>Users with external module in PPM mode should check and update PPM polarity setting.</b></li>
+</ul>
+
+
 <h2>Version 2.0.15 / 2015-01-13</h2>
 
 Nothing new (only Companion fixes)
@@ -14,7 +22,7 @@ Nothing new (only Companion fixes)
 <li>Added switch auto detection in Timer source</li>
 <li>Totally removed splash screen display if duration is set to ---</li>
 <li>Back-light turned on sooner at startup (Taranis +)</li>
-<li>Fixed reverse polarity on external CPPM output</li>
+<li><strike>Fixed reverse polarity on external CPPM output</strike> (update: this was not fixed in 2.0.14)</li>
 <li>Suppressed center beep for non-existing pots</li>
 <li>Fixed THt persistent timer running after model load (with Throttle at -100%)</li>
 <li>Fixed buffer overflow for automatic flight mode change audio announcements (if very long model and flight mode names were used)</li>
--- a/radio/src/Makefile
+++ b/radio/src/Makefile
@@ -307,6 +307,12 @@ TRACE_SD_CARD = NO
 TRACE_FATFS = NO
 TRACE_AUDIO = NO
 
+# Enable internal module PPM mode for Taranis 
+# Notice: enabling this only enable code in the driver, 
+# the menu selection is still not possible.
+# Values = NO, YES
+TARANIS_INTERNAL_PPM = NO
+
 #------- END BUILD OPTIONS ---------------------------
 
 # Define programs and commands.
@@ -712,6 +718,9 @@ ifeq ($(PCB), TARANIS)
   endif
   ifeq ($(DEBUG_TRACE_BUFFER), YES)
     CPPDEFS += -DDEBUG_TRACE_BUFFER 
+  endif
+  ifeq ($(TARANIS_INTERNAL_PPM), YES)
+    CPPDEFS += -DTARANIS_INTERNAL_PPM
   endif
   LDSCRIPT = targets/taranis/stm32_flash_bl.ld
   TRGT = arm-none-eabi-
--- a/radio/src/pulses/ppm_arm.cpp
+++ b/radio/src/pulses/ppm_arm.cpp
@@ -88,5 +88,10 @@ void setupPulsesPPM(unsigned int port)                   // Don't enable interru
   else if (port == EXTERNAL_MODULE) {
     set_external_ppm_parameters(rest, ppmDelay, !g_model.moduleData[EXTERNAL_MODULE].ppmPulsePol);
   }
+#if defined(TARANIS_INTERNAL_PPM)
+  else if (port == INTERNAL_MODULE) {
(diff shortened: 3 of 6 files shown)


bsongis:
What do you suggest? If people are upset/confused because we don't announce something that we don't do, it's a little bit problematic ;)

bsongis:
I reviewed the code and it's good (better than what we had before). I will merge it as soon as deadcode is completely removed. We will have more chance to have tests done after the merge than before.

Would you add this to the release note, we could have it in the first line, so that the users will see it quickly.

mpjf01:
The people that are referred to were advised of the change and they asked what had to be done by them to deal with it. There was no information provided about the likely impact. A change that has the capacity to render (does it though?) every model using an external module for every user inoperable should have a bit more fanfare about it's announcement than just an entry in the release notes.

It may have been a case where you didn't announce something that you didn't do, But that was an accident. You didn't announce something that you thought you did do, That's the point that Damjan is making.

Given that most users who upgrade do so when prompted by OTXCompanion I ask that a warning put there in the window that advises that new firmware is available etc.

Naturally if the change only affects a subset of users of external modules than that information should be provided as well.

We really don't want people getting to the flying field, selecting their favourite model on their radio, finding that something doesn't work, and not know why or how to fix it until they go home, do the research and find an entry in the firmware release notes.

projectkk2glider:
@kilrah in https://github.com/opentx/opentx/issues/1937#issuecomment-70954634 is saying that some people use Tranis internal module in PPM mode, but I am skeptical about that, because there is no provision in menus to choose PPM mode for internal module anywhere.

@bsongis says that this code shuld be removed. I can do a compromise and put some fidefs around this parts of the code, but the functionality of it will be untested. Or we could completely remove it. I need your decision.


Regarding notification for users: what @mpjf01 suggests:

>Given that most users who upgrade do so when prompted by OTXCompanion I ask that a warning put there in the window that advises that new firmware is available etc.

That additional window would have to be coded in and the message that would be displayed would have to come from the server and be linked somehow to the specific firmware version. I am afraid that it is not a simple feature and something that can be done very quickly.

ckleanth:
RE support for internal module PPM mode
what if someone hacks out the internal module to use a different module altogether
for example rip of the internal XJT and fit one of these instead
http://www.xtremepowersystems.net/products.php?cat=12

not everyone is using FrSky equipment and opentx is well open to any system

mpjf01:
Would it be possible to add a sentence to the existing message saying something like "If you are upgrading from 2.1.15 and if you are using an external module please check the operation of the module for each model before use. See release notes"

mpjf01:
Just for a few weeks until most users have done the upgrade.

mpjf01:
Adding a link to the release notes would be helpful too.

projectkk2glider:
>Adding a link to the release notes would be helpful too.

In Companion: Help -> Firmware Changes...

projectkk2glider:
I think I addressed all comments. 

The internal module PPM mode is now optional with TARANIS_INTERNAL_PPM (YES, NO) makefile option, but the menu selection is not possible.

The explicit notification dialog about PPM polarity change is still missing. Somebody else will have to do that.

bsongis:
Is there really somebody using PPM on the internal module? I didn't hear of this!

ckleanth:
I know at least one person.
http://www.rcgroups.com/forums/showpost.php?p=30554873&postcount=54232

bsongis:
ok, then we will keep it :)

projectkk2glider:
From the comments in the code it looks like PPM on internal module was used by developers before PXX protocol was implemented. It would be a shame to completely remove this code, we might need it in the future. I think current solution with ifdefs is reasonable.

bsongis:
Yes, it's ok, @ckleanth found a user for it, so we will keep it.

kilrah:
> A change that has the capacity to render (does it though?) every model using an external module for every user inoperable should have a bit more fanfare about it's announcement than just an entry in the release notes.

95% of things you might connect to the external module slot don't care about PPM polarity. If someone has a module that does then they'll already know about it and will know what to do. It's nothing big really.

> Adding a link to the release notes would be helpful too.

clipboard03

The "upgrade available" window has a Release notes button right there. I really wouldn't want to go as far as disabling the Yes button until the user has viewed the release notes... 

> We really don't want people getting to the flying field, selecting their favourite model on their radio, finding that something doesn't work, and not know why or how to fix it until they go home, do the research and find an entry in the firmware release notes.

We have repeated many times (and it's obvious common sense) that both reading the release notes AND testing every model after any upgrade is mandatory. Anybody can decide to skip it - but then don't come and cry :)
Hopefully if they do the mistake once they'll learn. 

> @kilrah in #1937 (comment) is saying that some people use Tranis internal module in PPM mode, but I am skeptical about that, because there is no provision in menus to choose PPM mode for internal module anywhere.

As I said they compiled the code themselves (obviously with the necessary menu mods to call the driver).

bsongis:
I just checked as well, @kilrah is right there is a Release Notes button there. Perhaps we update the text and add something about what needs to be checked before flashing a new firmware?

kilrah:
Yep why not...

PR #2036, opened 2015-01-21
Bsongis/issue1924 file copy rename in sdmanager

bsongis:
The 2k RAM saved won't be bad, I don't know exactly where we are but I am afraid that we need them, compared to RAM used in 'master' ...

### DIFF
--- a/radio/src/sdcard.cpp
+++ b/radio/src/sdcard.cpp
@@ -141,3 +141,47 @@ bool listSdFiles(const char *path, const char *extension, const uint8_t maxlen, 
 
   return s_menu_count;
 }
+
+const char *fileCopy(const char *filename, const char *srcDir, const char *destDir)
+{
+  FIL srcFile;
+  FIL dstFile;
+  char buf[256];
+  UINT read = sizeof(buf);
+  UINT written = sizeof(buf);
+
+  char path[2*CLIPBOARD_PATH_LEN+1];
+  char *tmp = strAppend(path, srcDir, CLIPBOARD_PATH_LEN);
+  *tmp++ = '/';
+  strAppend(tmp, filename, CLIPBOARD_PATH_LEN);
+
+  FRESULT result = f_open(&srcFile, path, FA_OPEN_EXISTING | FA_READ);
+  if (result != FR_OK) {
+    return SDCARD_ERROR(result);
+  }
+
+  tmp = strAppend(path, destDir, CLIPBOARD_PATH_LEN);
+  *tmp++ = '/';
+  strAppend(tmp, filename, CLIPBOARD_PATH_LEN);
+
+  result = f_open(&dstFile, path, FA_CREATE_ALWAYS | FA_WRITE);
+  if (result != FR_OK) {
+    return SDCARD_ERROR(result);
+  }
+
+  while (result==FR_OK && read==sizeof(buf) && written==sizeof(buf)) {
+    result = f_read(&srcFile, buf, sizeof(buf), &read);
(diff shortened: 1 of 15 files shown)


radio/src/sdcard.cpp
```diff
@@ -141,3 +141,47 @@ bool listSdFiles(const char *path, const char *extension, const uint8_t maxlen,
...
+  FIL srcFile;
+  FIL dstFile;
+  char buf[256];
+  UINT read = sizeof(buf);
+  UINT written = sizeof(buf);
+
+  char path[2*CLIPBOARD_PATH_LEN+1];
+  char *tmp = strAppend(path, srcDir, CLIPBOARD_PATH_LEN);
+  *tmp++ = '/';
+  strAppend(tmp, filename, CLIPBOARD_PATH_LEN);
```
projectkk2glider: CLIPBOARD_PATH_LEN is not the available space in tmp! 

I think: CLIPBOARD_PATH_LEN - (tmp - path)

The `strAppend()` also has a bug:
```c
char *strAppend(char *dest, const char *source, int len)
{
  while ((*dest++ = *source++)) {
    if (--len == 0) {
      *dest++ = '\0';
      //  ########## missing return!!! #########
    }
  }
  return dest - 1;
}

```

radio/src/sdcard.cpp
```diff
@@ -141,3 +141,47 @@ bool listSdFiles(const char *path, const char *extension, const uint8_t maxlen,
...
+
+  FRESULT result = f_open(&srcFile, path, FA_OPEN_EXISTING | FA_READ);
+  if (result != FR_OK) {
+    return SDCARD_ERROR(result);
+  }
+
+  tmp = strAppend(path, destDir, CLIPBOARD_PATH_LEN);
+  *tmp++ = '/';
+  strAppend(tmp, filename, CLIPBOARD_PATH_LEN);
```
projectkk2glider: Same as above, len wrong.

radio/src/sdcard.cpp
```diff
@@ -141,3 +141,47 @@ bool listSdFiles(const char *path, const char *extension, const uint8_t maxlen,
...
+
+  result = f_open(&dstFile, path, FA_CREATE_ALWAYS | FA_WRITE);
+  if (result != FR_OK) {
+    return SDCARD_ERROR(result);
```
projectkk2glider: Jut for completeness sake, here you forgot `f_close(&srcFile)`, but current FatFs implementation won't care.

bsongis: True, I will add it

radio/src/sdcard.cpp
bsongis: Obviously yes, I think I wanted to put a break or I wouldn't have written dest++ there!

bsongis: I don't get what you write about CLIPBOARD_PATH_LEN. The directory max len is CLIPBOARD_PATH_LEN, the same for for the file. Therefore the buffer is 2*CLIPBOARD_PATH_LEN and it remains at least CLIPBOARD_PATH_LEN bytes available in path (at position tmp), did I miss anything?

projectkk2glider: My bad, I missed that it was twice bigger.

bsongis:
What about porting this to 'master' as it saves 2k RAM?

projectkk2glider:
>What about porting this to 'master' as it saves 2k RAM?

I am for that, but we need to test this extensively before the new release. If you create a branch on top of the master, I promise to test it on my radio.

PR #2039, opened 2015-01-23
Fixes #2007: handle full SD card in openLogs()

### LINKED ISSUE #2007: When SD card overflow Taranis work very slow

projectkk2glider:
There are more changes than I really made, probably due to mixed line endings in some files. The debug traces for file access in simuspace are also improved.

I haven't tested how long `f_getfree()` would take on very big SD card, on mine (2G) seems fine. I've tested this on my radio, when SD card gets full you get a nice error message.

If accepted, I will port to next.


projectkk2glider:
I am not entirely happy with this solution, because when the first time `f_getfree()` is called it takes longer time. This produces audible pause in voice anouncement that I have configured to play when I enable SD logs.

I propose you wait with merge until I came up with better solution. I have couple of ideas:
* call `f_getfree()` when SD is initialized
* or write anew function that will return when first free cluster is found.

I have another idea to write/add diskio simulation to stand-alone simulator that would operate on disk image. This could then be used to trace/debug FatFs and to also analyze disk IO patterns. Maybe we can also solve why audio sometimes stutters.

projectkk2glider:
I have now implemented diskio simulation and got interesting results:
* there are many calls to `disk_status()` function. Fortunately our implementation is very short.

* when browsing bitmaps on SD card some take couple of disk sector reads while other take much more. Problematic bitmaps take many reads of FAT directory table to find their position. This are not fragmented files, just their name is deeply stacked in FAT structure. So every time such file is open the FAT has to be parsed from the beginning.

* as seen from the code FatFs has 512 bytes (1 sector) buffer for each file. Actual write to disk happens only when that buffer is full or if the sector boundary is crossed. That is good. Same applies to disk reads.

* When `f_getfree()` is called for the first time, it reads more then 300 sectors on my 2GB card. Subsequent calls don't read card at all.

projectkk2glider:
Standalone simu now supports new option DISKIO_SIMU if YES then FatFs is used and diskio functions are simulated. Simulated io uses `sdcard.image` file as block device - this file must contain all sectors of SD card. Easiest to create from existing SD card with `dd` command (copy whole device, not just first partition).

I've also implemented SD info menu for Taranis and moved couple of common functions to the new file `sdcard.cpp`.

radio/src/FatFs/ff.c
```diff
@@ -122,9 +122,10 @@
 #include "ff.h"			/* Declarations of FatFs API */
 #include "diskio.h"		/* Declarations of disk I/O functions */
 
-
-
-
+#include "debug.h"
```
bsongis: Would you remove changes in this file, I prefer we change it as little as possible!

radio/src/simu.cpp
```diff
@@ -396,7 +396,12 @@ int main(int argc,char **argv)
   simuSetSwitch(1, 0);
 #endif
 
-  return application.run();
+  int result = application.run();
+
+  StopMainThread();
+  StopEepromThread();
```
bsongis: It make me think we should have a thread for the mixer here

bsongis:
Perfect!

PR #2073, opened 2015-02-15
Bsongis/issue 1925 screenshots on taranis

bsongis:


bsongis:
No comment? Let's merge it then!

projectkk2glider:
Just tested in simulator, works excellent :+1: 

We just need to remind users to create SCREENSHOTS directory otherwise this won't work.

bsongis:
Normally no, it should be automatically created, just as [MODELS] or [LOGS]

projectkk2glider:
You are right, the problem is in simulator which does not return proper error on missing dir. Then the missing dir is not created. But as you say it should work on the radio.

projectkk2glider:
It will also work on simulator when #2093 is merged!

PR #2086, opened 2015-02-16
Projectkk2glider/issue 2084 lua insert input

### LINKED ISSUE #2084: Can't add new input with Lua model.insertInput()
Test script:
```lua
local function run(event)
  local ivalue={name="test", source=MIXSRC_Thr, weight=55, offset=3}
  model.insertInput(6, 0, ivalue)
  return 1
end

return {run=run}
```

projectkk2glider:
Closes #2084

Script for testing:
```lua


local function addInput(input, configuraton)
  local noInputs = model.getInputsCount(input)
  model.insertInput(input, noInputs, configuraton)
end

local function listInputs()
  for n = 0, 10, 1 do
    local noInputs = model.getInputsCount(n)
    print("input["..n.."] has "..noInputs.." lines")
    if noInputs > 0 then
      for m = 0, noInputs-1, 1 do
        local input = model.getInput(n, m)
        print("\tline "..m..": src="..input.source.." wt="..input.weight.." name="..input.name)
      end
    end
  end
end

local function deleteLastInputLine(input)
  local noInputs = model.getInputsCount(input)
  if noInputs > 0 then
    model.deleteInput(input, noInputs-1)
  end
end

local function run(event)

  print("Before:")
  listInputs()

  -- add inputs
  addInput(7, {name="test1", source=MIXSRC_Thr, weight=56, offset=3})
  addInput(1, {name="one", source=MIXSRC_Rud, weight=-30})
  addInput(100, {name="one", source=MIXSRC_Rud, weight=-30})  -- invalid input number, should do nothing

  print ("AFTER:")
  listInputs()

  -- delete one input
  deleteLastInputLine(7)
  print ("after delete:")
  listInputs()

  return 1
end

return {run=run}
```

If accepted, I will make the same changes to the `next` branch.

### DIFF
--- a/radio/src/lua.cpp
+++ b/radio/src/lua.cpp
@@ -673,6 +673,7 @@ static int luaModelGetInput(lua_State *L)
     lua_pushtableinteger(L, "source", expo->srcRaw);
     lua_pushtableinteger(L, "weight", expo->weight);
     lua_pushtableinteger(L, "offset", expo->offset);
+    lua_pushtableinteger(L, "switch", expo->swtch);
   }
   else {
     lua_pushnil(L);
--- a/radio/src/lua.cpp
+++ b/radio/src/lua.cpp
@@ -625,35 +625,31 @@ static int luaModelResetTimer(lua_State *L)
   return 0;
 }
 
-static int getFirstInput(int chn)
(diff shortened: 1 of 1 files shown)


bsongis:
Thanks! Would you add your tests to non regression tests?

PR #2088, opened 2015-02-17
Saves RAM on Taranis

bsongis:
Saves CPU (mixes calculated only when needed) and RAM (pulses arrays for PXX/DSM2/PPM shared) - Still not tested, needs a code review first!

### DIFF
--- a/radio/src/pulses/pulses_arm.cpp
+++ b/radio/src/pulses/pulses_arm.cpp
@@ -40,7 +40,7 @@ uint8_t s_pulses_paused = 0;
 uint8_t s_current_protocol[NUM_MODULES] = { MODULES_INIT(255) };
 uint16_t failsafeCounter[NUM_MODULES] = { MODULES_INIT(100) };
 uint8_t moduleFlag[NUM_MODULES] = { 0 };
-uint32_t nextMixerTime[NUM_MODULES] = { 0 };
+U64 nextMixerTime[NUM_MODULES] = { 0 };
 #define SCHEDULE_MIXER_START(module, delay) nextMixerTime[module] = (CoGetOSTime() + (delay)/2 - 1/*2ms*/)
 
 ModulePulsesData modulePulsesData[NUM_MODULES];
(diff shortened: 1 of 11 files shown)


radio/src/pulses/pulses_arm.cpp
```diff
@@ -40,6 +40,11 @@ uint8_t s_pulses_paused = 0;
 uint8_t s_current_protocol[NUM_MODULES] = { MODULES_INIT(255) };
 uint32_t failsafeCounter[NUM_MODULES] = { MODULES_INIT(100) };
 uint8_t moduleFlag[NUM_MODULES] = { 0 };
+uint32_t nextMixerTime[NUM_MODULES] = { 0 };
+#define SCHEDULE_MIXER_START(module, delay) nextMixerTime[module] = (CoGetOSTime() + (delay)/2 - 1/*2ms*/)
+
```
projectkk2glider: `CoGetOSTime()` returns 64 bit integer!

radio/src/tasks_arm.cpp
```diff
@@ -145,7 +148,14 @@ void mixerTask(void * pdata)
       if (t0 > maxMixerDuration) maxMixerDuration = t0 ;
     }
 
-    CoTickDelay(1);  // 2ms for now
+    int delay = 10; // 20ms default
+    if (GET_MIXER_DELAY(0) >= 0)
+      delay = min(delay, GET_MIXER_DELAY(0));
```
projectkk2glider: You have two calls to the GET_MIXER_DELAY() which don't get optimized (64 bit substraction), so we end up doing substraction two times. Maybe get the value in temporary variable and work with that value:
```c
    int delay = 10; // 20ms default
    int moduleDelay = GET_MIXER_DELAY(0);
    if (moduleDelay >= 0)
      delay = min(delay, moduleDelay);
#if NUM_MODULES >= 2
    moduleDelay = GET_MIXER_DELAY(1);
    if (moduleDelay >= 0)
      delay = min(delay, moduleDelay);
#endif
```

radio/src/pulses/pulses_arm.cpp
```diff
@@ -144,18 +149,22 @@ void setupPulses(unsigned int port)
   switch (required_protocol) {
     case PROTO_PXX:
       setupPulsesPXX(port);
+      SCHEDULE_MIXER_START(port, 9);
```
projectkk2glider: Are you confident with this schedules? Could we use heartbeat signal from XJT module for scheduling?

bsongis: The heartbeat could be used to synchronize the pulses, that's one step further, here we synchronize the mixer with the pulses

kilrah:
Hmm both PXX and DSM to external module are broken. PPM is OK, and PXX to internal is OK.All good in next.

Trainer out is OK, Trainer IN is broken (all modes), but doesn't work in next either.

bsongis:
Moved to another branch

PR #2101, opened 2015-02-20
Projectkk2glider/ram savings

projectkk2glider:
This saves about 3kB of RAM. I have tested this a while back on master branch and it worked good. This commits on next are untested on real radio, but it should work.

### DIFF
--- a/radio/src/gui/Taranis/view_statistics.cpp
+++ b/radio/src/gui/Taranis/view_statistics.cpp
@@ -149,12 +149,14 @@ void menuStatisticsDebug(uint8_t event)
   lcd_puts(lcdLastPos, MENU_DEBUG_Y_MIXMAX, "ms");
 
   lcd_putsLeft(MENU_DEBUG_Y_RTOS, STR_FREESTACKMINB);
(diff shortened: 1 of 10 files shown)


bsongis:
Then I merge it :) I am curious, would you do a comparaison between 'master' and 'next' for RAM usage (without double buffering)?

projectkk2glider:
Map comparison between `master` and `next`:

`__bss_end__` is at:
* master = 0x2000c1e0  (49632 bytes)
* next = 0x2000be00 (48640 bytes)

So next uses 992 bytes less for BSS and has 3k more HEAP space.

Used compilation command:
```
make PCB=TARANIS  HELI=YES GVARS=YES AUTOSWITCH=YES PPM_LIMITS_SYMETRICAL=YES PPM_CENTER_ADJUSTABLE=YES  LUA=YES  NANO=YES
```

Compiler:
```
$ arm-none-eabi-gcc --version
arm-none-eabi-gcc (GNU Tools for ARM Embedded Processors) 4.7.4 20130913 (release) [ARM/embedded-4_7-branch revision 202601]
```

bsongis:
Excellent :)

PR #2104, opened 2015-02-22
Mixers Monitor in the simulator

davxlw:
I had forgotten the simulator...


bsongis:
Hmm. no, I will have a lot of emails if I accept it!!!

davxlw:
Ok, I'll remove it but, could the simulator reflect the build options chosen in Companion ?

bsongis:
Yes it's possible, but a lot of work, we need to support DLLs / .so etc. which would be built on the compilation server just like our firmwares!

davxlw:
Couldn't be possible to build the simulator with all the options, and activate them at run time relative to the build options set in the profile settings ?

bsongis:
But you noticed that those options are compile options!

davxlw:
Yes, I tend to forget the close link between firmware and companion (and so simulator)...

PR #2132, opened 2015-03-09
Changelog links now open in external browser

projectkk2glider:
I think this is a cool feature when combined with the release notes that contain links to fixed issues.


bsongis:
Good idea

projectkk2glider:
Ported to next in 37a6f83f1b5d1441ef8a8e0e1fe33d10a2c3f648

Issue #2122, opened 2015-03-03
% in model name not saved


JoeB99:
Using companion, i named a model "MSR +125% o2" and the percent symbol is replaced with a space when the eepe is saved.

kilrah:
Special characters other than those you can enter on the radio (_ - . ,) aren't available in model names. Known limitation, was decided not to act on it but maybe we could also limit entry to supported chars in companion.

#651 is loosely related.

JoeB99:
Thanks- companion appears to prevent unsupported chars on other 'name' inputs, but not on the model name.

griebd:
Quite a simple solution, assuming the rules for model names are the same as for flight mode names....

in companion/src/modeledit/setup.cpp:
```
@@ -377,6 +377,8 @@ SetupPanel::SetupPanel(QWidget *parent, ModelData & model, GeneralSettings & gen
 
   ui->setupUi(this);
 
+  QRegExp rx(CHAR_FOR_NAMES_REGEX);
+  ui->name->setValidator(new QRegExpValidator(rx, this));
   ui->name->setMaxLength(IS_TARANIS(board) ? 12 : 10);
 
   if (firmware->getCapability(ModelImage)) {
```

though I use git, I'm new to GitHub.... tried to assign the issue, but I probably need permission for that, right?

projectkk2glider:
>though I use git, I'm new to GitHub.... tried to assign the issue, but I probably need permission for that, right?

Correct, you need to be a member of OpenTX github organization. But before you become a member you can fork OpenTX repository, make changes there and create a pull request against original OpenTX branch. You should base your work on OpenTX `master` branch which contains 2.0.x development.

PR #2136, opened 2015-03-12
Fix to #2122

griebd:
Limit input characters for model name in companion. Use the same rules as for flight mode names.
Allowed characters " A-Za-z0-9_.-," (space included).


bsongis:
Thanks!

griebd:
yes, no problem!

PR #2142, opened 2015-03-13
Fix #2122 on next branch

### LINKED ISSUE #2122: % in model name not saved

griebd:
Same fix done on master branch.


bsongis:
Thanks!

PR #2143, opened 2015-03-14
Fix #2091

### LINKED ISSUE #2091: [Bug][Companion 2.0.15]Various bug on model wizzard
Model wizzard allows only to select Channel 1 to 8 on Taranis even if Channel 1 to 32 are available.

bugwizard3

On "Tail" window, when Channel 1 to 8 are used for other control, the dropdown propose only Channel 1 and "Next" button has no effect (only "Prev" or "Cancel" buttons have an effect)

bugwizard1

griebd:
Just added a message on tail page if there is only one channel left to use. Nothing else changed.

Next branch has the same problem as master, but by looking at the wizard code I think we should think about some bigger changes for next. To start with, all pages are validated when "next" button is pressed and there is no error message on problems, it simple does nothing...

### DIFF
--- a/companion/src/wizarddialog.cpp
+++ b/companion/src/wizarddialog.cpp
@@ -1,6 +1,6 @@
 /*
  * Author - Kjell Kernen
- * 
+ *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License version 2 as
  * published by the Free Software Foundation.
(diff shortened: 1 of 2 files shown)


companion/src/wizarddialog.cpp
```diff
@@ -181,6 +181,16 @@ int StandardPage::nextFreeChannel(int channel)
   return -1;
 }
 
+int StandardPage::totalChannelsAvailable()
+{
+  int c = 0;
+  for(int i=0; i<WIZ_MAX_CHANNELS; i++)
```
bsongis: Cosmetics, but it's better to add { here

Issue #1471, opened 2014-07-08
read/write file capability for Lua 


ckleanth:
I was wondering if you can help.

I wanted to create a script/training tool that the user predefine a time history of stick input pattern (for example rolling harrier http://www.rcgroups.com/forums/showpost.php?p=28824025&postcount=10254) and the script will beep within some tolerance from the predefined path. 
 Also I wanted to implement a slow/speedup function

I have had a discussion with projectkk2glider how we would go about to do this using lua and basically (I copy/paste info here) we need read/write file capability to store reference stick movements that is to put them into some kind of log file.
The Lua script would then read this file and compare actual stick positions with recorded ones. You would of course have to have a very good algorithm that would detect if actual stick movements are ahead of behind the recorded stuff. But that is another problem altogether.

I'm an engineer, I can program but I'm no programmer and never came across lua up until it was introduced in OpenTx 2.0.

Can the developers please provide some help on how to implement this?

jermoo:
+1 for io.lua

bsongis:
Tests needed. I will add seek feature after first tests OK

projectkk2glider:
Done

PR #2145, opened 2015-03-16
#1471 - libio added to Lua. For tests, can be simplified and expanded

bsongis:


bsongis:
```lua
local function run(event)
  local f = io.open("/opentx.h")
  local str = io.read(f, 10)
  print(str)
  io.close(f)
  return 1
end

return { run=run }
'''

bsongis:
also io.write is implemented

bsongis:
and io.close

bsongis:
If it works it's easy to add the seek, I will add it if nobody else does it before

kilrah:
Hmm, with that script I just get a brief flash of the screen.

Changed it to:

```lua
local function run(event)
  local f = io.open("test.txt")
  local str = io.read(f, 10)
  lcd.clear()
  print(str)
  io.close(f)
  return 0
end

return { run=run }

The test.txt file (with a "hello world!" inside) seems to be found as I get a syntax error if not present, but I just get a blank screen.

bsongis:
Yes, this print was for simu (it only writes text on the console!). Would you try with lcd.drawText(...) ?

kilrah:
Hmm the "a" still seems to do the same as "w"

bsongis:
The "a" should first go to the end of file

PR #2151, opened 2015-03-18
Factory reset function (manual EEPROM format)

kilrah:
To be discussed for inclusion.
Did it mostly for myself for testing, but we did have some people request a way to reset their radio to factory defaults in the past.

Basically, the question is "do we trust users to understand they're erasing everything"...


radio/src/eeprom_rlc.cpp
```diff
@@ -1016,6 +1016,21 @@ void eeLoadModel(uint8_t id)
   }
 }
 
+void eeErase(bool warn)
```
bsongis: Not too much falsh added on stock 9X?

radio/src/translations/en.h.txt
```diff
@@ -925,6 +925,8 @@
 #define TR_INPUTS              "Inputs"
 #define TR_OUTPUTS             "Outputs"
 #define TR_EEBACKUP            "\004[ENTER Long] to backup the EEPROM"
+#define TR_FACTORYRESET        "\012[MENU Long]: Factory reset"
```
bsongis: I would see a contextual menu + a confirmation, no?

bsongis:
Yes I think it's a needed feature, at least until the EEPROM format v217 is in stone

radio/src/translations/en.h.txt
kilrah: See line just below :)

radio/src/eeprom_rlc.cpp
kilrah: Didn't check yet. Figured it would add "a few bytes"...

kilrah: It actually saves 2 bytes on my compiler version!

radio/src/translations/en.h.txt
bsongis: Yes the confirmation is there, but not a contextual menu. I am not sure what is best. Perhaps it's good to keep this as it is, but if we have a new entry to add, we put everything in a menu.

kilrah: Oh that, OK. Yes until something else is needed...

bsongis:
Perhaps we add an option in companion, I already introduced something like this for the Bind / Range Check in the special functions.

bsongis:
Ok go, if somebody is not happy we will add a Makefile option

PR #2166, opened 2015-03-24
Fix to allow Companion to compile on Windows using Cygwin (posix) Issue 2164

jtaylor2:


bsongis:
Perfect, thanks!
If you have time, would you update the wiki for this compilation method?

PR #2167, opened 2015-03-25
Bsongis/telemetry totalize

kilrah:
Should be quite good now...


bsongis:
Perhaps the tts script as well? At least french :)

PR #2170, opened 2015-03-27
Fix switch handling on X9E, cleanup/harmonize code for all revisions

kilrah:


radio/src/gui/Taranis/helpers.cpp
```diff
@@ -38,14 +38,8 @@
...
-  if (source <= 5*3)
-    return MIXSRC_FIRST_SWITCH + (source-1) / 3;
-  else if (source <= 17)
-    return MIXSRC_SF;
-  else if (source <= 20)
-    return MIXSRC_SG;
-  else
-    return MIXSRC_SH;
+  div_t qr = div(source-1, 3);
+  return qr.quot+MIXSRC_FIRST_SWITCH;
```
bsongis: Nice idea!

PR #2171, opened 2015-03-28
Companion/Sim: Add more fields to telemetry simulator UI

mpaperno:
Hello, is this the right place/branch for this?  I don't really use the "next" branch.  LMK if there's a better way.

Anyway, this just adds a bunch more telemetry values that can be sent to the simulator (F4 feature in Sim GUI).  Can use hex notation for values, and blanks are now skipped instead of sending zero value.

Cheers, 
-Max

simu-values

### DIFF
--- a/companion/src/simulation/telemetrysimu.cpp
+++ b/companion/src/simulation/telemetrysimu.cpp
@@ -60,32 +60,141 @@ void generateSportPacket(uint8_t * packet, uint8_t dataId, uint8_t prim, uint16_
 void TelemetrySimulator::generateTelemetryFrame()
 {
   static int item = 0;
+  bool ok = true;
   uint8_t buffer[FRSKY_SPORT_PACKET_SIZE];
+  memset(buffer, 0, sizeof(buffer));
 
   switch(item++) {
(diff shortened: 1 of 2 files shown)


kilrah:
Seems good, thanks!
But before I merge could you please just fix indents to our standard? No tabs, 2 spaces per indent.

mpaperno:
My bad, fixed.

kilrah:
Thanks, merged for 2.0.16. If you want it to also be in the upcoming 2.1 version you should also create a similar pull request based on the next branch.

projectkk2glider:
@mpaperno nice :+1:

Awapp:
Very nice. Is it possible to add "Cell" since this is my major alarm.

mpaperno:
I believe that the value for "cell" (the cell with the lowest voltage) is derived from the "CELLS" telemetry value(s) received.

-Max

kilrah:
Correct. The simulator expects a 32-bit value containing all cells according to the telemetry protocol, and the firmware then calculates Cell and Cells form those values.

PR #2173, opened 2015-03-28
Companion/Sim: [next] Add more fields to telemetry simulator UI

mpaperno:
Same as #2171 but for 'next' branch.

Thanks!
-Max


kilrah:
The #if defined(XJT_VERSION_ID) needs to stay!

mpaperno:
Good catch, sorry!

Issue #2129, opened 2015-03-08
Raw input marking


mhotar:
Hello,
I feel a little lack of graphics to distinguish the physical inputs/formatted inputs.
At least raw gimbal inputs need some icon :) 

![snapshot_21](https://cloud.githubusercontent.com/assets/6114100/6547378/5147ed6a-c5d6-11e4-8963-0d093c554971.png)

bsongis:
Really nice!

mhotar:
Perfect ;)

Let's continue then :)  The timer icon, and sound/voice icons.

I am not sure if they are to be used globally or to use them in the custom language font only, for use in the languages where it is not easy to find short enough expression for particular sentence.

![snapshot_02](https://cloud.githubusercontent.com/assets/6114100/6580314/0c6e837c-c750-11e4-80dd-2ff15200a7e8.png)
![snapshot_03](https://cloud.githubusercontent.com/assets/6114100/6580318/1147b6d4-c750-11e4-873a-4c83cff97df3.png)

![snimek obrazovky porizeny 2015-03-10 18 16 23](https://cloud.githubusercontent.com/assets/6114100/6580537/9b146e42-c751-11e4-9e2d-3aa200e9f2bc.png)

kilrah:
Don't agree with those. 
The graphics IMO are nice when they are there to complement info you already have (like with the sources), not when they replace the data altogether as they are often confusing on their own (just like the graphical menu discussion).

mhotar:
Ooops, Kilrah is back. So am I at least successful in 50% here?

I agree here, It is better to use text, when you can translate it to fit to given space meaningfully.

bsongis:
I prefer to remain as it is now. The sticks oees are especially good to differenciate from inputs with the same name. And I like a lot the trim one, I was happy to remove the "tRl" "tRr" "tEd" "tEu" "tTd" "tTu" "tAl" "tAr"

mhotar:
You both convinced me again :)

dkisselev:
I was working on another change with custom functions and found that these changes broke how trims are rendered on the 9XR:

![snapshot_01](https://cloud.githubusercontent.com/assets/945848/6882831/86b9a7d4-d550-11e4-9282-3f989479304d.png)

Just wondering what the best way to fix it would be, a conditional define that renders the old tRl style labels, or is it possible to port the icon font over?

PR #2175, opened 2015-03-28
Fix trim switch labels on 9X

dkisselev:
The changes in #2129 broke how trim switches render on the 9X, this adds the taranis compile function to render the old tRl style labels on other devices.


kilrah:
Oh yes. Can you also fix the other languages while you're at it? Thanks!

dkisselev:
Sure, I've standardized the changes across all the languages

bsongis:
Thanks for your help!

PR #2179, opened 2015-03-31
[ARM] Pulses calculated just in time

bsongis:


bsongis:
Well, not sure we are ready for this one ;)

projectkk2glider:
This looked like a very sensible solution. Why did you ditched it?

bsongis:
Nobody tested it! And last problems with B02 radios makes me a little bit more reticent about such changes. Well, this one is not so dangerous, as there is a maximum timeout, I hesitate ...

kilrah:
Need to fix the PXX first anyhow ;)

bsongis:
True!

projectkk2glider:
After looking closely at this code, I think it does not work as intented.

The mixer task is already sleeping when the interrupt sets the time when mixes should be calculated next time. That time won't be used by the mixer thread, because it is already sleeping.

I have another but similar idea, we can use function that sets sleep time of already sleeping task. I will test and prepare this solution.

projectkk2glider:
Final implementation https://github.com/opentx/opentx/commit/29325ad54f12aa373271e77e85afd7157d747b92 is similar, but slightly different. The mixer thread wakes up every tick and checks if it needs to do a mixers calculation. There is no other simple way. The promising CoOS function `CoResetTaskDelayTick()` can not be called from the interrupt.

Issue #1679, opened 2014-09-06
Constant "beep" special function disappeard.


batzkass:
I'm using 9XR 128m 2.0.8.
Many times ago it was possible to set a constant steady beeping special function. Now it's only possible to get a short beep every 10, 20, 30, 40, 50 or 60 secs. However, it was useful to set for example a model low-voltage alarm, especially for people who don't have "Audio" nor "Voice" nor "Haptic" functions.
Thanks.

batzkass:
Hi, at this time the feature I suggested (constant beep) is not implemented. Why is this issue closed ? I can develop it but I just need a little bit of help please : where may I find the beep special function code ? (the one which can beep every 0, 10, 20, ... 60 secs ?
After that, I plan to develop different beep (monotone) melodies.

kilrah:
It's not closed, another identical issue was closed for being a duplicate.

Would probably be quicker to rewrite appropriate code than to look for how it was done as it's pretty much impossible to find when it became inactive.

batzkass:
Sorry, I misunderstood the "closed" status.
I'm not looking for the old version of this function, but rather at the actual function that only make short beeps. I already tried to find, but the code is huge for someone new.

kilrah:
See buzzer.cpp and the macros in buzzer.h.

dkisselev:
+1, just got telemetry working and the single beeps make it difficult to tell between a voltage dip just from gunning the throttle, and actual low voltage.

thatjoshguy:
Any progress on this at all?  Those of us holding out on the Taranis could really use this functionality.

PR #2181, opened 2015-04-02
Add a continuous beep mode for radios w/ buzzers

dkisselev:
This PR isn't ready to merge yet, but I'd like to submit it for some help and a code review.

This is my proposed solution for #1679,

I'd like to add a menu option right beside the 'Beep' label in Custom Functions to allow users to choose a beep type. (like how the Trainer mode comes with a channel choice)

I got the basics of the menu implemented, but a problem that I've run into is that for the buzzer, we want to have both the beep type selection menu and a timer option. This specific combination hasn't been implemented anywhere else, and no matter what I do I can't disconnect the two (changing CFN_PLAY_REPEAT also changes CFN_BEEP_INDEX even though they're in separate columns). I'm wondering if this is even possible to do or if anyone has suggestions for other ways to accomplish the general idea. (The Trainer selection is a checkbox, not a time setting, so it does work)

I wrote the function handler code such that we can also add more beep tones later on too, even though it's just beep once and continuous for now. BeepOnce is the current implementation and lets you choose those 10 second intervals. BeepCont beeps every second (and ignores the time setting)

Thanks!


kilrah:
The CFN_BEEP_INDEX() macro you added is accessing the same variable than CFN_PLAY_REPEAT() (CustomFunctionData.all.param), so it's normal changing one changes the other.

You should rather use CFN_PARAM() (which points to CustomFunctionData.value) in this case, as it is not used in the situation your improvement is useful (when AUDIO isn't defined i.e. no audio mod done).

dkisselev:
Thanks, that makes everything much more clear.

It looks like companion builds the 9XR sim with audio enabled, so it's not possible to see/tweak the buzz settings, is that the case or am I missing something?

I'll test my changes on hardware in a few days but I think the functionality is complete now.

kilrah:
You can use either the firmware simu (make simu [options] in radio/src) but that needs libfox on a linux system, or normally removing the AUDIO definition in the companion simu config should work.

dkisselev:
Yeah, I'm using the firmware simu for testing but I meant for the user side of things.

Looks like the simulators for each device are hard-compiled into the application, so users would never see a AUDIO=NO BUZZER=YES simulator inside the software unless they compiled it themselves, right?

I guess the more specific question should be, should I implement anything for this feature on the companion side of things, or would users never see it anyway?

kilrah:
Users won't see it in the simulator, but they need to be able to see/configure it in the model editor :)

dkisselev:
@kilrah, is there a way to tell if the radio/firmware uses the buzzer instead of an audio module?

On the firmware side, the state is determined by the BUZZER and AUDIO defines are set at compile-time.

But I can't find any specific code that I could hook into on the Companion side. 

Without any changes to companion,  

BeepOnce in my implementation appears in Companion as "Play Sound Beep 1"
and
BeepCont appears as "Play Sound Beep 2"

Which makes sense because the buzzer just hjacks the audio functions to beep the buzzer instead of playing sounds. This is basically the same as the old functionality (the existing beep function shows up in Comp as "Play Sound Beep 1", but I can play round with dynamically editing the options while I'm already here if it's possible.

From what I've found, I can alter things based on the radio model, but then the sound editing options go away for those that have done the audio mod...

kilrah:
Yes you should read the selected options in the firmware settings. If audio is checked then the audio mod if done.

dkisselev:
Ok, from what I've gathered, that'd done through the firmware->getCapability function? (That's how it's done for showing/hiding the telemetry display)

Which capability would I want to hook into? SoundMod and HasBeeper are the two relevant ones that I can see, but aren't don't seem equivalent to the firmware-side behavior.

kilrah:
> Ok, from what I've gathered, that'd done through the firmware->getCapability function?

Yes correct.

> Which capability would I want to hook into?

Seems there isn't one yet. You would likely want to add one, like HasAudioMod, and use id.contains("audio") to query it.

dkisselev:
Will do, thanks for the all the help, I really appreciate it.

radio/src/functions.cpp
```diff
@@ -507,14 +507,45 @@ void evalFunctions()
 #else
           case FUNC_PLAY_SOUND:
           {
+#if defined(AUDIO)
             tmr10ms_t tmr10ms = get_tmr10ms();
             uint8_t repeatParam = CFN_PLAY_REPEAT(cfn);
-            if (!functionsContext.lastFunctionTime[i] || (repeatParam && (signed)(tmr10ms-functionsContext.lastFunctionTime[i])>=1000*repeatParam)) {
+            if (!functionsContext.lastFunctionTime[i] || (repeatParam && (signed)(tmr10ms-functionsContext.lastFunctionTime[i])>=100*repeatParam)) {
```
bsongis: Is this change intentional?

radio/src/gui/9X/menu_model_custom_functions.cpp
```diff
@@ -161,6 +165,14 @@ void menuCustomFunctions(uint8_t event, CustomFunctionData * functions, CustomFu
             break;
           }
 #endif
+#if !defined(AUDIO)
+          else if (func == FUNC_PLAY_SOUND) {
+          	maxParam = 1;
```
bsongis: Cosmetics, we are used to indent with 2 chars in OpenTX

radio/src/functions.cpp
dkisselev: I don't think it was sorry. Thanks for keeping such a close eye on it.

Fixed in my working copy.

radio/src/gui/9X/menu_model_custom_functions.cpp
dkisselev: Will fix with next commit

dkisselev:
Is there any planned eta/deadline for 2.1?

I'm nearly done the changes to the companion side, but am now in the middle of school exams so haven't had a chance to work on it. I'll make sure to get things done by any deadline you guys specify though, really looking forward to seeing this live in 2.1

bsongis:
Still no deadline, it will be release in the same time than X9E

dkisselev:
Just to add a note,

I'm away from home for the summer and didn't bring an AVR programmer with me, so I'm not sure if I'll be able to finish this off for a while (I might just buy a new one, but there's no ETA).

If someone else is willing to test and finish up any last bits, that would be awesome, otherwise I'll do it when I can.

bsongis:
Deadline will come really soon, do we wait 2.1.1?

dkisselev:
Hey Bertrand, I think 2.1.1 would be easier.

It's ultimately not a huge feature, especially for the 9X series which doesn't see much development anymore. I think I have access to an AVR at work so I'll try to finalize the changes in the next couple of weeks, but probably not before 2.1

bsongis:
Perfect!

thatjoshguy:
was this not implemented in 2.1.1?  I'm on 2.1.7 and see no change.

dkisselev:
Unfortunately I've more-or-less abandoned my work on this, last summer was crazy busy and I haven't flown this season yet. I might pick it back up if I feel inspired and have some time, but I've got a lot of other projects brewing right now so it's hard to get into it (late march/early april will probably be the first opportunity for me to try). I'll try my best since it really is an important feature but can't make any promises.

if you're willing to use an outdated alpha version of OpenTX and are willing to compile it yourself, I'm pretty sure the code in this PR is mostly working, I was just trying to get the companion side of things fully synced (so that the drop downs reflected the tx side).

bsongis:
Not in 2.2

Issue #1732, opened 2014-09-19
Wrong joystick channel mapping.


Sacre100:
Version 2.0.12.

Using the USB cable, the mapping of the channel is wrong : 
Ch1 >> Ch1       -  OK
Ch2 >> Ch2       -  OK
Ch3 >> Ch3       -  OK
Ch4 >> Ch6       -  WRONG (4 and 6 are inverted)
Ch5 >> nothing  -  LOST
Ch6 >> Ch4       -  WRONG (4 and 6 are inverted)
Ch7 >> nothing  -  LOST
Ch8 >> Ch5       -  WRONG

bsongis:
I don't know this part of the code. I see this:
if (axis>=0 && axis<=8) 
So, 9 channels?

bsongis:
Oh but ...
int jsmap[8];
At least there is a bug

bsongis:
The Joystick object can send signals:
-  axisValueChanged
-  buttonValueChanged
-  hatValueChanged

Only axisValueChanged is catched, I am afraid in your case one of the 2 other signals is needed.

bsongis:
Out of curiosity, what is a hat? Google translate doesn't help ;)

projectkk2glider:
http://en.wikipedia.org/wiki/Joystick#Hat_switch

davxlw:
@bsongis,
> Out of curiosity, what is a hat? Google translate doesn't help ;)

![aviatorlogitechjse-orig](https://cloud.githubusercontent.com/assets/7569433/5685801/4146c42e-983e-11e4-98c5-f24aeb2278eb.jpg)

jtaylor2:
I took a look at this today with the thought of trying to fix it.  Verified that it works correctly on self compiled next on OpenSUSE 13.2 Linux and fails as reported on Windows 7 with both downloaded 2.0.15 and self compiled next compiled with Cygwin.

Discovered that on Windows  SDL_JoystickNumAxes(joystick); in joystick.c was returning 6 axes instead of 8.

Looking at the source for SDL-1.2.15/src/joystick/win32/SDL_mmjoystick.c found this statement:

#define MAX_AXES 6 /* each joystick can have up to 6 axes */

[...]

PR #2182, opened 2015-04-03
Proper SDL library initialization for Audio and Joystick subsystems

projectkk2glider:
Now that we use two SDL subsystems (joystick and audio) the SDL initialization has to be corrected. 

This pull might also affect and/or fix https://github.com/opentx/opentx/issues/1732#issuecomment-87364197

If accepted I will prepare a PR for `next` branch.

### DIFF
--- a/companion/src/companion.cpp
+++ b/companion/src/companion.cpp
@@ -48,6 +48,9 @@
 #include <QSplashScreen>
 #include <QThread>
 #include <iostream>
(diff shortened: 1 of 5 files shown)


bsongis:
Thanks!

projectkk2glider:
Ported to next in https://github.com/opentx/opentx/compare/df2318a45d...eac8c0a535

Issue #2188, opened 2015-04-06
Add support for replacement LCD ERC12864-2


universam1:
Assume there is a general interest in a good replacement for the LCD of the 9X & 9XR - had a hard time to find a fitting substitue. I've no interest in advertising, just if someone needs it's almost same size 2,6", hardware compatible (backlight needs pin swap) with IMHO better quality. Search for ERC12864-2 in eBay or buydisplay.com

Unfortunatelly with current init its not usable, almost no contrast. After lots of tests the manufacturer explained that the init sequence https://github.com/opentx/opentx/blob/097befac8f73c9854534a3d378e8ba59f7ab9af3/radio/src/targets/stock/lcd_driver.cpp#L73-L104 needs some changes:

"We checked your initalizer,a few instructions need to be modified,See below please;
0x25, //Select int resistance ratio R2 R1 R0 =5_____Modify 0x25 to 0x27
0x22, // 24 SV5 SV4 SV3 SV2 SV1 SV0 = 0x18_____Modify 0x22 to 0x2D
g_eeGeneral.contrast = 0x22;______Modify 0x22 to 0x2D
Bias(0xA2, // Select LCD bias=0____Modify 0xA2 to 0xA3)."

With these changes it works excellent. Would love to get general support here!?

![2015-04-06 08 42 51b](https://cloud.githubusercontent.com/assets/10851289/7001512/a0a52dda-dc3a-11e4-91be-d332c43812cc.jpg)

bsongis:
Yes, with an option in Companion, would you provide the patch?

2015-04-06 8:57 GMT+02:00 universam1 <notifications@github.com>:

kilrah:
From your photo that would be only for 9XR, right? As far as I know the LCDs are not interchangeable between 9X and 9XR due to a different connector.

universam1:
Connector actually is the same, I've tried it on a 9X. Unfortunately I
don't have access to a 9X any more, so a schematic would be needed. Anyone?
Am 06.04.2015 10:00 schrieb "Andre Bernet" <notifications@github.com>:

kilrah:
I think the connector is the same but mirrored (i.e. the LCD would end up facing the board unless you twisted the flat cable, which you can't).

universam1:
Likely, how should the define be named?

2015-04-06 9:32 GMT+02:00 Bertrand Songis <notifications@github.com>:

universam1:
[...]

PR #2189, opened 2015-04-07
Add support for LCD ERC12864FSF

universam1:
common replacement LCD needs different init data

### DIFF
--- a/radio/src/targets/stock/lcd_driver.cpp
+++ b/radio/src/targets/stock/lcd_driver.cpp
@@ -77,12 +77,22 @@ const static pm_uchar lcdInitSequence[] PROGMEM =
    0xa1, //ADC = 1: reverse direction(SEG132->SEG1)
    0xA6, //REV = 0: non-reverse display
    0xA4, //EON = 0: normal display. non-entire
+#ifdef LCD_ERC12864FSF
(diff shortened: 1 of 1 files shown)


universam1:
Question to the manufacturer:
1. Would you be able to have this display available for a longer time period?
2. Please name any other display from you that has the same 30Pin connector and uses the same "different" type of init commands so we can support any of these.

"Here is our answer about the below questions;

1.This module ERC12864FSF-2 is the one of our popular product,It will be keep on sale at recent years,We will inform the customer in advance if we plan to upgrade this module,So please don't worry about it.

2.Our ERC12864FSF-4 could reach your requirement,It has same 30Pin connector and same programd as our ERC12864FSF-2,only difference is the voltage,Please refer to the datasheet for more details at http://www.ebay.com/itm/12864-128x64-GLCD-Graphic-LCD-Module-Display-SPI-Serial-ST7565P-3V-Black-White-/291022273441?pt=LH_DefaultDomain_0&hash=item43c2477ba1
"

universam1:
this commit resolves #2188

bsongis:
No option in the Makefile?

universam1:
Would you mind for help out on the makefile since I'm not confident with that

PR #2201, opened 2015-04-13
[companion][log viewer]fix performance decrease if selection (fixes #2199)

### LINKED ISSUE #2199: log viewer plotting time changes when there is some selection
If some data is selected, like a fly session, the plotting time is noticeable longer than when no data is selected and all data is used.

Didn't find the specific details but it's related to Qt "isSelected" implementation...

I will put a pull request soon with some changes to not use "isSelected". This way there is no change in performance anymore. (this code is already on next, on my last pull request #2191)

griebd:
Performance is now the same, with or without selection.
Removed some unused stuff too.

### DIFF
--- a/companion/src/logsdialog.cpp
+++ b/companion/src/logsdialog.cpp
@@ -1,7 +1,6 @@
 #include "logsdialog.h"
 #include "appdata.h"
 #include "ui_logsdialog.h"
-#include "qcustomplot.h"
 #include "helpers.h"
 #if defined WIN32 || !defined __GNUC__
 #include <windows.h>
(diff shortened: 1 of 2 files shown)


companion/src/logsdialog.cpp
```diff
@@ -128,12 +120,11 @@ void logsDialog::legendDoubleClick(QCPLegend *legend, QCPAbstractLegendItem *ite
 {
   // Rename a graph by double clicking on its legend item
 
-  Q_UNUSED(legend)
   if (item) // only react if item was clicked (user could have clicked on border padding of legend where there is no item, then item is 0)
   {
     QCPPlottableLegendItem *plItem = qobject_cast<QCPPlottableLegendItem*>(item);
     bool ok;
-    QString newName = QInputDialog::getText(this, "QCustomPlot example", "New graph name:", QLineEdit::Normal, plItem->plottable()->name(), &ok);
+    QString newName = QInputDialog::getText(this, "Graph Name Change", "New graph name:", QLineEdit::Normal, plItem->plottable()->name(), &ok);
```
bsongis: Missing tr(...)

griebd:
Quite true, probably in next too... Will add here and make a new pull request for next, ok?

bsongis:
Perfect, thanks!

PR #2209, opened 2015-04-18
EEPROM on Sky9x

bsongis:
I changed the EEPROM file system on Sky9x because the actual max size for a file is around 4k
With the telemetry refactoring I need around 5k.
Here are some random bits as an explanation: 
- we have 4kbits = 512kbytes available
- we have 64 files of 8kbytes each (2blocks) : 1 FAT + 1 General + 60 Models + 2 Temp
- only 128bytes are needed for the FAT, 1st time the FAT is written, the 1st block is ERASED, then 128 first bytes written ; second time, the 128 first bytes are overwritten with 0 (because we can only write 0 there if we don't want to ERASE the block each time), and the real FAT is written in bytes [128:256]
- when writing a file we write it in a Temp file, and at the end we update the FAT (as explained above) with the new references of the blocks used

I am not sure I am clear enough. This explanation could be written in the eeprom_arm.cpp file as a doxygen comment, feel free to reword it, I am not native english!


bsongis:
@MikeBland Would you have a look to this review please? Thanks in advance!

bsongis:
Interesting comment from Mike:
"Bertrand,
Generally looks OK, although I would handle the FAT slightly differently. I'd use the incrementing ID method to mark up the latest FAT, rather than write zeroes over them. It would also use half the number of writes to each byte. I would also probably include a checksum in the FAT.
It also means a previous FAT is still available, so , in the event of a write failing, you still have the last FAT to go back to."

MikeBland:
A 32-bit value for the ID, if it gets incremented once per second on average, will take 136 years before it gets to 0xFFFFFFFF, so there is no need to consider it wrapping round.

bsongis:
@MikeBland Thanks for your review, I implemented your idea, it works perfectly.

Issue #2163, opened 2015-03-23
Show Now button (Request) for Companion Log Viewer


HuntTheShunt:
The next time someone is working on the Companion Log Viewer could a button be placed by the Reset button (or anywhere the developer thinks appropriate) the would activate the current selected available fields.

Currently as soon as you select an available field the program locks out the user because it is working in the background processing the users selection. 
The problem is the user has not completed his multi-selection yet and cannot do so because of the background processing.
The above spoils an excellent open-tx tool.

If the default status of the graph was to show nothing until the user pressed a button marked/entitled "Show" or a long one with "Display selected fields" .
 Yes "Plot" would be better.

The above would make great tool easier to use.

Note: Apologies if this request has been placed incorrectly.

kilrah:
Hmm IMO the root issue is rather the poor performance.
Anytime I open a log and it's slow to respond to a selection I just close it and use Excel instead.

Allowing a multi-selection then clicking a button that then has you wait for 5 minutes is not really a solution. If it's too slow to process my multi-selection it's too slow to use at all.

projectkk2glider:
I agree about the slowness, it has to be fixed. I had a plan to look at this slowness issue, but haven't found the time yet.

griebd:
I am looking at this code, to fix #1863..., and the code is already running a bit faster, but I need a bigger test file (mines are short, my heli only flies like 5-8 min... ;). And I only make logs of it (my planes have no log) because the LiPo cells, so #1863 is a fix for me too actually.
Anyhow, I would tend to agree that we should make our selection and then click "plot"....
And I think we should limit the selection to what we have available, 2 y axis only. This way you wouldn't be able to select 'cells', 'temps' and 'rpm' at the same time, their ranges are too far different. But you could select all 12 'cells' at once and include all 'temps' in the same graph...

HuntTheShunt:
Let me have your email via PM and I will send you the longest I have.
Not sure how that is done in here. You can contact me by PM on rcgroups.com to TrueBuld

TrueBuild:
griebd ,
I can send you some log files if you contact me via email.

jtaylor2:
>And I think we should limit the selection to what we have available, 2 y axis only. This way you wouldn't be able to select 'cells', 'temps' and 'rpm' at the same time, their ranges are too far different. But you could select all 12 'cells' at once and include all 'temps' in the same graph, one y axis to 'cells' and the other to 'temps'...

[...]

PR #2215, opened 2015-04-21
log viewer performance #2163

### LINKED ISSUE #2204: plot is wrong when hand picking data in log viewer
If you select the rows one by one the points are ordered by the order you select them, not by time as expected.

One line fix on the way... :wink:

griebd:
Fixes #2204 and #2199 on next! (already fixed on master)

Try to solve #2163 on next. (easily to port back to master if accepted)

Updated QCustomPlot to latest version 1.3.0 (December 27, 2014). And put it in a sub-directory.

Follow suggestions from QCustomPlot to improve performance:
- pen width <= 1.0 (was 1.5)
- setNoAntialiasingOnDrag to true (as name suggests, disable anti-aliasing on drag only)

Function moveLegend() is commented out as it's a bit more work to migrate to new QCustomPlot and I couldn't see it working in the old version, so there is no lose of functionality... (correct me if I'm wrong)

When opening a new file:
- try to improve performance while loading the data on ui->logTable. There is some improvement, but unnoticeable. So far the only solution I did find is to change from QTableWidget to QTableView and that means a bigger change. And that is why I decided to put all this in next and them start working on this last change.
- improved performance on columns resize by changing the way its done.
- changed the parsing of log files completely. NEEDS TEST as I have only a couple of test files!!!! (but reduces the loading time by 50%)

Changed the way fly sessions are handled and now you get no loading time when you change the fly session (but if you have some fields selected the plotting time still exists).

Many changes, I know. So, review and tell me what needs to be fixed or changed or whatever.... :wink:
At least for me, it's a lot faster to use "log viewer" now... I hope everyone else feels the same!!!!


companion/src/logsdialog.cpp
```diff
@@ -399,24 +407,25 @@ void logsDialog::removeAllGraphs()
...
-    bool ok;
-    int dataInt = contextAction->data().toInt(&ok);
-    if (ok)
-    {
-      ui->customPlot->legend->setPositionStyle((QCPLegend::PositionStyle)dataInt);
-      ui->customPlot->replot();
-    }
-  }
-}
+// void logsDialog::moveLegend()
```
bsongis: Please remove the code if needed, unless it's a TODO and you keep the code for later?

companion/src/qcustomplot/GPL.txt
```diff
@@ -0,0 +1,674 @@
+                    GNU GENERAL PUBLIC LICENSE
```
bsongis: Was it a thirdparty? If yes it should be in its own directory (as I did for FatFS / Lua / CoOS / etc.), no?

griebd: yes, it is a third party lib, and it's now in its own directory. (the last version wasn't)

companion/src/logsdialog.cpp
griebd: well, thing is I don't know if it's a TODO... will remove then, if someone finds the need I implement it later!

bsongis:
Thanks a lot for this one!

PR #2216, opened 2015-04-23
Fixes #2195

### LINKED ISSUE #2195: "Battery meter range" setting is missing in Companion
The "General Settings" of Companion include the "Battery Warning", but no "Battery meter range" setting.

griebd:
Please review and tell me if I didn't understood something.... Supposedly fixes #2195!
Implemented missing Battery Meter Range for Taranis in companion General Settings.
Tested only in simulator!!!
From what I did understood I just needed to add the ui and some attributions to make it work, all rest was already done.

Sorry, for the change in indentation in one of the file, it was double indented and was auto-magically corrected... if that is a problem, just tell me and I will redo the changes forbidding the auto-magically editor to do its work.

### DIFF
--- a/companion/src/generaledit.cpp
+++ b/companion/src/generaledit.cpp
@@ -13,53 +13,43 @@
 #define BEEP_VAL_SHIFT   3
 
 GeneralEdit::GeneralEdit(RadioData &radioData, QWidget *parent) :
-    QDialog(parent),
-    ui(new Ui::GeneralEdit),
-    radioData(radioData),
-    g_eeGeneral(radioData.generalSettings)
+QDialog(parent),
+ui(new Ui::GeneralEdit),
+radioData(radioData),
+g_eeGeneral(radioData.generalSettings)
 {
-    ui->setupUi(this);
-    this->setWindowIcon(CompanionIcon("open.png"));
+  ui->setupUi(this);
+  this->setWindowIcon(CompanionIcon("open.png"));
 
-    QString firmware_id = g.profile[g.id()].fwType();
-    ui->tabWidget->setCurrentIndex( g.generalEditTab() );
-    QString name=g.profile[g.id()].name();
-    if (name.isEmpty()) {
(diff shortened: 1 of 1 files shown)


companion/src/generaledit.cpp
```diff
@@ -13,53 +13,43 @@
 #define BEEP_VAL_SHIFT   3
 
 GeneralEdit::GeneralEdit(RadioData &radioData, QWidget *parent) :
-    QDialog(parent),
```
bsongis: Please next time, try not to reindent in the same commit than modifications, it's easier to understand what changed!

romoloman:
I don't like to much to have battery range under calibration as it something more related to display than to hardware calibration itself, i would like it much more in "Setup" tab

companion/src/generaledit.cpp
griebd: yep, thought about that later on... really sorry! learning to develop in a team....

griebd:
anyway,

@romoloman, I was in doubt about that one... but it's not so much related at display as it's to hardware! You need this if you use a different battery type, with different voltage, in your Taranis. In my mind I think you need to calibrate your software/firmware to know your battery voltage range.

In the end,

@bsongis, do I redo this with a separated commit because of the re-indent?

all others, do you prefer this config in setup or calibration tab?

anything else need here?

griebd:
ok, re-indent separated!

romoloman:
In the radio is not in the calibration menu.... I prefer to keep stuff in the same section as in the radio...

griebd:
@romoloman, with that I do agree completely!!!!!! making the change... :wink:

romoloman:
You can merge the ui changes from by branch....

bsongis:
I think I can delete this pull request, right? Thanks for the work anyhow, and yes the reindent could be done in a separate pull request, it's easier for reviewers. Next time ...

PR #2218, opened 2015-04-23
Romolo/companion batt meter range 2195

romoloman:
Like Griebd one but range in Setup tab


bsongis:
This will be needed in 'next' I think!

PR #2219, opened 2015-04-23
Romolo/toolfixes20

romoloman:
Some fixes for open bugs in comparedialog and in printdialog, and enhancements to generalsetting and curvedialog


bsongis:
Thanks!

PR #2210, opened 2015-04-18
TELEMETRY_MOD=14051_SWAPPED

xmxwx:
TELEMETRY_MOD=14051 (by gerardv @ forum.turnigy9xr.com) makes USART0 available
by reallocating AIL and THR onto pins 14 and 15 of the 14051 multiplexer.

TELEMETRY_MOD=14051_SWAPPED is based on TELEMETRY_MOD=14051, but uses USART1
instead of USART0, freeing USART0 for other purposes (debugging, interfacing
via AVRISP port). USART1 is freed up by reallocating LVert Trim Dn/Up onto
14051's 12 and 1 pins respectively.

Additionally, this commit cleans a few things up: moves code related to this
telemetry mod to board_stock and replaces a few magic numbers with enums.

PR #2220, opened 2015-04-23
TELEMETRY_MOD=14051_SWAPPED

xmxwx:
TELEMETRY_MOD=14051 (by gerardv @ forum.turnigy9xr.com) makes USART0 available
by reallocating AIL and THR onto pins 14 and 15 of the 14051 multiplexer.

TELEMETRY_MOD=14051_SWAPPED is based on TELEMETRY_MOD=14051, but uses USART1
instead of USART0, freeing USART0 for other purposes (debugging, interfacing
via AVRISP port). USART1 is freed up by reallocating LVert Trim Dn/Up onto
14051's 12 and 1 pins respectively.

Additionally, this commit cleans a few things up:
 * moves AVR ADC-related code to adc_driver.cpp,
 * provides #defines to have a convenient way to access either USART0 or USART1
   registers depending on static (build-time) configuration
 * moves code related to this telemetry mod to board_stock
 * replaces a few magic numbers with enums.

Thanks to @bsongis for review and suggestions!

### DIFF
--- a/radio/src/Makefile
+++ b/radio/src/Makefile
@@ -415,7 +415,7 @@ ifeq ($(PCB), $(filter $(PCB), STD 9X 9XR))
   INCDIRS += targets/stock
   CPPDEFS = -DF_CPU=$(F_CPU)UL -DPCBSTD -DCPUM64 -DEEPROM_VARIANT=$(shell echo ${EEPROM_VARIANT} | bc)
   BOARDSRC = main_avr.cpp targets/stock/board_stock.cpp
-  EXTRABOARDSRC = targets/common_avr/telemetry_driver.cpp
+  EXTRABOARDSRC = targets/common_avr/adc_driver.cpp targets/common_avr/telemetry_driver.cpp
   ifeq ($(GUI), YES)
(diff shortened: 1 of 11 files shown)


xmxwx:
Squash, fix and rebase of https://github.com/opentx/opentx/pull/2210 onto 'next' branch.

bsongis:
Excellent work, thanks! I would just like to have a fix for #2221 before merging!

xmxwx:
There's a tiny issue with this mod: if the splash screen is disabled, switch warning test on boot-up gives a false negative when only THR and/or AIL is ON.

bsongis:
Would you please take care of this? I think it will be annoying and somebody will open an issue. Thanks!

xmxwx:
> Would you please take care of this? I think it will be annoying and somebody will open an issue. Thanks!
I took care of this in 31820a0520af39060161641d04300922fd352a98 and pushed it directly to the repo. Is it OK not to open a pull request?

BTW. This feels like a workaround. A clean solution would require a small refactoring of ADC support.

PR #2228, opened 2015-04-25
Fixes #2221 (LCD contents shifted upwards by 6 pixels on 9XR).

### LINKED ISSUE #2221: LCD contents shifted upwards by 6 pixels (with wraparound) on 9XR
'next' branch, 41f78440db2230a20bb980c54640c406a65e36cf

xmxwx:


bsongis:
Thanks! I wonder if it's not better to have a different lcd_driver.cpp for this board. It depends on the #ifdef count, if they are too many, the code is rather difficult to read ...

PR #2229, opened 2015-04-25
[companion][logsdialog]Backport fix for #2163 from next to master.

griebd:
Improve performance of graphs plotting on master, same code from next. (QCustomPlot updated as well)


projectkk2glider:
This code is extremelly fast! Good work @griebd 

But I noticed a bug, using large log file http://wikisend.com/download/803076/ShadowF5X-2015-04-12.csv containing three sessions:
* when the whole file is displayed, all is OK
* first session is displayed OK
* second session has wrong time stamps in the first part, then has a blank range somewhere in the middle, something is wrong here.

projectkk2glider:
screenshot

Example of above bug: 
* the part with the constant altitude from about 10:32:00 is wrong
* the time at beginning should be 10:37:53

griebd:
Thanks @projectkk2glider! The problem was that the X axis values weren't selected "from the selection" in fly sessions.

projectkk2glider:
All good now, merging. Thanks!

PR #2230, opened 2015-04-25
Fix bug found by @projectkk2glider

griebd:
When back-porting logsdialog improvements to master @projectkk2glider found a bug, this is the fix on next!


bsongis:
I will let @projectkk2glider review it then

projectkk2glider:
Bug was found and described in #2229

PR #2235, opened 2015-04-28
[companion][logsdialog]Small glitch when loading a second file.

griebd:
Fixes a mostly unnoticeable glitch when loading a second file in Log Viewer.
sessions_CB was cleared too early. If you had something plotted, that means a new plot was done.


projectkk2glider:
:+1:

PR #2237, opened 2015-04-28
Fixes #2154: USB mass storage problem with cards bigger than 4GB

### LINKED ISSUE #2154: File system corruption with fullish large SD card
Summary
-------

When a known-good largish SD card (16GB) more than half full is plugged into the Taranis (X9D or X9DP running OTX 2.0.13+), and the Taranis is plugged into a PC, (from bootloader mode), CHKDSK reports many errors on that removable drive (and attempts to read certain files from the drive fail). But take this same SD card and plug it back into the PC using a standard SD card adapter (not the Taranis), then CHKDSK reports zero errors (and all files can be accessed without errors). Note that because of this bug, the files used to fill up the disk must be copied using the external card adapter - attempts to do so when connected via the Taranis won't work.

Background
----------

I'm getting ready to release "TaraniJuke", which is a collection of some LUA scripts which implement a rather useful background music player. It works just fine with a 2GB card, but I started noticing some problems when I moved up to a larger SD card (16GB). Once the SD card starts filling up, the Taranis starts experiencing problems accessing the SD card. These problems accessing the card only appear when the SD card is connected to the computer using the Taranis - these problems do not occur when the SD card is connected using a standard SD card adapter. It is unknown if this error manifests under normal operating conditions when the Taranis is not plugged into a computer.

How to reproduce
----------------

Download the following ZIP file:

ftp://ftp.xmission.com/pub/users/j/jmr/outgoing/TARANI_FS_BUG_TEST.zip

MD5: D3D4884F0CA7A1DFF1EC68EA8C42D7C3
SHA-1: E9BE01D7EA6228AE6D8173B8086E8CF988744F1B

This zip archive contains 3 files:
DUPWAV.BAT - Batch file to make 1000 copies of WAV file in the SOUNDS\en folder.
README.TXT - Description of how to reproduce the error.
TEST.WAV - A 32-kHz, mono Taranis compatible WAV file that is roughly 10MB in size.

Here are the contents of the README.TXT file from the archive:

README.TXT
----------

How to reproduce the Taranis FatFs bug:

1. Place a 16GB MicroSD card in an external USB adapter and plug it into a Windows PC. It 
will show up as a removable drive "X:" in Windows Explorer (however, the drive letter on 
your system will likely be different from X). Do NOT use the Taranis as the USB adapter at 
this point.
2. Open Windows Explorer and right click on the removable drive X:. Format the SD card as 
FAT32 using the default settings (cluster size = 8KB). Do NOT select the "quick format" 
option.
3. Copy the standard OpenTX files and directories to the SD card. It should look something 
like the following:
    Drive X: Directory tree:  
        +---BMP  
        +---EEPROMS  
        +---FIRMWARES  
        +---LOGS  
        +---MODELS  
        +---SCRIPTS  
        |   +---MODEL01  
        |   +---FUNCTIONS  
        |   +---MIXES  
        |   \---WIZARD  
        \---SOUNDS  
            \---en  
                \---system  
    (The only mandatory folder to be copied is: SOUNDS\en)  
4. Copy DUPWAV.BAT and TEST.WAV (from this zip archive) to the root directory of the SD 
card, i.e. "X:\".
5. From Windows Explorer, run X:\DUPWAV.BAT by double clicking on it. This batch file simply 
makes 1000 copies of the TEST.WAV file (which is 12MB) to the SOUNDS\en folder giving each 
copy of the file a unique name: _TST0001.WAV, _TST0002.WAV, ... _TST1000.WAV. Note that this 
process takes a LONG TIME ~1 hour. (Go take a break!)
6. Once the batch file finishes making all the wav file copies, verify that the file system 
is A-Ok; Double click on X:\SOUNDS\en\_TST1000.WAV and verify that it plays OK. Open a DOS 
Command Prompt Window and type: "CHKDSK X:" (Where "X" is the drive letter for the SD card.) 
Verify that CHKDSK reports no errors.
7. Dis-mount the SD card by clicking on the "Safely Remove HArdware" icon in the system tray 
(for WIn XP -  On Win7 and higher, right click the removable drive X; in Explorer and 
select: "Eject"). Remove the MicroSD card and insert it into the Taranis SD card slot.
9. Connect the Taranis to the PC by using the 3-finger-salute bootloader method. The SD card 
in the Taranis shows up as removable drive "Y:"  (But once again, the drive letter on your 
system will likely be different from Y:). At this point, things should start going wonky.
10. Repeat the tests described in step 6 to test the file system integrity. At this point, 
Windows should experience a variety of problems accessing the drive - it will be unable to 
play the later WAV files and will report many errors when running CHKDSK from the command 
line. Do NOT modify any of the files on the SD drive.
11. Dismount the Taranis, remove the SD card and remount it using the external adapter. The 
card should once again be A-Ok.

projectkk2glider:
This attempts to fix large (>4GB) SD card problem without resorting to 64 bit variables and arithmetics.

I have tested this with my radio and 8GB SD card and looks like it is working. I insist that somebody else also tests this, before we merge into `master`!

Closes #2154

### DIFF
--- a/radio/src/targets/taranis/STM32_USB-Host-Device_Lib_V2.1.0/Libraries/STM32_USB_Device_Library/Class/msc/src/usbd_msc_scsi.c
+++ b/radio/src/targets/taranis/STM32_USB-Host-Device_Lib_V2.1.0/Libraries/STM32_USB_Device_Library/Class/msc/src/usbd_msc_scsi.c
@@ -511,7 +511,6 @@ static int8_t SCSI_Read10(uint8_t lun , uint8_t *params)
     }
     
     MSC_BOT_State = BOT_DATA_IN;
-    SCSI_blk_addr *= SCSI_blk_size;
     SCSI_blk_len  *= SCSI_blk_size;
     
     /* cases 4,5 : Hi <> Dn */
@@ -583,7 +582,6 @@ static int8_t SCSI_Write10 (uint8_t lun , uint8_t *params)
       return -1; /* error */      
     }
     
-    SCSI_blk_addr *= SCSI_blk_size;
     SCSI_blk_len  *= SCSI_blk_size;
     
     /* cases 3,11,13 : Hn,Ho <> D0 */
@@ -672,7 +670,7 @@ static int8_t SCSI_ProcessRead (uint8_t lun)
   
   if( USBD_STORAGE_fops->Read(lun ,
                               MSC_BOT_Data, 
-                              SCSI_blk_addr / SCSI_blk_size, 
+                              SCSI_blk_addr, 
                               len / SCSI_blk_size) < 0)
   {
     
@@ -687,7 +685,7 @@ static int8_t SCSI_ProcessRead (uint8_t lun)
              len);
   
   
-  SCSI_blk_addr   += len; 
+  SCSI_blk_addr   += len / SCSI_blk_size; 
   SCSI_blk_len    -= len;  
   
   /* case 6 : Hi = Di */
@@ -715,7 +713,7 @@ static int8_t SCSI_ProcessWrite (uint8_t lun)
   
   if(USBD_STORAGE_fops->Write(lun ,
                               MSC_BOT_Data, 
-                              SCSI_blk_addr / SCSI_blk_size, 
+                              SCSI_blk_addr, 
                               len / SCSI_blk_size) < 0)
   {
     SCSI_SenseCode(lun, HARDWARE_ERROR, WRITE_FAULT);     
@@ -723,7 +721,7 @@ static int8_t SCSI_ProcessWrite (uint8_t lun)
   }
   
   
-  SCSI_blk_addr  += len; 
+  SCSI_blk_addr  += len / SCSI_blk_size; 
   SCSI_blk_len   -= len; 
   
   /* case 12 : Ho = Do */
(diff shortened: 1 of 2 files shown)


bsongis:
Thanks a lot for this one!

projectkk2glider:
Did you test it?

radio/src/targets/taranis/STM32_USB-Host-Device_Lib_V2.1.0/Libraries/STM32_USB_Device_Library/Class/msc/src/usbd_msc_scsi.c
```diff
@@ -687,7 +685,7 @@ static int8_t SCSI_ProcessRead (uint8_t lun)
              len);
   
   
-  SCSI_blk_addr   += len; 
+  SCSI_blk_addr   += len / SCSI_blk_size;
```
bsongis: Are we sure the rest of the division will be 0 ?

projectkk2glider: Pretty sure because:
* len is some number multiplied by SCSI_blk_size 
* MSC_MEDIA_PACKET is multiple of SCSI_blk_size (which is equal to BLOCKSIZE which is 512 in our case)

bsongis:
I didn't test it yet. But I would like it is included ASAP so that we have feedback tomorrow. Just in time for a release, I will delay it again by one day (tomorrow if everything is good) to give a little bit time for the tests.

Issue #2080, opened 2015-02-16
Enhancement: Set "Backup Folder" as part of the Radio Profile instead of Application Settings


CoyotteDundee:
Hi All,
I wonder if the Companion's "Backup Folder" parameter could be moved from the Application Settings to the Radio Profile, allowing having a backup folder per Radio Profile.
This could allow people having multiple models to not mix backups.

Best regards,

Coyotte

LapinFou:
:+1: Very good idea !

LapinFou:
Can't wait to try the first stable version of 2.1.
:)

romoloman:
I would prefer to have both with profile backup folder overriding the default application one if set.
If this is accepted by all members I will implement it

CoyotteDundee:
Hello,
This 'll allow usual operation mode, as also a specific mode if needed by the user. 
The best of both worlds ;-)
Thanks for the proposal.

Coyotte

bsongis:
Yes of course, thanks @romoloman for this one, I will have more time for other issues!

romoloman:
Done in profilebackupfolder branch, give it a test.

romoloman:
Isn't this issue closed ?
Feedbacks are welcome

bsongis:
You just forgot to write "Fixes #2080" in your commit :)
Thanks for your this one!

PR #2243, opened 2015-05-01
Implement profile specific settings for backup folder.

romoloman:
Profile backup path if set overrides general one, in any case the backup will be performed if either profile specific setting or general settings ones are enabled.
Disabling backup in profile will not disable backups if generic settings are enabled

### DIFF
--- a/companion/src/apppreferencesdialog.cpp
+++ b/companion/src/apppreferencesdialog.cpp
@@ -18,10 +18,6 @@ AppPreferencesDialog::AppPreferencesDialog(QWidget *parent) :
   ui->setupUi(this);
   updateLock=false;
   setWindowIcon(CompanionIcon("apppreferences.png"));
-  ui->backupPathClear->setIcon(CompanionIcon("clear.png"));
-  ui->pBackupPathClear->setIcon(CompanionIcon("clear.png"));
-  ui->libraryPathClear->setIcon(CompanionIcon("clear.png"));
-  ui->sdPathClear->setIcon(CompanionIcon("clear.png"));
 
   initSettings();
   connect(ui->downloadVerCB, SIGNAL(currentIndexChanged(int)), this, SLOT(baseFirmwareChanged()));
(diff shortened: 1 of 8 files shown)


companion/src/apppreferencesdialog.cpp
```diff
@@ -164,6 +166,19 @@ void AppPreferencesDialog::initSettings()
   ui->stickmodeCB->setCurrentIndex(g.profile[g.id()].defaultMode());
   ui->renameFirmware->setChecked(g.profile[g.id()].renameFwFiles());
   ui->sdPath->setText(g.profile[g.id()].sdPath());
+  ui->profilebackupPath->setText(g.profile[g.id()].pBackupDir());
+  if (!g.profile[g.id()].pBackupDir().isEmpty()) {
+    if (QDir(g.profile[g.id()].pBackupDir()).exists()) {
+      ui->profilebackupPath->setText(g.backupDir());
```
projectkk2glider: Here you overwrite profile backup folder with the application global backup folder. Why?

companion/src/mainwindow.cpp
```diff
@@ -1336,16 +1352,36 @@ void MainWindow::writeFlash(QString fileToFlash)
...
+        else {
+          backupEnable = false;
         }
       }
+      if ((g.enableBackup() || g.profile[g.id()].penableBackup()) && !(backupEnable) && !IS_TARANIS(GetEepromInterface()->getBoard())) {
+        if (!QDir(backupPath).exists()) {
+          QMessageBox::warning(this, tr("Backup is impossible"), tr("The backup dir set in preferences does not exist"));
+        }      
+      } 
       else {
```
projectkk2glider: Can we put this section and one in MainWindow::writeBackup() in a new function? They look very similar.

companion/src/mainwindow.cpp
```diff
@@ -1379,13 +1415,35 @@ void MainWindow::writeFlash(QString fileToFlash)
...
+        }
+        else {
+          backupEnable = false;
+        }
+      }
+      if ((g.enableBackup() || g.profile[g.id()].penableBackup()) && !(backupEnable)) {
+        if (!QDir(backupPath).exists()) {
+          QMessageBox::warning(this, tr("Backup is impossible"), tr("The backup dir set in preferences does not exist"));
+        }      
       }
```
projectkk2glider: Same as previous comment, we have three same sections of code.

companion/src/mdichild.cpp
```diff
@@ -516,18 +516,35 @@ QString MdiChild::strippedName(const QString &fullFileName)
...
+    }
+    else {
+      backupEnable = false;
+    }
+  }
+  if ((g.enableBackup() || g.profile[g.id()].penableBackup()) && !(backupEnable)) {
+    if (!QDir(backupPath).exists()) {
+      QMessageBox::warning(this, tr("Backup is impossible"), tr("The backup dir set in preferences does not exist"));
+    }      
   }
```
projectkk2glider: Four :)

projectkk2glider:
Just to cross-reference: this pull fixes #2080 on master branch

companion/src/apppreferencesdialog.cpp
romoloman: This is a mistake, thanks for reviewing... the code was already duplicated across files, i only added some more.
Truly we can set a function to get the proper backup path.

bsongis:
@projectkk2glider May I let you press the button?

romoloman:
I think that as in 2.1 things changes a lot makes little sense to move out of mainwindows.
As is mostly called there.

PR #2252, opened 2015-05-03
Italian translations updated and obsolete removed.

romoloman:
Missing strings, they are also missing in other languages.
I also manually launched lupdate with -no-obsolete to clean up the file from obsolete no longer used strings.

### DIFF
--- a/companion/src/translations/companion_it.ts
+++ b/companion/src/translations/companion_it.ts
@@ -65,49 +65,46 @@
         <translation>Modifica Impostazioni</translation>
     </message>
     <message>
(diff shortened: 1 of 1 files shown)


griebd:
@bsongs, about translations, is it o.k. to include a new (almost incomplete) translation? I'm asking, because maybe, if the file is available, someone else (that knows no programming) decides to do the full translation. The language in question is Brazilian Portuguese, which is quite different from Portugal one.

bsongis:
Yes please open a pull request, only for 'next' branch!

PR #2264, opened 2015-05-05
Fixes #2186

### LINKED ISSUE #2186: Feature request: Allow different receiver numbers when using two modules.
When using the internal and external modules in Opentx on a Taranis, the receiver numbers must be the same for both receivers.  (Setting one automatically changes the other in the TX. Note that Companion allows you to select different numbers but they revert to the same number in the TX.)

My request is to allow the use of different receiver numbers when using both the internal and external modules.  This would allow receivers that use "Model Match" like X series Frsky and Spectrum compatible DSM2 and DSMX to be used simultaneously with different receiver numbers.

bsongis:
Also provides a lua function to retrieve module settings

### DIFF
--- a/radio/src/lua_api.cpp
+++ b/radio/src/lua_api.cpp
@@ -584,7 +584,6 @@ static int luaModelGetInfo(lua_State *L)
   lua_newtable(L);
   lua_pushtablezstring(L, "name", g_model.header.name);
   lua_pushtablenzstring(L, "bitmap", g_model.header.bitmap);
-  lua_pushtableinteger(L, "id", g_model.header.modelId);
   return 1;
 }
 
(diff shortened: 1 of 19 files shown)


bsongis:
For comments / review. I will squash those commits into one when everything will be ok.

radio/src/lua_api.cpp
```diff
@@ -603,15 +602,56 @@ static int luaModelSetInfo(lua_State *L)
...
+    ModuleData & module = g_model.moduleData[idx];
+    luaL_checktype(L, -1, LUA_TTABLE);
+    for (lua_pushnil(L); lua_next(L, -2); lua_pop(L, 1)) {
+      luaL_checktype(L, -2, LUA_TSTRING); // key is string
+      const char * key = luaL_checkstring(L, -2);
+      if (!strcmp(key, "rfProtocol")) {
+        module.rfProtocol = luaL_checkinteger(L, -1);
+      }
+      else if (!strcmp(key, "modelId")) {
+        g_model.header.modelId[idx] = luaL_checkinteger(L, -1);
```
projectkk2glider: Maybe you forgot:
```c
modelHeaders[g_eeGeneral.currModel].modelId = g_model.header.modelId[idx];
```

radio/src/opentx.cpp
```diff
@@ -414,8 +422,9 @@ void modelDefault(uint8_t id)
 #endif
 
 #if defined(PXX) && defined(CPUARM)
-  modelHeaders[id].modelId = g_model.header.modelId = id+1;
-  checkModelIdUnique(id);
+  modelHeaders[id].modelId[1] = g_model.header.modelId[1] = 0;
+  modelHeaders[id].modelId[0] = g_model.header.modelId[0] = id+1;
```
projectkk2glider: Is the size of  modelHeaders[id].modelId array always 2 here? Or should you use NUM_MODULES?

radio/src/lua_api.cpp
bsongis: Good catch!

radio/src/opentx.cpp
bsongis: Yes it will be better to use NUM_MODULES

bsongis:
Following discussion with @kilrah I changed my mind and initialize all modules with the same index (not 0 on second module)

kilrah:
Tested and good!

PR #2273, opened 2015-05-07
Bsongis/issue 2183 failsafe settings on individual channels

bsongis:
For review. The companion part is needed. I will squash all commits into one at the end


bsongis:
It's not so nice, I will rely on @mhotar for a nice idea :)

bsongis:
[Enter] switch to edition mode
then setup the value with the stick / mix / etc.
[Enter] save the value
or [Exit] back to the previous value

When not in edition mode:
[Enter Long] Hold
[Enter Long] No pulses
[Enter Long] back to the channel value
in a loop ...

kilrah:
Tested and working!

mhotar:
@bsongis, Nice job as usual !
I would only center these texts if possible.
snapshot_06
Please also consider to add manual setup mode using UP and DOWN buttons directly instead of channel value. Simply when you enter the edit mode, values will be taken from the channel, when you press UP or DOWN (in edit mode active), values will be changeable only by these buttons and the CHxx will change to [see the picture]

kilrah:
I don't like the dots, too heavy ;) Not sure about manual edit, is that really useful?

projectkk2glider:
I've glanced over the code changes. Nothing bad spotted.

bsongis:
@mhotar Then I propose [Enter Long] in edit mode takes the channel value, while otherwise [Up] and [Down] allow the user to manually modify the value.

kilrah:
Yes I prefer that too!

bsongis:
Here it is, comments welcome!

kilrah:
Great! :+1:

mhotar:
Yes, clever and nice!

bsongis:
Ok I will squash everything in one commit tomorrow and let you merge into 'next'

PR #2274, opened 2015-05-08
Bsongis/issue 2266 set failsafe custom function

bsongis:
For review. Still needed the Companion part. I will squash all commits into one at the end of the review.


bsongis:
Needs the other review to be accepted and closed before this one.

PR #2282, opened 2015-05-09
Romolo/simulator s3

### LINKED ISSUE #1917: S3 Pot management
Please if possible implement in Companion the possibility to manage S3 pot.
Now only S1 and S2 are manageable

TIA
Mirco

romoloman:
S3 and multipos pots in Taranis simu
fixes #1917


companion/src/simulation/simulatordialog.h
```diff
@@ -27,8 +27,14 @@ namespace Ui {
...
-#define SIMULATOR_FLAGS_NOTX            1
-#define SIMULATOR_FLAGS_STICK_MODE_LEFT 2
+#define SIMULATOR_FLAGS_NOTX              1
+#define SIMULATOR_FLAGS_STICK_MODE_LEFT   2
+#define SIMULATOR_FLAGS_S3                4
+#define SIMULATOR_FLAGS_S4                8 //reserved for the future
+#define SIMULATOR_FLAGS_S1_MULTI         16
+#define SIMULATOR_FLAGS_S2_MULTI         32
+#define SIMULATOR_FLAGS_S3_MULTI         64
+#define SIMULATOR_FLAGS_S4_MULTI        128 //reserved for the future
```
bsongis: Hex values for this!

companion/src/simulation/simulatordialog.cpp
```diff
@@ -170,6 +170,27 @@ SimulatorDialogTaranis::SimulatorDialogTaranis(QWidget * parent, SimulatorInterf
...
+    ui->dialP_2->setPageStep(2048/5);
+  }
+  if (!(flags & SIMULATOR_FLAGS_S3)) {
+    ui->dialP_5->hide();
+    ui->dialP_5_lbl->hide();
+  } else {
+    if (flags & SIMULATOR_FLAGS_S3_MULTI) {
+      ui->dialP_5->setValue(-1024);
+      ui->dialP_5->setSingleStep(2048/5);
+      ui->dialP_5->setPageStep(2048/5);
```
bsongis: Isn't it possible to have a for loop here?

kilrah:
If we change that it would ideally be good to make it generic (I started scratching the surface some time ago in my kilrah/simucontrols branch). With the X9E we will have 4 pots as standard, and there can be up to 4 more based on user config in the "Hardware" page. Same for switches as on the X9E, there can be anything between 0 and 17 so the simu should also load them based on the settings in the hardware tab. 
In my branch I started something but got stuck as to how to access the general settings early enough to read the hardware settings at the time the dialog is created, they weren't loaded at that point.

bsongis:
The pots could be created at startup, but hidden later!

kilrah:
Oh yes of course... damn

romoloman:
To be perfect we should non only hide pots but also we should name switches properly (if user renamed),  also we would need to access simulator eeprom settings at runtime (so if people modify settings from the simulator we modify the interface).

I'm also thinking about a new GUI for X9E simulator resembling the real radio. 
Creating POTS at runtime in my opinion make difficult to control the exact geometries of the gui, so I prefer to design in QTDesigner and then hide at runtime.
Thinking about a new interface for X9E what is done for SimulatorTaranis is imho sufficient (we can improve later).

kilrah:
Just tested and fixed a couple of little errors.

2 things to note:
- Doesn't work at all in standalone simu (none of the Sx pots appear)
- Multipos always assumes 6 positions, when it can be anywhere between 2 and 6 depending on calibration data

companion/src/simulation/simulatordialog.cpp
```diff
@@ -169,7 +169,24 @@ SimulatorDialogTaranis::SimulatorDialogTaranis(QWidget * parent, SimulatorInterf
   //restore switches
   if (g.simuSW())
     restoreSwitches();
-
+  QDial * pots[] = { ui->dialP_1 , ui->dialP_2 , ui->dialP_5};
```
bsongis: I am adding a findChildren here

companion/src/simulation/simulatordialog.cpp
```diff
@@ -169,7 +169,24 @@ SimulatorDialogTaranis::SimulatorDialogTaranis(QWidget * parent, SimulatorInterf
   //restore switches
   if (g.simuSW())
     restoreSwitches();
-
+  QDial * pots[] = { ui->dialP_1 , ui->dialP_2 , ui->dialP_5};
+  QLabel * potslabel[] = { ui->dialP_1_lbl , ui->dialP_2_lbl , ui->dialP_5_lbl};
+  int potsflag[] = {SIMULATOR_FLAGS_S1, SIMULATOR_FLAGS_S2, SIMULATOR_FLAGS_S3};
+  int potsmflag[] = {SIMULATOR_FLAGS_S1_MULTI, SIMULATOR_FLAGS_S2_MULTI, SIMULATOR_FLAGS_S3_MULTI};
+
+  for (int i=0; i<3; i++) {
```
bsongis: Will be something pots.count()

bsongis:
Please don't merge it tonight. I have work in progress on it.

bsongis:
Merged in 'next' after squashing all commits into one

PR #2286, opened 2015-05-09
What about this?

bsongis:


bsongis:
I hope I didn't break everything with boyscout work ;)

romoloman:
Andrè: looking to the sort of comments I received on other branches, I see we are looking for perfection here. So I would also like a UI perfection and not only a code one :)

bsongis:
Agree about perfection on both sides :)

kilrah:
Yep. We had 2 branches, one with good code and one with good looks, as people don't always master both we just need to merge everything together ;)

bsongis:
@romoloman Thanks it's really nice now. Even one POT now renders fine, the pot is centered!

PR #2288, opened 2015-05-10
Bsongis/ccm memory on f4

bsongis:
Ready for merge, it saves 45k RAM!


radio/src/opentx.cpp
```diff
@@ -1453,7 +1453,7 @@ void flightReset()
 }
 
 #if defined(THRTRACE)
-uint8_t  s_traceBuf[MAXTRACE];
+uint8_t  s_traceBuf[MAXTRACE] _CCM;
```
bsongis: Obviously this one is not reset to 0

radio/src/telemetry/telemetry.cpp
```diff
@@ -1,6 +1,6 @@
 #include "../opentx.h"
 
-TelemetryItem telemetryItems[MAX_SENSORS];
+TelemetryItem telemetryItems[MAX_SENSORS] _CCM;
```
bsongis: I am afraid the constructor is not called

bsongis:
We need a way to initialize variables in CCM or reset them to 0 if not initialized, or call the constructors if any!

bsongis:
Also I would like that all variables come to CCM except those marked as _NOCCM

bsongis:
I will test this one as soon as I receive the X9E

bsongis:
The radio is here. I can tell you, it is really nice!

projectkk2glider:
> The radio is here. I can tell you, it is really nice!

Lucky you :grin:  ... now, for the punishment, you have to make sure that everything works.

bsongis:
First flight done in my hattic :)

PR #2289, opened 2015-05-10
avoid very long vertical scrolls

romoloman:
Calibration tab is incredibly long,
Trying to make it shorter.


bsongis:
Yes it's better now. I would have put PPM settings at the left though, no?

romoloman:
Battery and volt are always present for any radio so there is no problem of having an empty space on the left, but for me there is really no problem in swapping them if you prefer.
The important thing for me is to avoid infinite scrolling with data in a single column.

bsongis:
@projectkk2glider made a change to use QTableWidget. It will allow us to have the header always visible.

PR #2290, opened 2015-05-11
Fixes #2266

### LINKED ISSUE #2266: Feature Request-One click setting of Failsafe settings
I saw in "[Enhancement] Increase Failsafe security #2117" that you were discussing setting Failsafe options.  Setting the failsafes are obviously a really smart idea but my personal challenge with them is knowing what to set for each channel.  There is no way of knowing how much up elevator is enough to slow the decent or fly level without having the airplane dive or climb and stall, etc.  I was thinking a great way to address this would be to use the momentary switch in the same manner as the instant trim function to automatically set the failsafe settings for all channels instantly.  This way you could take off, climb to a safe altitude and configure the airplane in its landing mode (ie gear down, flaps down, proper amount of throttle and elevator to ensure a slight decent or level flight, etc).  Once the airplane is behaving as you want it to, flicking the momentary switch would then "save" those settings to the failsafe settings for that model.  That would be a very quick and accurate way of making sure all channels are correct and nothing is missed.  Thoughts?

bsongis:
SetFailsafe Custom Function


bsongis:
For review / comments.
Still missing the Companion part.

bsongis:
I modified my previous commit as I don't want a new EEPROM conversion - all my models are already converted to 2.1 ;)

bsongis:
I need tests for this one, I would like it's really included in 2.1.0 Thanks!

PR #2298, opened 2015-05-11
Fixes #2188

### LINKED ISSUE #2188: Add support for replacement LCD ERC12864-2

bsongis:
Just the Makefile option added

### DIFF
--- a/radio/src/Makefile
+++ b/radio/src/Makefile
@@ -57,6 +57,10 @@ PCB = 9X
 # PCB revision
 PCBREV =
 
+# LCD
+# Values: DEFAULT, ST7565P, ST7565R, ERC12864FSF
+LCD = DEFAULT
+
 # Enable JETI-Telemetry or FrSky Telemetry reception on UART0
 # For this option you need to modify your hardware!
 # More information at [insertURLhere]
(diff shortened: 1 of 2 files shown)


bsongis:
Overwritten by #2325

PR #2302, opened 2015-05-13
Early Brasilian Portuguese translation

griebd:
Well, first thing is: language letters are the same for Portugal and Brazil. The correct way to specify the Brazilian one is pt-BR, so I did it! Is it ok?!
Second thing is, I don't believe I will have the time and patience to do it, I just wish to make easier for someone else.... (of course I can help a non-programmer on doing it)
If both things are O.K., just merge and delete.


griebd:
forget it, just saw the compile error....

PR #2316, opened 2015-05-17
Fixes #2084: wrong constant used to limit maximum number of expos

### LINKED ISSUE #2084: Can't add new input with Lua model.insertInput()

projectkk2glider:
This should fix the problem described in https://github.com/opentx/opentx/issues/2084#issuecomment-102593774

### DIFF
--- a/radio/src/lua.cpp
+++ b/radio/src/lua.cpp
@@ -689,7 +689,7 @@ static int luaModelInsertInput(lua_State *L)
   unsigned int first = getFirstInput(chn);
   unsigned int count = getInputsCountFromFirst(chn, first);
 
-  if (chn<MAX_INPUTS && getExpoMixCount(1)<MAX_INPUTS && idx<=count) {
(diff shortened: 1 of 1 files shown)


radio/src/lua.cpp
```diff
@@ -689,7 +689,7 @@ static int luaModelInsertInput(lua_State *L)
   unsigned int first = getFirstInput(chn);
   unsigned int count = getInputsCountFromFirst(chn, first);
 
-  if (chn<MAX_INPUTS && getExpoMixCount(1)<MAX_INPUTS && idx<=count) {
+  if (chn<MAX_INPUTS && getExpoMixCount(1)<MAX_EXPOS && idx<=count) {
```
bsongis: Yes! I think the constant name could be changed to something better!

PR #2320, opened 2015-05-19
Special EU firmware for Taranis

bsongis:
Only D16 mode is EU-compatible. This version will hide both D8 and LR12 modes.

### DIFF
--- a/radio/src/Makefile
+++ b/radio/src/Makefile
@@ -319,7 +319,6 @@ NANO = NO
 # NO - use default Lua allocator
 USE_BIN_ALLOCATOR = NO
 
-
 # Enable trace of events into Trace Buffer for debugging purposes
 # Activating any of these options also activates DEBUG and DEBUG_TRACE_BUFFER
 # Values = NO, YES
@@ -342,6 +341,10 @@ LCD_DUAL_BUFFER = NO
 # Enable internal module PPM mode for Taranis
 # Values = NO, YES
 TARANIS_INTERNAL_PPM = NO
+
+# Support for D16-EU only (no D8, no LR12 which are not EU compatible)
+# Value = NO, YES
+SUPPORT_D16_EU_ONLY = NO
 
 # Enable disk IO simulation in simulator
 # Simulator will use FatFs module and simulated IO layer that
@@ -873,6 +876,9 @@ ifeq ($(PCB), TARANIS)
   endif
   ifeq ($(TARANIS_INTERNAL_PPM), YES)
     CPPDEFS += -DTARANIS_INTERNAL_PPM
+  endif
+  ifeq ($(SUPPORT_D16_EU_ONLY), YES)
+    CPPDEFS += -DMODULE_D16_EU_ONLY_SUPPORT
   endif
   LDSCRIPT = targets/taranis/stm32_flash_bl.ld
   TRGT = arm-none-eabi-
--- a/radio/src/gui/Taranis/menu_model_setup.cpp
+++ b/radio/src/gui/Taranis/menu_model_setup.cpp
@@ -187,38 +187,43 @@ int getSwitchWarningsCount()
 }
 #endif
 
+#if defined(TARANIS_INTERNAL_PPM)
+  #define IF_INTERNAL_MODULE_ON(x)        (g_model.moduleData[INTERNAL_MODULE].type == MODULE_TYPE_NONE ? HIDDEN_ROW : (uint8_t)(x))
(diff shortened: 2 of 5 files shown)


bsongis:
New *SUPPORT_D16_EU_ONLY* compilation option added (default is NO)

ckleanth:
I understand this is to comply with legislation for new products, since I own D8 rx's doest that means that I cannot use the rx if I load the new/EU firmware? lets say I use the normal complication opentx firmware and I use the eu firmware on my x8 rx, would this still work?
bottom line why do we need this option? its already complicated having two rx firmware

kilrah:
Radios shipped by FrSky to European dealers will have RF module firmware that does not support those modes at all, so having them show up in the original firmware menus would just be confusing. This option is there for them to use. 

The firmware we supply through companion will still have them enabled, BUT to use them you'll of course have to flash module firmware that supports them.

ckleanth:
I sure hope you guys implement Mike's maintenance mode soon as I now see this being essential (instead as a nice to have option)
source is here:
http://openrcforums.com/forum/viewtopic.php?f=45&t=6847&sid=2b7d5b6008a92a96a0bf7b65b700cd43

kilrah:
Smart port flashing is already implemented in 2.1, but not as a "maintenance mode".

Johnex:
I updated to the EU firmware on my taranis and my X8R and after doing it all my telemetry converted from a teensy from a pixhawk works, except for current. For whatever reason i get the current and total current on my OSD but i get the value 0 in the taranis. I am on 2.0.17. The code on the teensy to do the telemetry conversion wasn't updated, and neither was the arudcopter version, just the radio and taranis firmware. Any ideas what could be wrong?

kilrah:
Try the previous OpenTX version you had, if it still doesn't work then it isn't in our control.

There could either be a bug in the latest receiver/module firmware, or the teensy firmware data format isn't exactly correct and some changes in the receiver/module FW makes it not work anymore.

Johnex:
Ok Andre thanks for the help )
You are always quick to answer. Can't wait for v2.1 btw guys, you rock!

vieuxdruide:
bonjour si j'ai bien compris les nouvelles taranis vendues en europe avec le nouveau firmware ne supportenet plus les mode D8 T L12
mais est t'il possible de les flasher avec le logiciel non EU
pour récupérer les modes D8 et L12?qui restent très intéressants je trouve.
merci

fred99:
Addon to vieuxdruide: The default value of this new directive  is set to NO. 
Does it mean this mode is not hidden by default?
Will we have a new compilation option in companion to choose to hidden or not ?
Sorry to bother you but normaly EU is not "retroactif"

kilrah:
This is intended only to provide sellers with a version that does not introduce confusion by listing modes not supported by the RF module firmware that is loaded.

AFAIK firmware downloaded from companion would always come with them enabled, you should know better if your radio supports them or not ;)
But addition of an option in companion could be discussed of course.

fred99:
I am sorry for the rework, I haven't seen the comment already done above, thanks for your kindness.
Without a new compilation option in companion, a new customer with a new Taranis loaded with openTX 2.1, upgrading a 2.1.x firmware, will be surprise to see the full listing mode to appear.
Old customer using D8 Tx (it’s legal if bought before 1 of 2015) will be anxious to know if the hidden directive SUPPORT_D16_EU_ONLY is set to YES or NO at each new version of openTX!

fred99:
Thanks for your fast and open answer.

Is there no way to dynamicaly build the list of RF modes after a probing of the RF section by the mainboard?

This will avoid a lot of questions: we have to cope with the version of OpenTX, the version of RF section and the version of the receiver!

 

De : Andre Bernet [mailto:notifications@github.com] 
Envoyé : jeudi 18 juin 2015 20:07
À : opentx/opentx
Cc : fred99
Objet : Re: [opentx] Special EU firmware for Taranis (#2320)

 

This is intended only to provide sellers with a version that does not introduce confusion by listing modes not supported by the RF module firmware that is loaded.

AFAIK firmware downloaded from companion would always come with them enabled, you should know better if your radio supports them or not ;)
But addition of an option in companion could be discussed of course.

kilrah:
As far as we know there is no way to query the RF module for its capabilities.
Besides, someone recently reported that the RF module firmware that is supposed not to have the D8 and LR12 modes actually has them...

fred99:
Thanks for the clarification, my questions are solved.

Issue #2265, opened 2015-05-05
Profiling/benchmarking the code on the target platform


xmxwx:
I plan to propose a code refactor (particularily #2257, but it applies to any bigger refactor), however, I'd like to be sure that the changes I made are not only correct but also don't introduce any performance degradation.

I think it would be nice to have a possibility to benchmark some parts of the code on the target platform. I think about adding an optional compilation parameter which would enable instrumentation code which would use CPU timer to measure how long it takes to execute some routines. The measurements could be performed once on power-on, manually triggered via menu or remote triggered via serial port, with results displayed on LCD, sent through serial port or stored on EEPROM to be read via bootloader or avrdude.

What do you think?

projectkk2glider:
I think I remember having something like this in one of my branches at home. It is a light weight implementation of timers much like the existing timers for mixers duration for example. I will post my implementation later.

xmxwx:
@projectkk2glider did you manage to find some of your code which would help to achieve this?

projectkk2glider:
Sorry, forgot about this. 

Take a look at commit 328439525f2b3c6b3be2cf7466a2db96cb5bef15 from my branch https://github.com/projectkk2glider/opentx/tree/projectkk2glider/issue1849_delay_measuring

It contains a class DebugTimer which can measure the time of code execution. For short times (less than 30ms) it uses high precision timer and for higher times a low precision system tick timer.

xmxwx:
Thanks! I've used your code, extended it to support AVR platform, increased debug_timer_t resolution to 10 us (on AVR it exceeds the 64 us resolution of the internal ~16 kHz timer) and used it to measure the fix for #2319. If #2322 gets accepted, I think this issue could be closed as well. If not, we can cherry-pick only the DebugTimer-related stuff.

Issue #2247, opened 2015-05-02
Jenkins build scripts


xmxwx:
Where can I find scripts used for building plans on [OpenTX's Jenkins service](http://jenkins.open-tx.org:8080/)?

Before posting a pull request I try building the project after my changes using my own script. However, it would be much better if I've done it exactly the same way Jenkins does. Recently I broke Jenkins build because my script didn't include compiling a simulator build. Of course, CI service like Jenkins is there to catch such mistakes, but if the scripts were available it could save us a bit of hassle.

kilrah:
cd radio/src/

# OpenTX on 9X stock
make clean
make PCB=9X EXT=FRSKY HELI=YES
make simu PCB=9X EXT=FRSKY HELI=YES 
make gtests EXT=FRSKY HELI=YES
./gtests

# OpenTX on Taranis Plus
make clean
make PCB=TARANIS PCBREV=REVPLUS HELI=YES LUA=YES
make simu PCB=TARANIS PCBREV=REVPLUS HELI=YES LUA=YES
make gtests PCB=TARANIS PCBREV=REVPLUS HELI=YES LUA=YES
./gtests

# OpenTX on Taranis
make clean
make PCB=TARANIS HELI=YES LUA=YES
make simu PCB=TARANIS HELI=YES LUA=YES
make gtests PCB=TARANIS HELI=YES LUA=YES
./gtests

# OpenTX on Sky9x
make clean
make PCB=SKY9X EXT=FRSKY HELI=YES
make simu PCB=SKY9X EXT=FRSKY HELI=YES
make gtests PCB=SKY9X EXT=FRSKY HELI=YES
./gtests

# OpenTX on Gruvin9x
make clean
make PCB=GRUVIN9X EXT=FRSKY SDCARD=YES
make simu PCB=GRUVIN9X EXT=FRSKY SDCARD=YES

# OpenTX on MEGA2560
make clean
make PCB=MEGA2560

# Companion
cd ../../buildir/
make

xmxwx:
Thank you very much!

xmxwx:
BTW. In order to get gtests compiling on my system (Ubuntu 14.04 Trusty Tahr) I had to add -I/usr/include/qt4 -fPIC to gtests' Makefile rule. It currently seems to be impossible to pass these options to make without modifying Makefile. Add another issue for that?

bsongis:
I am really afraid of somebody adding a rm * in a commit which would break everything ...

[...]

PR #2322, opened 2015-05-19
LCD AVR: Fix for #2319 + gtests improvements

### LINKED ISSUE #2319: LCD AVR: strange y-coordinate dependent bounding line
I couldn't find any explanation for this line of code: https://github.com/opentx/opentx/blob/next/radio/src/gui/9X/lcd.cpp#L1255

I wrote a simple test:

```
TEST(Lcd, InversSteps)
{
  lcd_clear();
  int x = 0, y;
  for(y = 0; y < 16; y++) {
    lcd_putcAtt(x, y, 'B', INVERS);
    x += FW + 2;
  }

  EXPECT_TRUE(checkScreenshot("invers_steps"));
}
```

...and it yields strange result on AVR:
invers_steps_128x64_post_ef28d7e

I've traced this piece of code with git blame and here's how the test result looks like before it was edited in ef28d7e:
invers_steps_128x64_pre_ef28d7e

CPUARM variants (9XRPRO and TARANIS) use the new code which render without such a strange effect:
invers_steps_128x64_9xrpro
invers_steps_212x64_taranis

The mentioned line of code looks like a broken code refactored to an even more broken code.

xmxwx:
Not really a pull request yet. Rather a request for review and comments for now.

Fixes #2319 by adding only 12 bytes.

I feel that before merging it would be good to compare it with the old code by benchmarking (see #2265). Maybe it would be also good to think about some corner cases and introduce more gtests.

Some more thoughts (not necessarily important ones):
- It is only an implementation for STDSIZE. Maybe DBLSIZE could be implemented using STDSIZE?
- How fast and how compact would be lcdPutPattern() on AVR? Maybe it could be optimized for AVR? Maybe it could be optimized so that both AVR and ARM would benefit from it?


bsongis:
Nice one, but why did you remove some tests (and screenshots)?

bsongis:
I added you in the team, would you use a branch from the project, you would have automatic tests on all platforms!

xmxwx:
> Nice one

Thanks. Still working on it, though. I really like the clean approach of ARM implementation of lcd_putcAtt() using lcdPutPattern(). On the other hand this code calls lcd_plot() for every single pixel, which I am afraid might be slow for AVR. That's why I need #2265 (thanks to @projectkk2glider, it's almost there) to see the real performance. If I manage to implement lcdPutPattern() via bit operations like in c46c659, not only AVR could benefit from it, but I hope also ARM. 9XRPRO/SKY9X seems straightforward, but Taranis might make it a little bit more tricky due to more bits per pixel. We'll see :-).

> but why did you remove some tests (and screenshots)?

I've removed Invers_0_0 and Invers_0_1 because InversSteps covers more cases in a single test.

I've also compacted pairs of screenshot sizes (128x64, 212x64) into single screenshots (128x64) in cases where size of the screen didn't matter.

But now I think that the way I didn't was not the best one - I've reduced the canvas on which LCD content is painted, which effectively ignores whatever is painted outside the 128x64 region. I should have leave it LCD_W x LCD_H, compare only 0 <= x < LCD_W, 0 <= y < LCD_H with the reference screenshot and explicitly check that regions 128 <= x < LCD_W and 64 <= y < LCD_H are empty.

xmxwx:
> I added you in the team, would you open a branch in the project, you would have automatic tests on all platforms!

First I've pushed it as LCD and Jenkins started the build. But then I noticed that most branches are in the username/branchname form, so I pushed it as xmxwx/LCD. But then I didn't see Jenkins building it. Does Jenkins build only branches without slash inside their name?

BTW. Thanks to discussion in #2247 I can gtest (hopefully) all the platforms (and add my own build configurations) on my PC. But - of course - doing it also in a separate clean environment like Jenkins is the way go!

bsongis:
It is not normal. I am afraid you have to remove the branch with your name, perhaps it masks all other with your name as a directory

bsongis:
I removed this branch, Jenkins was frozen since it was pushed!

bsongis:
Are we ready to merge this one?

xmxwx:
Well, I've finally measured the original code and my modifications. My current code (c46c659, correct for all values of y % 8, adds only 12 bytes) is ~70% slower in the common case (y % 8 == 0), but ~10% faster in the less common case (y % 8 != 0). The actual test was writing almost whole LCD screen with text. The mentioned 70% degradation is 4 ms -> 6.8 ms.

Compact code? Performance? Correctness? Choose two. :-)

Still, I think that implementing lcdPutPattern() optimized for AVR should be the way to go for the future.

> Are we ready to merge this one?

I've added #2265 with implementation for AVR.

I think it's OK to merge it. But please review as always!

mhotar:
> adds only 12 bytes) is ~70% slower in the common case
I cant say if it is acceptable or not. If there are resources I am sure OK to fix it. In the other hand it adds/fixes feature that we will probably never use for AVR boards.

kilrah:
> In the other hand it adds/fixes feature that we will probably never use for AVR boards.

That. I feel like we're trying to fix something that isn't broken in the first place.

bsongis:
There are very good things in this patch, especially on the readability side, but I do agree with others, it's difficult to accept the 70% slower, even if it only adds 12bytes!

xmxwx:
> My current code (c46c659, correct for all values of y % 8, adds only 12 bytes) is ~70% slower in the common case

> it's difficult to accept the 70% slower, even if it only adds 12bytes!

Well, it in fact is slower for y % 8 == 0, but... I was wrong saying that y % 8 == 0 is the common case. Places where there is the largest amount of text to display is the menu and menu is actually shifted by one, i.e. y % 8 == 1 (the case for which the top bounding line is displayed correctly, see #2319). And for that case my code is faster. I mentioned the average ~10%, but for y % 8 == 1 the performance boost is as high as 24%: 6.6 ms -> 5 ms.

xmxwx:
A sidenote: I've compiled the ARM version of lcd_putcAtt() (i.e. using lcdPutPattern) for AVR. It is 10 times slower (52 ms instead of 4-6.6 ms).

xmxwx:
> Places where there is the largest amount of text to display is the menu and menu is actually shifted by one, i.e. y % 8 == 1. And for that case my code is faster. (...) the performance boost is as high as 24%: 6.6 ms -> 5 ms.

So... how about merging this one?

mhotar:
Yes, only top line is written at y % 8 == 0. So, if three is nice speed gain, better readability and costs only 12b, I am positive to merge It. Good job.

bsongis:
Would you do a recap of the timings? I am a little bit lost! (timings for main view, timings for modelselect)

xmxwx:
For the sake of the recap I've just implemented measuring the actual time it took to render the screen (milliseconds, average of 10 renders, written in the bottom right corner of the screen) and LCD screenshotting over UART for 9XR with TELEMETRY_MOD=14051_SWAPPED. Here are results for a few screenshots before and after applying #2322:

Screenshot name | #2322 applied | #2322 not applied
--- | --- | ---
Model sel | patched_modelsel | unpatched_modelsel
Servos | patched_servos | unpatched_servos
Switches | patched_switches | unpatched_switches
Main: Outputs+values | patched_view_outputs_values | unpatched_view_outputs_values

As you can see, among these screenshots the slight performance boost is only visible on switches page where there is a lot of text written with y % 8 == 1. Other screens have either too little of the y % 8 == 1 text or don't have it at all (only y % 8 == 0), hence a slight performance degradation is seen.

bsongis:
DIfficult choice. In one hand the change is nice, in the other hand the screens which are displayed during the flight are a performance degradation of around 15%

xmxwx:
I've implemented a bitshifting lcdPutPattern(), however, without further optimizations (like static versions for particular y%8) it runs slower than the current algorithm. I don't want to dedicate more time to that now. Maybe in future.

I've added also timings of the simple 44 B fix (posted by @mhotar in #2319).

If you don't want any of the fixes we can still cherry-pick some other interesting parts from this branch (some of the gtests cleanups, DebugTimers - #2265, screenshoting over UART).

PR #2323, opened 2015-05-20
Special EU firmware for Taranis - For Master

bsongis:
Only D16 mode is EU-compatible. This version will hide both D8 and LR12
modes.


bsongis:
Tested OK by the german dealer

PR #2328, opened 2015-05-22
log viewer enhancement, now for next

griebd:
Just the same on next branch, as Log Viewer is exactly the same in next and master... ;)


bsongis:
Thanks!

PR #2344, opened 2015-05-27
Projectkk2glider/issue 2260 simulator volume gain

### LINKED ISSUE #2260: Add setting for volume multiplier for simulator audio
Users usually have lower than max volume set on their radios. This means that simulator audio volume (which respects those settings) is usually quite low. This multiplier would be used to multiply the desired volume before it is applied to audio samples. 

Needed on master and next.

projectkk2glider:
I've left this in a separate commits to aid the review. I can squash them into one commit after it is approved. 

Closes #2260 

The new setting **Simulator volume gain** in the radio profile is used to multiply the volume set by the radio in the simulator. The multiplier is only applied up to resulting maximum volume to avoid audio clipping.

Standalone simu uses default gain of 1.0. It would require addition of option parsing to be able to set the gain from command-line when starting. I think it is not needed, almost nobody uses it anyway.

The standalone Companion simulator uses gain set in Companion.

I will port this to `next` if accepted.

### DIFF
--- a/radio/src/targets/simu/simpgmspace.cpp
+++ b/radio/src/targets/simu/simpgmspace.cpp
@@ -396,8 +396,15 @@ void StopMainThread()
 }
 
 #if defined(CPUARM)
-int audio_volumeGain;
-int Volume;
+
+struct SimulatorAudio {
+  int volumeGain;
+  int currentVolume;
+  uint16_t leftoverData[AUDIO_BUFFER_SIZE];
+  int leftoverLen;
+  bool threadRunning;
+  pthread_t threadPid;
+} simuAudio;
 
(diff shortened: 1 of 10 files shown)


radio/src/targets/simu/simpgmspace.cpp
```diff
@@ -396,7 +396,16 @@ void StopMainThread()
 }
 
 #if defined(CPUARM)
-int Volume = volumeScale[VOLUME_LEVEL_DEF];
+
+struct SimulatorAudio {
```
bsongis: Yes, thanks!

bsongis:
Perfect, thanks!

PR #2345, opened 2015-05-29
Bsongis/i2c hardware on taranis

bsongis:
Tests needed. I don't know if it's a good idea to increase speed over 240kHz


bsongis:
@MikeBland Mike would you review it please? Thanks!

MikeBland:
Bertrand,

From what I can see, it looks OK.

On another subject, Jhsa on the forum just asked me about the push
button as the power switch on the X9E, from a safety point of view. His
concern is a software failure might power the Tx off, rather than get a
watchdog reboot.
If a software failure turned off the soft power switch, then this is a
possibility.
I would suggest, as a safety operation, when the watchdog is
re-triggered, perhaps the soft power output pin should be re-configured
as an output and set active. This would mean a software error that
turned the soft power off, would be corrected, or the watchdog would
reset the CPU.

Mike.


Bertrand Songis wrote:

  @MikeBland Mike would you review it please?
Thanks!
  —
Reply to this email directly or view
it on GitHub.

bsongis:
I need tests from you all
- REV4a max speed is 100kHz. That I am sure, I have it.
- REVPLUS is OK at 400kHz (speed is almost twice actual speed)
- NORMAL Taranis (non plus) should be ok at 400kHz, but I need your tests for being sure! I keep 200kHz for now.

bsongis:
I left some code to write/read/check a 4k block, so that tests are easy, thanks for your feedback!

projectkk2glider:
Testing on Taranis revision B01 (PCBREV=VER4), production date 23.12.2013 (enabled test in ` eepromOpen()`):

| I2C_SPEED | Result |
|----------------|----------|
|  200000      |  **Works** |
|  400000      |  **Also works!** |

bsongis:
Oh that's good news! Anyone else?

MikeBland:
The early circuit diagram we have for the first boards shows 10K pullup
resistors on the I2C bus.
Current production shows 4K7.
It is likely that with 10K, it is not possible to run the bus much
above 100kHz.
I found a reference recommending 4K7 or lower, and I seem to recall
seeing 2K2 as an expected value.

Mike.


Bertrand Songis wrote:

  Oh that's good news! Anyone else? 
  —
Reply to this email directly or view
it on GitHub.

bsongis:
Oh thanks @MikeBland  for this explanation, it makes sense now! I will set 400kHz on all boards except REV4a then!

PR #2349, opened 2015-05-31
Fixes #2260: added volumeGain in simulator (ported from master)

### LINKED ISSUE #2260: Add setting for volume multiplier for simulator audio

projectkk2glider:
Closes #2260 

Had to do some adaptations to the #2344 commits. The profile settings were no longer available in simulator interface. Therefore I changed the interface a little bit to pass on the whole profile to the simulator interface. Some code deduplication also done.

@bsongis  On a related note: it seems that files in https://github.com/opentx/opentx/tree/next/companion/src/firmwares/er9x, https://github.com/opentx/opentx/tree/next/companion/src/firmwares/th9x and several other are obsolete - ie not used anymore. Can they be removed?

### DIFF
--- a/companion/src/appdata.cpp
+++ b/companion/src/appdata.cpp
@@ -295,6 +295,7 @@ QString Profile::fwName()        const { return _fwName;        }
 QString Profile::fwType()        const { return _fwType;        }
 QString Profile::name()          const { return _name;          }
 QString Profile::sdPath()        const { return _sdPath;        }
+int     Profile::volumeGain()    const { return _volumeGain;    }
 QString Profile::pBackupDir()    const { return _pBackupDir;    }
 QString Profile::splashFile()    const { return _splashFile;    }
 bool    Profile::burnFirmware()  const { return _burnFirmware;  }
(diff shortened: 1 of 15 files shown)


bsongis:
I prefer we keep them for now, they should be needed later (perhaps not the interface, is it what you meant?)

bsongis:
Thanks!

projectkk2glider:
>I prefer we keep them for now, they should be needed later (perhaps not the interface, is it what you meant?)

Not sure if they used at all. They are certainly not used for simulation. Are they  used for eeprom import?

bsongis:
they should be used in simulator!

projectkk2glider:
How is that possible? I certainly can not select (for example) er9x or th9x firmware in radio profile settings.

bsongis:
No it would be dlls used in the standalone simulator!

projectkk2glider:
I have created a problem, the Windows build is broken. http://jenkins.open-tx.org:8080/job/OpenTX%20Unstable%20-%20Nightly%20builds/316/

The reason is that I introduced dependency on `class Profile` (which is in common lib) in simulator dynamic libraries. It works on Linux because it supports undefined symbols in .so libraries, but not on Windows.

Since the simulator dynamic libraries were meant to be very independent of other parts of Companon, I decided to revert back to the old system. I will add a new method to the simulator interface that will set volume gain. Hold on...

projectkk2glider:
Fixed in 67cd3214000ee7dd10ca22caaba0ed06c34248fa

Issue #503, opened 2014-01-04
Unexpected behaviour when Diff and Trim are specified in the same mix.


RC-SOAR:
Firmware r2768 - board: taranis

Specifying Diff and Trim in the same mixer line leads to unexpected behaviour
- Diff is currently applied to the Trim value. This leads to lower trim sensitivity on one side than the other. It will also produce unwanted camber changes when applying trim on a dual-aileron setup.
- Servo movement is incorrect on one side if trim applied
- Crossover point for diff is incorrect

Seems that OpenTx does not take into account trim offset when processing diff. I'm working around this by splitting the Diff and Trim into separate mixes, but it's a bit untidy and requires extra mixer lines.

PS thanks for all the good work :-)

JivaroFAD:
I think you should re-opent this issue.

Please read the following test (works with curve or diff, that's the same) with 2.0.99 :
Stick Aileron, ratio = 25%
CH1 : source = aileron, mixratio = 100%, diff = 50%
CH2 : source = aileron, mixratio = -100%, diff = 50%

Put aileron stick on max right/left : CH1 = +25/-12.5, CH2 = -12.5/+25
=> OK, differential works normaly with stick alone
Now put aileron trim on max right : CH1 = +25, CH2 = -12.5
=> OK, differential works normaly with trim alone
Put again aileron stick on max right/left : CH1 = +50/0, CH2 = -25/0
=> KO, differential works abnormaly with trim + stick
      we should expect : CH1 = +50/+12.5, CH2 = -25/+12.5

Now, put trim at 0 and add +25 mixoffset to each : CH1 = +25, CH2 = +25
=> OK, offset works normaly
Put again aileron stick on max right/left : CH1 = +50/0, CH2 = 0/+50
=> KO, differential works abnormaly with stick + mixoffset
      we should expect : CH1 = +50/+12.5, CH2 = +12.5/+50

In my mind, normal operation should be :
Ouput = (MixOffset + Trim * +/-Diff + Stick * StickRatio * +/-Diff) * MixRatio

Or, a little better, assuming that trim effect can be reduce proportionnaly to stick ratio :
Ouput = (MixOffset + (Trim * +/-Diff + Stick * +/-Diff) * StickRatio) * MixRatio

Note : would I prefer applying MixOffset to input (instead of ouput as actualy) because it makes more sense for uncentered mixes (like gaz or airbrakes to elevator compensation). Moreover, if ouput needs to be offseted, it can easily be done with an additionnal line mix.

PR #2361, opened 2015-06-13
Jivarofad, offset and differential improvement in mixer

JivaroFAD:
Curve and differentiel are now applied to stick and trim separatly, so real differential correspond exactly to programmed differential when trim is different from 0.
See description in issue #503

In the same time, offset is now applied to source, it makes more sense for un-centered compensations (like throttle or airbrake to elevator) then applying offset at the end of the mix (that can simply be done with another line mix, unlike source offsetting).


radio/src/mixer.cpp
```diff
@@ -838,38 +841,66 @@ void evalFlightModeMixes(uint8_t mode, uint8_t tick10ms)
           v = (tact >> DEL_MULT_SHIFT);
         }
       }
+      if (apply_offset_and_curve) {
+        if (!(mode & e_perout_mode_notrims)) {
```
bsongis: Indentation problem?

radio/src/mixer.cpp
```diff
@@ -838,38 +841,66 @@ void evalFlightModeMixes(uint8_t mode, uint8_t tick10ms)
...
+        v -= trim;  
+      }
+#else
+      int8_t mix_trim = md->carryTrim;
+      if (mix_trim == THR_STICK && g_model.throttleReversed)
+        v += trim;
+      else
+        v -= trim;
+#endif
+        }
```
bsongis: Indentation here as well

JivaroFAD:
Full testing have been done this day, and I discover an inappropriate ouput if y <> 0 if x=0 (curve is offsed by result of trim calculation).
It works properly if curve crosses 0 when x=0, so a solution would be to make a subtraction of y value when x=0. I'm not sure it is worth the pain, so I decided to dedicate this new feature only for differential.

JivaroFAD:
Here is a graphical representation of the new mixer formalism :

image

Ready for merge ? :-)

bsongis:
I will first squash all commits into one, ok?

JivaroFAD:
No problem :)

bsongis:
This one needs to be checked as well and integrated into 2.1.0 before the release!

projectkk2glider:
The calculation for mixes was:
* in pre 2.0 days `Output = (Source + Offset) * Weight`
* in 2.0 changed to `Output = Source * Weight + Offset`

Now this changes it back **again**. Is this a smart move?

JivaroFAD:
I think revert to "pre 2.0" it is better because : 
- applying an offset after is still easy (just add a complementary mixer line) 
- uncentered compensation (like throttle or airbrakes to elevator) is much simplier if offset is applied before (because you don't have to change the offset when adjusting the weight)

kilrah:
The change for 2.0 was made after numerous requests and follows a better "logic" so it shouldn't be reverted.

JivaroFAD:
I read the discussions provides by projectkk2glider, and I don't agree with you when you say "As offset depends on weight the calculation is complicated". It is exactly the inverse in my exemples listed before.
Take the very common "throttle to elevator" compensation :
- with offset applied before (to input) : just put +100% to offset (so, compensation = 0% when throttle low) and adjust weight to have the expected compensation with full throttle
- with offset applied after (to ouput) : each time you change the weight, you have to change the offset to have compensation = 0 when throttle low. Not so simple.

If this change does disturb you too much, no problem, I will add it into a flash option, something like "SymTrim" or include it in "SymLimits" (usefullness is similar).

Please note that this patch does not only concern the offset, but also the differential behaviour with trim, that solves a big issue for gliders.

projectkk2glider:
Maybe a crazy idea, but what if we have two offsets:
```Output = (Source + Offset1) * Weight + Offset2```

Then again it might just confuse users too much.

**Comment by JivaroFAD** [comment]
Not crasy, but already existing :) Precisely, with "=" option in servo menu, subtrim acts exactly as Offset2 and allow to simply (without changing limits) align a control surface (for example).

**Comment by kilrah** [comment]
Not at all, because "Offset2" would apply to inputs and mixer i.e. not used for the same purpose at all. 

The most common use for Offset in the inputs/mixer is to say "I want my control to cause a response of -60 to +40"... and it's easy with the current way to figure out that you need a weight of 50% and an offset of -10%,  while with the old method it was a pain to calculate.

**Comment by JivaroFAD** [comment]
The most common use I know are : 
- offset = 0 for common centered mix
- offset = +/-100% of stick for uncentered compensation
For others special needs, like -60% / +40%, a curve should be used, or  Input with different rates when x>0 and and x<0.

**Comment by kilrah** [comment]
> a curve should be used

Why not use a curve for your case instead?

>  Input with different rates when x>0 and and x<0.

That wouldn't work as the center would be forced to 0:0.

So to me either the 2.0 behavior should be kept, or we can do like ersky9x and add a choice whether offset is applied before or after, but that means an eeprom change for which it is too late as the format for 2.1 is now frozen... so 2.0 way it is.

We got many complaints about the pre-2.0 offset and none since it has changed until now, so...

**Comment by JivaroFAD** [comment]
Don't you think it could be because 2.0 has introduced another major improvement (the graphical view of the offset + weight combination), much more than the change of the offset's application ?

So...  I propose a simple flash option, that will don't change the eeprom. OK ?

**Comment by bsongis** [comment]
Yes, no choice for the offset calculation.

**Comment by kilrah** [comment]
Umm no because that graphical thingy isn't actually present in 2.0, only in 2.1 ;)

And I'm not for a firmware option either, because how will you handle that in companion? If someone simulates his model without making very sure the option is selected correctly in companion according to the firmware he has on the radio he will have completely different behavior than on the radio and might not catch it before it's too late.

**Comment by bsongis** [comment]
There is one small possibility, with more work for Franck :)
I have seen there is one bit unused in M2650 / M2561 MixData structure, I
am ok with a change on M2560 / M2561, but not on Taranis / 9XR-PRO (it will
wait next release)

2015-06-29 18:09 GMT+02:00 Andre Bernet <notifications@github.com>:

**Comment by JivaroFAD** [comment]
Thanks Bertrand :)
That's a good deal as I really need this feature for all my applications (some examples : glider : airbrakes to elevator, camber to elevator, plane : throttle to elevator, flap to elevator, heli : throttle to tail), and the new system "offset after weight" is very frustrating as it do not allows to use GVAR on weight to set theses compensation in flight (as it moves the compensation at low end of input, where it should be 0) :(

**Comment by mhotar** [comment]
While reading this discussion, I can see one compromise that can be usefull here. What about to have checkbox "Zero at the side" to change source behaviour to 0-100. Does it make sense? It can be simpler to understand than two types of offset.

**Comment by mhotar** [comment]
Sorry, I meant 0 - 200 instead of 0 - 100.

**Comment by kilrah** [comment]
A bit too specific, no?

**Comment by JivaroFAD** [comment]
I don't have any example of application that needs an input offset different of 100.

A checkbox makes sense for all my applications, which use raw input of throttle stick. 0-100 should be the right way, assuming that 100 is the max travel usable for a compensation (equal to full throw of control surface).
The only issue I can see with this system is that the 0 side of the stick is not always the same, it can be -100 or +100, depending of application. So we should have 2 checkbox to manage the side, that's a little tricky.

Just to make a banchmark of other radios brands : most (probably all...) use offset at input and, for throttle stick, many have a system that recognise and memorise its position. That makes exactly the same thing as writing the value in an input field, but is more "user friendly".

**Comment by JivaroFAD** [comment]
> I have seen there is one bit unused in M2650 / M2561 MixData structure

I understand that we can have the choice for each mixer line. If the case, we can do something that will please everyone :)

So, why not adding a checkbox in each mixer  :
- unchecked : offset after
- checked : offset before

Or a field with two possibilites :
- S : offset applied to Source of mix
- O : offset applied to Ouput of mix

**Comment by kilrah** [comment]
> So, why not adding a checkbox in each mixer :
unchecked : offset after
checked : offset before

That's what er9x does, and I'm fine with it given your scenarios.
But again too late now, it will have to wait for 2.2. If you had made it clear in a comment that you had changed the offset when committing 2 weeks ago it could have been added :(

**Comment by JivaroFAD** [comment]
I thought this block diagram was good enough : https://github.com/opentx/opentx/pull/2361#issuecomment-112151466

But no problem, the most important for me is that we have a consensus :)

For the differential mod, are you ok ?

File: radio/src/mixer.cpp [hunk 3]
```diff
@@ -767,13 +767,20 @@ void evalFlightModeMixes(uint8_t mode, uint8_t tick10ms)
 #endif
       }
 
+      /*========== OFFSET BEFORE ========      Reserved for new option "OffsetOnInput"
```
bsongis: I would use a #if 0 here with a TODO in 2.2.0

radio/src/mixer.cpp
```diff
@@ -851,25 +854,62 @@ void evalFlightModeMixes(uint8_t mode, uint8_t tick10ms)
...
+      //========== OFFSET AFTER ========= 
+      if (apply_offset_and_curve) { 
+        int16_t offset = GET_GVAR(MD_OFFSET(md), GV_RANGELARGE_NEG, GV_RANGELARGE, mixerCurrentFlightMode); 
+        if (offset) dv += int32_t(calc100toRESX_16Bits(offset)) << 8; 
       }
 
       //========== DIFFERENTIAL =========
 #if !defined(XCURVES)
       if (md->curveMode == MODE_DIFFERENTIAL) {
+        //Revert trim
```
bsongis: You revert the trim here, but what if there is a SLOW on the mixer line?

radio/src/targets/mega2560/board_mega2560.h
```diff
@@ -82,8 +82,8 @@
 #define RESUME_PPMIN_INTERRUPT()   TIMSK3 |= (1<<ICIE3)
```
bsongis: I think this cosmetic change is not intentionnally done?

radio/src/mixer.cpp
```diff
@@ -793,11 +800,7 @@ void evalFlightModeMixes(uint8_t mode, uint8_t tick10ms)
...
       }
-
-      // saves 12 bytes code if done here and not together with weight; unknown reason
-      int16_t weight = GET_GVAR(MD_WEIGHT(md), GV_RANGELARGE_NEG, GV_RANGELARGE, mixerCurrentFlightMode);
-      weight = calc100to256_16Bits(weight);
-
+ 
       //========== SPEED ===============
       // now its on input side, but without weight compensation. More like other remote controls
       // lower weight causes slower movement
```
bsongis: I wonder if SLOW shouldn't work on weight instead of the input value

radio/src/mixer.cpp
```diff
@@ -793,11 +800,7 @@ void evalFlightModeMixes(uint8_t mode, uint8_t tick10ms)
 #endif
         }
       }
-
```
bsongis: 2) This code should remain here (so that SLOW could modify weight)

NB: perhaps weight will have to be shifted to avoid precision lost

radio/src/mixer.cpp
```diff
@@ -851,25 +854,62 @@ void evalFlightModeMixes(uint8_t mode, uint8_t tick10ms)
...
 
       //========== DIFFERENTIAL =========
 #if !defined(XCURVES)
       if (md->curveMode == MODE_DIFFERENTIAL) {
+        //Revert trim    
+        if (apply_offset_and_curve) {
+          if (!(mode & e_perout_mode_notrims)) {
+#if defined(VIRTUALINPUTS)
+            if (md->carryTrim == 0) {
+              dv -= dtrim;
```
bsongis: 3) Differential would also play on weight (and obviously this code should be executed before applying weight to the input)

bsongis:
Ok I am not sure I am clear enough in my explanations, I guess I should give it a try before telling you what should be done... Let-me know!

JivaroFAD:
No problem :) I asked myself exactly the same questions, and was worried by gtests failure when I try to re-arrange the order of application in something more logical.

So, for the moment, I kept the same order as in the "historic mixer" (with diff after weight whereas curve is before), all is working correctly on the radio and with gtets, including with slow and delay.

Then, for 2.2, I planed to rework all of this in a much better way... I really do not like to leave things half-finished ! But, if we give some more work together, we can we something much better and very quickly :)

bsongis:
I think we are not ready. Both my mods and Franck's mods don't pass a new google test that Franck just wrote. I delay this pull request to 2.1.1 (differential fix) and 2.2.0 for the offset before/after (because of the EEPROM change)

bsongis:
Not ready, I close it

PR #2365, opened 2015-06-16
Fixes #2364: use fixed analog value for non-connected pots

### LINKED ISSUE #2364: Taranis Plus S2 pot problem
Hi guys!

I have the latest revision Taranis Plus with the latest software from FrSky. When I try to set my S2 pot as an input source by turning it while the source option is highlighted, the transmitter sometimes detects it as S3. This mostly happens close to the center point of S2. In my settings S2 is set as pot with detent and S3 as none. Also calibrated the pot multiple times. Here is a video someone posted on youtube about this behaviour: http://youtu.be/NeWSy73GsEs
Anyone has any idea why this could be happening? Otherwise the pot works fine when I select it as source manually. S3 is not installed in my radio.

Thanks for any info!

projectkk2glider:
**WARNING** this is completely untested! I have no radio with S3 pot.

Closes #2364 

If accepted I will port it to the `next`

### DIFF
--- a/radio/src/targets/taranis/adc_driver.cpp
+++ b/radio/src/targets/taranis/adc_driver.cpp
@@ -136,6 +136,12 @@ void adcRead()
       Analog_values[i] = 0;
     }
 #endif
+    if (IS_POT(i) && !IS_POT_AVAILABLE(i)) {
+      // Use fixed analog value for non-existing and/or non-connected pots.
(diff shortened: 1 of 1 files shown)


bsongis:
I didn't test, but obviously this test is needed! Would you port it to 'next'? Thanks!

PR #2366, opened 2015-06-16
Fixes #2326: Google Earth export GPS glitch filtering and lat/lon syn…

### LINKED ISSUE #2326: Google Earth export
Lat/lon data tends to be out of sync. The export from logs seems to pick samples when there is a change in lat/lon, even if just one of them has changed. However, when the next coordinate get updated in the next sample, this seems to be ignored. 
This leads to stepped tracks. Some filtering would be nice. Like; if just one coordinate has changed, pick the next sample (which seems to contain both, even at 10Hz sampling).

Log file:
http://www.rcgroups.com/forums/showatt.php?attachmentid=7887955&d=1432153024

ge

projectkk2glider:
…chronization

Closes #2326 

Done a couple of related things:
* a little code refactoring
* fixes #2326 
* filters out any GPS glitches and zero lat/lon location
* fixed icon path

### DIFF
--- a/companion/src/helpers.cpp
+++ b/companion/src/helpers.cpp
@@ -1097,3 +1097,53 @@ bool caseInsensitiveLessThan(const QString &s1, const QString &s2)
 {
   return s1.toLower() < s2.toLower();
 }
+
+bool gpsGlitchFilter::isGlitch(double latitude, double longitude) 
+{
+  if ((fabs(latitude) < 0.1) && (fabs(longitude) < 0.1)) {
+    return true;
+  }
+
+  if (lastValid) {
+    if (fabs(latitude - lastLat) > 0.01) {
+      // qDebug() << "gpsGlitchFilter(): latitude glitch " << latitude << lastLat;
+      if ( ++glitchCount < 10) {
+        return true; 
+      }
+    }
+    if (fabs(longitude - lastLon) > 0.01) {
+      // qDebug() << "gpsGlitchFilter(): longitude glitch " << longitude << lastLon;
+      if ( ++glitchCount < 10) {
+        return true; 
+      }
+    }
+  }
+  lastLat = latitude;
+  lastLon = longitude;
+  lastValid = true;
+  glitchCount = 0;
+  return false;
+}
+
+bool gpsLatLonFilter::isValid(QString latitude, QString longitude)
+{
+  if (lastLat == latitude) {
+    return false;
+  }
+  if (lastLon == longitude) {
+    return false;
+  }
+  lastLat = latitude;
+  lastLon = longitude;
+  return true;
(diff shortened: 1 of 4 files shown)


companion/src/helpers.h
```diff
@@ -162,4 +162,31 @@ QSet<QString> getFilesSet(const QString &path, const QStringList &filter, int ma
 
 bool caseInsensitiveLessThan(const QString &s1, const QString &s2);
 
+
+class gpsGlitchFilter
```
bsongis: First letter should be uppercase!

companion/src/helpers.cpp
```diff
@@ -1097,3 +1097,53 @@ bool caseInsensitiveLessThan(const QString &s1, const QString &s2)
...
+  lastLat = latitude;
+  lastLon = longitude;
+  return true;
+}
+
+double toDecimalCoordinate(QString value) 
+{
+  double temp = int(value.left(value.length()-1).toDouble() / 100);
+  double result = temp + (value.left(value.length() - 1).toDouble() - temp * 100) / 60.0;
+  if ((value.right(1) == "S") || (value.right(1) == "W")) {
```
bsongis: It would be better to compare a char, not a QString!

companion/src/helpers.h
```diff
@@ -162,4 +162,31 @@ QSet<QString> getFilesSet(const QString &path, const QStringList &filter, int ma
...
+  int glitchCount;
+  double lastLat;
+  double lastLon;
+};
+
+class gpsLatLonFilter
+{
+public:
+  gpsLatLonFilter() {};
+  bool isValid(QString latitude, QString longitude);
```
bsongis: Those args should be const QString & (unless they are modified). It will avoid a couple of memcpy ...

bsongis:
Thanks, the new map is really nice!

companion/src/helpers.cpp
```diff
@@ -1097,3 +1097,55 @@ bool caseInsensitiveLessThan(const QString &s1, const QString &s2)
...
+}
+
+double toDecimalCoordinate(const QString & value) 
+{
+  if (value.isEmpty()) return 0.0;
+  double temp = int(value.left(value.length()-1).toDouble() / 100);
+  double result = temp + (value.left(value.length() - 1).toDouble() - temp * 100) / 60.0;
+  QChar direction = value.at(value.size()-1);
+  if ((direction == 'S') || (direction == 'W')) {
+    result = -result;
```
projectkk2glider: @bsongis is this what you had in mind? I am not sure if it is any faster.

bsongis: Perfect, and yes I really think it's faster!

PR #2370, opened 2015-06-17
Fixes #2364: use fixed analog value for non-connected pots (ported fr…

### LINKED ISSUE #2364: Taranis Plus S2 pot problem

projectkk2glider:
…om master)

Closes #2364 

Again this was **NOT TESTED**

### DIFF
--- a/radio/src/targets/taranis/adc_driver.cpp
+++ b/radio/src/targets/taranis/adc_driver.cpp
@@ -165,6 +165,12 @@ void adcStop()
 
 uint16_t getAnalogValue(uint32_t index)
 {
+  if (IS_POT(index) && !IS_POT_AVAILABLE(index)) {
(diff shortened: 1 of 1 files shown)


JivaroFAD:
This feature will be a great addition for M2560 board :)
I can do the test on this board.

bsongis:
Thanks!

PR #2386, opened 2015-06-26
Fixes #2382: new atribute onlyPositive for telemetry sensor

### LINKED ISSUE #2382: Vario only showing positive values
As title. Taranis Plus. Nightly 23/06/2015. Frsky D4R-11 and OpenXVario. In sensor page (13/13) only positive values are shown for Vspd.

projectkk2glider:
This is an implementation of the third option in https://github.com/opentx/opentx/issues/2382#issuecomment-115352055

Consider this as a possible solution for telemetry values that should only be positive. The name of the new option and its translations could probably be improved. Comments welcome.

Closes #2382


bsongis:
Right, the system is more flexible with this new attribute, thanks!

PR #2388, opened 2015-06-28
Fixes #2381: time based V2 params were not formated properly. Time un…

### LINKED ISSUE #2381: When V1 is a timer the value in V2 stays decimal
When V1 is Timer, the V2 changes in hours, minutes, seconds in Companion 2.0.x  (logical switches)
In Companion 2.0.99 it stays decimal (seconds?)

This may be intended.
I just notice it's different

I'm using nightly build 23-06-2015

projectkk2glider:
…it also added.

Maybe we need a new enum for SOURCE_TYPE_SPECIAL telemetry items.

Closes #2381


bsongis:
Thanks!

PR #2445, opened 2015-07-09
Fixes #2444: Wrong scroll direction in simulators for X9E

### LINKED ISSUE #2444: Wrong scroll direction in simulators for X9E
Affected programs: Companion, Standalone simulator, simu.

Steps to reproduce:
* open X9E radio simulation
* scroll trough menus with + an - keys

Bug: the scroll direction is wrong, but not in all cases, for example in name editor the direction is good. The problem is caused by the fact that rotary encoder on X9E works in other direction.

projectkk2glider:
Closes #2444


bsongis:
Thanks!

PR #2482, opened 2015-07-17
Enables side by side installation of Companion 2.1 and 2.0 on Linux

### LINKED ISSUE #2500: Simulators missing from Companion installer on Linux platforms
Two problems:
 * The simulator dynamic libraries libpentx-*-simulator.so are missing from RPM and DEB packages. 
 * The Companion looks in current working directory for those files, which is not the standard way on Linux, the libs should be in /usr/local/lib/companion.

Perhaps similar problem is also present on MAC.

This issue is solved in pull #2482, which also includes a change that enables Companion 2.0 and 2.1 install at the same time (both packages can't have the same files).

@bsongis and @kilrah I urge you to consider that pull request before the 2.1.1 is released!

projectkk2glider:
Adds missing simulator libraries on Linux
Installs Standalone Simulator icon on Linux

I've only tested this on Linux, I might have broken something on other platforms.

The simulator library search path is hard coded, which is not ideal solution. But I have no better idea. Without this change the simulation does not work and in fact the official RPM package for 2.1 is broken.

This also enables one to have installed at the same time both the old Companion 2.0 and the new one 2.1 which I think is important.

I am not a cmake expert, so improvements/comments are welcome.

Closes #2500


companion/src/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-PROJECT( companion )
```
bsongis: Why would we have a different project name? It should not be dependant of the release!

projectkk2glider: It is a left over of the conversion process and can be changed back. It was used as a target name before, now I introduced COMPANION_NAME that is used for target name.

The original problem was that one can not rename a target when installing (`INSTALL( TARGETS `) like it is possible when installing normal files (`INSTALL( FILES`).

projectkk2glider: Hmm, I was wrong above the PROJECT_NAME is used also in the name of DEB package and it must have different value from the Companion 2.0. We used `companion` for 2.0, I would like to use `companion2.1` for 2.1. The package name must be different if we want to have both versions installed at the same time!

bsongis: Isn't it possible to have a different name for the package / directories, without changing the project name? It seems not good to me to do so!

projectkk2glider:
The project now changed back to "companion". But this does not change the fact that this branch now creates two targets:
 * companion2.1 (Companion2.1.exe on Windows)
 * simulator2.1 (Simulator2.1.exe on Windows)

I see no other way of enabling side by side install of 2.1 and 2.0 on linux where both executables must be installed in the same directory /usr/local/bin. The same principle is used for example for Python.

companion/src/CMakeLists.txt
```diff
@@ -9,6 +9,8 @@ CMAKE_MINIMUM_REQUIRED( VERSION 2.8 )
 
 SET(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
 
+SET(COMPANION_NAME "companion2.1")
```
bsongis: Perhaps reuse the MAJOR and MINOR constants now?

projectkk2glider:
Still not ideal, we have two hard coded dependencies that will brake if version changes:
 * values in `companion/src/companion.desktop` and `companion/src/simulator.desktop`

They should be pre-procesed by cmake it think.

kilrah:
Speaking of that, I tried building the package on my Ubuntu and get this:

> Run CPack packaging tool...
CPack: Create package using DEB
CPack: Install projects
CPack: - Run preinstall target for: companion
CPack: - Install project: companion
CMake Error at /home/andre/opentx/build/cmake_install.cmake:92 (FILE):
  file INSTALL cannot find
  "/home/andre/opentx/code/companion/src/simulator.desktop".
CPack Error: Error when generating package: companion2.1

projectkk2glider:
Yes I forgot that file, hold on...

projectkk2glider:
Should be fixed now. 

Next I will try to get read of fixed library search path https://github.com/opentx/opentx/pull/2482/files#diff-2d9a347e3d44c40b10cb0c89f5883e36R49

kilrah:
Better!

I do get a warning when installing the package though:

ubuntu 14 04 x86-2015-07-19-13-51-35

projectkk2glider:
That file (control) is generated by cmake, not sure we can influence that.

kilrah:
http://stackoverflow.com/questions/17796670/cpack-deb-generator-control-file-has-bad-permissions-mdsums-0644-0644

Seems it will be fixed in a cmake update, until then a custom script is needed.

projectkk2glider:
Funny, my system produces good permissions: 
```
companion2.1_2.1.0_amd64/DEBIAN $ stat control 
  File: ‘control’
  Size: 484       	Blocks: 8          IO Block: 4096   regular file
Device: 811h/2065d	Inode: 18088715    Links: 1
Access: (0644/-rw-r--r--)  Uid: ( 1000/  damjan)   Gid: ( 1000/  damjan)
Access: 2015-07-19 14:00:34.729753345 +0200
Modify: 2015-07-19 13:54:20.000000000 +0200
Change: 2015-07-19 14:00:33.573747298 +0200
 Birth: -
```

projectkk2glider:
This branch was rebased and squashed into one commit.

projectkk2glider:
I am pretty happy whit this solution **for Linux** now, even 
```
cmake -DCMAKE_INSTALL_PREFIX=/opt/foo ../src/
```
Works as intended. 

**But** I am not sure (I haven't tested) about other platforms. Somebody needs to do a test on Windows and MAC. I think that on MAC the simulator libs are still missing!

bsongis:
Ok I merge and test under VC++

kilrah:
OK, on Windows it breaks everything. The executables are now companion2.1.exe and simulator2.1.exe, while all the packaging and distribution scripts are coded to companion.exe and simulator.exe.

There are 2 solutions, either update the scripts or keep the old naming for Windows.

IMO the latter is the best. Have the version in the name for platforms that require it (linux and mac) but not in Windows where the separation is made by the different directory.

projectkk2glider:
Ok, I will try to get my Windows setup going and fix this. Hold on...

projectkk2glider:
@kilrah  would you try now, the fix https://github.com/opentx/opentx/commit/7b490b03d942f781333c7c2ee61d70ae9e995c74 commited to master.

kilrah:
Works now :)

I guess we'll take care of Mac later.

projectkk2glider:
Good. I will not touch MAC with a ten foot pole :grin:

PR #2496, opened 2015-07-18
Should fix #2492 - review please, but should be good, radio now match…

### LINKED ISSUE #2492: [Taranis B] Heli settings -> Swashring do not work as expected
Taranis B OpenTx 2.1 English, haptic ,timer3, LUA, sqt5font, mixmonitor

In Taranis Open attached file MODEL17-2015-07-18.bin
https://www.dropbox.com/s/5807mj4tfjpcd4g/MODEL17-2015-07-18.bin?dl=0

and move Aileron stick left and right, look in Channel monitor
there you see that cyc2, cyc3 dont have the same max throw Ail - left <-> right

change Heli settings -> Swashring from 100 to 0 and do the same test, all is ok

This problem is only in Taranis 2.1 not companion 2.1 simulator

kilrah:
…es the simu and seems to behave OK

### DIFF
--- a/radio/src/mixer.cpp
+++ b/radio/src/mixer.cpp
@@ -482,13 +482,14 @@ void evalInputs(uint8_t mode)
         }
       }
 
-#if defined(HELI)
+#if defined(VIRTUALINPUTS)
+      calibratedStick[ch] = v;
+#else
+ #if defined(HELI)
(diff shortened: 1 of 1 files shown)


bsongis:
But 3 non regression tests are KO!

bsongis:
Squashed into one commit and pushed to 'master' and 'next'

PR #2501, opened 2015-07-19
Fixes #2404: model name played when model is loaded

### LINKED ISSUE #2404: OpenTx 2.1 enhancement for model-name.wav
I always have the model name spoken as well as displayed on the tx.  It would be very convenient if a model_name(.wav) could be included on the setup page just as the model image(.bmp) is.

Then a Global function to play model-name or just automatically played or with a paly/no-play checkbox.

projectkk2glider:
Plays `SOUNDS/…<language>/<model_name>/name.wav`

Closes #2404


radio/src/eeprom_raw.cpp
```diff
@@ -382,6 +382,10 @@ void eeLoadModel(uint8_t id)
     LOAD_MODEL_BITMAP();
 
     SEND_FAILSAFE_1S();
+
+#if defined(SDCARD)
```
bsongis: the ifdef is not needed, you defined it in audio_arm.h
instead I would see it also defined as nothing in audio_avr.h and beep.h

projectkk2glider:
>Does it work?

It works, tested in simu for Taranis target.

bsongis:
With your last commit it's perfect :)

bsongis:
Would you port it to 'next'? Thanks!

projectkk2glider:
Ported to next in c0fbd5cc2d5a2c47bfc7a8d0262b3dccd1eab76c

bsongis:
Thanks Damjan!

rdeanchurch:
Thank you Gentlemen!
Dean

PR #2522, opened 2015-07-20
Fixes #2519: removed telemetry values from available selection in Glo…

### LINKED ISSUE #2519: Companion 2.1.99 crashes when I open general settings
companion-crash
Companion 2.1.99 crashes when I open general settings.
It doesn't do this when it's empty...

It does so when I open General settings after opening an eepe or reading the settings from my radio.
I just tested it using remote desktop on my machine at work and it does the same.
It seems to be related to my radio's setup, so I will attach the eepe
20jul2015 eepe

projectkk2glider:
…bal Functions for "Play Value"

Closes #2519

### DIFF
--- a/companion/src/helpers.cpp
+++ b/companion/src/helpers.cpp
@@ -681,8 +681,6 @@ void populateSourceCB(QComboBox *b, const RawSource & source, const GeneralSetti
     }
     else {
       for (int i=0; i<(flags & POPULATE_TELEMETRYEXT ? TELEMETRY_SOURCES_STATUS_COUNT : TELEMETRY_SOURCES_COUNT); i++) {
-        if (i==TELEMETRY_SOURCE_RSSI_TX && IS_TARANIS(board))  // TODO this condition is never true since else clause is never taken for ARM boards!!!
-          continue;
         if (i==TELEMETRY_SOURCE_TX_TIME && !GetCurrentFirmware()->getCapability(RtcTime))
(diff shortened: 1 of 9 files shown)


companion/src/helpers.cpp
```diff
@@ -681,7 +681,7 @@ void populateSourceCB(QComboBox *b, const RawSource & source, const GeneralSetti
     }
     else {
       for (int i=0; i<(flags & POPULATE_TELEMETRYEXT ? TELEMETRY_SOURCES_STATUS_COUNT : TELEMETRY_SOURCES_COUNT); i++) {
-        if (i==TELEMETRY_SOURCE_RSSI_TX && IS_TARANIS(board))
+        if (i==TELEMETRY_SOURCE_RSSI_TX && IS_TARANIS(board))  // TODO this condition is never true since else clause is never taken for ARM boards!!!
```
projectkk2glider: This is just something I found while working nearby. I think the whole condition it is obsolete and could be removed.

bsongis:
Thanks a lot! Would you port it to 'next' ? For users on 2.1.99 of course ;)

projectkk2glider:
Way ahead of you :smile:  done.

bsongis:
oh ok, sorry :)

projectkk2glider:
No, no, don't be sorry. It is easy to forget something. The reminders are always welcome.

PR #2523, opened 2015-07-20
Fixes #2513: set value of duplicated telemetry sensors

### LINKED ISSUE #2513: "Copy sensor" function on telemetry item popup
The system is made so that you can have multiple copies of a sensor with different attributes, but at this point you can only get to it if you enter everything manually. That's all good although cumbersome for things like Alt, but we've got a problem for a few sensor types that have hidden attributes that can't be entered manually.

projectkk2glider:
I haven't checked if there is any other check that prevents creation of duplicate sensor in gui.

### DIFF
--- a/radio/src/telemetry/telemetry.cpp
+++ b/radio/src/telemetry/telemetry.cpp
@@ -482,12 +482,19 @@ int lastUsedTelemetryIndex()
 
 void setTelemetryValue(TelemetryProtocol protocol, uint16_t id, uint8_t instance, int32_t value, uint32_t unit, uint32_t prec)
 {
+  bool available = false;
+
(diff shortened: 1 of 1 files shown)


projectkk2glider:
I replaced previous commit with the new one!

bsongis:
I was completely lost for those merge, perhaps we could only open pull requests for 'master' and port them to 'next' without another review, it will simplify things, at least for me :)

projectkk2glider:
Yes I agree. I even think you perhaps created the `next` branch prematurely. Until we develop some new feature for 2.2 i see no reason for it to exist. Current situation just produces more work - duplicate commits.

PR #2571, opened 2015-07-26
Fixes #2533: menu was wrong if all items on the end were hidden

### LINKED ISSUE #2533: Sensor menu has two lines with "name"
One can find sensor name at the top and at the bottom of the sensor setup menu. The bottom one is I think a bug.

projectkk2glider:
This fixes problem on the sensor menu.  Closes #2533.

**But other menus** uses the same kind of code and could have the same issue.


bsongis:
Ah yes. Perhaps we should write a macro for this code ...

projectkk2glider:
I will leave this one to you.

bsongis:
ok later :)

PR #2593, opened 2015-07-28
Added support for Trainer signal alarm (issue #2579)

### LINKED ISSUE #2579: Checking trainer input validity
Gentlemen,

As of now, I've seen that the validity of the trainer input signal is tested properly in the code (ppmInValid is set in such case), and the trainer mode is deactivated properly, thus returning the control to the instructor. However, the issue tracking this feature (#36) was referring to the possibility to trigger some sound in such a case.

The only thing I've seen so far to implement this feature was to define a couple of logical switches checking for TR1 to TR2 being 0. Apart from the fact that this event is not really singular (could happen by accident, just leave the sticks in the middle ;-)), it is probably not really reliable.

So now my question: is there any way to directly check in a valid trainer input is available or not? I have no idea what the best way to implement that would be, btw. what the correct and logical way would be. Maybe a logical switch condition? I could implement it myself if you guys just point me into the right direction.

raphaelcoeffic:
Here a very simplistic implementation to add an alarm signal when the trainer is lost, or after it gets recovered. It does not take into account whether or not the trainer mode and/or function is enabled or not. Please advise if anything need to be done differently or be taken into account.

Closes  #2579

### DIFF
--- a/radio/util/tts.py
+++ b/radio/util/tts.py
@@ -1,4 +1,4 @@
-﻿#!/usr/bin/env python
+#!/usr/bin/env python
 # -*- coding: utf-8 -*-
 
 
@@ -186,6 +186,8 @@ def ttsEn():
                     (u"radio antenna defective", "swr_red", NO_ALTERNATE),
                     (u"telemetry lost", "telemko", NO_ALTERNATE),
                     (u"telemetry recovered", "telemok", NO_ALTERNATE),
+                    (u"trainer signal lost", "trainko", NO_ALTERNATE),
+                    (u"trainer signal recovered", "trainok", NO_ALTERNATE),
                    ]:
         systemSounds.append((s, filename(f, a)))
     for i, (s, f) in enumerate([
@@ -282,6 +284,8 @@ def ttsFr():
                     (u"Antenne défectueuse", "swr_red", NO_ALTERNATE),
                     (u"Plus de télémétrie", "telemko", NO_ALTERNATE),
                     (u"Télémétrie retrouvée", "telemok", NO_ALTERNATE),
+                    (u"Signal écolage perdu", "trainko", NO_ALTERNATE),
+                    (u"Signal écolage retrouvé", "trainok", NO_ALTERNATE),
                    ]:
         systemSounds.append((s, filename(f, a)))
     for i, (s, f) in enumerate([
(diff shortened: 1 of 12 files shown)


radio/util/tts.py
```diff
@@ -282,6 +284,8 @@ def ttsFr():
                     (u"Antenne défectueuse", "swr_red", NO_ALTERNATE),
                     (u"Plus de télémétrie", "telemko", NO_ALTERNATE),
                     (u"Télémétrie retrouvée", "telemok", NO_ALTERNATE),
+                    (u"Signal écolage perdu", "trainko", NO_ALTERNATE),
```
bsongis: Are you french?

raphaelcoeffic: Oui :-) And living in Germany for 15 years now...

bsongis:
On first glance it seems perfect to me. I will wait others feedback. Did you test your mod on the Tx?

raphaelcoeffic:
Only on my X9D+

projectkk2glider:
@raphaelcoeffic I've taken the liberty to add `Closes #2579` into your opening comment. This way this pull is linked to the original issue and when merged it will automatically close it.

raphaelcoeffic:
Thx! I'm quite yet used to github, as I must confess.

projectkk2glider:
One general comment, that is not related to work in this issue, but it might as well be fixed here.

The name `ppmInValid` (and derived names) can be read (if one is not careful) as **ppmInvalid**. It think the name is otherwise very descriptive, but it can be interpreted wrong. I propose we change it, my proposal:  ppmInputValid

raphaelcoeffic:
In fact, it's more of a timer. Maybe something like 'ppmInputTimer', or 'ppmInputValidityTimer' if you like long names.

bsongis:
Yes I prefer the long name. Perhaps we could avoid the ppmInState with a special value of this timer?

bsongis:
Did you have time to do it? I would like to release the 2.1.1 during the week-end. No issue if we delay it to 2.1.2 of course!

raphaelcoeffic:
Yep, will be done shortly.

raphaelcoeffic:
Not sure this was the right move, but I pushed a new version to opentx:raphaelcoeffic/issue2579. Now I don't know whether I should delete this pull request and create a new one, or just repost to my fork, so that it gets tracked on the same request. Any advise?

projectkk2glider:
Yes, create new pull and close this one. Just mention this pull in the new one, so it is easier to find related discussion later.

PR #2594, opened 2015-07-28
Added support for SBUS failsafe bit

raphaelcoeffic:
Proposed in issue #2579. When the receiver (at least the X8R & X4R do) goes into failsafe mode, it continues sending SBUS frames with the failsafe bit set. I think we should reject those frames as being an invalid trainer signal, so that the control can return to the instructor.

### DIFF
--- a/radio/src/sbus.cpp
+++ b/radio/src/sbus.cpp
@@ -36,6 +36,11 @@
 
 #include "opentx.h"
 
+#define SBUS_FLAGS_IDX 23
+
+#define SBUS_FRAMELOST_BIT   2
+#define SBUS_FAILSAFE_BIT    3
+
 Fifo<32> sbusFifo;
 uint8_t SbusFrame[28] ;
 uint16_t SbusTimer ;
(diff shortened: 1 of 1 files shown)


raphaelcoeffic:
done :-)

bsongis:
I gave you commit rights. It's interesting to open branches inside our project (username/branchname) because you will have non regression tests executed before the merge is done!

raphaelcoeffic:
Another small question here: is there actually any reason for not doing more of the SBUS processing in the UART ISR? I'm currently implementing iBus (used by a new flysky rx of mine) and would love to get rid of the active polling in the main loop (for SBUS & iBus).

bsongis:
We always try to keep the interrupts as small as possible!

raphaelcoeffic:
ok, then I'll try to find something better to only poll what is currently needed for the trainer input (depending on trainer mode).

kilrah:
I have just tested this with an X4R-SB connected to the module pins.
The Trainer function gets successfully disabled (and the telemetry lost alarm played) when I disconnect the X4R, but does NOT when the slave radio is turned off (= sbus failsafe).

raphaelcoeffic:
Unfortunately, I can first look at it on Monday. I hope this does not block anything.

kilrah:
Fixed, you were just looking one byte too far. I also added a test of the frame lost bit as AFAIK one should consider the data to be potentially invalid in this case.

raphaelcoeffic:
Well, the flight controller guys keep the "lost frames", and trigger return-home on failsafe.

PR #2606, opened 2015-07-30
Fixes #2579: Added support for Trainer signal alarm

### LINKED ISSUE #2579: Checking trainer input validity

raphaelcoeffic:
- added alarm “Trainer signal lost” & “Trainer signal recovered”.
- moved PPM capture code to trainer_input.{h,cpp}

This is the second try (see earlier push request #2593).
(Fixes #2579)

### DIFF
--- a/companion/src/simulation/simulatorimport.h
+++ b/companion/src/simulation/simulatorimport.h
@@ -108,7 +108,7 @@ return main_thread_error;
 
 #ifdef SETTRAINER_IMPORT
 #undef SETTRAINER_IMPORT
-  ppmInValid = 100;
-  g_ppmIns[inputNumber] = LIMIT< ::int16_t>(-512, value, 512);
+  ppmInputValidityTimer = 100;
+  ppmInput[inputNumber] = LIMIT< ::int16_t>(-512, value, 512);
 #endif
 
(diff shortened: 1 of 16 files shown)


projectkk2glider:
Good, now you can see that the build of yor branch failed. Follow  details link and look at console output to find the reason.

raphaelcoeffic:
quite nice indeed!

projectkk2glider:
Jep, Jenkins is there to help us.

raphaelcoeffic:
By the way, I tested the trainer input on heartbeat port (CPPM), but not on the other interfaces, and only on the X9D+. I just don't have any trainer jack here...

raphaelcoeffic:
Any chance we can still get this one into 2.1.1?

bsongis:
Excellent work, thanks!

PR #2622, opened 2015-08-02
Fixes #2607: Added support for 0.01V resolution to D series VFAS ID 0x39

### LINKED ISSUE #2607: VFAS ID 0x39 Selectable precision
ref #2531
Can the interpretation of this value be based on the selected precision for the sensor (default; 0.1V)?
This would give support for both 0.1V and 0.01V sensors.

projectkk2glider:
Closes #2607 

Tested in simu with the help of gdb (manually calling function `call processHubPacket(0x39, 14087)`) and it seems to work.


bsongis:
Thanks! If you have time would you add one or two non regression tests?

flarssen:
Version 2.1.1N403:
 I can confirm that when sending ID 0x39 in 0.01V, VFAS changed from 98.40V to 9,84V after I added 2000 to the data. Dropped in 10mV intervals as I advanced throttle.
Thanks!

projectkk2glider:
>If you have time would you add one or two non regression tests?

Done in https://github.com/opentx/opentx/commit/5d8c86923b535292ecfcb37667dcfcb1f18123c6

PR #2653, opened 2015-08-06
Fixes #2638: header line for logs fixed for 9XE

### LINKED ISSUE #2638: X9E logs more values than header fields
Using latest versions of both companion and firmware as downloaded by companion.

Companion cannot open the log files the TX has created, no problem opening in Excel and the data looks good and "clean", error message in Companion is "selected file contains X invalid lines out of X total lines". Same result if tge log file is transferred to desktop first then opened.

projectkk2glider:
Closes #2638

Not pretty but it works (tested in simu).


bsongis:
Yes, thanks! Ideally the switches string should also be dynamic, later ...

kilrah:
>Yes, thanks! Ideally the switches string should also be dynamic, later ...
Was about to say it. Later indeed, will open an issue :)

PR #2654, opened 2015-08-06
Fixes #2652:  Add missing presets to frskyDSetDefault()

### LINKED ISSUE #2652: Add missing presets to frskyDSetDefault()
The `frskyDSetDefault()` is missing some presets that are defined in `frskySportSetDefault()`. Related conversation in https://github.com/opentx/opentx/issues/2646#issuecomment-128309838

projectkk2glider:
Closes #2652

### DIFF
--- a/radio/src/telemetry/frsky_sport.cpp
+++ b/radio/src/telemetry/frsky_sport.cpp
@@ -267,7 +267,6 @@ void frskySportSetDefault(int index, uint16_t id, uint8_t instance)
     uint8_t prec = min<uint8_t>(2, sensor->prec);
     telemetrySensor.init(sensor->name, unit, prec);
     if (id == RSSI_ID) {
-      telemetrySensor.filter = 1;
       telemetrySensor.logs = true;
     }
     else if (id >= ADC1_ID && id <= BATT_ID) {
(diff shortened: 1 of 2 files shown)


radio/src/telemetry/frsky_sport.cpp
```diff
@@ -267,6 +267,7 @@ void frskySportSetDefault(int index, uint16_t id, uint8_t instance)
     uint8_t prec = min<uint8_t>(2, sensor->prec);
     telemetrySensor.init(sensor->name, unit, prec);
     if (id == RSSI_ID) {
+      telemetrySensor.filter = 1;
```
kilrah: I remember filter was disabled by default on purpose for smart port telemetry. We receive too few samples in sport and it was introducing too much lag.

projectkk2glider: I wondered about that, will fix immediately.

bsongis:
Yes, thanks, I forgot them.
I think that the conversion m=>ft according to the imperial parameter should not be done here, but in telemetry.cpp instead (it's common to all telemetry protocols)

kilrah:
Correct!

PR #2666, opened 2015-08-07
A few lcd_outdezNAtt() fixes

xmxwx:
There were a few bugs affecting printing numbers with non-STDSIZE fonts:
- TINSIZE on non-Taranis ARM was broken
- some bugs caused by improper calculation of decimal separator and minus sign widths
- printing decimal separator for SMLSIZE was incorrect

Additionally:
- checkPartialScreenshot() has been introduced which allows to reduce the number of reference images (and this number has been, in fact, reduced for already existing LCD tests)
- check*Screenshot() calls now work differently, they compare the images pixel by pixel instead of comparing the contents of compressed files
- in case of LCD test failure in addition to the LCD buffer contents, the check*Screenshot() routines produce an image showing differences between the obtained and the reference image


radio/src/gui/9X/lcd.cpp
```diff
@@ -181,7 +181,7 @@ void lcd_putcAtt(coord_t x, coord_t y, const unsigned char c, LcdFlags flags)
     lcdPutPattern(x, y, q, 5, 6, flags);
   }
   else if (fontsize == TINSIZE) {
-    q = &font_3x5[((uint16_t)c-0x2d)*3];
+    q = &font_3x5[((uint16_t)c-0x20)*3];
```
bsongis: I am almost sure I fixed it already last week, am I wrong?

radio/src/gui/9X/lcd.cpp
```diff
@@ -402,19 +402,40 @@ void lcd_outdezNAtt(coord_t x, coord_t y, lcdint_t val, LcdFlags flags, uint8_t
...
-      if (mode > 0)
-        x += 2;
-    }
 #if defined(BOLD_FONT) && !defined(CPUM64) || defined(EXTSTD)
     if (flags & BOLD) fw += 1;
 #endif
   }
 
   if (flags & LEFT) {
+    if (mode > 0) {
```
bsongis: I have to check the flash usage on 9X (AVR, 64k flash)

bsongis:
This is really good job! I am only afraid of 9x flash usage (AVR, 64k). If it consumes more than before, I propose we have a different function for both platforms.

radio/src/gui/9X/lcd.cpp
kilrah: He based his branch on next instead of master so if it's fixed he wouldn't have noticed it... should rebase on master

xmxwx:
> He based his branch on next instead of master so if it's fixed he wouldn't have noticed it

Master doesn't have TINSIZE font layout changed so the fix is not needed (actually, it would break the display again).

> I am almost sure I fixed it already last week, am I wrong?

Maybe what you remember is the commit 059ea49 (mentioned in the git message) which introduced this change for Taranis.

Thanks for the review! I will also check the flash usage on AVR 64k and report it back here.

bsongis:
It was this commit:
https://github.com/opentx/opentx/commit/5c4b12dfc220c6a565598ee7d64efdbcaf0f8b4f

PR #2682, opened 2015-08-09
Fixes #2598: added USB serial port 

### LINKED ISSUE #2598: Add USB serial interface
Related issues #2008 and #2595.

I will take a stab at implementing this now.

projectkk2glider:
Closes #2598 

Use `USB=SERIAL` to enable. If `DEBUG=YES` then debug output also goes to the USB serial port.

The CLI is not implemented, the current example only returns all received characters with a `+` character injected between each one. The buffer sizes should also be tuned.

The USB peripheral is now initialized and ready to go from the start. The reason is stated in the comments. There is now a slight chance that this could affect the radio operation in a negative way even with the USB cable unconnected. But I think it will be OK.


bsongis:
I have the cli working in a local branch. I will add it as soon as possible.

radio/src/targets/taranis/uart3_driver.cpp
```diff
@@ -109,6 +109,10 @@ void debugPutc(const char c)
 {
   if (uart3Mode == UART_MODE_DEBUG) {
     uart3Putc(c);
+#if defined(USB_SERIAL)
```
bsongis: Why inside the bracket? It shouldn't depend on UART3 mode, no?

bsongis:
Are you ok if I add the cli in your branch before merging?

radio/src/targets/taranis/uart3_driver.cpp
projectkk2glider: It seemed like a good idea at the time.

projectkk2glider:
Sure, go ahead.

PR #2696, opened 2015-08-11
Bsongis/issue2595 cli

bsongis:
It works here, and the USB serial works perfectly as well (Debian Stretch)


projectkk2glider:
I am still missing a check for the "fullness" of USB TX buffer. You should stop printing chars from CLI thread and wait some time before continuing if TX buffer is full.

bsongis:
I think it could be inside the driver, not the cli!

bsongis:
exactly as it is done here:
void serial2Putc(const char c)

projectkk2glider:
I agree.

projectkk2glider:
Something is not right. This command stalls the CLI output:
```
print 0x20000fe6 200
```

I guess when the out buffer (512 bytes) is filled the check inside usbSerialPutc() goes into infinite loop.

projectkk2glider:
I've got it, fix coming soon...

PR #2691, opened 2015-08-10
Fixes #2623: several fixes:

projectkk2glider:
Disable menu for [..],
Refresh file list after delete,
Copy destination can be a directory,
Prevent copy/paste into the same directory,
Enabled renaming of directories and files without extension,
Removed delete menu option for directories,
Several simulator filesystem functions improved

Closes #2623 

Please test thoroughly before merging!

PR #2724, opened 2015-08-17
Fixes #2623: several fixes (ported from 2.1):

### LINKED ISSUE #2623: [Bug][OpenTX 2.0.17] No refesh after deleting a file
Hi folks,

From the SD CARD 2/8 menu when I'm deleting a file, it doesn't disappear from the list. I need to move to a parent folder then going back in the folder to see that it has been effectively removed.

projectkk2glider:
Disable menu for [..],
Refresh file list after delete,
Removed delete menu option for directories,
Several simulator filesystem functions improved

Closes #2623 

This is the the port of #2691 to the 2.0 branch. I urge you to test this thoroughly. There are some changes compared to the 2.1 solution, namely the `REFRESH_FILES()` is different: `m_posVert = 1` where on 2.1 the `m_posVert = 0` is used.

### DIFF
--- a/radio/src/gui/menu_general.cpp
+++ b/radio/src/gui/menu_general.cpp
@@ -984,12 +984,12 @@ void menuGeneralSdManager(uint8_t _event)
       else
 #endif
       {
-#if defined(CPUARM)
         uint8_t index = m_posVert-1-s_pgOfs;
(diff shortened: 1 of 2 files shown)


bsongis:
Thanks!

PR #2729, opened 2015-08-18
Fixes #2728: edge switch could trigger right after the logicalSwitche…

### LINKED ISSUE #2728: Edge {false boolean} [00:--]  --- 2.0   becomes true upon selecting a model
I created a switch that has to be true 2 seconds after leaving FM1

Edge FM1 [00:--] --- 2.0

That switch will put my model in FM4
FM4 triggers "Play Track launch"

When I choose my model it will go into "launch" (FM4) and play the track "launch"

I tested other input values instead of FM1 (ones that are for certain false) and they also make this switch become true for 2 seconds.
Further behaviour is as expected...

I'm on nightly and it does this on the radio and on the simulator

projectkk2glider:
…sReset() was called

Re #2728

Please test this fix. If this is accepted solution, I will also port it back to the 2.0 branch.


bsongis:
Thanks!

Issue #2615, opened 2015-08-01
Improve use of "ratio" parameter in telemetry sensors


kilrah:
"Ratio" currently has the same effect as the "range" parameter for analog sensors on pre-2.1, i.e. "how much the output should be when the input is 255".

It is still required for analog sensors nowadays (yet could be improved for S.port type that have a full scale of 330 instead of 255), but it unfortunately makes no sense at all for all other telemetry sources.

The use of that parameter needs to be differentiated between analog sensors (no change) and the others (a simple multiplying factor, with PREC2 so as to allow for pretty precise conversion factors both when >1 and <1.

frobert0:
Ok works, only the determination of the required value is difficult.

In time was 43 instead of 50

Set I was ordered 25 ????????????????

kilrah:
This issue is precisely about making that better, for now the number is "arbitrary".

projectkk2glider:
I think 2.2 would be a better target for this - it might require an EEPROM change.

PR #2735, opened 2015-08-19
Re #2615: ratio telemetry sensor parameter broken into multiplier and…

projectkk2glider:
… divisor, added rounding for divisions Re #2615.

Here is my take on the ratio subject.

The `ratio` parameter is now replaced by two values: 
- multiplier
- divisor

The `multiplier` is just that a number with a range of 1-16000. The `divisor` is a power of 10 divisor. Possible values are 1, 10, 100 and 1000. 

The new values are used as:

```
       out_value = in_value * multiplier / 10^divisor
```

Together both new parameters give following resulting multipliers:
- 1 - 16000 in steps of 1
- 0.1 - 1600 in steps of 0.1
- 0.01 - 160 in steps of 0.01
- 0.001 - 16 in steps of 0.001

I think this is big enough range and also precise enough. 

This new parameters are used also for A1 and A2 values, they give ample adjustment resolution if divisor 1000 is used. The default value for them is again set to accommodate the  1:4 divider in FrSky receivers.

Tell me if you like the way this is going or perhaps we should take some other route. Code is not yet cleaned up, this is just a proof of concept. Do not merge yet!

### DIFF
--- a/companion/src/firmwares/opentx/simulator/opentxsimulator.cpp
+++ b/companion/src/firmwares/opentx/simulator/opentxsimulator.cpp
@@ -24,6 +24,7 @@
 #define COMPANION
 #define SIMU
 #define SIMU_EXCEPTIONS
+#define SIMU_TELEMETRY
 #define GUI
 #define SPLASH
 #define FRSKY
--- a/radio/src/gui/9X/menu_model_telemetry.cpp
+++ b/radio/src/gui/9X/menu_model_telemetry.cpp
@@ -267,7 +267,7 @@ bool isSensorAvailable(int sensor)
 #define SENSOR_PREC_ROWS       (sensor->isConfigurable() ? (uint8_t)0 : HIDDEN_ROW)
 #define SENSOR_PARAM1_ROWS     (sensor->unit >= UNIT_FIRST_VIRTUAL ? HIDDEN_ROW : (uint8_t)0)
 #define SENSOR_PARAM2_ROWS     (sensor->unit == UNIT_RPMS || sensor->unit == UNIT_GPS || sensor->unit == UNIT_DATETIME || sensor->unit == UNIT_CELLS || (sensor->type==TELEM_TYPE_CALCULATED && (sensor->formula==TELEM_FORMULA_CONSUMPTION || sensor->formula==TELEM_FORMULA_TOTALIZE)) ? HIDDEN_ROW : (uint8_t)0)
-#define SENSOR_PARAM3_ROWS     (sensor->type == TELEM_TYPE_CALCULATED && sensor->formula < TELEM_FORMULA_MULTIPLY) ? (uint8_t)0 : HIDDEN_ROW
+#define SENSOR_PARAM3_ROWS     (sensor->type == TELEM_TYPE_CUSTOM || (sensor->type == TELEM_TYPE_CALCULATED && sensor->formula < TELEM_FORMULA_MULTIPLY)) ? (uint8_t)0 : HIDDEN_ROW
 #define SENSOR_PARAM4_ROWS     (sensor->type == TELEM_TYPE_CALCULATED && sensor->formula < TELEM_FORMULA_MULTIPLY) ? (uint8_t)0 : HIDDEN_ROW
 #define SENSOR_AUTOOFFSET_ROWS (sensor->isConfigurable() ? (uint8_t)0 : HIDDEN_ROW)
 #define SENSOR_FILTER_ROWS     (sensor->isConfigurable() ? (uint8_t)0 : HIDDEN_ROW)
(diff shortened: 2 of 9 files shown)


jtaylor2:
I'm not sure it would be understood by the majority of users (unless it is documented really well, with examples).  I had to read it several times before I understood it, if I actually do understand it.  For example on my self compiled 2.1.2

My current A1 4-1 offsets are 13.1 to 13.2 and I could use use a multiplier of 13.2 and a divisor of 1?
My current 3 cell A2's would be the same
My current 4 cell (6-1) A2's are currently 19.8 to 20.1 would be multiplier of 19,8 and divisor of 1?
(My offsets differ from the actual 3.3 times divider to correct for voltage divider resistor impression)  

Is this correct?  And if I wanted more precision or range I could change them accordingly.

Looking at the code I can't figure out what you are doing with the current users configuration.  Does  the change take the current offsets and make them the multiplier with a divisor of 1 automatically or would the user have to go in and fix them?  If the change is transparent to the existing settings, great.  But if it requires the user to fix the values I think it would be better done in 2.2 instead of 2.1.x.

Jim

projectkk2glider:
Maybe the names for the new parameters are not appropriate but both new parameters together form a `combined multiplier` that is applied to the incoming telemetry value:

So `combined_multiplier` = `multiplier / divisor` 

and `telemetry_value = combined_multiplier * input_value`.

The `multiplier` is an integer number with a range of 0 (disabled) to 16000. If you want to make it more precise the use `divider` that is different from 1. For example if you set divider to 1000, then the telemetry value gets multiplied by `multiplier/1000` giving you a possible range of 0.001 to 16.000 in steps of 0.001.

So in effect we have folowing `combined_multiplier` ranges with different resolutions:
* from 0.001 to 16 in steps of 0.001
* from 16 to 160 in steps of 0.01
* from 160 to 1600 in steps of 0.1
* from 1600 to 16000 in steps of 1

As far as A1 and A2 this has changed how they are handled completely. The old `ratio` parameter should be translated to the new `multiplier` and `divisor` values. The correct setting for the D series receivers with 1:4 divider would be: `multiplier` = 5176 and `divisor` = 1000 giving a `combined multiplier` of 5.176. 

How come the 5.176 is the correct combined_multiplier? The receiver send a value of 255 for a voltage of 3.3V. The default resistor divider for A1 is 1:4. If we want to store that value with a PREC2 (two decimal places) we have to multiply it by 100. So the formula is: `3.3 * 4* 100 / 255 = 5.176`

For Sport receivers the multiplier is different: the receiver sends 330 when it measures 3.3V at its input. If the 1:4 resistor divider is used again the combined_multiplier would be `3.3 * 4 * 100 / 330 = 4`. So I would set  `multiplier` = 4000 and `divisor` = 1000. This then gives you a possibility to fine tune (calibrate) the voltage reading with for example multiplier 4.0001 or 3.999 (very precise).

This pull does not have any conversions from the old `ratio` to the new parameters yet. Only if proposed solution is accepted we could add it. The Companion part is also not done yet.

projectkk2glider:
I am not completely satisfied with this solution for the very small `combined_multiplier` values. For example:
* smallest combined_multiplier can be 0.001
* the next one is 0.002

The first one is an equivalent of `input_value / 1000` and the second one is `input_value / 500`. So the step is quite huge, one can not for example divide input value by 700.

So here is an idea that could improve above problem. Lets make the `multiplier` signed. We loose one bit for sign so the available range becomes from -8000 to 8000. But now we have two ways of using it:
* if `multiplier` is positive, we calculate (as before) `combined_multiplier = multipler/divider`
* if `multiplier` is negative, we calculate inverted `combined_multiplier = divider/abs(multiplier)`

So this will give following possible ranges for `combined_multiplier`:
* 0 to 8 in steps of 0.001
* 8 to 80 in steps of 0.01
* 80 to 800 in steps of 0.1
* 800 to 8000 in steps of 1

and if negative `multiplier` is used we get `combined_divisor` which has the same ranges as above!

jtaylor2:
I had said offsets in my previous comment when I actually meant ratios, I have corrected it.

In any case thanks for the explanation about how A1, A2 will change.  I personally think that it is too complicated for the average user to understand.

Before Taranis people using DFT/DJT had trouble understanding the 4:1, 6:1 3.3v 255 analog ratios and figuring out what numbers to plug into the alarms..  It got better with OpenTX 2.0 when you could program the range and alarms directly.  You only had to know the max input to the analog ports and ratio to calculate the range (3.3 X 4 = 13.2) and you could program the alarms directly.   It got more confusing at 2.1 where the range now goes in ratio (13.2 not 4) and you have to program the alarms with logical switches.  Looking at your explanation of how to calculate the A1, A2, etc. values with this change I don't think the average user stands a chance of doing it correctly.  Perhaps a screen in Companion or a LUA wizard on the radio where you can just fill in the blanks to get the value might be the answer.

flarssen:
Will there be anything done regarding this for 2.1.x, like a possibility to do something simple as dividing data by 10 or 100?

bsongis:
Can't be done in 2.2, to be continued later in 3.0

PR #2751, opened 2015-08-23
Fixes #2635 - Not tested yet

### LINKED ISSUE #2635: Add haptic option for countdown timer
Is it possible to add in haptic options for the countdown timer? Ideally I would like the option to have the option to enable haptic as well as beeps/voice, or instead of them. This would really helpful when flying in a noisy environment. Many thanks!

bsongis:


bsongis:
Tested on 9XR-PRO

PR #2763, opened 2015-08-25
Fixes #2620. Return false if version is bigger than 216.

### LINKED ISSUE #2620: Companion 2.0 opens 2.1 eeproms but obviously with a corrupt result
Should give an invalid eeprom error, or even better say it's a 2.1 eeprom.

griebd:
Should fix #2620.
Please review as I'm getting to know the code now...
Only tested with Taranis eeprom.

### DIFF
--- a/companion/src/firmwares/opentx/opentxinterface.cpp
+++ b/companion/src/firmwares/opentx/opentxinterface.cpp
@@ -344,6 +344,13 @@ bool OpenTxEepromInterface::load(RadioData &radioData, const uint8_t *eeprom, in
     return false;
   }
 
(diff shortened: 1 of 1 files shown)


companion/src/firmwares/opentx/opentxinterface.cpp
```diff
@@ -344,6 +344,13 @@ bool OpenTxEepromInterface::load(RadioData &radioData, const uint8_t *eeprom, in
     return false;
   }
 
+  // we do not want to support OpenTX 2.1 ans later on Companion 2.0
```
bsongis: I think that this check should be in 
'''
bool OpenTxEepromInterface::checkVersion(unsigned int version)
'''

PR #2766, opened 2015-08-25
Do not accept eeprom(s) newer than 216 in companion 2.0

### LINKED ISSUE #2620: Companion 2.0 opens 2.1 eeproms but obviously with a corrupt result

griebd:
Changed accordingly @bsongis.
Fixes #2620.

I was mislead by "std::cout << " not open9x\n";". It should be OpenTX in place of open9x, no?!
By the way, in this part of the code there are a few cout(s), shouldn't they be enabled just for debug purposes? The way it is they are enable in release too.


bsongis:
Yes your comments are right, this code is a little bit old, don't hesitate
to apply the boyscout rule there!
Le 25 août 2015 2:34 PM, "Adriano Grieb" <notifications@github.com> a
écrit :

griebd:
Did the changes and noticed something else...
opentxinterface.cpp already #include <QMessageBox> (from what I can see it is there as a dirt way to show some messages, only two actually). I can then alert the user about trying to open a 2.1 eeprom in 2.0.
BUT there will be two alerts then, the new one about the version and then the old one about invalid file format.

Maybe we can just leave 2.0 as is with those changes already done and do some refactoring in 2.1 to be able to send messages back from the interface file to the GUI.

griebd:
By the way, by using #ifndef NDEBUG, the 'cout's are only removed if you use cmake -DCMAKE_BUILD_TYPE=Release. All other options enable the messages. Maybe that is not the solution for this environment (from what I could get so far the downloadable companion is build without specifying CMAKE_BUILD_TYPE).

kilrah:
> Maybe we can just leave 2.0 as is with those changes already done and do some refactoring in 2.1 to be able to send messages back from the interface file to the GUI.

Yeah. What I'd really like is a warning when a user opens a 2.0 eeprom in 2.1 saying it will be converted and he should "save as" instead of overwriting. Maybe even change the filename straight away.

griebd:
Two changes:
1) changed cmake to define DEBUG when CMAKE_BUILD_TYPE=Debug (case sensitive) and changed the #ifndef NDEBUG to #ifdef DEBUG. (deleted previous comment as it's invalid now)
2) added message through QMessageBox::warning when trying to open a 2.1 eeprom. BUT, because the way the code is there will be two equal messages. Should we let this way or only the original unable to load.

companion/src/firmwares/er9x/er9xinterface.cpp
```diff
@@ -91,7 +94,9 @@ inline void applyStickModeToModel(Er9xModelData & model, unsigned int mode)
 
 bool Er9xInterface::loadxml(RadioData &radioData, QDomDocument &doc)
 {
+  #ifdef DEBUG
```
bsongis: Why not qDebug() << ...
If I am not mistaken it won't display anything when in RELEASE mode, no?

griebd: because I thought there where no QT libs on those files... just tested and it works, I was mistaken... :wink:

griebd:
I will change to qDebug then, all else is ok as is? the double alert?
I will make a new clean branch if all is ok...

bsongis:
The double alert is not good, you are right, feel free to change this code,
otherwise I think it's good!
Le 26 août 2015 8:21 PM, "Adriano Grieb" <notifications@github.com> a
écrit :

griebd:
looks like I'm doing something wrong or you were wrong on this... :wink:
qDebug is executed and display messages in any mode!
anyway, I think it's good to have a DEBUG defined, no?

bsongis:
Isn't the QT_NO_DEBUG_OUTPUT working?
Le 27 août 2015 2:00 AM, "Adriano Grieb" <notifications@github.com> a
écrit :

projectkk2glider:
@griebd  qDebug() is disabled in release mode. But you can not just change the cmake command-line parameters in the existing directory. The changed cmake cmdline won;t change anything. You must start with the fresh (empty) directory if you want to use different cmake command line!

griebd:
So there is something strange here...
I know a new clean directory is needed, I have 3 for my testings at this moment:
- one with cmake -DCMAKE_BUILD_TYPE=Debug ../companion/src/
- second with cmake ../companion/src/
- and third with cmake -DCMAKE_BUILD_TYPE=Release ../companion/src/
qDebug() prints messages in all of then!!!!

griebd:
QT_NO_DEBUG_OUTPUT? is it added as a -D compile option? From my tests it's not defined in any mode...
anyway, it would be the same as NDEBUG, we would need to use the double negative (#ifndef NDEBUG)

PR #2770, opened 2015-08-25
Alert about duplicated receiver id on copy. #2187

### LINKED ISSUE #2187: "Copy model" should warn when using same receiver number
I'm using a DIY DSMX module. When I copy a model the receiver number is copied to the new model slot. But it's the same number as in the original, which will have confusing effects since (AFAICT) the module stores settings keyed by receiver number and will overwrite the original's settings when binding for the copy. It would be nice to show a warning or even offer to pick an unused number.

griebd:
Fixes #2187.

Again, review with care and point me in the right direction.

### DIFF
--- a/radio/src/gui/menu_model.cpp
+++ b/radio/src/gui/menu_model.cpp
@@ -439,7 +439,7 @@ void menuModelSelect(uint8_t event)
           s_editMode = 0;
           break;
         }
-        // no break;        
+        // no break;
 #endif
       case EVT_KEY_LONG(KEY_ENTER):
       case EVT_KEY_BREAK(KEY_ENTER):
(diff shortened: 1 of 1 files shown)


radio/src/gui/menu_model.cpp
```diff
@@ -479,6 +479,10 @@ void menuModelSelect(uint8_t event)
             eeDirty(EE_GENERAL);
           }
 
+#if defined(CPUARM)
+          checkModelIdUnique(g_eeGeneral.currModel);
```
bsongis: Does it mean you check that the receiver id is unique each time you switch from a model to another?
If this is what we want, I would prefer a check as we do for throttle / switch / ... warnings, no?

griebd: no, we check only when you copy and paste a model!

bsongis: yes you are right, it seems good to me

Issue #2681, opened 2015-08-09
Test the X9E watchdog reset


projectkk2glider:
Have we tested what happens with the 9XE power when the watchdog reset happens? Does it survive the reboot and remains powered up? Does the value of additional capacitor from the RTC issue #2626 affect this in any way?

bsongis:
Yes I tested it on mine, but I don't have the capa, even no 100uF!

projectkk2glider:
Here is a special OpenTX build for 9XE that includes a test for a watchdog reset and instructions on how to use it:

**WARNING** this version of OpenTX is for **testing only**. **Do not fly with it**. Provided bin file is for Taranis 9XE only. I will not work on other Taranis radios!

First **make sure you have a backup of your current radio firmware**. You will need it to revert back to your existing setup after the test is done. You should also have a **backup of your current models and settings** (EEPROM) in case things go south.

Steps to test:
* make backups of your current firmware and settings
* flash the test firmware (link below)
* do the test (instructions below)
* report back the result of the test (here)
* flash back your backed-up firmware and if neeeded restore models and settings back.


The provided bin file is in English language and includes the default options. It is build from a pre-release version of 2.1.2 (branch https://github.com/opentx/opentx/tree/projectkk2glider/issue_2681_watchdog_test)

Watchdog test instructions:
* start the radio normally and go to the main screen
* long press `ENT` key and select option `Statistics`
* scroll to the screen `DEBUG`
* long press `MENU` key, the warning will apear.
  * pressing `EXIT` key will abort the test.
  * pressing `ENT` key will **preform the test**. 

Result of the test:

The radio should reset in about one second and then turn or again and resume the operation without any switch/pots/throttle warnings. The sound will also not work. There should be an asterisk icon beside the clock indicating that the radio was reset by the watchdog. After the normal power off, power on sequence the radio should work normally (warnings and sounds back, asterisk gone).

If the radio is powered off by the test, the the test **FAILED**.

Please report back the result of the test and include:
* result
* any mods done to the 9XE radio (like the C100 cap for RTC, etc..)

Here is the bin file for the testing: http://downloads-21.open-tx.org/test/opentx-9xe-2.1.2-watchdog-test.bin

lichtl:
Hi,

just performed the testing with your provided bin file. (I'm uploading a video to youtube and share the link later)

X9E000017A01
E02 BE MS
17/07/2015
Stock Radio (no RTC Mod or any changes)

Test: **FAILED**

[...]

PR #2772, opened 2015-08-25
Projectkk2glider/issue 2681 watchdog fix

projectkk2glider:
Phew, this is a result of two days of constant compile/flash/test cycling.

We have two fixes:
* first one is for 9XE watchdog issue #2681. This was tested on three different gcc versions and all work. But also do the test yourself (use `WATCHDOG_TEST=YES` compile option).
* second one is a power off sequence refactoring. The RTC reset prevention fix from Mike is also enabled for 9XE. It works on my 9XE board (the power off/on part). I have not tested the RTC function!

### DIFF
--- a/radio/src/Makefile
+++ b/radio/src/Makefile
@@ -361,6 +361,10 @@ SIMU_DISKIO = NO
 # Values = NO, YES
 SIMU_AUDIO = NO
 
+# Enable Watchdog test (DANGEROUS)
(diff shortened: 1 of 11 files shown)


bsongis:
Thanks!
I will test it as soon as I am back and release the next day.
Meanwhile it would be nice to have as much feedback as we can!

PR #2775, opened 2015-08-25
Fix #2774 (out of bounds access to trim array when the source of an input is not a stick)

### LINKED ISSUE #2774: Output jitter - visable on mixer screen and RX output
I am experiencing quite a lot of jitter on the outputs from the RX. On closer inspection the gitter is visible on the channels and mixer output screen as a bluring of the last digit. This is the case even on channels which are sourced from a switch if that switch is then applied through a curve.

I use various high channels as mix channels and when they are applied to the RX output channels via curves I then get the jitter. I would speculate that the problem is related to the output from the curves.

Running 2.1 on X9E and 2.1.1 on X9D and both are the same. Down-grading the X9D to 2.0.X removes the problem. 

Although the jitter is only slight it manifests itself in digital servos in the form of buzzing and chatter which was not present in previous versions of the firmware. Measured using a logic analyser the output gitters by around 1~2 microseconds.

kilrah:
Looks good to me and it didn't fail tests, but I'd still appreciate a review...

### DIFF
--- a/radio/src/mixer.cpp
+++ b/radio/src/mixer.cpp
@@ -530,8 +530,13 @@ int getSourceTrimValue(int source, int value=0)
 {
   if (source >= MIXSRC_Rud && source <= MIXSRC_Ail)
     return getStickTrimValue(source - MIXSRC_Rud, value);
-  else if (source >= MIXSRC_FIRST_INPUT && source <= MIXSRC_LAST_INPUT)
-    return getStickTrimValue(virtualInputsTrims[source - MIXSRC_FIRST_INPUT], value);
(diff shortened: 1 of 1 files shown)


projectkk2glider:
I've looked over the code. It solves the problem. 

I also have on suggestion: perhaps a better location for the negative source check would be inside the `getStickTrimValue()` function.

bsongis:
True, also the parameter 'value' could be called 'stickValue' (my mistake),
otherwise it's difficult to understand what it is without the whole code
Le 26 août 2015 10:21 PM, "Damjan Adamic" <notifications@github.com> a
écrit :

Issue #2778, opened 2015-08-26
Refactoring between eeprom interface and GUI, so we can have more info on error messages.


griebd:
From #2766.
Quoting @kilrah:
> What I'd really like is a warning when a user opens a 2.0 eeprom in 2.1 saying it will be converted
> and he should "save as" instead of overwriting. Maybe even change the filename straight away.

griebd:
This one is for 2.1.X. It's the inverse case, when using companion 2.1 and opening a eeprom/backup from 2.0. It is referenced from the commit because the solution will be based on the same idea.

PR #2789, opened 2015-08-27
for review: proposed solution for #2620 and #2778

griebd:
after #2763 and #2766, here it's again... :wink:
I will focus on the two issues at hand, the part of the cout's I'll do later, after finishing this one.

It still needs a lot of changes on the existing code but the new part is done! For a taranis eeprom it's working already.

What do you think of this solution?

### DIFF
--- a/companion/src/eeprominterface.cpp
+++ b/companion/src/eeprominterface.cpp
@@ -1364,14 +1364,67 @@ void UnregisterFirmwares()
   }
 }
 
-bool LoadEeprom(RadioData &radioData, const uint8_t *eeprom, const int size)
+void ShowEepromErrors(QWidget *parent, QString title, QString mainMessage, unsigned long errorsFound)
 {
-  foreach(EEPROMInterface *eepromInterface, eepromInterfaces) {
-    if (eepromInterface->load(radioData, eeprom, size))
-      return true;
+  std::bitset<NUM_ERRORS> errors(errorsFound);
+  QString errorsList(QT_TRANSLATE_NOOP("EepromInterface", "Possible causes for this:\n"));
+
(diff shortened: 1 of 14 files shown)


companion/src/eeprominterface.cpp
```diff
@@ -1357,21 +1357,74 @@ QList<FirmwareInterface *> firmwares;
...
-void UnregisterFirmwares() 
+void UnregisterFirmwares()
 {
   foreach (FirmwareInterface * f, firmwares) {
     delete f;
   }
 }
 
-bool LoadEeprom(RadioData &radioData, const uint8_t *eeprom, const int size)
+void ShowEepromErrors(QWidget *parent, QString title, QString mainMessage, unsigned long errorsFound)
```
bsongis: const QString & here!

companion/src/eeprominterface.cpp
```diff
@@ -1357,21 +1357,74 @@ QList<FirmwareInterface *> firmwares;
...
+{
+  std::bitset<NUM_ERRORS> errors(errorsFound);
+  QString errorsList(QT_TRANSLATE_NOOP("EepromInterface", "Possible causes for this:\n"));
+
+  if (errors.test(UNSUPPORTED_NEWER_VERSION)) { errorsList += QT_TRANSLATE_NOOP("EepromInterface", "\n- Eeprom is from a newer version of OpenTX"); }
```
bsongis: Instead of the \n in all strings, I would see a QStringList with join("\n"), no?

bsongis:
Why not, I will merge the pull request and you will fix the little things after, ok? It will avoid merge problems ...

griebd:
ok, working to finish it then...

bsongis:
Is it possible to release in this state? Or do you prefer we wait some more days?

griebd:
I think there will be problems as it is... I didn't finish it yet because a friend of my got shot at work a few days ago, he is recovering now and I'm coming back to normal life! I should be able to finish it this week, I think, so, wait a few days, please.

kilrah:
Ouch!

PR #2801, opened 2015-08-28
Fixes #2790: power on/off lop if big VCC capcitor was used

### LINKED ISSUE #2790: 9XE power on/off loop possibility
The recent commit 77d3eae4d608f4a65fee786d4f31771daec36452 from the pull #2772 which was done to help battle the RTC reset issue #2626 can/is causing problems if the additional capacitor on VCC line is too big. After the power off, the radio will go into a cycle of power on - power off (couple of times a second).

Let me explain what is happening:
* The radio is on
* The user presses the power button and holds it enough time to trigger the power down.
* the `boardOff()` function is called
* he MCU_PWR_ON output is driven low.
* while the power button is still pressed the watchdog is periodically reset, when the power button is finally released the code continues.
* in order to fight the RTC reset issue the radio now enters the low power mode (a recommendation/solution by Mike). In this mode the watchdog is not poked (reset)!
* now comes the gist of the problem:
  * the radio by now is without battery power, the FET connecting the battery to the switching power supply is OFF.
  * the VCC is falling, the rate of fall depends on the radio consumption and the value of additional capacitor.
  * if the **rate of VCC fall** too smal, then the watchdog resets the radio **before** the VCC falls below the brownout reset value.
  * this reset then restarts the radio firmware and the first thing after the reset is the turning on of the MCU_PWR_ON output which then connects the battery to the power supply. The radio power is restored.
  * then later radio finds out that power switch is not pressed and starts to turn of by calling `boardOff()`. 
  * the cycle repeats.
 
This issue is not happening on stock 9XE development board, but is shows itself if I solder 1000uF capacitor to the VCC line. I haven't tried other values. FrSky plans to use 470uF. The tolerance for electrolytic caps is huge (like 20%). So some radio could be affected and some not.

There are two possible solutions:
* to (again) disable the part of the code that puts the CPU in low power mode.
* to devise a system to poke the watchdog even in low power mode.

projectkk2glider:
Solves #2790.

### DIFF
--- a/radio/src/loadboot.cpp
+++ b/radio/src/loadboot.cpp
@@ -100,13 +100,19 @@ void _bootStart()
   __ASM volatile ("nop");
   __ASM volatile ("nop");
 
-  // Turn soft power ON now
-  GPIOD->BSRRL = 1;                                  // set PWR_GPIO_PIN_ON pin to 1
-  GPIOD->MODER = (GPIOD->MODER & 0xFFFFFFFC) | 1;    // General purpose output mode
+  // Turn soft power ON now, but only if we got started because of the watchdog
+  // or software reset. If the radio was started by user pressing the power button
+  // then that button is providing power and we don't need to enable it here.
+  //
+  // If we were to turn it on here indiscriminately, then the radio can go into the 
+  // power on/off loop after being powered off by the user. (issue #2790)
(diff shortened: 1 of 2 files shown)


bsongis:
Thanks! I will be back to home tomorrow, and I will be able to test myself.

Issue #2780, opened 2015-08-26
Lua getValue() not working for special telemetry values


projectkk2glider:
The #2773 reminded me that we are missing a way for Lua to get the values of all *special* telemetry sensors. That pull fixes the case for GPS coordinates. But other sensors types like cells and datetime also need this special handling. 

Also related to the request to have all individual cell voltages avalilable in Lua #1799

dsbeach:
I might be in a position to take this on.  Is there a list of 'special'
sensors, or should I just find all the enumeration values for
TelemetrySensor.unit?

projectkk2glider:
>should I just find all the enumeration values for TelemetrySensor.unit?

Yes that is the right path. I would also like to change the flow of `luaGetValueAndPush()`: there is no need to call `getValue(src)` for these special telemetry values. It might as well be done along with this issue.

dsbeach:
Should be closed by #2826

PR #2826, opened 2015-09-02
Issue2780 - getValue() for special telemetry fields

dsbeach:
Changes to address issue #2780.  Tested on Taranis hardware using two different GPS units and 2-cell and 3-cell lipos with FrSky Lipo voltage sensor.

Note: this change also includes modification of getValue for UNIT_GPS to use table members 'lat' and 'lon' (not capitalized) following the convention of the luaGetDateTime() function.

Due to problems related to variable scope, I opted to comment the initial call to getValue(src) indicating that it is ignored for special telemetry values rather than restructure the use of the variable 'value'.

David

### DIFF
--- a/radio/src/lua_api.cpp
+++ b/radio/src/lua_api.cpp
@@ -182,7 +182,7 @@ static void luaGetValueAndPush(int src)
   if (src >= MIXSRC_FIRST_TELEM && src <= MIXSRC_LAST_TELEM) {
     src = (src-MIXSRC_FIRST_TELEM) / 3;
     // telemetry values
-    if (TELEMETRY_STREAMING() /* && telemetryItems[src].isAvailable() */) {
+    if (TELEMETRY_STREAMING() && telemetryItems[src].isAvailable()) {
       TelemetrySensor & telemetrySensor = g_model.telemetrySensors[src];
       switch (telemetrySensor.unit) {
         case UNIT_GPS:
--- a/radio/src/lua_api.cpp
+++ b/radio/src/lua_api.cpp
@@ -119,7 +119,7 @@ static int luaGetTime(lua_State *L)
 static void luaPushDateTime(lua_State *L, uint32_t year, uint32_t mon, uint32_t day,
                             uint32_t hour, uint32_t min, uint32_t sec)
 {
-  lua_newtable(L);
+  lua_createtable(L, 0, 6);
   lua_pushtableinteger(L, "year", year);
   lua_pushtableinteger(L, "mon", mon);
   lua_pushtableinteger(L, "day", day);
@@ -150,7 +150,7 @@ static void luaPushLatLon(TelemetrySensor & telemetrySensor, TelemetryItem & tel
   lon = gpsLon / 1000000.0;
   if (telemetryItem.gps.longitudeEW == 'W') lon = -lon;
 
-  lua_newtable(L);
+  lua_createtable(L, 0, 2);
   lua_pushtablenumber(L, "lat", lat);
   lua_pushtablenumber(L, "lon", lon);
 }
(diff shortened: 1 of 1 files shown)


radio/src/lua_api.cpp
```diff
@@ -144,31 +150,57 @@ static void luaPushLatLon(TelemetrySensor & telemetrySensor, TelemetryItem & tel
...
+  luaPushDateTime(L, telemetryItem.datetime.year + 2000, telemetryItem.datetime.month, telemetryItem.datetime.day,
+                  telemetryItem.datetime.hour, telemetryItem.datetime.min, telemetryItem.datetime.sec);
+}
+
+static void luaPushCells(TelemetrySensor & telemetrySensor, TelemetryItem & telemetryItem)
+{
+  if (telemetryItem.cells.count == 0)
+    lua_pushinteger(L, (int)0); // returns zero if no cells
+  else {
+    lua_createtable(L, 0 /*telemetryItem.cells.count*/, 0);
```
projectkk2glider: Why did you comment out the `telemetryItem.cells.count`, it should help with memory allocation.

projectkk2glider:
Looks good. I have not tested this.

radio/src/lua_api.cpp
dsbeach: The change was made after my initial testing when I noticed that the convention for this module seemed to me using the lua_newtable() macro (which expands to lua_createtable(L, 0, 0)).

There are 13 references to lua_newtable() in this file.  Should they all be changed to explicit allocations?

David

projectkk2glider: >There are 13 references to lua_newtable() in this file. Should they all be changed to explicit allocations?

Well probably no, but when you explicitly call lua_createtable()  you can optimize the call by providing end table size. Its not a big issue in any case.

dsbeach:
Now using pre-allocated table space via lua_createtable() for GPS, DATETIME, and CELLS.

bsongis:
Thanks!

galosre:
Maybe not relevant to this thread but for 4 and 6 cells I think there is an improvement but it not completely fixed as for me the value now between 4.01 and 4.09 still do not play correctly it omitted the point and the middle digit I use this script modified by a Swiss guy (Sacre100) but before it use to do that even for all value above 4.00v up to 4.19 anyway below 3.99 everything is correct

    --[[Quick and dirty two decimal place voice
    announcement for Cels or other telemetry cell
    voltages that are reported at 0.01V resolution

    Put the script in SD/SCRIPTS/FUNCTIONS folder.
    Specify the input channel name and logical or
    physical switch that triggers script below.
    Then set up a SF to play this script when  
    mySwitch is activated.]]

    myVoltage="Cels"

    local delay_val = 500 --minimum time between announcements (x10ms)
    local wake_time = 0

    local function sum_cels(cels)
      local val = 0
      for i, v in pairs(cels) do
        val = val + v
      end
      return val
    end

    local function announce(voltage)
      local cell = getValue(voltage)
      if "table" == type(cell) then
        cell = sum_cels(cell)
      end
      if cell and cell ~= 0 then
        playNumber(math.floor(cell * 10), 0 ,PREC1)
        if math.floor(cell * 10 % 10) == 0 then
          playFile("SYSTEM/0180.wav")
        end
        playNumber(cell * 100 % 10, 1)
      else
        print("Le capteur " .. voltage .. " n'existe pas")
      end
      wake_time = getTime() + delay_val
    end

    local function run(event)
      if getTime() >= wake_time then
        announce(myVoltage)
      end
    end

    return {run = run}



whole discussion below

 https://translate.google.com/translate?sl=fr&tl=en&js=y&prev=_t&hl=en&ie=UTF-8&u=http%3A%2F%2Ffrskytaranis.forumactif.org%2Ft3984-nouveau-ici-du-bout-du-monde&edit-text=

3djc:
You need to discuss that with the script author./maintainer

kilrah:
That script hardcodes "SYSTEM/0180.wav" as being "point zero". That's correct for the French voice pack on 2.1 for which this "quick and dirty" script as labelled was intended for, but it doesn't handle the fact that other languages and other versions use different numbering.
So modify the script to your setup.

galosre:
I knew it was not really relevant but thousand times thank you @ kilrah  for putting on the right way in order for me to be able to correct this error.
je ne le ferai plus!

PR #2827, opened 2015-09-02
fix for issue 2758 lua support for latitude and longitude'

dsbeach:
This has been implemented in branch master.  Do I need to generate the pull request for branch next as well?

David


bsongis:
No! 'next' will start from 'master' when 'master' will be more stable (and we will need an eeprom change)

PR #2828, opened 2015-09-03
Tmrttmrt/fixes#2750  mega2560 st7920 mixer freq too slow

tmrttmrt:
Now against the master branch.


radio/src/main_avr.cpp
```diff
@@ -44,6 +44,9 @@ void perMain()
 #if defined(SIMU)
   doMixerCalculations();
 #endif
+#if defined(LCD_ST7920)
+	uint8_t lcdstate=0;
```
bsongis: Indentation problem

radio/src/mixer.cpp
```diff
@@ -1097,7 +1098,7 @@ void evalMixes(uint8_t tick10ms)
     }
   }
 
-#if defined(PCBGRUVIN9X) && defined(DEBUG) && !defined(VOICE)
+#if (defined(PCBMEGA2560) || defined(PCBGRUVIN9X)) && defined(DEBUG) && !defined(VOICE)
```
bsongis: The option CPU2560 should simplify your line!

radio/src/main_avr.cpp
```diff
@@ -129,6 +132,9 @@ void perMain()
   const char *warn = s_warning;
   uint8_t menu = s_menu_count;
 
+#if defined(LCD_ST7920)
+  if(0==lcdstate){//No need to redraw until lcdRefresh_ST7920(0) below completely refreshes the display.
```
bsongis: It's difficult to read, would you add a definition, something like IS_LCD_REFRESH_ALLOWED() that would remove the #ifdef here?

bsongis:
@JivaroFAD I cannot test, I just did a code review, would you press the buttons once all comments reviewed? Thanks!

@tmrttmrt Thanks for your work, it's nice to see this new MEGA2560 target making progress!

radio/src/main_avr.cpp
tmrttmrt: Thank you for reviewing. Would you be please more specific. Where should I define the macro?
I would also appreciate some guidance regarding the procedure from this point. Should I create another commit and pull request?

bsongis: Somewhere in board_9x.h you define a new macro like this:
```
#if defined(LCD_ST7920)
#define IS_LCD_REFRESH_ALLOWED() (0==lcdstate)
#else 
#define IS_LCD_REFRESH_ALLOWED() (1)
#endif 
```
So that here you can remove all these new #ifdef with:
```
if (IS_LCD_REFRESH_ALLOWED()) {
...
}
```

bsongis:
No progress on this pull request @tmrttmrt is there a problem? or time is just missing?

tmrttmrt:
Just too busy with another project to find time for testing.

tmrttmrt:
Please check if it is OK now.

JivaroFAD:
Sorry for, I did not either take the time to test it :(

ch-nry:
i made a quick test. using the master branch and a ST7920 screen : servos movement are so slow it's unusable. 
with this branch :https://github.com/tmrttmrt/opentx/commit/11753450ae4c0137e3b35f761c8305578da60c89
everything look normal.
servos mouvement is smooth, and screen is working.

JivaroFAD:
Good, and many thanks for this test :)
@tmrttmrt : ok for merge ?

tmrttmrt:
OK from my side.

JivaroFAD:
Done :)

tmrttmrt:
Thanks!

bsongis:
The problem is now the indentation as written in one in my remarks above !

tmrttmrt:
I have fixed that?

JivaroFAD:
Here, I think : https://github.com/tmrttmrt/opentx/commit/11753450ae4c0137e3b35f761c8305578da60c89

radio/src/main_avr.cpp
```diff
@@ -129,31 +132,36 @@ void perMain()
...
-  if (menuEvent) {
-    m_posVert = menuEvent == EVT_ENTRY_UP ? g_menuPos[g_menuStackPtr] : 0;
-    m_posHorz = 0;
-    evt = menuEvent;
-    menuEvent = 0;
-    AUDIO_MENUS();
-  }
-  g_menuStack[g_menuStackPtr]((warn || menu) ? 0 : evt);
+  if(IS_LCD_REFRESH_ALLOWED()){//No need to redraw until lcdRefresh_ST7920(0) below completely refreshes the display.
+      lcd_clear();
```
bsongis: Here

radio/src/targets/stock/lcd_driver.cpp
```diff
@@ -150,17 +150,65 @@ void lcdSetRefVolt(uint8_t val)
...
-  for (uint8_t y=0; y<64; y++) {
+  uint8_t result;
+  uint8_t *p;
+  if(full!=0){
+    yst=0;
+    yend=64;
+    state=0;
+  } 
+  else{
+    switch (state){//Since writing to ST7920 is too slow we need to split it to five bands.
```
bsongis: It would be better to have a const array for these 3 values

JivaroFAD:
Something like that (please verify, I do not have a ST7920 LCD to test) ?

  uint8_t yst[6]={0,0,13,26,39,52};
  uint8_t yend[6]={64,13,26,39,52,64}; 
  
  if(full!=0){
    state=0;
  } 
  else{ //Since writing to ST7920 is too slow we need to split it to five bands
    if (state==5){
      state=1;
    }
    else{
      state++;
    }    
  }

  for (uint8_t y=yst[state]; y<yend[state]; y++) {

JivaroFAD:
Or, another way:

  uint8_t yst;
  uint8_t yend;
  uint8_t y_table[6]={0,13,26,39,52,64}; 
  
  if(full!=0){
    yst=0;
    yend=64;
    state=0;
  } 
  else{ //Since writing to ST7920 is too slow we need to split it to five bands
    yst=y_table[state];
    yend=y_table[state+1];
    if (state==4){
      state=0;
    }
    else{
      state++;
    }    
  }

  for (uint8_t y=yst; y<yend; y++) {

tmrttmrt:
I appreciate your comments. I had some doubts whether to do a "loop" or a case. The case is certainly more human readable, while the loop is more extendable and flexible.  Since I do not see any need for extension in the foreseeable future I opted for the case. Moreover, I personally do not like to put a table on the  stack so it should be defined statically, which I try to avoid within functions.  I propose to leave it like this.

PR #2832, opened 2015-09-05
Fixes #2810 - Reject new sensor option

### LINKED ISSUE #2810: Ghost sensor:   On one of my models I suddenly have a telemetry entry 1110
In the past I had duplicate Vario's in one of my models.
I suspect it happened because I used another one of my 4 vario's

To prevent this from happening again I enabled the option "ignore ID's"  (is that a correct solution, BTW?)

Now I all of a sudden have this extra telemetry sensor.
Most of the time I edit only 1 model and copy that model to the other 3 after I do a restructuring of the setup.
Afterwards I will edit the Name, outputs, RxNum, and GVARs  (I'm telling you all this to give you additional information).

I regularly backup as an eepe and it turns out this sensor appeared after I flew that model last Friday. I started flying with

Addition:

I added that same model (of Thursday's setup) without the extra sensor to my setup using Companion and made a backup of the EEPROM.
I then connected it to the model.
This extra sensor didn't appear.
No telemetry sensors were swapped this weekend.

eeprom-2015-08-30-10-14-31 bin
eeprom-2015-08-30-09-56-53 bin
eeprom-2015-08-30-10-16-31 bin

bsongis:


projectkk2glider:
Looks good (did only code review), you could make code even simpler if you used a checkbox for ITEM_TELEMETRY_DISCOVER_SENSORS.

PR #2834, opened 2015-09-05
Jtaylor2/issue  2762

jtaylor2:
Add ability to select the sensor instance  to which the  telemetry simulator sends data,
Add default data to detected sensors (RSSI, SWR, A1, A2, rxBAT) in Companion Simulator to prevent alarms programmed with LS/SF from alarming constantly while in the simulator.
Fixes issue 2762.

### DIFF
--- a/radio/src/telemetry/frsky.cpp
+++ b/radio/src/telemetry/frsky.cpp
@@ -583,6 +583,37 @@ void telemetryReset()
 #endif
 #endif
 
+/*Add some default sensor values to the simulator*/
+#if defined(CPUARM) && defined(SIMU)
+  for (int i=0; i<MAX_SENSORS; i++) {
+    const TelemetrySensor & sensor = g_model.telemetrySensors[i];
+    printf("Sensor ID %x Instance %i \n", sensor.id, sensor.instance);
+      switch (sensor.id)
+      {
+        case RSSI_ID:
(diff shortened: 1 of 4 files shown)


radio/src/telemetry/frsky.cpp
```diff
@@ -583,6 +583,32 @@ void telemetryReset()
...
+        setTelemetryValue(TELEM_PROTO_FRSKY_SPORT, SWR_ID, sensor.instance, 30, UNIT_RAW, 0);
+        break;
+      case BATT_ID:
+        setTelemetryValue(TELEM_PROTO_FRSKY_SPORT, BATT_ID, sensor.instance, 100, UNIT_RAW, 0);
+        break;
+    }
+  }
+#endif
+
+
```
kilrah: This should either replace or be better combined with the code below that does just the same but statically!

jtaylor2: The static code does not add default values to the Companion simulator, which is what needs them to stop the alarms.  I didn't understand why Companion was excluded so I didn't mess with that code.

I can add the rest of the sensors from the static list to my dynamic list (I don't have all the sensors to test, but I can copy the values) and delete the static code.  But then I don't think it will provide values to simu,  Not sure how to better combine, suggestions?

Issue #1914, opened 2014-12-10
More informative Lua error messages for telemetry scripts


projectkk2glider:
For example in `master` (`next` doesn't even have this error message) user gets "Syntax error" for both true syntax error and out of memory error. At least these two should be separated. Ideally user could also see the actual reason for error (ie line number, etc) for problematic telemetry scripts.

projectkk2glider:
Also separate error for wrong number of returned values would be beneficial.

dsbeach:
I started looking at this today, but have not yet identified the code that implemented this in 2.0.x.

Can anyone point me in the right direction?

David

bsongis:
![snapshot_04](https://cloud.githubusercontent.com/assets/5603504/10106385/a57c2a5a-63b4-11e5-875a-3a29c28d4b10.png)

PR #2864, opened 2015-09-11
2.1.x implementation of displayLuaScriptError()

dsbeach:
This is an initial fix for issue #1914 and implements the 2.0.x changes for lua telemetry screen errors.

Still outstanding is whether or not to attempt to capture and display more meaningful lua error messages.

David

### DIFF
--- a/radio/src/gui/Taranis/view_telemetry.cpp
+++ b/radio/src/gui/Taranis/view_telemetry.cpp
@@ -182,11 +182,53 @@ void displayCustomTelemetryScreen(uint8_t index)
   displayNumbersTelemetryScreen(screen);
 }
 
+#if defined(LUA)
+
+#define LUA_SCRIPT_ERROR_Y    3*FH
+// TODO more visual appealing implementation and text translations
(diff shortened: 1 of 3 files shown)


radio/src/gui/Taranis/view_telemetry.cpp
```diff
@@ -182,11 +182,53 @@ void displayCustomTelemetryScreen(uint8_t index)
   displayNumbersTelemetryScreen(screen);
 }
 
+#if defined(LUA)
+
+#define LUA_SCRIPT_ERROR_Y    3*FH
+// TODO more visual appealing implementation and text translations
```
bsongis: You may use the addtr.py script to save time!

bsongis:
Seems good to me otherwise, thanks!

bsongis:
Is it possible you put it in a dsbeach/issueXXXX_display_lua_script_error? The pull request would be checked by Travis CI this way

Also I think the translations are still missing, would you take care of them? Thanks in advance!

bsongis:
I finished this one which is now in 'master' branch, thanks for your help!

PR #2888, opened 2015-09-17
New interface between eeprominterface and gui.

griebd:
With alert when openning an old eeprom version.
Please review and test with different files, I have only a few ones from Taranis.


griebd:
ignore it, I will push again...

bsongis:
ok! would you open in branch inside the project so that we have jenkins and travis checking as well?

PR #2890, opened 2015-09-17
New interface between eeprominterface and gui.

griebd:
With alert when opening an old eeprom version.
Please review and test with different files, I have only a few ones from Taranis.

First commit is the same as 2.0.
Second one is new code to have the alert.


bsongis:
ok I didn't see this one, thanks!

PR #2901, opened 2015-09-19
Fixes #2896: draw curves without using temporary disk files

### LINKED ISSUE #2896: [Companion 2.1.3][Bug] Model comparison issue: curves
Hi folks,

The new comparison engine is much better, but there is still one issue with the curves. The data are ok, but not the graphic curves.
See the snapshot below:
issuecurves

projectkk2glider:
Closes #2896.

Perhaps this change will fix the original issue. The model curves are no longer created on disk, but embedded directly into the document.


bsongis:
Excellent idea, thanks!

projectkk2glider:
I will port it to the next branch.

projectkk2glider:
Ported in 69a6a1853ae15a88ad92f42625bd1da0316c7a25

PR #2903, opened 2015-09-19
Fixes #2902

### LINKED ISSUE #2902: PXX pulses - Wrong maths when PPM_CENTER applied
Found when fixing #2880 - PPM_CENTER with DSM2

bsongis:


projectkk2glider:
I recommend you also add automatic conversion from the old to the new ppm center value when converting old eeprom.

bsongis:
Oh yes this is a good idea, thanks!

PR #2909, opened 2015-09-20
Fixes #2950: Mavlink build option was not working

### LINKED ISSUE #2950: [OpenTX 2.1.3][Bug][9XR-Pro] Battery gauge bar shows only 1/2 battery when 100% full
If using the graphical battery gauge bar option on a 9XR Pro, the same number of bars is displayed for a full battery as on a Taranis, but Taranis has a smaller battery symbol. So the "full" charge does not "fill" the battery icon - it looks half empty.

projectkk2glider:
Closes #2950

Please check my changes in myeeprom.h carefully!

### DIFF
--- a/radio/src/Makefile
+++ b/radio/src/Makefile
@@ -1273,7 +1273,7 @@ endif
 
 ifeq ($(EXT), MAVLINK)
   CPPDEFS += -DMAVLINK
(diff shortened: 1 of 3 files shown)


bsongis:
It seems perfect, thanks!

PR #2929, opened 2015-09-25
Fixes #2925: added coordinate check in lcd_line()

### LINKED ISSUE #2925: Display freeze OpenTX 2.1.3
I changed some telemetry lua scripts to use the new data names, but my Taranis Plus freezes the display when no telemetry data is available, for example when you do reset flight. 
Exit button and off switch do not respond, I have to pull the battery.
Flight controls remain working.

projectkk2glider:
I tried fixing input parameters to safe values, but the code was too complicated. Instead a simple check is done, if start or end of the line is outside the LCD the drawing is skipped altogether.

This pull probably closes #2925, but we need to verify that.

### DIFF
--- a/radio/src/gui/Taranis/lcd.cpp
+++ b/radio/src/gui/Taranis/lcd.cpp
@@ -45,6 +45,11 @@
   display_t displayBuf[DISPLAY_BUF_SIZE] __DMA;
 #endif
 
+inline bool lcdIsPointOutside(coord_t x, coord_t y)
+{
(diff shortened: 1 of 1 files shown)


projectkk2glider:
@bsongis why are the `coord_t` signed variables in 2.1? They were unsigned in 2.0 and that prevented this problem there.

bsongis:
I can't remember why I changed to signed, perhaps for drawing horizontal / vertical lines that start on a negative X or Y but could be considered as valid. Not sure though.

Issue #2870, opened 2015-09-13
9XR PRO PPM2 support?


Links2004:
Hardware: 9XR PRO
Version: 2.1.4N453

is there a way of enabling the PPM2 port of the 9XR PRO.

may this helps:

![image](https://cloud.githubusercontent.com/assets/974709/9836267/f90265d2-5a12-11e5-9a67-26bf09f18cce.png)

http://openrcforums.com/forum/viewtopic.php?f=122&t=5659

projectkk2glider:
@MikeBland you seem to know 9XR-PRO the best. Is this possible?

kilrah:
Pretty sure we already have the code in opentx and it's just not enabled for 9xr-pro.

MikeBland:
The PPM2 port was always on the SKY board and the 'PRO is basically a copy of that.
I thought PPM2 was already supported, and even open9x seems to have support for PPM2 on the SKY board. It's certainly been in ersky9x for both SKY and 'PRO from when they became available.

bsongis:
Yes it's only me, I didn't know PPM2 port was available in 9XR-PRO and I disabled this port. It should only some #define / #ifdef to change.

Links2004:
may this one?
https://github.com/opentx/opentx/blob/6c0a9714fc85237af9d17b0b44ee1348c315f08d/radio/src/gui/9X/menu_model_setup.cpp#L704

what is the meaning of the different ```REVX```  and ```AR9X``` for the ```PCBSKY9X```?

MikeBland:
There is the original SKY board, then there is the 9XR-PRO (REVX), which is slightly different, and there is a new board, not yet fully for sale, called the AR9X.

PR #2931, opened 2015-09-25
enable PPM2 support for 9XR PRO + allow 16CH on PPM2 #2870

Links2004:
enable PPM2 support for 9XR PRO
allow 16CH on PPM2
see #2870


bsongis:
Good job!
There was only a merge problem so I cherry-picked your mods. Also I fixed reverted some permissions on python scripts (in the util directory)

PR #2946, opened 2015-10-04
Python 2/3 compatibility (and PEP 8)

flacjacket:
The first commit makes the Python code base run in both Python 2 and Python 3, also reverting the changes to the radio `Makefile` to find Python 2 only. This will need Python >=3.3 on the Python 3 side to support `u""` unicode string literals.

The second commit is a bit of bike-shedding, I went in and PEP 8'ed all of the Python code, making a handful of minor changes along the way. I don't know which Python versions you target, but I used some context managers (e.g. `with open(...) as f:`) that are Python 2.7+. I can pull the PEP 8 stuff if this isn't something you guys want.


projectkk2glider:
Wow, a lot of changes, thanks.

I looked over the changes, they seem reasonable, but I haven’t yet tested any of them. In fact testing every changed script might be a challenge, since some of them are not used in build process. Therefore I am a bit wary about this pull. I do not want us to break anything now only to discover the breakage some time down the line.

@flacjacket did you preform any tests on your changes?

flacjacket:
I've tested building the firmware and companion with both of these commits, but I'm not sure how to use all of the scripts. If you can refer me to somewhere to look on how to test them, definitely let me know.

The first commit is the more important one 8e5218deae2c53a149a226ea7efdc2ed45736b4d, and is a more reasonable +122/-92 change. The second commit is purely style changes, fixing whitespace, code conventions, etc. If it's too hard to check all the changes, I can remove that commit to make this PR a little more manageable.

flacjacket:
Or leave the first commit here and move the second commit to another PR until it can be properly vetted.

bsongis:
Really good job, thanks! I like almost everything here. I have a few remarks to make the merge easier and quicker:
- Is it possible to have one commit per file to allow us not to take everything in one shot, we have changes in progress, especially the Makefile ...
- The compile.py is a very old script, we can remove it completely


Also the tts.py is too long, it grew with languages additions, while you are making big changes, is it possible for you to split it? I am sure you will do clean work.

flacjacket:
No problem, I'll see what I can do splitting everything up and reworking tts.py.

flacjacket:
Alright, I think I've broken everything out, now the commits are as follows:

* 7a084ab makes everything Python 3 compatible
* 55c5276 updates the `Makefile` to use the system Python
* 4212d95 removes `compile.py`
* ecf6cb7..17af3b7 (17 commits) PEP8's each file individually
* 5e188f5 PEP8's tts.py and splits each language off into its own file, which is imported from tts.py
* 172684a use `with open() as` rather `open()`

bsongis:
I will merge this pull request once the Horus branch is in 'next'

flacjacket:
Great, thanks for the heads-up.

bsongis:
Thanks! Would you port it to 'next'?

flacjacket:
Sure, I just opened up a new PR where I moved all this over to `next`, #3144

Issue #2953, opened 2015-10-09
Ambiguity in ZSTR_TEMP referring to both TEMP1_ID and TEMP2_ID


ptsneves:
This ambiguity causes problems with the scripts' `getValue()`. 

A simple rename is unadvised because the user doesn't know which of the "discovered" sensors represents the intended data.

Also there is an inconsistency in frsky_sport.cpp where sport sensors for `ZSTR_TMP`

 is valid for the range `T1_FIRST_ID, T2_LAST_ID` when the `T1_FIRST_ID, T1_LAST_ID` and  `T1_FIRST_ID, T1_LAST_ID` already exist.

I have not created a pull request because this patch is very small and applicable to 2.1.3 tag and the master and next branch are already too far ahead and with merge conflicts in frsky_d.cpp. 

If anybody can guide me where the changes of frsky_d.cpp are applicable in the `next` or `master` branch let me know and I will make a pull request

```
diff --git a/radio/src/telemetry/frsky_d.cpp b/radio/src/telemetry/frsky_d.cpp
index 49e3e9c..5abcb8e 100644
--- a/radio/src/telemetry/frsky_d.cpp
+++ b/radio/src/telemetry/frsky_d.cpp
@@ -513,8 +513,8 @@ const FrSkyDSensor frskyDSensors[] = {
   { D_A2_ID, ZSTR_A2, UNIT_VOLTS, 1 },
   { RPM_ID, ZSTR_RPM, UNIT_RPMS, 0 },
   { FUEL_ID, ZSTR_FUEL, UNIT_PERCENT, 0 },
-  { TEMP1_ID, ZSTR_TEMP, UNIT_CELSIUS, 0 },
-  { TEMP2_ID, ZSTR_TEMP, UNIT_CELSIUS, 0 },
+  { TEMP1_ID, ZSTR_TEMP1, UNIT_CELSIUS, 0 },
+  { TEMP2_ID, ZSTR_TEMP2, UNIT_CELSIUS, 0 },
   { CURRENT_ID, ZSTR_CURR, UNIT_AMPS, 1 },
   { ACCEL_X_ID, ZSTR_ACCX, UNIT_G, 3 },
   { ACCEL_Y_ID, ZSTR_ACCY, UNIT_G, 3 },
diff --git a/radio/src/telemetry/frsky_sport.cpp b/radio/src/telemetry/frsky_sport.cpp
index b2f6930..1df75ad 100644
[...]

PR #2955, opened 2015-10-09
Corrected ambiguity in ZSTR_TEMP referring to both TEMP1_ID and TEMP2…

ptsneves:
Answer to Issue #2953 
This ambiguity causes problems with the scripts' getValue().

A simple rename is unadvised because the user doesn't know which of the "discovered" sensors represents the intended data.

Also there is an inconsistency in frsky_sport.cpp where sport sensors for ZSTR_TMP

is valid for the range T1_FIRST_ID, T2_LAST_ID when the T1_FIRST_ID, T1_LAST_ID and T1_FIRST_ID, T1_LAST_ID already exist.

I have not created a pull request because this patch is very small and applicable to 2.1.3 tag and the master and next branch are already too far ahead and with merge conflicts in frsky_d.cpp.

If anybody can guide me where the changes of frsky_d.cpp are applicable in the next or master branch let me know and I will make a pull request


radio/src/telemetry/frsky_sport.cpp
```diff
@@ -64,7 +64,8 @@ const FrSkySportSensor sportSensors[] = {
   { A3_FIRST_ID, A3_LAST_ID, ZSTR_A3, UNIT_VOLTS, 2 },
   { A4_FIRST_ID, A4_LAST_ID, ZSTR_A4, UNIT_VOLTS, 2 },  
   { BATT_ID, BATT_ID, ZSTR_BATT, UNIT_VOLTS, 1 },
-  { T1_FIRST_ID, T2_LAST_ID, ZSTR_TEMP, UNIT_CELSIUS, 0 },
+  { T1_FIRST_ID, T1_LAST_ID, ZSTR_TEMP1, UNIT_CELSIUS, 0 },
+	{ T2_FIRST_ID, T2_LAST_ID, ZSTR_TEMP2, UNIT_CELSIUS, 0 },
```
projectkk2glider: Indentation problem.

projectkk2glider:
Looks good. Do not worry about the `next` branch for now.

bsongis:
I am ok as well. It's consistent with A1, A2 etc. I don't like Tmp1 but no choice, we only have 4 chars available

ptsneves:
The Tmp was due to the 4 character limitations indeed. There are other references to Tmp in the general code so it is not terribly inconsistent. 

In Monday I will correct the indentation problem, if necessary.

ptsneves:
The Indentation problem is corrected.

bsongis:
Thanks!

PR #2957, opened 2015-10-10
Fixes #2936: 3POS type removed from SH and SF switches (hw only suppo…

### LINKED ISSUE #2936: Taranis X9E SWF and SWH don't seem to recognise 3 pos on-off-on switch
Hi,
    I've fitted an On-Off-On three position switch to both SWF and SWH on my Taranis X9E and set 3 POS in the hardware menu. The main screen display page only shows a 2 positions when the switch is toggled. I note that the switch port connectors header markings are in bold (like the analogue inputs) SWH rather than SW-H so is this an analogue "switch" port that I need to feed 0v - Vcc/2 -Vcc to get the three positions or is it a firmware bug??

Any thoughts gratefully received.

Barry M

projectkk2glider:
…rts 2POS type)

Closes #2936

### DIFF
--- a/companion/src/generaledit/calibration.cpp
+++ b/companion/src/generaledit/calibration.cpp
@@ -1,7 +1,7 @@
 #include "calibration.h"
 #include "ui_calibration.h"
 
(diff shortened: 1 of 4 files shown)


bsongis:
Thanks!

PR #2979, opened 2015-10-18
Out of bounds protection for telemetryItems[] array access and some c…

projectkk2glider:
…ode refactoring for telemetry value precision

Added several checks that should prevent out of bounds array access for telemetyItems[].


bsongis:
All good, thanks!

PR #3011, opened 2015-10-29
Projectkk2glider/lua io issues

### LINKED ISSUE #3001: 2 of 4 related issues to LUA io library
io.open("filename","w") only succeeds when the file already exists on Taranis.
No problem on PC simulator.
I tried -w

### LINKED ISSUE #3003: 4 of 4 related issues to LUA io library
io.write(fh,"any string") is OK
io.write(fh,stringvarname) is OK
io.write(fh,numvarname) writes .14g to the file
io.write(fh,2) writes .14g
io.write(fh,2.2) writes .14g

This is only on Taranis.  All good on PC sim.  That's how I know my syntax is correct?

projectkk2glider:
I have tested this on my radio (Taranis) and in simu. Both work. The test script:
```lua
local function run(event)
  print("c1")
  local f = io.open("foo.bar", "a")
  io.write(f, "first line\r\nsecond line\r\n")
  io.write(f, 4, "\r\n", 35.6778, "\r\n")  -- one can write multiple items at the same time
  local foo = -4.45
  io.write(f, foo, "\r\n")
  io.close(f)
  return 1    -- this will end the script execution
end

return { run=run }
```

Perhaps we should wait for @sbright33 to test this, the result of his test is expected in #3003.

Closes #3003 and closes #3001

### DIFF
--- a/radio/src/targets/simu/simpgmspace.cpp
+++ b/radio/src/targets/simu/simpgmspace.cpp
@@ -819,10 +819,10 @@ FRESULT f_open (FIL * fil, const TCHAR *name, BYTE flag)
     fil->fsize = tmp.st_size;
     fil->fptr = 0;
   }
-  fil->fs = (FATFS*)fopen(realPath, (flag & FA_WRITE) ? "wb+" : "rb+");
+  fil->fs = (FATFS*)fopen(realPath, (flag & FA_WRITE) ? ((flag & FA_CREATE_ALWAYS) ? "wb+" : "ab+") : "rb+");
   fil->fptr = 0;
   if (fil->fs) {
-    TRACE("f_open(%s) = %p", path, (FILE*)fil->fs);
+    TRACE("f_open(%s, %x) = %p", path, flag, (FILE*)fil->fs);
     return FR_OK;
   }
   TRACE("f_open(%s) = error %d (%s)", path, errno, strerror(errno));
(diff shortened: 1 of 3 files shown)


radio/src/targets/simu/simpgmspace.cpp
```diff
@@ -856,6 +856,19 @@ FRESULT f_lseek (FIL* fil, DWORD offset)
...
+{
+  if (fil && fil->fs) {
+    long curr = ftell((FILE*)fil->fs);
+    fseek((FILE*)fil->fs, 0, SEEK_END);
+    long size = ftell((FILE*)fil->fs);
+    fseek((FILE*)fil->fs, curr, SEEK_SET);
+    TRACE("f_size(%p) %u", fil->fs, size);
+    return size;
+  }
+  return 0;
```
bsongis: Wow, is it the best solution?

radio/src/thirdparty/FatFs/ff.h
```diff
@@ -238,7 +238,11 @@ TCHAR* f_gets (TCHAR* buff, int len, FIL* fp);						/* Get a string from the fil
 #define f_eof(fp) ((int)((fp)->fptr == (fp)->fsize))
 #define f_error(fp) ((fp)->err)
 #define f_tell(fp) ((fp)->fptr)
-#define f_size(fp) ((fp)->fsize)
+#if !defined(SIMU)
+	#define f_size(fp) ((fp)->fsize)
+#else
+	UINT f_size(FIL* fil);
+#endif
```
bsongis: Would it be possible to write something like this in simpgmspace.h (to avoid mods in thirdparty):

```
#undef f_size
UINT f_size(FIL* fil);
```

bsongis:
All good, thanks!

radio/src/thirdparty/FatFs/ff.h
projectkk2glider: Sure, good idea.

projectkk2glider: I can't do it the way you suggest. Everywhere that we have `#include "ff.h"` the macro gets used before we undefine it. So the function f_size() is never called.

bsongis: ok!

Issue #3024, opened 2015-11-02
Companion 2.1 (windows) - Lua 'execute' fails in non-root directory


dsbeach:
Using the SD browser, select a .lua file, select 'Execute'.

Debug window reveals:
```
findTrueFileName(\SCRIPTS\ONETIME/play-dur.lua)
	not found
f_open(\SCRIPTS\ONETIME/play-dur.lua) = INVALID_NAME
Error in script \SCRIPTS\ONETIME/play-dur.lua: cannot open \SCRIPTS\ONETIME/play-dur.lua: Bad file descriptor
```

Apparently the companion doesn't like the mixed forward and backward slashes.

I'll create a pull request for the following addition to menu_general_manager.cpp line 241
```
#if defined(SIMU)
    for (int i = 0; lfn[i] != 0; i++) {
        if (lfn[i] == '\\') { lfn[i] = '/'; }
	}
#endif
```

bsongis:
In your commit message, would you write something like "Fixes #3024" so that the issue is automatically marked as fixed when it's merged into 'master'? Thanks!

dsbeach:
I did not know about that feature. Can do.

kilrah:
Are you sure that's a problem? There's always been a mix of forward and backslashes on Windows but it was never an issue before.

dsbeach:
It's definitely an issue on my MSVC build. I'll test against an official build today and add #if defined(MSVC) provided that is appropriate.

dsbeach:
Thanks for the link to the style guide, that will be useful.  I've adjusted my Visual Studio settings so at least the indentation and tabs -> spaces should be automatic from now on.

I reinstalled Companion 2.1.3 this morning and confirmed it is an issue in that release on my Windows 10 system.

Thanks for being patient, I think this is good to go.

kilrah:
> It's definitely an issue on my MSVC build. I'll test against an official build today and add #if defined(MSVC) provided that is appropriate.

Still wondering because I've been building and using on MSVC build since I started playing with OpenTX 3 years ago.
Not even talking of the officially released builds that are built the same way, and we never had a report of a related issue.

dsbeach:
I just confirmed the same issue on a Win 7 (32-bit and 64-bit).  Start the simulator, long press Menu, short press Page, navigate to SOUNDS, navigate to EN, navigate to a .wav file, press ENT, select PLAY.

Trace from debugger window:
```
f_opendir(.) = OK
[...]

PR #3025, opened 2015-11-02
Issue #3024 - Companion 2.1 (windows) - Lua 'execute' fails in non-root directory

dsbeach:
~~Conversion of backward slash to forward slash for companion.~~

Fixes #3025 

Solution was to modify convertSimuPath() to append fully qualified path name for requests starting with both forward and backward slashes.

### DIFF
--- a/radio/src/targets/simu/simpgmspace.cpp
+++ b/radio/src/targets/simu/simpgmspace.cpp
@@ -697,7 +697,7 @@ FATFS g_FATFS_Obj;
 char *convertSimuPath(const char *path)
 {
   static char result[1024];
-  if (path[0] == '/' && strcmp(simuSdDirectory, "/") != 0)
+  if (((path[0] == '/') || (path[0] == '\\')) && (strcmp(simuSdDirectory, "/") != 0))
     sprintf(result, "%s%s", simuSdDirectory, path);
   else
     strcpy(result, path);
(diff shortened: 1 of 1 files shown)


projectkk2glider:
I think the change should be done in simpgmspace.cpp as it has nothing to do with the radio.

dsbeach:
I'll investigate and see if I can find the right spot.

radio/src/targets/simu/simpgmspace.cpp
```diff
@@ -1006,6 +1006,12 @@ FRESULT f_getcwd (TCHAR *path, UINT sz_path)
 
   // remove simuSdDirectory from the cwd
   strcpy(path, cwd + strlen(simuSdDirectory));
+
+  // convert backslashes to forward - issue #3024
+  for (int i = 0; path[i] != 0; i++) {
+	  if (path[i] == '\\') { path[i] = '/'; }
```
bsongis: Code style not ok!

dsbeach: Please elaborate. Is there a style guide?

bsongis: Have a look in other files, you will see, here the indentation is KO (we use 2 chars) and a new line is missing

projectkk2glider: I've started a style guide here https://github.com/opentx/opentx/wiki/OpenTX-Code-Style-Guide

PR #3054, opened 2015-11-11
SIMU Lua compile on load enhancement ( file.lua.src -> file.lua)

dsbeach:
Modifications for simulator only (via #ifdef SIMU). When a lua script is loaded it will look for a file with ".src" appended and attempt to compile it without debugging symbols. For large scripts, this results in significant run time memory savings.


projectkk2glider:
I will close this, slightly modifed code from this pull is in https://github.com/opentx/opentx/commit/754b875d81b072e2e18f66694960b5c35deacc0e

PR #3060, opened 2015-11-12
Fixes #3059: added unit conversions from km/h

### LINKED ISSUE #3059: Add conversions from km/h
Needed for air speed sensor which reports its speed in km/h

projectkk2glider:
Haven't tested it, so please give it a quick look before merging.

Closes #3059


projectkk2glider:
Hold on! I have a better idea, I will put all conversions in a table and do a lookup.

projectkk2glider:
I think this is better now. I have also quickly tested it in Companion simulator.

PR #3065, opened 2015-11-14
Projectkk2glider/autogen lua docs 2

### LINKED ISSUE #2947: Lua environment status function, Enhancement request.
It would be helpful in finding bugs while developing a Lua script if Lua could determine its environment.
That is, is it operating a PC simulator or on an actual tx?
This would allow an if statement to decide if something needs to have a simulated value set....such as TxVoltage., SWR, RSSI,....

projectkk2glider:
This goes hand in hand with https://github.com/opentx/lua-reference-guide/pull/9

Also closes #2947

### DIFF
--- a/radio/src/lua_api.cpp
+++ b/radio/src/lua_api.cpp
@@ -452,7 +452,7 @@ static int luaGetValue(lua_State *L)
 /*luadoc
 @function playFile(name)
 
-Plays a track with given name
+Plays a track with a given name
 
 @param name (string) file name (including path) to play. If the path is not 
 absolute (name starts with the `/` character), then the given name is used relative to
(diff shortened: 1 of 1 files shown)


radio/src/lua_api.cpp
```diff
@@ -104,12 +104,72 @@ void hook(lua_State* L, lua_Debug *ar)
   }
 }
 
+#if defined(PCBTARANIS) && defined(REV9E)
+  #define RADIO "taranisx9e"
```
bsongis: This will have to be removed when the switch to CMake will be completely done!

bsongis:
Thanks!

radio/src/lua_api.cpp
projectkk2glider: I am not happy whit this part of the code also, but I needed to hack up something. Do you have an example of the right solution?

bsongis: It will easily fixed in a next release, this is already done somewhere else

PR #3068, opened 2015-11-15
Projectkk2glider/lua split files

projectkk2glider:
@bsongis I had to nuke the stamp-opentx.h generation in the Cmake, it was interfering with the build of Companion (I could not  find why that became a problem), but the contents of this file look like they were not used. Please check in case I broke anything.


dsbeach:
Any chance of getting 011c265945254f8959473f0d64c987890a5c3bdd included while you are doing this?

projectkk2glider:
@dsbeach I am completely happy with that change. Too complicated. My idea:
* we define new compile option, lets say LUA_COMPILE
* when LUA_COMPILE=YES and SIMU is defined we dump compiled script to <script_name>.luac whenever a script is loaded (no need for special extensions).

dsbeach:
I don't care how it is done, but I do need the ability to create compiled scripts.  My GPS triangle racing script is too big to compile and run in a single pass.  I was wondering about adding the functionality through a new 'Lua' menu option in Companion, but that got too involved.

projectkk2glider:
@dsbeach I have used your code and modified it slightly. If compiled with LUA_COMPILE=YES, then the simu produces .luac file (which is LUA byte-code) for every loaded file (execpt if the .luac file already exists).

If you want this functionality in Companion, you just have to add to the `opentxsimulator.cpp`:
```
#define LUA_COMPILE
```

Perhaps in the future we can make this choice in the runtime, we add a setting to the Companion and a if statement in luaCompileAndSave()

dsbeach:
I'd like to see this option available to everyone, not just those who can build companion. But this works for me and I'm glad to see it in the production source base.

Comparing this approach to my original, I see a few drawbacks:
  -  Every script is compiled (I suspect there are only a few that need it)
  -  Compile is aborted if .luac exists (could we at least compare timestamps and overwrite if newer)
  -  Running the compiled code requires renaming the file (and possibly mistakenly overwriting the source)

As long as it's a companion build option, why not take the original approach of always compiling when a xxx.lua.src exists?  Is there an advantage to creating files with .luac extensions?

projectkk2glider:
The only drawback (for me) with your approach was that you had to have both foo.lua and foo.lua.src before the complier was run. Both files were Lua source at this time. Then after the run the foo.lua will be overwriten by the Lua byte-code. Isn’t that a bit dangerous too? 

That was the reason I decided to wrote the compiled code in a foo.luac file. But I can be persuaded into many things :smile: you just have to try hard enough.

dsbeach:
The typical flow for me was to start development with a .lua file.

Get something you like, and try it on a radio, but it bombs for lack of memory.

Now copy the file to .lua.src. Back to companion and it will compile, then load the compiled file.  Need a change, just change the .lua.src file.  Easy!

You won't be very likely to successfully edit the compiled file.

projectkk2glider:
Sorry, I need a second opinion.

So question for @bsongis and @kilrah We are adding in an ability for the simulators to dump the compiled Lua code into a file. There are two approaches and we need to decide which one to use:

Approach A:
* when a script is loaded (I will use `foo.lua` in this examples), check if `foo.lua.src` exists.
* if it does, then  load `foo.lua.src` file, compile it and overwrite the `foo.lua` with the compiled version (Lua byte-code).

Approach B:
* when a script `foo.lua` is loaded, check if `foo.luac` exists.
* if it doesn't, then load `foo.lua`, compile it and write compiled code into `foo.luac` file.

bsongis:
I am open to those 2 solutions, but I wonder if it is a job for the simu. Why not an external tool (possibly called inside Companion such as the SD synchronization tool)?

dsbeach:
My first thought was also an external tool.  I just wasn't up to diving into Qt to get it done.  I've learned enough new 'stuff' this week.

Am I correct that the Lua functions are only available in the simulator dll's?  What approach would you take to get Lua into Companion?  Would you bring in a whole environment with a syntax sensitive editor?

bsongis:
I don't hink you need to dive into Qt for this tool! But yes a whole environment would be a nice feature, perhaps in a second time?

bsongis:
I just reviewed the changes, it's a nice job, it was really needed!

kilrah:
I'd ideally vote for a "precompile script" functionality in companion, it's what makes most sense, but is just a lot more work integrating lua into companion than simply dumping the simu's output.

PR #3077, opened 2015-11-16
Update flasheepromdialog.ui

indavis:
Corrected the tool tip for the backupBeforeWrite checkbox. It incorrectly had the text from Check Firmware Compatibility. I tried to word the tooltip as descriptive as possible.


bsongis:
Would you port it to the 'next' branch? Thanks!

indavis:
Will do.

PR #3079, opened 2015-11-16
Modifications to luaCompileAndSave()

dsbeach:
As scripts are loaded it will check for existing file <file>.lua.src and compile it to <file>.lua if found.

This functionality still requires the build option LUA_COMPILE=YES


projectkk2glider:
Looks OK to me.

projectkk2glider:
Thanks!

PR #3081, opened 2015-11-17
Update flasheepromdialog.ui, Fixed backupBeforeWrite toolTip.

indavis:
backupbeforewrite_tooltip
Updating the backupBeforeWrite toolTip. It was incorrectly set to the checkFirmwareCompatibility text.


bsongis:
Thanks!

PR #3087, opened 2015-11-20
Fixes #2469: Allow usage of PLUS, MINUS, MENU and ENT keys in telemet…

### LINKED ISSUE #2469: OpenTX 2.1 killEvents not working?

projectkk2glider:
…ry scripts

I think this is the right solution:
* all keys are seen by the Lua scripts
* Lua scripts can not use killEvents() on ENT and EXIT
* if a telemetry script has run, filter out the keys PLUS, MINUS, ENT (only short),  EXIT (only short) and MENU from the menu system. These keys can now be used freely in the telemetry scripts.

Tested in simu.

Closes #2469. Replaces #3051

Updated description (added EXIT short)

### DIFF
--- a/radio/src/functions.cpp
+++ b/radio/src/functions.cpp
@@ -323,13 +323,13 @@ void evalFunctions()
             newActiveFunctions |= (1 << FUNCTION_INSTANT_TRIM);
             if (!isFunctionActive(FUNCTION_INSTANT_TRIM)) {
 #if defined(GUI)
-              if (g_menuStack[0] == menuMainView
+              if (menuHandlers[0] == menuMainView
 #if defined(FRSKY)
-                || g_menuStack[0] == menuTelemetryFrsky
+                || menuHandlers[0] == menuTelemetryFrsky
 #endif
 #if defined(PCBTARANIS)
-                || g_menuStack[0] == menuMainViewChannelsMonitor
-                || g_menuStack[0] == menuChannelsView
+                || menuHandlers[0] == menuMainViewChannelsMonitor
+                || menuHandlers[0] == menuChannelsView
 #endif
               )
 #endif
--- a/radio/src/gui/9X/menu_general_diaganas.cpp
+++ b/radio/src/gui/9X/menu_general_diaganas.cpp
@@ -79,24 +79,24 @@ void menuGeneralDiagAna(uint8_t event)
   adcBatt = ((adcBatt * 7) + anaIn(TX_VOLTAGE)) / 8;
   uint32_t batCalV = (adcBatt + adcBatt*(g_eeGeneral.txVoltageCalibration)/128) * 4191;
   batCalV /= 55296;
-  putsVolts(LEN_CALIB_FIELDS*FW+4*FW, MENU_HEADER_HEIGHT+1+4*FH, batCalV, (m_posVert==1 ? INVERS : 0));
+  putsVolts(LEN_CALIB_FIELDS*FW+4*FW, MENU_HEADER_HEIGHT+1+4*FH, batCalV, (menuVerticalPosition==1 ? INVERS : 0));
 #elif defined(PCBGRUVIN9X)
   lcd_putsLeft(6*FH-2, STR_BATT_CALIB);
   // Gruvin wants 2 decimal places and instant update of volts calib field when button pressed
   static uint16_t adcBatt;
   adcBatt = ((adcBatt * 7) + anaIn(TX_VOLTAGE)) / 8; // running average, sourced directly (to avoid unending debate :P)
   uint32_t batCalV = ((uint32_t)adcBatt*1390 + (10*(int32_t)adcBatt*g_eeGeneral.txVoltageCalibration)/8) / BandGap;
-  lcd_outdezNAtt(LEN_CALIB_FIELDS*FW+4*FW, 6*FH-2, batCalV, PREC2|(m_posVert==1 ? INVERS : 0));
+  lcd_outdezNAtt(LEN_CALIB_FIELDS*FW+4*FW, 6*FH-2, batCalV, PREC2|(menuVerticalPosition==1 ? INVERS : 0));
 #else
   lcd_putsLeft(6*FH-2, STR_BATT_CALIB);
-  putsVolts(LEN_CALIB_FIELDS*FW+4*FW, 6*FH-2, g_vbat100mV, (m_posVert==1 ? INVERS : 0));
+  putsVolts(LEN_CALIB_FIELDS*FW+4*FW, 6*FH-2, g_vbat100mV, (menuVerticalPosition==1 ? INVERS : 0));
 #endif
-  if (m_posVert==1) CHECK_INCDEC_GENVAR(event, g_eeGeneral.txVoltageCalibration, -127, 127);
+  if (menuVerticalPosition==1) CHECK_INCDEC_GENVAR(event, g_eeGeneral.txVoltageCalibration, -127, 127);
 
 #if defined(TX_CAPACITY_MEASUREMENT)
   lcd_putsLeft(6*FH+1, STR_CURRENT_CALIB);
-  putsValueWithUnit(LEN_CALIB_FIELDS*FW+4*FW, 6*FH+1, getCurrent(), UNIT_MILLIAMPS, (m_posVert==2 ? INVERS : 0)) ;
-  if (m_posVert==2) CHECK_INCDEC_GENVAR(event, g_eeGeneral.txCurrentCalibration, -49, 49);
+  putsValueWithUnit(LEN_CALIB_FIELDS*FW+4*FW, 6*FH+1, getCurrent(), UNIT_MILLIAMPS, (menuVerticalPosition==2 ? INVERS : 0)) ;
+  if (menuVerticalPosition==2) CHECK_INCDEC_GENVAR(event, g_eeGeneral.txCurrentCalibration, -49, 49);
 #endif
 
 #if defined(PCBSKY9X)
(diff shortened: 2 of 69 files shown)


dsbeach:
I'll test this with my scripts.  However, I suspect it will cause me some grief in that I have 2.0 Lua telemetry scripts that are dependent on being able to process the EXIT key. I can code them a different way, but what is the reasoning behind disallowing EXIT when it was available in 2.0?

In my case the 'top level' telemetry screen uses the ENT key to enable modification of run-time parameters.  When desired parameters are set, the EXIT key returns to the 'top level' telemetry screen.

projectkk2glider:
The script will see the EXIT key, but the main firmware will also see it. That means that in case of telemetry screen the radio will exit telemetry screen and return to the main menu. That is what @kilrah proposed in https://github.com/opentx/opentx/issues/2469#issuecomment-122671830 and how the radio reacts in other normal telemetry screens.

kilrah:
Actually I'd also quite like EXIT short. 
I was considering toggling between display and config using PAGE, but it actually makes more sense using MENU to go to config and back with EXIT as PAGE could get confusing if you have other screens configured.

projectkk2glider:
OK, I will add EXIT short. But the EXIT LONG has no effect in the main firmware when we are in a telemetry view! Should I add it and make it exit from the telemetry views to the main screen?

dsbeach:
Great! - I'm testing now with allowing EXIT, and all is well. :+1: 

Regarding EXIT LONG - I don't care, but my suggestion is to maintain compatibility with 2.0 if that is an option.

projectkk2glider:
EXIT LONG must stay at the control of the main firmware, because it is the only way to exit from the Lua telemetry view it if miss-behaves and no other telemetry views are defined!

dsbeach:
Something not quite right with the latest commit.

When ENT LONG brings up the telemetry reset menu, it is non-responsive if a Lua script is underneath.  This occurs regardless of whether or not the Lua script is processing key events.

Telemetry reset behaves normally when invoke on top of a telemetry NUMS screen.

projectkk2glider:
I see, there are other issues with popup menus as well. I will try to fix that and make the code more readable. Hold on...

projectkk2glider:
I've gone far, maybe too far :smile: @bsongis  please check and approve my menu renaming. I had to do it, to better understand the menu functioning. 

The code is not complete, there are a few traces still to be removed, but for testing they can stay in (it is easier to see what is happening).

This now works like we discussed above (unless I made some mistake) and looks like it is working OK in simu.

@bsongis please check out the https://github.com/opentx/opentx/commit/9796eab94d9b52dab569ef4b4b3e268c7a58584c especially carefully. I think we don't need `pgm_read_byte()` here. Am I wrong?

radio/src/functions.cpp
```diff
@@ -323,13 +323,13 @@ void evalFunctions()
             newActiveFunctions |= (1 << FUNCTION_INSTANT_TRIM);
             if (!isFunctionActive(FUNCTION_INSTANT_TRIM)) {
 #if defined(GUI)
-              if (g_menuStack[0] == menuMainView
+              if (menuHandlers[0] == menuMainView
```
bsongis: good!

radio/src/gui/9X/menu_general_diaganas.cpp
```diff
@@ -79,24 +79,24 @@ void menuGeneralDiagAna(uint8_t event)
   adcBatt = ((adcBatt * 7) + anaIn(TX_VOLTAGE)) / 8;
   uint32_t batCalV = (adcBatt + adcBatt*(g_eeGeneral.txVoltageCalibration)/128) * 4191;
   batCalV /= 55296;
-  putsVolts(LEN_CALIB_FIELDS*FW+4*FW, MENU_HEADER_HEIGHT+1+4*FH, batCalV, (m_posVert==1 ? INVERS : 0));
+  putsVolts(LEN_CALIB_FIELDS*FW+4*FW, MENU_HEADER_HEIGHT+1+4*FH, batCalV, (menuVerticalPosition==1 ? INVERS : 0));
```
bsongis: good!

radio/src/gui/9X/menu_general_sdmanager.cpp
```diff
@@ -121,8 +121,8 @@ void menuGeneralSdManager(uint8_t _event)
   fno.lfsize = sizeof(lfn);
 
 #if defined(SDCARD)
-  if (s_warning_result) {
-    s_warning_result = 0;
+  if (warningResult) {
+    warningResult = 0;
```
bsongis: good

radio/src/gui/9X/menu_general_sdmanager.cpp
```diff
@@ -157,50 +157,50 @@ void menuGeneralSdManager(uint8_t _event)
     case EVT_KEY_FIRST(KEY_RIGHT):
     case EVT_KEY_FIRST(KEY_ENTER):
     {
-      if (m_posVert > 0) {
-        vertpos_t index = m_posVert-1-s_pgOfs;
+      if (menuVerticalPosition > 0) {
+        vertpos_t index = menuVerticalPosition-1-s_pgOfs;
```
bsongis: s_pgOfs needs you ;)

projectkk2glider:
>Will it work correctly on gruvin9x board? I am afraid no! 

Please elaborate, I can't find why!

radio/src/gui/9X/popups.cpp
```diff
@@ -139,28 +139,27 @@ void (*popupFunc)(uint8_t event) = NULL;
...
 
-  uint8_t display_count = min<uint8_t>(s_menu_count, MENU_MAX_LINES);
+  uint8_t display_count = min<uint8_t>(popupMenuNoItems, POPUP_MENU_MAX_LINES);
   uint8_t y = (display_count >= 5 ? MENU_Y - FH - 1 : MENU_Y);
   drawFilledRect(MENU_X, y, MENU_W, display_count * (FH+1) + 2, SOLID, ERASE);
   lcd_rect(MENU_X, y, MENU_W, display_count * (FH+1) + 2);
 
   for (uint8_t i=0; i<display_count; i++) {
-    lcd_putsAtt(MENU_X+6, i*(FH+1) + y + 2, s_menu[i], s_menu_flags);
+    lcd_putsAtt(MENU_X+6, i*(FH+1) + y + 2, popupMenuItems[i], 0);
```
bsongis: Here if you give the flag BSS you will display a string from AVR RAM, but if you don't you will display a string at the same address, but from AVR FLASH!

bsongis: Of course your change is good on ARM boards, because the addresses values for FLASH and RAM cannot be the same!

projectkk2glider: I see now, thanks for the explanation, I will fix it later.

bsongis:
May I change a little bit this pull request? It's difficult to read. I would like to commit directly the names change and then review the real differences.

radio/src/gui/9X/menu_model_limits.cpp
```diff
@@ -130,16 +130,16 @@ void menuModelLimits(uint8_t event)
...
   for (uint8_t i=0; i<LCD_LINES-1; i++) {
     coord_t y = MENU_HEADER_HEIGHT + 1 + i*FH;
-    uint8_t k = i+s_pgOfs;
+    uint8_t k = i+menuVerticalOffset;
 
     if (k==NUM_CHNOUT) {
       // last line available - add the "copy trim menu" line
       uint8_t attr = (sub==NUM_CHNOUT) ? INVERS : 0;
-      lcd_putsAtt(CENTER_OFS, y, STR_TRIMS2OFFSETS, s_noHi ? 0 : attr);
+      lcd_putsAtt(CENTER_OFS, y, STR_TRIMS2OFFSETS, NO_HIGHLIGHT() ? 0 : attr);
```
bsongis: good!

radio/src/gui/9X/menu_model_limits.cpp
```diff
@@ -130,16 +130,16 @@ void menuModelLimits(uint8_t event)
...
       if (attr) {
         s_editMode = 0;
         if (event==EVT_KEY_LONG(KEY_ENTER)) {
-          s_noHi = NO_HI_LEN;
+          START_NO_HIGHLIGHT();
```
bsongis: :+1:

radio/src/gui/9X/menus.cpp
```diff
@@ -36,43 +36,43 @@
...
-void chainMenu(MenuFuncP newMenu)
+void chainMenu(menuHandlerFunc newMenu)
 {
-  g_menuStack[g_menuStackPtr] = newMenu;
+  menuHandlers[menuLevel] = newMenu;
   menuEvent = EVT_ENTRY;
 }
 
-void pushMenu(MenuFuncP newMenu)
+void pushMenu(menuHandlerFunc newMenu)
```
bsongis: First letter should be upper case

radio/src/gui/Taranis/lcd.cpp
```diff
@@ -225,20 +225,11 @@ void lcd_putsnAtt(coord_t x, coord_t y, const pm_char * s, uint8_t len, LcdFlags
...
-    unsigned char c;
-    switch (flags & (BSS+ZCHAR)) {
-      case BSS:
-        c = *s;
-        break;
-#if !defined(BOOT)
-      case ZCHAR:
-        c = idx2char(*s);
-        break;
+#if defined(BOOT)
```
bsongis: Yes it is better!

radio/src/main_arm.cpp
```diff
@@ -99,6 +99,40 @@ void checkEeprom()
   }
 }
 
+void handleGui(uint8_t event) {
```
bsongis: I did something similar in 'next', it would be good to compare both versions!

radio/src/sdcard.cpp
```diff
@@ -96,22 +96,24 @@ bool listSdFiles(const char *path, const char *extension, const uint8_t maxlen,
     memset(reusableBuffer.modelsel.menu_bss[0], 0, MENU_LINE_LENGTH);
   }
 
-  s_menu_count = 0;
-  s_menu_flags = BSS;
+  popupMenuNoItems = 0;
+#if !defined(CPUARM)
+  popupMenuFlags = BSS;
```
bsongis: It could be a #define to avoid the #if here

bsongis:
I reviewed everything. The remarks I had are only cosmetics.

PR #3091, opened 2015-11-22
Use new TableLayout for Logical switches, Outputs and Special functions

### LINKED ISSUE #2861: Companion: let the LSs’ parameters line always visible
Th parameters line doesn’t stay visible if you have to use the ‘elevator’ to go down in the list of LSs. Personally I tend to have a hesitation about the last two items when I can’t see them (which one is duration, which one is delay?).

Visible:
parameters line

Not visible:
parameters line 2

projectkk2glider:
Closes #2861

### DIFF
--- a/companion/src/helpers.cpp
+++ b/companion/src/helpers.cpp
@@ -995,3 +995,38 @@ QStringList extractLatLon(const QString & position)
   }
   return result;
 }
+
+TableLayout::TableLayout(QWidget * parent, int rowCount, const QStringList & headerLabels)
+{
+  tableWidget = new QTableWidget(parent);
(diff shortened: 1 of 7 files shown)


projectkk2glider:
There is one downside to this new system, it is slower than the current on. The time to open model edit window is longer. 

I have spent some time with the profiler and basically the main slowdown is when adding items into the QTableWidget. I can't figure out what to do to improve it.

griebd:
I don't remember exactly anymore, but in logsdialog.cpp I tried a lot to improve QTableWidget (logTable) when filling it up. Look at lines 485 forward... :wink:

bsongis:
Thanks!

PR #3097, opened 2015-11-23
Projectkk2glider/lua additions

### LINKED ISSUE #2921: Simulator 2.1.3 - getValue("pilot-latitude") and getValue("pilot-latitude") return always 0
In Companion Simulator 2.1.3, getValue("pilot-latitude") and getValue("pilot-latitude") return always 0 even with valid simulated GSP latitude and longitude.

GPS coord 1 : 46.515826, 6.999142 -> converted for : 27909495 -2143284163
GPS coord 2 : 46.510178, 6.992867 -> converted for : 27906106 -2143287928

Distance between point : 790m (OK in simulator).

### LINKED ISSUE #3062: Close any Lua open files when radio powers off
A script might have a file open (by using io library). That file should be closed before the radio powers off.

### LINKED ISSUE #3095: Lua getDateTime() returns integer values. I am suggesting an enhancement.
When we use the getDateTime() routine in lua scripts, the table that is created is integer values. The means that if we are putting together a string like YYYYMMDD, there are no leading zeros on values below 10. January 1, 2015 would come together as 201511. The question is, do we want to fix this using lua, which I have a routine written for, or do we want to fix this in the opentx model code. I have written and tested the following code in radio/src/lua/api_general.cpp. 

```c
static void luaPushDateTimeStr(lua_State *L, uint32_t year, uint32_t mon, uint32_t day,
                            uint32_t hour, uint32_t min, uint32_t sec)
{
  char stryear[5];
  char strmon[3];
  char strday[3];
  char strhour[3];
  char strmin[3];
  char strsec[3];
  sprintf(stryear, "%lu", year);
  if (mon<10) sprintf(strmon, "0%lu", mon); else 	sprintf(strmon, "%lu", mon);
  if (day<10) sprintf(strday, "0%lu", day); else sprintf(strday, "%lu", day);
  if (hour<10) sprintf(strhour, "0%lu", hour); else sprintf(strhour, "%lu", hour);
  if (min<10) sprintf(strmin, "0%lu", min); else sprintf(strmin, "%lu", min);
  if (sec<10) sprintf(strsec, "0%lu", sec); else sprintf(strsec, "%lu", sec);
  lua_newtable(L);
  lua_pushtablestring(L, "year", stryear);
  lua_pushtablestring(L, "mon", strmon);
  lua_pushtablestring(L, "day", strday);
  lua_pushtablestring(L, "hour", strhour);
  lua_pushtablestring(L, "min", strmin);
  lua_pushtablestring(L, "sec", strsec);
}

/*luadoc
@function getDateTimeStr()

Returns current system date and time that is kept by the RTC unit as a string with leading zeros

@retval table current date and time, table elements:
 * `year` year
 * `mon` month
 * `day` day of month
 * `hour` hours
 * `min` minutes
 * `sec` seconds
*/

static int luaGetDateTimeStr(lua_State *L)
{
  struct gtm utm;
  gettime(&utm);
  luaPushDateTimeStr(L, utm.tm_year + 1900, utm.tm_mon + 1, utm.tm_mday, utm.tm_hour, utm.tm_min, utm.tm_sec);
  return 1;
}
```

As a secondary issue, I am compiling this on a 32 bit system:
```
isaac@isaac-ThinkPad-T61:~/myopentx/opentx/companion/lbuild$ uname -m
i686
isaac@isaac-ThinkPad-T61:~/myopentx/opentx/companion/lbuild$ getconf LONG_BIT
32
```

If I use %lu in the sprintf command above, I get a warning when I compile Companion:
```
/home/isaac/myopentx/opentx/companion/src/../../radio/src/lua/api_general.cpp:168:31: warning: format ‘%lu’ expects argument of type ‘long unsigned int’, but argument 3 has type ‘uint32_t {aka unsigned int}’ [-Wformat]
```
and no warning when I compile the Model.

If I change it to %u, I get no warning when compiling Companion and this warning when compiling Model (make PCB=TARANIS LUA=YES):
```
lua/api_general.cpp:168:30: warning: format '%u' expects argument of type 'unsigned int', but argument 3 has type 'uint32_t {aka long unsigned int}' [-Wformat]
```

Looks like there are two issues here.
1) Do we want to solve the leading zero problem on the getDateTime with a new getDateTimeStr function in the radio code, or would we rather keep the solution on the lua side. I am honestly not sure which would save more space, but know that we want to keep the solution in the safest most efficient place. If we decide to go with this, I will request a pull with my new code after completing my testing. It's working in the simulator but I want to test it in my radio. 

2) What is the proper sprintf conversion for uint32_t? I can only find %u and %lu, and either one or the other compiles throws a warning.

projectkk2glider:
Closes #3095, closes #3062, closes #2921.

**WARNING**: none of this was tested on the radio! I only did the test in the `simu` and there everything worked (I only tried `string.format()` for string library).

### DIFF
--- a/radio/src/Makefile
+++ b/radio/src/Makefile
@@ -399,7 +399,12 @@ IMG2LBM = -$(PYTHON) ../util/img2lbm.py
 BIN2LBM = -$(PYTHON) ../util/bin2lbm.py
 FONT2PNG = -$(PYTHON) ../util/font2png.py
 TRANSLATE = -$(PYTHON) ../util/translate.py
-VERSION = 2.1.6$(OPENTX_VERSION_SUFFIX)
+
+VERSION_MAJOR = 2
+VERSION_MINOR = 1
+VERSION_REVISION  = 6
+
+VERSION = $(VERSION_MAJOR).$(VERSION_MINOR).$(VERSION_REVISION)$(OPENTX_VERSION_SUFFIX)
 
 THIRDPARTY = thirdparty
 COOSDIR = $(THIRDPARTY)/CoOS
@@ -1538,6 +1543,10 @@ stamp_header:
 	@echo "#define DATE_STR \"`date +%Y-%m-%d | tr -d '\r\n'`\"" >> stamp-opentx.h
 	@echo "#define TIME_STR \"`date +%H:%M:%S | tr -d '\r\n'`\"" >> stamp-opentx.h
 	@echo "#define VERS_STR \"$(VERSION)\"" >> stamp-opentx.h
+	@echo "" >> stamp-opentx.h
+	@echo "#define VERSION_MAJOR     $(VERSION_MAJOR)" >> stamp-opentx.h
+	@echo "#define VERSION_MINOR     $(VERSION_MINOR)" >> stamp-opentx.h
+	@echo "#define VERSION_REVISION  $(VERSION_REVISION)" >> stamp-opentx.h
 
 opentx-stamp:
 	@echo "#define DATE_STR \"`date +%Y-%m-%d`\"" > stamp-opentx.txt
--- a/radio/src/Makefile
+++ b/radio/src/Makefile
@@ -1547,11 +1547,13 @@ stamp_header:
 	@echo "#define VERSION_MAJOR     $(VERSION_MAJOR)" >> stamp-opentx.h
 	@echo "#define VERSION_MINOR     $(VERSION_MINOR)" >> stamp-opentx.h
 	@echo "#define VERSION_REVISION  $(VERSION_REVISION)" >> stamp-opentx.h
+	@echo "#define GIT_STR \"$(shell git rev-parse HEAD)\"" >> stamp-opentx.h
 
 opentx-stamp:
 	@echo "#define DATE_STR \"`date +%Y-%m-%d`\"" > stamp-opentx.txt
 	@echo "#define TIME_STR \"`date +%H:%M:%S`\"" >> stamp-opentx.txt
 	@echo "#define VERS_STR \"$(VERSION)\"" >> stamp-opentx.txt
+	@echo "#define GIT_STR \"$(shell git rev-parse HEAD)\"" >> stamp-opentx.txt
 	@cat stamp-opentx.txt
 
 translations/en.h: translations/en.h.txt
--- a/radio/src/Makefile
+++ b/radio/src/Makefile
@@ -1036,7 +1036,7 @@ ifeq ($(PCB), TARANIS)
     CPPSRC += lua/interface.cpp lua/api_general.cpp lua/api_lcd.cpp lua/api_model.cpp
     LUASRC = $(LUADIR)/lapi.c $(LUADIR)/lcode.c $(LUADIR)/lctype.c $(LUADIR)/ldebug.c $(LUADIR)/ldo.c $(LUADIR)/ldump.c $(LUADIR)/lfunc.c $(LUADIR)/lgc.c $(LUADIR)/llex.c $(LUADIR)/lmem.c \
            $(LUADIR)/lobject.c $(LUADIR)/lopcodes.c $(LUADIR)/lparser.c $(LUADIR)/lstate.c $(LUADIR)/lstring.c $(LUADIR)/ltable.c $(LUADIR)/lrotable.c $(LUADIR)/ltm.c $(LUADIR)/lundump.c $(LUADIR)/lvm.c $(LUADIR)/lzio.c \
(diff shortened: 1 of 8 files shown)


radio/src/Makefile
```diff
@@ -399,7 +399,12 @@ IMG2LBM = -$(PYTHON) ../util/img2lbm.py
 BIN2LBM = -$(PYTHON) ../util/bin2lbm.py
 FONT2PNG = -$(PYTHON) ../util/font2png.py
 TRANSLATE = -$(PYTHON) ../util/translate.py
-VERSION = 2.1.6$(OPENTX_VERSION_SUFFIX)
+
+VERSION_MAJOR = 2
+VERSION_MINOR = 1
+VERSION_REVISION  = 6
+
+VERSION = $(VERSION_MAJOR).$(VERSION_MINOR).$(VERSION_REVISION)$(OPENTX_VERSION_SUFFIX)
```
bsongis: Yes it's a good change, but I am almost sure it breaks the compilation server!

radio/src/Makefile
```diff
@@ -1538,6 +1543,10 @@ stamp_header:
 	@echo "#define DATE_STR \"`date +%Y-%m-%d | tr -d '\r\n'`\"" >> stamp-opentx.h
 	@echo "#define TIME_STR \"`date +%H:%M:%S | tr -d '\r\n'`\"" >> stamp-opentx.h
 	@echo "#define VERS_STR \"$(VERSION)\"" >> stamp-opentx.h
+	@echo "" >> stamp-opentx.h
+	@echo "#define VERSION_MAJOR     $(VERSION_MAJOR)" >> stamp-opentx.h
+	@echo "#define VERSION_MINOR     $(VERSION_MINOR)" >> stamp-opentx.h
+	@echo "#define VERSION_REVISION  $(VERSION_REVISION)" >> stamp-opentx.h
```
bsongis: Yes, it's good. It makes me think we should add the git hash, date / time are not enough!
Something like:
```
@echo "#define GIT_STR \"$(shell git rev-parse HEAD)\"" >> $@
```

bsongis:
Thanks for this pull request. I would like to know how much RAM it uses to add the strings lib, but I am rather confident it will be OK

*Also the compilation server needs to be adapted before I merge (both Windows and Linux)*

radio/src/Makefile
projectkk2glider: I am sure of that too!

projectkk2glider:
Initial test show that at least nothing is broken and strings library as well as io close is working: https://github.com/opentx/opentx/issues/3095#issuecomment-159314512

projectkk2glider:
The new version.sh is on the jenkins server (only in nightly-21). Old file saved in version-old.sh. It should work on master (aka old version) and in the new branch.

projectkk2glider:
The new stuff does not use any more RAM. Tested master and this branch by running a simple script that does not use new stuff (also one mix script running) and the free ram is in both cases 56320b.

dsbeach:
@projectkk2glider - How is free ram monitored? I'd love to have that capability.

bsongis:
@projectkk2glider Thanks for your confirmation on the RAM side!

bsongis:
I will fix the Windows server now

bsongis:
Ok it should be good. I did something compatible with both 2.1.x and 2.2.x
Would you port all these changes to 'next'?

indavis:
I tested a couple of things tonight. 

1) String functions in lua are working just fine. Same test code I was using in the simulator.
```
local ver, radio, maj, minor, rev = getVersion()
io.write(file, "Version="..ver.."\nRadio Version="..radio.."\nVersion Major="..maj.."\nVersion Minor="..minor.."\nRevision Number="..rev.."\nVersion Len="..string.len(ver).."\nLeft 2 of ver="..string.sub(ver,1,3).."\nUpper of Radio Version="..string.upper(radio).."\n")

Result:

Version=2.1.6
Radio Version=taranis
Version Major=2
Version Minor=1
Revision Number=6
Version Len=5
Left 2 of ver=2.1
Upper of Radio Version=TARANIS
```
2) Closing of open files. I purposely left the file open from the initial write and powered off the radio. Works perfectly. 

I need to rewrite some code to test the pilot lat and lon and will test that as soon as I can.

projectkk2glider:
>How is free ram monitored? I'd love to have that capability.

Very unscientific: The radio reports amount of the free heap memory on statistics page (long press ENT on main scree, Statistics, key PLUS). 

I use that value and do two identical runs on the radio one with older firmware, one with new. You must also remember that when the memory gets taken from the heap it is never returned back to the system - it is under memory allocator supervision. That is why the free heap number can never increase in the firmware lifetime, even if all allocated memory is freed.

indavis:
Just another update. I have been testing this all week in my Taranis, and everything has been working as expected. The addition of the string library has allowed me to write a file in csv format (on or off the radio), and then read it back into the radio and parse out the values. This is really nice functionality.

PR #3105, opened 2015-11-25
Lua addition ported from master to next

### LINKED ISSUE #2921: Simulator 2.1.3 - getValue("pilot-latitude") and getValue("pilot-latitude") return always 0

### LINKED ISSUE #3062: Close any Lua open files when radio powers off

### LINKED ISSUE #3095: Lua getDateTime() returns integer values. I am suggesting an enhancement.

### LINKED ISSUE #3104: Port #3097 to the `next` branch
Port #3097 to the `next` branch

projectkk2glider:
Git commit ID added to stamp
Fixes #2921: getValue() now also returns pilot position
Fixes #3062: Close any Lua open files when radio powers off
Fixes #3095: Lua string library added (only 6k bytes of flash used)
Re #3095: luaGetVersion() also returns version as numbers
(Ported from master 7ef790d3..4d0bd14ad and squashed into one commit)

Closes  #3104

All commits from #3097 squashed into one and applied to the next branch. 

I've also added back `opent-stamp` Makefile target that I foolishly removed in 32d0e3890af7bc1c2fb23e48a7a4ddf68400d655 (I thought it was not used, but now I see that it is used in the release process and by Companion when checking for new version).

The Jenkins version.sh needs updating, but I have no permission to nighlty-22 right now. There is also git id macro that needs to be tested on Windows (see comment below)


companion/src/cmake/Macros.cmake
```diff
@@ -24,3 +24,15 @@ MACRO (NOW RESULT)
     ENDIF (WIN32)
 ENDMACRO (NOW)
 
+MACRO (GIT_ID RESULT)
+    IF (WIN32)
+        EXECUTE_PROCESS(COMMAND "cmd" " /C git rev-parse HEAD" OUTPUT_VARIABLE ${RESULT})
```
projectkk2glider: This command was not tested!

companion/src/cmake/Macros.cmake
```diff
@@ -24,3 +24,15 @@ MACRO (NOW RESULT)
...
+        string(REGEX REPLACE "(..):(..).*" "\\1:\\2:00" ${RESULT} ${${RESULT}})
```
projectkk2glider: This also needs adaptation!

companion/src/cmake/Macros.cmake
bsongis: Just a cosmetics comment here, I prefer lowercase for commands such as MACRO or IF or EXECUTE_PROCESS (but not COMMAND)

radio/src/Makefile
```diff
@@ -404,7 +404,11 @@ else
 endif
 TRANSLATE = -$(PYTHON) ../util/translate.py
 
-VERSION = 2.2.0$(OPENTX_VERSION_SUFFIX)
+VERSION_MAJOR    = 2
```
bsongis: Aie, only one space of it will break the compilation server!

companion/src/cmake/Macros.cmake
projectkk2glider: I just copied your code :grimacing:

radio/src/Makefile
projectkk2glider: I should not, the version will be parsed from stamp.h (like it is in nightly-21 already).

companion/src/cmake/Macros.cmake
bsongis: Yes I know. I copied the code from somewhere else to avoid duplicate, but didn't change the style. This will be done later, don't worry

radio/src/Makefile
bsongis: Yes but I meant for the windows server, where cut etc are not available so easily!

PR #3106, opened 2015-11-25
Avoid conditional directives that split up parts of statements.

flaviommedeiros:
A suggestion to compile entire statements and expressions, as suggested by code style guidelines from the Linux Kernel and practitioners.

- https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/CodingStyle#n892
- https://www.cqse.eu/en/blog/living-in-the-ifdef-hell/

It might improve code understanding, maintainability and error-proneness.

### DIFF
--- a/radio/src/thirdparty/GCS_MAVLink/include_v1.0/mavlink_helpers.h
+++ b/radio/src/thirdparty/GCS_MAVLink/include_v1.0/mavlink_helpers.h
@@ -261,6 +261,7 @@ MAVLINK_HELPER uint8_t mavlink_parse_char(uint8_t chan, uint8_t c, mavlink_messa
 	mavlink_message_t* rxmsg = mavlink_get_channel_buffer(chan); ///< The currently decoded message
 	mavlink_status_t* status = mavlink_get_channel_status(chan); ///< The current decode status
 	int bufferIndex = 0;
+	int checker = 0;
 
 	status->msg_received = 0;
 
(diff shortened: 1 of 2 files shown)


radio/src/thirdparty/STM32_USB-Host-Device_Lib_V2.1.0/Libraries/STM32_USB_Device_Library/Class/msc/src/usbd_msc_scsi.c
```diff
@@ -439,13 +439,15 @@ extern uint8_t lunReady[] ;
...
-  if (lun < 2) 
+  check_lun = (lun < 2); 
 #else
-  if (lun < 1) 
-#endif
-    {
+  check_lun = (lun < 1); 
+#endif
+  if (check_lun)
+  {
```
bsongis: No newline here

bsongis:
I am not sure we have a lot of these, but you are right!

flaviommedeiros:
There only a few across the source files and some of them in third party code with similar patterns, such as:

- https://github.com/flaviommedeiros/opentx/commit/8c45305b47899c1e4e1ed8a7d9eca166e024684c

radio/src/thirdparty/GCS_MAVLink/include_v1.0/mavlink_helpers.h
```diff
@@ -278,13 +279,13 @@ MAVLINK_HELPER uint8_t mavlink_parse_char(uint8_t chan, uint8_t c, mavlink_messa
...
+		checker = status->msg_received; 
 /* Support shorter buffers than the
    default maximum packet size */
 #if (MAVLINK_MAX_PAYLOAD_LEN < 255)
-				|| c > MAVLINK_MAX_PAYLOAD_LEN
+		checker = checker || c > MAVLINK_MAX_PAYLOAD_LEN;
 #endif
-				)
+		if (checker)
 		{
```
bsongis: No newline here as well

kilrah:
Umm... All of these are in third-party libraries, and shouldn't we be leaving those alone? It's not our job to change them, and would likely cause massive conflicts should we update them.

bsongis:
Kilrah is right, we change thirdparties to the minimum, but at least the first fix is in some code already modified.

Issue #2919, opened 2015-09-22
companion 2.1.3 telemetry sensors simulator


wolkstein:
how can i simulate telemetry sensors (values) with taranis simulator in companion2.1.3. i noticed under 2.0x that sensors you request by lua also was filled with an value. if you open this scripts in taranis simulator with 2.1.3 all fields are empty. i know that sensors under 2.1.x need other getValu() calls. but this is the crux. it would be more easy to debug and port lua scripts to 2.1.x if all possibly telemetry sensor data (tables) could be filled by default.

/g
wolke

dsbeach:
I've started on this.  I plan to address GPS, Vario, and FLVS (lipo sensor) to start.

Are there other sensors you need?  If so, in what priority?

FYI - My goal is to be able to have Companion 'play back' sensor values from a saved log file.

bsongis:
Excellent idea, thanks for your help on this!

dsbeach:
I have a design question regarding the use of instance on the Companion Telemetry Simulator dialog box.

As currently implemented there is a single instance variable with a default value of '2'.

Option 1 - Use that instance for all simulated telemetry values

Option 2 - Add an instance text box for each simulated sensor.

I'm leaning toward option 2, but would like to identify appropriate default values and store them in a QSettings object.  Do you agree?

Status wise, the changes are looking good thanks to source code cherry picked from OpenXSensor. GPS, Vario, and Flvss data is flowing nicely!

bsongis:
@dsbeach obviously option 2) is the best :)
In the gtests directory you will find some code as well, don't hesitate to reuse it!

jtaylor2:
@dsbeach I think option 2 is better.  I added the current single instance variable as a quick fix for the D series rssi, swi, A1 and A2 being instance 25 in 2.1.  I planned on coming back and doing your option 2 (and some other improvements), but hadn't gotten to it yet.   I used a default of 2 because that is what was hard coded before my change.  Glad to see you working on it :smile:.

kilrah:
I would personally load all sensors from the simulated model with the ID/instance/name as configured, with a couple of extra sets of boxes you can fill up manually to test discovery etc. 
Then an enable checkbox per sensor, and a global simulator enable.

dsbeach:
I thought about that last night, and agree 100%. I'll investigate that approach today.

Similarly, I hope to make the process for parsing log file data like dates, and GPS positions dependent on the radio settings.

[...]

PR #3124, opened 2015-12-03
Dsbeach/issue2919 telemetry sensors simulator

dsbeach:
Looks like this feature is ready for review, testing, and hopefully acceptance into master.

I opted to use metric values for keyboard input, but the log file replay feature will support imperial formats as input.  I've updated all the dialog field labels to match the transmitter default values.

Please have a look and let me know if you find things that need attention.

David


jtaylor2:
Compiled and tested on openSUSE 13.2.  Very Nice.....  Just a couple of comments.
If I click on the Log Replay Row# Scroll bar position indicator without a log file loaded Companion crashes with an uncaught exception:

```
Qt has caught an exception thrown from an event handler. Throwing
exceptions from an event handler is not supported in Qt. You must
reimplement QApplication::notify() and catch all exceptions there.

Qt has caught an exception thrown from an event handler. Throwing
exceptions from an event handler is not supported in Qt. You must
reimplement QApplication::notify() and catch all exceptions there.

terminate called after throwing an instance of 'std::exception'
  what():  std::exception
Aborted
```

GUI nit - The Log Replay "Stop"  QPushButton always shows depressed.  If you press Play, the Play button goes depressed momentarily and the Stop button goes not depressed momentarily  but then the Play button shows not depressed and the Stop button shows depressed even though it is still playing.

Also I would prefer to have the default for the Simulate checkbox not checked the way it is now so that if you have "Discover New Sensors" running you don't get new sensors created when you bring up the dialog from the three that have default non zero values.  But that's just my opinion and no biggie.

Very nice, a great improvement.

dsbeach:
@jtaylor2 Thanks for testing.  The first two issues you identified appear to be a side-effect of Qt on that platform. On Windows the buttons are disabled until a file is loaded, and the play button stays depressed.  I've got a Fedora virtual machine I can test on today.  It it displays the same behavior it will be easy to address, otherwise I may need your testing assistance to get it resolved.

For my testing, I found it more convenient to default to simulation 'checked'. Does anyone else have an opinion on what the default should be?

jtaylor2:
A1 and A2 are limited to 13.2 V and 1 decimal.  Since they are typically used with a voltage divider to measure pack voltage the maximum should be at least as high as a 6S pack (if not higher 12S?) and decimal 2 (since the sensor can be precision 2).  Interestingly cell, which should never be much over 4.2 has a maximum of 99.99, but of course having too much range doesn't hurt anything.

I compiled your branch on my Windows 7 laptop using Cygwin (POSIX) using Cygwin's QT 4.8.6 and ran it under on Windows 7 using X.  It has the same problems as on my OpenSUSE 13.2 system, crashes Companion if you click on the scrollbar position indicator (you have to click on the position indicator, not anything else on the widget), and the Stop and Play buttons act the same way.

Several things going on right now, but if I get a chance later I'll take a look at it and see if I can figure out why.

dsbeach:
@jtaylor2 - I've fixed the scroll bar press crash (and a couple of other issues related to not having a log file loaded).  Thanks for finding those.

The problem with the stop button was me (and the same on Windows I just didn't recognize it). That will be fixed with the next commit in my branch.

I don't use A1 and A2 and wasn't sure how to handle them.  Based on some experimentation I've decided to add a ratio field.  I will retrieve the ratio when the telemetry dialog is initialized and use that to convert the displayed value to a percentage of range.  I'll have an update with that change before I shut down for the day.

Thanks again for testing - I suspect there are more issues to be found.

jtaylor2:
@dsbeach Not sure just what you mean by ratio for A1 A2 but it seems like overkill.  The only thing that really needs to be done is change the input maximum to 99.99 and the decimal to 2, the way it is for cells.  Everything with it works great, it's just that you can't simulate a voltage higher than 13.2 and one decimal the way it is now.

Jim

jtaylor2:
Forget my last comment about A1, A2, I just looked at the code and saw the way you did it.  Using 19.39 it has to be limited to 13.2 V to stay under 255.

dsbeach:
Today's changes have been committed.  Looking at A1 and A2 I should probably do the same thing for RxBt.  Maybe tomorrow...

jtaylor2:
Just did a compile and quick test on my Windows laptop.  Looks good, I like the way you did the ratio...nice.  Play and Stop buttons look good.  No position indicator when file not loaded so can't click and crash.  Just one question, should all the values (from sensors that are supported) play back from the log?  In my quick test it didn't look like A1 and A2 were playing back.  When I get to my Linux desktop later I'll compile it there and do a better test.

jtaylor2:
@dsbeach Compiled your branch on my OpenSUSE 13.2 system and did some more testing.  A1 and A2 do not replay from log.  Did a little troubleshooting and it looks like the problem is that the column to function name map for A1 and A2 does not match my log header.  My log header is A1(V) and  A2(V).  If I make the change shown in the diff below A1 and A2 replay from my logs. 

Looking at radio/src/logs.cpp it looks like the user selected units get appended to the sensor name:

```
strcat(label, "(");
        strncat(label, STR_VTELEMUNIT+1+3*sensor.unit, 3);
        strcat(label, ")");
```

It looks like the user selected units get appended to the sensor name.


```
diff --git a/companion/src/simulation/telemetrysimu.cpp b/companion/src/simulation/telemetrysimu.cpp
index 935608c..65485e7 100644
--- a/companion/src/simulation/telemetrysimu.cpp
+++ b/companion/src/simulation/telemetrysimu.cpp
@@ -567,8 +567,8 @@ TelemetrySimulator::LogPlaybackController::LogPlaybackController(Ui::TelemetrySi
   colToFuncMap.insert("RxBt(V)", RXBT_V);
   colToFuncMap.insert("RSSI(dB)", RSSI);
   colToFuncMap.insert("SWR", SWR);
-  colToFuncMap.insert("A1", A1);
-  colToFuncMap.insert("A2", A2);
+  colToFuncMap.insert("A1(V)", A1);
+  colToFuncMap.insert("A2(V)", A2);
   colToFuncMap.insert("A3", A3);
   colToFuncMap.insert("A4", A4);
   colToFuncMap.insert("Tmp1(@C)", T1_DEGC);
```

dsbeach:
Testing anyone?  I think I have a reasonable layout for the simulator dialog in both Windows and Linux.   Some UI feedback on Apple hardware would be good.

kilrah:
Pretty good, although I'd still have preferred what I mentioned in the comments for https://github.com/opentx/opentx/issues/2919 :)

One thing, one can click/move the slider of the log player when no log is loaded, and that will cause a crash.

kilrah:
Sorry, disregard the note about the crash, my branch wasn't up to date...

dsbeach:
Can I get a volunteer to review this and to the merge?  I've got nothing to change at this point except keep this branch up with master.

bsongis:
Thanks!

bsongis:
Will you port it to 'next'?

dsbeach:
You bet.  I just setup a next clone.  Might even add a tab for 'Custom'! Thanks.

Issue #3138, opened 2015-12-09
9XR Pro: Inverter on serial input not configured


mikeller:
I came across a (quiet weird looking) bug last night:

Set up: 9XR pro with OrangeRX openLRS module, modified to output serial (telemetry) on pin 5 of the module connector.

I had had openTX 2.1.6 running on my TX, with everything (including telemetry) working fine.

I then flashed er9sky (r217) onto the TX to check something out.

When I reflashed openTX, config was restored with everything working, except 'No Data' for telemetry.

After tracing all connections involved, I reflashed er9sky, went into telemetry config, and ticked 'invert com1'. Reflashed openTX, restored config. Now everything including telemetry works just like it had before I had flashed er9sky for the first time.

It looks like openTX just accepts whatever value (on/off) the inverter on the internal serial input was set to by the previous firmware, and runs with it.

As a suggestion, what about adding an additional option 'FrSky D (inverted)' to the telemetry protocol list, and switch the inverter on or off based on the user's choice?

If nobody is working on this already, I am happy to sink some time into it.

kilrah:
>It looks like openTX just accepts whatever value (on/off) the inverter on the internal serial input was set to by the previous firmware, and runs with it.

I don't see how that could be possible. So you have a transistor inverter in your mod? i.e. the standard mod to make it actually compliant with smart port?

mikeller:
No, I don't, all I've got is a pull down on the connection between the serial Tx and pin 5. The 9XR pro has a configurable inverter on the serial input on pin 5 that can be enabled and disabled in the menu. The code that actually switches it is:

```
void writeMFP()
{
        NVIC_DisableIRQ(TWI0_IRQn) ;
        Rtc_write_pending |= 2 ;
        i2c_check_for_request() ; 
        NVIC_EnableIRQ(TWI0_IRQn) ;
}

void setMFP()
{
        MFPsetting = 0x80 ;
        writeMFP() ;
}

void clearMFP()
{
        MFPsetting = 0 ;
        writeMFP() ;
}
```

It is in `radio/ersky9x/src/sky/sound.cpp`, of all places.

kilrah:
OK, that probably never was included in OpenTX. Nobody in the team actually uses a 9XR-Pro and only one person has one, so it doesn't get much love.

>As a suggestion, what about adding an additional option 'FrSky D (inverted)' to the telemetry protocol list, and switch the inverter on or off based on the user's choice?

Sounds like a good way to implement it :)

[...]

PR #3139, opened 2015-12-10
Invert serial rx

mikeller:
Hi there.

As discussed earlier today, I have added the option for the 9XR pro to receive FrSky D telemetry that is inverted on pin 5 of the module connector, and feed it through its internal inverter to get it to the right format.

I have made this option only available for the 9XR pro. This might be too restrictive, as other sky9x based boards (or potentially the Taranis) might have a switchable on-board inverter too, but I have no way to verify this.

hope this helps, cheers
  Michael Keller


radio/src/Makefile
```diff
@@ -52,7 +52,7 @@ DELAY = 5
 # PCB version
 # Values: 9X, 9X128, 9X2561, 9XR, 9XR128, 9XR2561, GRUVIN9X, MEGA2560, SKY9X, 9XRPRO, AR9X, TARANIS
 # Old values STD and STD128 are mapped to 9X and 9X128 for compatibility
-PCB = 9X
+PCB = 9XRPRO
```
kilrah: Your personal makefile changes shouldn't be committed! Hint: use command line instead for your builds, preferably in a script for convenience

radio/src/Makefile
```diff
@@ -784,6 +784,12 @@ ifeq ($(PCB), $(filter $(PCB), SKY9X 9XRPRO AR9X))
     FLAVOUR = 9xrpro
     CPPDEFS += -DREVX
     LDSCRIPT = targets/sky9x/sam3s8c_flash.ld
+  ifeq ($(NANO), YES)
+    # use newlib-nano for linking
+    NEWLIB_NANO_FLAGS = --specs=nano.specs -u _printf_float
+    CPPDEFS += -DNANO
+  endif
+
```
kilrah: Can somebody comment on NANO for sky9x? If good, why only for 9XR-Pro and not the entire sky9x family?

radio/src/telemetry/frsky.cpp
```diff
@@ -69,7 +69,7 @@ FrskyData frskyData;
 
 #if defined(CPUARM)
 uint8_t telemetryProtocol = 255;
-#define IS_FRSKY_D_PROTOCOL()      (telemetryProtocol == PROTOCOL_FRSKY_D)
+#define IS_FRSKY_D_PROTOCOL()      (telemetryProtocol == PROTOCOL_FRSKY_D || telemetryProtocol == PROTOCOL_FRSKY_D_INVERTED)
```
kilrah: This breaks on all ARM platforms but sky9x (the reason Travis check failed) because you defined `PROTOCOL_FRSKY_D_INVERTED` only for REVX. Same at line 616 below

radio/src/Makefile
```diff
@@ -850,6 +856,8 @@ ifeq ($(PCB), $(filter $(PCB), SKY9X 9XRPRO AR9X))
   ifeq ($(RTCLOCK), YES)
     CPPDEFS += -DRTCLOCK
     CPPSRC += rtc.cpp targets/sky9x/rtc_driver.cpp
+  else ifeq ($(PCB), SKY9X)
+    CPPSRC += targets/sky9x/rtc_driver.cpp
```
kilrah: I would not add this but rather complete the `#if defined(REVX)` you added for this PR by `#if defined(REVX) && defined(RTCLOCK)`

radio/src/gui/9X/menu_model_telemetry.cpp
```diff
@@ -606,7 +606,7 @@ void menuModelTelemetry(uint8_t event)
     switch (k) {
 #if defined(CPUARM)
       case ITEM_TELEMETRY_PROTOCOL_TYPE:
-        g_model.telemetryProtocol = selectMenuItem(TELEM_COL2, y, STR_TELEMETRY_TYPE, "\017FrSky S.PORT\0  FrSky D\0       FrSky D (cable)", g_model.telemetryProtocol, PROTOCOL_TELEMETRY_FIRST, CASE_PCBSKY9X(PROTOCOL_FRSKY_D_SECONDARY) attr, event);
+        g_model.telemetryProtocol = selectMenuItem(TELEM_COL2, y, STR_TELEMETRY_TYPE, "\017FrSky S.PORT\0  FrSky D\0       FrSky D (inv)\0 FrSky D (cable)", g_model.telemetryProtocol, PROTOCOL_TELEMETRY_FIRST, CASE_PCBSKY9X(PROTOCOL_FRSKY_D_SECONDARY) attr, event);
```
kilrah: Good job with understanding how the length works, everybody usually gets that wrong at first :)

bsongis: I don't get why you add a choice here, does it mean there is an FrSky D non inverted protocol?

radio/src/Makefile
bsongis: I think it's not a bad idea to have it available on 9XR-PRO, but I wonder if it's possible to avoid these 4-5 lines in all ARM platforms, perhaps it would be good to write this line elsewhere, no? By the way on 9XR-PRO malloc is only used for printf if I am not mistaken. The debate NANO/No NANO comes when we speak about Lua scripts ...

radio/src/targets/sky9x/rtc_driver.cpp
```diff
@@ -280,3 +280,25 @@ void rtcInit()
   coprocReadData();
 #endif
 }
+
```
bsongis: So it means I didn't put this code anywhere else?

radio/src/telemetry/frsky.cpp
```diff
@@ -629,6 +629,14 @@ void telemetryInit(uint8_t protocol)
   else {
     telemetryPortInit(FRSKY_SPORT_BAUDRATE);
   }
+
+#if defined(REVX)
+  if (g_model.telemetryProtocol == PROTOCOL_FRSKY_D_INVERTED) {
+    setMFP();
+  } else {
```
bsongis: New line here, cf wiki page from Damjan on code style

radio/src/gui/9X/menu_model_telemetry.cpp
kilrah: Well, it's an inverted inversion...
If you use a genuine FrSky module it would work as it does now with (cable), and it supplies inverted serial.
His module is DIY-modded and provides non-inverted serial, which is thus *inverted from the standard* and that's what the new choice means.
At least that's what I understood.

radio/src/targets/sky9x/rtc_driver.cpp
kilrah: I had a quick look and didn't find it. `Rtc_write_pending & 2` is handled in i2cCheck() but never set anywhere.

radio/src/gui/9X/menu_model_telemetry.cpp
bsongis: OK. Shouldn't the new choice be the last one in the list? Or it might break existing setups

radio/src/Makefile
mikeller: Sorry, that was a misunderstanding on my part of how github does pull requests. I assumed the pull request would only cover the changes in the branch, but not in master. So I added my build config to master, and all the changes for the feature in my branch.

radio/src/Makefile
```diff
@@ -784,6 +784,12 @@ ifeq ($(PCB), $(filter $(PCB), SKY9X 9XRPRO AR9X))
     FLAVOUR = 9xrpro
     CPPDEFS += -DREVX
     LDSCRIPT = targets/sky9x/sam3s8c_flash.ld
+  ifeq ($(NANO), YES)
+    # use newlib-nano for linking
+    NEWLIB_NANO_FLAGS = --specs=nano.specs -u _printf_float
```
mikeller: I suspect this has slipped in when I moved my build config between master and next and back. As I said above, please disregard all changes in the Makefile, except for the one in the last changeset.

radio/src/Makefile
mikeller: Won't this make my build fail because `targets/sky9x/rtc_driver.cpp` gets included twice?

radio/src/gui/9X/menu_model_telemetry.cpp
mikeller: Yes, @bsongis is spot on with the why, and the _inverted inversion_. As to being last on the list: I thought the order would be more logical if it was

[protocol/port 1] [protocol/port 2] [protocol/port 2 (inverted)] [protocol/port 3] 

than

[protocol/port 1] [protocol/port 2] [protocol/port 3] [protocol/port 2 (inverted)]

But open for discussion.

radio/src/targets/sky9x/rtc_driver.cpp
mikeller: Yes, as far as I could see, the 2 and 4 flags for `Rtc_write_pending` are not (yet) used anywhere.

radio/src/telemetry/frsky.cpp
mikeller: Yes, you are right, I forgot to make these conditiona. Will fix.

radio/src/telemetry/frsky.cpp
mikeller: Noted. Will be a hard one, as it is different from how it is done at my work. If I start using eclipse, auto format will be very handy.

radio/src/gui/9X/menu_model_telemetry.cpp
kilrah: The problem with not putting it as last item is that anyone who had his mode configured as FrSky D (cable) will see his choice now being interpreted as FrSky D (inverted), his setup won't work anymore, and he won't understand why. 
If you want to be clean and group the settings then your should target your PR for the next major version (2.2), base your branch on `next`, and implement an eeprom conversion routine that handles the shift when upgrading the eeprom format.
Note that if you change it as last item you of course have to move the new item in myeeprom.h too.

radio/src/Makefile
kilrah: It would have worked that way if you had created your branch *before* commtting your makefile changes on master (or alternatively rebased it to a point of master before these changes were committed after the fact)

radio/src/Makefile
kilrah: Yes it seems `rtc_driver.cpp` and `coproc_driver.cpp` actually contain a bunch of duplicated or nearly similar code, not sure why `rtc_driver.cpp` even exists for sky9x. If you're up for cleaning this up, removing dupes and putting the rest in `coproc_driver.cpp`it would be welcome. 
The `#if defined(RTCLOCK)` would then become `#if defined (COPROCESSOR)` and it would all be more clear and clean. Might require a few fixes in different places.
If you decide to fix that it might be wise to do it as a separate step in another branch after this one is merged.

radio/src/gui/9X/menu_model_telemetry.cpp
mikeller: You are right, I had not considered that. Will change. Cleaning up as part of the move to 2.2 sounds like a good plan too.

When I tried to build `next` with my settings it failed, that's why I wrote my patch against `main`. Is there documentation on how features are merged / grafted between `main` and `next`, or what is a good place to ask some questions about that, and the release process?

As an alternative to the implementation of 'FrSky D (inverted)' as a separate 'protocol', there would be the option to add a checkbox 'invert internal serial port' and make this independent of the protocol selection. Thoughts?

radio/src/Makefile
mikeller: On 11/12/15 18:46, Andre Bernet wrote:
>
> It would have worked that way if you had created your branch /before/
> commtting your makefile changes on master (or alternatively rebased it
> to a point of master before these changes were committed after the fact)

Definitely. That was not what I wanted though, I wanted to make the 
changes to the makefile permanent in my own branch, so that I could 
build with my settings.

Having the makefile under revision control is a bit problematic anyway, 
or at least it would be nice for people building from source to have a 
config file where they could store their build settings that did not 
interfere with version control.

(I have not done much with git in the past, so I will first have to get 
a grip on git-ish features like rebase.)


-- 
GCS$/CC/E/IT d- s+ a C++ UL+++/S++ P L++ E-
W++ N o? K? w O(++) M-- V+ PS+ PE+ Y? PGP+ t
5? X R tv b++ DI++ D++ G e+++ h---- r+++ y+++

radio/src/Makefile
mikeller: Oh. When I mentioned `coproc_driver.cpp` I meant this as a suggestion for renaming `rtc_driver.cpp`, I wasn't aware that this file already exists. ;-)

Happy to do some cleaning up, and yes definitely *not* as part of this patch. I would think the cleanup will probably best be done against `next` anyway? (I really first need to understand more about the release process for this project.)

kilrah: >  I would think the cleanup will probably best be done against next anyway?

Yes absolutely, seems you got it quite well ;)

radio/src/Makefile
kilrah: It will obviously always be a mess to keep mods in the makefile. Whch is again why it's not a good idea to do it... it also limits you to a single config.

I instead use multiple scripts for the various configs I use regularly in the form of 
```
make clean
make PCB=TARANIS PCBREV=REVPLUS HELI=YES GVARS=YES TRANSLATIONS=EN DEBUG=NO LUA=YES MIXERS_MONITOR=YES TIMERS=3 USB=JOYSTICK CLI=NO DANGEROUS_MODULE_FUNCTIONS=NO TARANIS_INTERNAL_PPM=NO

dfu-util-static -a 0 -d 0483:df11 --dfuse-address 0x08000000 -D opentx.bin
```
for example for the Taranis, that cleans, builds the version I want, then flashes it automatically.

radio/src/gui/9X/menu_model_telemetry.cpp
kilrah: There is no hard rule, but a compromise between the limited resources and timing... We don't have enough devs to have branch maintainers who backport things or otherwise do work twice so we try to avoid. So for some time after a major release we'll continue working in `master`, until `next` starts having too many differences to be able to be regularly rebased on current master and thus needs duplication of commits like now. Anything that changes eeprom storage format needs to be done in next, so that pretty much decides when it happens.

Making an "Invert serial port" checkbox would probably be better in the long run indeed, but the ability to do that now on `master` precisely depends on whether we have a free bit in the model data structure to store that setting. If not, then doing it this way would need to target `next`

Of course we want to avoid including "temporary behaviors" where the same thing gets done differently in 2 successive versions to avoid confusing users, so we don't want to have it done with a new protocol choice now, and then change to the independent checkbox in 2.2.

bsongis: I would vote for the checkbox personnally. Perhaps reusing a bit in channelsStart?

radio/src/Makefile
bsongis: coproc_driver.cpp doesn't only manage the RTC!

kilrah: Yeah, but on Sky9x RTC is always managed by/part of the coproc...
Although it might actually change if we ever implement support for optional RTC on AR9x.

Still, the currently duplicated code isn't very clean...

bsongis: I have to admit I didn't take care too much of the sky9x!

radio/src/telemetry/frsky.cpp
```diff
@@ -283,7 +286,14 @@ void telemetryWakeup()
 {
 #if defined(CPUARM)
   uint8_t requiredTelemetryProtocol = MODEL_TELEMETRY_PROTOCOL();
+#if defined(REVX)
+  uint8_t requiredSerialInversion = g_model.moduleData[EXTERNAL_MODULE].invertedSerial;
+  if (telemetryProtocol != requiredTelemetryProtocol
```
bsongis: On the same line, or it's more difficult to read!

bsongis:
It's all good for me, just the little cosmetics issue I saw, perhaps you may fix it directly in 'master'? Thanks for this good job!

radio/src/telemetry/frsky.cpp
mikeller: Argh, again my work habits showing through. (We have a rule that no line may be longer than screen width.)
Will try to improve my context switching.

mikeller:
You're welcome. Happy to see that I can add some value to this project.

One more question, just to be absolutely clear: The development process is that now these changes will have to be manually grafted into `next`? I am happy to do the work, I just want to make sure I am not messing doing work that cannot be used, or messing up a merge down the line.

bsongis:
Yes, manually, the 2 branches have diverged a lot now

projectkk2glider:
This pull is for #3138

PR #3142, opened 2015-12-13
Ported simulator enhancements to next

dsbeach:
Tested on both Taranis (looks OK) and Horus simulators (GPS issue - assumed not related to telemetry simulator) in companion.


bsongis:
Thanks!

PR #3148, opened 2015-12-14
Changes to send ASpd as knots

### LINKED ISSUE #3147: OpenTX 2.1 - telemetry simulator support for Aspd as knots
Update the telemetry simulator to handle Aspd as knots.

dsbeach:
Fixes #3147 - telemetry simulator changes to send ASpd as knots in telemetry frame. Simulator input (text box) will be kph.  Added log playback support for logging in knots, kph, or mph.

### DIFF
--- a/companion/src/simulation/telemetrysimu.cpp
+++ b/companion/src/simulation/telemetrysimu.cpp
@@ -314,7 +314,7 @@ void TelemetrySimulator::generateTelemetryFrame()
 
     case 17:
       if (ui->aspeed->value() > 0)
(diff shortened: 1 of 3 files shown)


projectkk2glider:
The code looks good to me.

projectkk2glider:
@dsbeach I would prefer (in the future) if you also used issue reference in the commit messages like so:
```
Fixes #3147: Changes to send ASpd as knots
```
or even (if the commit is related to the issue but does not solve it):
```
Re #3147: Changes to send ASpd as knots
```

This makes life much easier when we update the changelog. Thanks.

dsbeach:
My mistake, I put it in the pull request, but not in the commit message. I'm guessing it is best to have it in both.

PR #3153, opened 2015-12-14
Graft of 'invert_serial_rx' onto next.

mikeller:
I wasn't sure if we'd want to leave

```uint8_t failsafeMode:7;
uint8_t invertedSerial:1;```

the way it is, or move `invertedSerial` into a better spot and do a conversion. Let me know.

### CODE REVIEW

File: radio/src/telemetry/frsky.cpp [hunk 1]
```diff
@@ -622,6 +631,15 @@ void telemetryInit(void)
   else {
     telemetryPortInit(FRSKY_SPORT_BAUDRATE);
   }
+
+#if defined(REVX)
```
bsongis: I remember it introduced a compilation error on VC++, it should be:
```
#if defined(REVX) && !defined(SIMU)
```

mikeller: Drats I can totally see why. 'learn how to build and use the simulator' is somewhere on my TODO list.

bsongis:
Thanks!

mikeller:
De rien.

PR #3169, opened 2015-12-21
Switch to CMake

bsongis:
We can move little after little, the Makefile will remain operational until everything is ported to CMake.
All comments welcome!


projectkk2glider:
Wow, looks like a lot of work! Did you write Cmake files by hand or do you have any helper tool? 

Looks nice, but I haven't done any tests yet, just skimmed over the changes.

bsongis:
EVerything by hand, yes. I don't see how this could be automatically done!

bsongis:
CLion has nice helpers for CMake, the automatic formatting is done automatically. Also it is able to hide all portions of code which are not used (#ifdef), then you change the context, or modify a CMake option and it hides other parts.

PR #3174, opened 2015-12-26
Dsbeach/issue3141 fx key to reload scripts

### LINKED ISSUE #3141: Simulator 2.1.6 adding a Fx key for reloading it
For a faster testing when making scrips could it be possible to add a Fx key for reloading simulator in place of quiting cliquing retying testing values .... ?

dsbeach:
Fixes #3141 Fx key to reload Lua scripts

Added PF7 - Reload Lua Scripts

Implemented via one liner - luaState = INTERPRETER_RELOAD_PERMANENT_SCRIPTS which forces a re-initialization of the Lua environment.

Very handy for script modification and debugging.


dsbeach:
Thanks - will port to 'next' today.

PR #3193, opened 2016-01-02
First version of multi module support for Taranis

schwabe:
Since DSM2 uses the same framing/serial protocol (Serial with 125k)
the multi module code heavily reuses that code and depends on it being enabled.

This first version only implements the protocol on the taranis.

Telemetry code is currently untested


schwabe:
I do not expect this code to merge as is. It is something I coded for fun and try to make it works. However I would like have feedback on if or in what state this code would be considered for inclusion.

The hardware module to make this work is here: https://github.com/pascallanger/DIY-Multiprotocol-TX-Module

bsongis:
Nice project! I would prefer to include this in the 'next' branch (2.2), rather than 'master' (2.1) though

radio/src/Makefile
```diff
@@ -228,7 +228,11 @@ PXX = NO
 
 # DSM2 (Spektrum) protocol
 # Values = NO, SERIAL, PPM
-DSM2 = NO
```
bsongis: Don't change this!

radio/src/Makefile
```diff
@@ -1051,6 +1055,9 @@ ifeq ($(PCB), TARANIS)
   ifeq ($(DSM2), PPM)
     CPPSRC += pulses/dsm2_arm.cpp
   endif
+  ifeq ($(MULTIMODULE), YES)
+    CPPSRC += pulses/multi_arm.cpp
```
bsongis: The CPPDEFS could be added here as well

radio/src/Makefile
```diff
@@ -228,7 +228,11 @@ PXX = NO
 
 # DSM2 (Spektrum) protocol
 # Values = NO, SERIAL, PPM
-DSM2 = NO
+DSM2 = PPM
+
+# Multimodule support
+# (http://www.rcgroups.com/forums/showthread.php?t=2165676)
+MULTIMODULE = NO
```
bsongis: You will need to add the option in the radio/src/CMakeLists.txt (in the 'next' branch)

radio/src/gui/Taranis/menu_model_setup.cpp
```diff
@@ -204,12 +204,18 @@ int getSwitchWarningsCount()
 #define IS_D8_RX(x)                       (g_model.moduleData[x].rfProtocol == RF_PROTO_D8)
 #define IF_EXTERNAL_MODULE_XJT(x)         (IS_MODULE_XJT(EXTERNAL_MODULE) ? (uint8_t)x : HIDDEN_ROW)
 #define INTERNAL_MODULE_CHANNELS_ROWS     IF_INTERNAL_MODULE_ON(1)
-#define EXTERNAL_MODULE_CHANNELS_ROWS     IF_EXTERNAL_MODULE_ON((IS_MODULE_DSM2(EXTERNAL_MODULE) || IS_MODULE_CROSSFIRE(EXTERNAL_MODULE)) ? (uint8_t)0 : (uint8_t)1)
+#define EXTERNAL_MODULE_CHANNELS_ROWS     IF_EXTERNAL_MODULE_ON((IS_MODULE_DSM2(EXTERNAL_MODULE) || IS_MODULE_CROSSFIRE(EXTERNAL_MODULE) || IS_MODULE_MULTIMODULE(EXTERNAL_MODULE)) ? (uint8_t)0 : (uint8_t)1)
```
bsongis: We should add another definition, something like IS_CHANNEL_COUNT_MODIFIABLE(protocol) to avoid to duplicate this test which becomes longer and longer ...

radio/src/gui/Taranis/menu_model_setup.cpp
```diff
@@ -204,12 +204,18 @@ int getSwitchWarningsCount()
...
 #define TRAINER_CHANNELS_ROWS()           IF_TRAINER_ON(1)
 #define PORT_CHANNELS_ROWS(x)             (x==INTERNAL_MODULE ? INTERNAL_MODULE_CHANNELS_ROWS : (x==EXTERNAL_MODULE ? EXTERNAL_MODULE_CHANNELS_ROWS : TRAINER_CHANNELS_ROWS()))
 #define FAILSAFE_ROWS(x)                  (HAS_RF_PROTOCOL_FAILSAFE(g_model.moduleData[x].rfProtocol) ? (g_model.moduleData[x].failsafeMode==FAILSAFE_CUSTOM ? (uint8_t)1 : (uint8_t)0) : HIDDEN_ROW)
 #define TIMER_ROWS                        2|NAVIGATION_LINE_BY_LINE, 0, CASE_PERSISTENT_TIMERS(0) 0, 0
-#define EXTERNAL_MODULE_MODE_ROWS       (IS_MODULE_XJT(EXTERNAL_MODULE) || IS_MODULE_DSM2(EXTERNAL_MODULE)) ? (uint8_t)1 : (uint8_t)0
+#ifdef MULTIMODULE
```
bsongis: #if defined

radio/src/myeeprom.h
```diff
@@ -422,7 +422,8 @@ PACK(typedef struct {
   int8_t  rfProtocol:4;
   uint8_t channelsStart;
   int8_t  channelsCount; // 0=8 channels
-  uint8_t failsafeMode:7;
+  uint8_t failsafeMode:3;
+  uint8_t subType:3;
   uint8_t invertedSerial:1; // telemetry serial inverted from standard
```
bsongis: One bit is missing somewhere!

radio/src/myeeprom.h
```diff
@@ -2052,6 +2056,27 @@ enum DSM2Protocols {
...
+    MM_RF_PROTO_FLYSKY=-8,
+    MM_RF_PROTO_FIRST=MM_RF_PROTO_FLYSKY,
+    MM_RF_PROTO_HUBSAN,
+    MM_RF_PROTO_FRSKY_D8,
+    MM_RF_PROTO_HISKY,
+    MM_RF_PROTO_V2X2,
+    MM_RF_PROTO_DSM2,
+    MM_RF_PROTO_DEVO,
+    MM_RF_PROTO_YD717,
+    MM_RF_PROTO_V977,
```
bsongis: You want this one as default, right?

radio/src/myeeprom.h
```diff
@@ -422,7 +422,8 @@ PACK(typedef struct {
   int8_t  rfProtocol:4;
```
bsongis: I can't remember why it is an int rather than an uint, and I am not sure we use more than 7 rfProtocols today. If I am not mistaken, this could be changed to an uint8_t

radio/src/targets/taranis/pulses_driver.cpp
```diff
@@ -612,7 +612,12 @@ extern "C" void TIM8_CC_IRQHandler()
     EXTMODULE_TIMER->DIER |= TIM_DIER_CC2IE ;  // Enable this interrupt
   }
 #if defined(DSM2)
-  else if (s_current_protocol[EXTERNAL_MODULE] >= PROTO_DSM2_LP45 && s_current_protocol[EXTERNAL_MODULE] <= PROTO_DSM2_DSMX) {
+  else if ((s_current_protocol[EXTERNAL_MODULE] >= PROTO_DSM2_LP45 && s_current_protocol[EXTERNAL_MODULE] <= PROTO_DSM2_DSMX)
```
bsongis: You defined a IS_MULTIMODULE_PROTOCOL macro that would be nice here ;)

radio/src/telemetry/frsky.h
```diff
@@ -556,6 +556,10 @@ void frskyUpdateCells(void);
 #if defined(CROSSFIRE)
     if (g_model.moduleData[EXTERNAL_MODULE].type == MODULE_TYPE_CROSSFIRE)
       return PROTOCOL_PULSES_CROSSFIRE;
+#endif    
+#ifdef MULTIMODULE
```
bsongis: #if defined

radio/src/translations.cpp
```diff
@@ -1,714 +1,725 @@
-/*
```
bsongis: Line endings problem? The whole file is marked as modified!

bsongis:
Nice pull request, a couple of minor remarks to be addressed, but it could be merged quickly into 'next'

radio/src/myeeprom.h
kilrah:
```
enum RFProtocols {
  RF_PROTO_OFF = -1,
  RF_PROTO_X16,
```
So we get X16 by default...

bsongis: Oh yes I remember why I put a int there, thanks! I should add comments I know ...

radio/src/myeeprom.h
schwabe: Not really, I did not want to break the eeprom format and had only 4bit available.

radio/src/translations.cpp
schwabe: Yeah. There are some line ending problems in the file. The license has only \n while most of the rest of the file has \r\n. Should this file be DOS (\r\n) or Unix (\n) style? Some of the other files have this problem has well.

bsongis: Oh ok, it's because I used a script to modify all licenses in one shot
Is it possible you don't change it? I mean, you only add in this commit the real changes, another commit will follow with unification of all line endings

schwabe: Yeah, I am reworking the patch to fix all the issues and will only add the lines that are changed.

bsongis: Cool, but please on 'next', the port shouldn't be too difficult

projectkk2glider:
Closing this one in favour of #3198

PR #3212, opened 2016-01-07
Projectkk2glider/issue 2939 edge problem

### LINKED ISSUE #2939: Edge in 2.1.x behaves different than in 2.0.17
http://www.rcgroups.com/forums/showpost.php?p=32799773&postcount=1291

Someone posted this on rcgroups and I asked him to post it on github as I don't have my radio with me.

But I was just able to test this with the simulator and even use a 1-liner to make it easier to reproduce

2.1
Edge   SFup  0,0   (no release)   SAup

2.0

Edge SFup  0.0  (instant)   SAup

In 2.0 that LS can only become true if SF is up
In 2.1 that LS also becomes true if you move SA up and leave SF down

projectkk2glider:
Closes #2939 in 2.0 branch. 

This will have to be ported to master and next also!


bsongis:
I thought it would be a bigger change :) Thanks for this fix!

PR #3213, opened 2016-01-07
Fixes #3207: proper sensor instance generation in Telemetry Simulator

### LINKED ISSUE #3207: Telemetry Simulator
the altitude is not populated  opentx 2.1.7
untitled-2

projectkk2glider:
Closes #3207

### DIFF
--- a/companion/src/simulation/telemetrysimu.cpp
+++ b/companion/src/simulation/telemetrysimu.cpp
@@ -146,34 +146,34 @@ void TelemetrySimulator::closeEvent(QCloseEvent *event)
 void TelemetrySimulator::showEvent(QShowEvent *event)
 {
 
-  ui->rxbt_inst->setText(QString::number(simulator->getSensorInstance(BATT_ID)));
+  ui->rxbt_inst->setText(QString::number(simulator->getSensorInstance(BATT_ID, 1)));
   ui->rxbt_ratio->setValue(simulator->getSensorRatio(BATT_ID) / 10.0);
-  ui->rssi_inst->setText(QString::number(simulator->getSensorInstance(RSSI_ID)));
-  ui->swr_inst->setText(QString::number(simulator->getSensorInstance(SWR_ID)));
-  ui->a1_inst->setText(QString::number(simulator->getSensorInstance(ADC1_ID)));
+  ui->rssi_inst->setText(QString::number(simulator->getSensorInstance(RSSI_ID, 25)));
+  ui->swr_inst->setText(QString::number(simulator->getSensorInstance(SWR_ID, 25)));
+  ui->a1_inst->setText(QString::number(simulator->getSensorInstance(ADC1_ID, 1)));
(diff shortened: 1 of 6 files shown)


dsbeach:
I'm OK with this but as long as you are going to default to a valid sensor id value, shouldn't the default match the FrSky hardware sensor default values?  The reason I chose zero originally was I thought it made more sense to use an invalid value that would get inadvertently 'discovered'.

I was thinking that the best place to document the behavior of the telemetry simulator was in the Lua reference.  Maybe that would avoid some of the confusion regarding its usage. I don't really understand #3207 is a bug, nor how this fixes it.

projectkk2glider:
The `Instance` field (dataId) has a special format: bits from 0-4 carry value (valid range is 0 - 0x01B), bits 5-7 have a sort of CRC value that is dependent on lower 5 bits. 

When you had 0 in the instance fields, that got passed as 0xFF (0 -1 -> converted to unsigned) to the function that generates SPORT packets. When that dataId was seen by the radio firmware it used only lower 5 bits (0x1F = 31), giving instance 32. 

Valid range for instance id (as shown to the user) is 1-28. But `buffer[0] == 0` is a **valid** instance number! It is instance id 1.

kilrah:
Agreed that the default value should be the normal value for that sensor, not just always 1.

BTW you've used "defaultVaue" everywhere instead of "defaultVa**l**ue"

projectkk2glider:
>Agreed that the default value should be the normal value for that sensor

Do we know which is the default value? Its easy to change.

Nice one, spotting the typo. I will fix it.

kilrah:
> Do we know which is the default value?

Well these, &0x1F + 1
https://github.com/opentx/opentx/blob/master/radio/src/telemetry/frsky.h#L170

companion/src/simulation/telemetrysimu.cpp
```diff
@@ -193,13 +195,30 @@ void setSportPacketCrc(uint8_t * packet)
   //TRACE("crc set: %x", packet[FRSKY_SPORT_PACKET_SIZE-1]);
 }
 
-void generateSportPacket(uint8_t * packet, uint8_t dataId, uint8_t prim, uint16_t appId, uint32_t data)
+uint8_t bit(uint8_t position, uint8_t value) 
+{
+  if (value & (uint8_t)(1 << position)) {
```
bsongis: I would prefer one line for this test!

dsbeach:
@projectkk2glider, thanks for the explanation. Bit by bit I'm getting more familiar with OpenTX from the inside. It is an amazing open source accomplishment.

projectkk2glider:
@dsbeach I am doing this for about two years and there are still parts of the code I have jet to discover.

projectkk2glider:
Anything else to be done before the merge? @dsbeach @bsongis

dsbeach:
Looks good to me.  Thanks.

PR #3228, opened 2016-01-13
Fixes #2837: better representation of EDGE logical switch (sam as the…

### LINKED ISSUE #2837: Edge representation is different in print model / OpenTX University / Companion Setup
When I look at Edge in Companion the setup page, Companion print model, the radio itself and the explanation on http://open-txu.org/home/continuing-education/logical-switch-functions/ 

I will get different terminology for what is called "instant" on OpenTX university.
In "print model" this is -0.1
IMHO this should be "<<"

It is true I'm not that familiar with Edge.
In this case the explanation on OpenTX university becomes confusing.

Where OpenTX-U speaks of "instant" it is "no release)"  without the opening brace in Companion

projectkk2glider:
… radio)

Closes #2837

Result:
screenshot
screenshot

### DIFF
--- a/companion/src/modeledit/logicalswitches.cpp
+++ b/companion/src/modeledit/logicalswitches.cpp
@@ -244,20 +244,11 @@ void LogicalSwitchesPanel::edited()
         updateTimerParam(cswitchOffset[i], model->logicalSw[i].val2, 0.1);
         break;
       case LS_FAMILY_EDGE:
-        cswitchOffset2[i]->setSpecialValueText(tr("(instant)"));
         if (sender() == cswitchOffset[i]) {
           model->logicalSw[i].val2 = TimToVal(cswitchOffset[i]->value());
-          updateTimerParam(cswitchOffset[i], model->logicalSw[i].val2, 0.0);
         }
         else {
           model->logicalSw[i].val3 = TimToVal(cswitchOffset2[i]->value()) - model->logicalSw[i].val2;
-        }
-        updateTimerParam(cswitchOffset2[i], model->logicalSw[i].val2+model->logicalSw[i].val3, ValToTim(TimToVal(cswitchOffset[i]->value())-1));
-        if (model->logicalSw[i].val3 == 0) {
-          cswitchOffset2[i]->setSuffix("(infinite)");
-        }
-        else {
-          cswitchOffset2[i]->setSuffix("");
         }
         break;
       default:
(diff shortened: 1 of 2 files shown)


companion/src/modeledit/logicalswitches.cpp
```diff
@@ -244,21 +244,12 @@ void LogicalSwitchesPanel::edited()
...
           model->logicalSw[i].val3 = TimToVal(cswitchOffset2[i]->value()) - model->logicalSw[i].val2;
         }
-        updateTimerParam(cswitchOffset2[i], model->logicalSw[i].val2+model->logicalSw[i].val3, ValToTim(TimToVal(cswitchOffset[i]->value())-1));
-        if (model->logicalSw[i].val3 == 0) {
-          cswitchOffset2[i]->setSuffix("(infinite)");
-        }
-        else {
-          cswitchOffset2[i]->setSuffix("");
-        }
         break;
```
projectkk2glider: These lines look like they are not needed, the same is happening in printLogicalSwitchLine() which is called from this function later.

projectkk2glider:
@fraternl how do you like it now? the `0,0 -- (infinite)` can't be done without the number I am afraid.

fraternl:
If the number can't be omitted then I would prefer it like

0.0:0.0 ---

projectkk2glider:
If we remove `infinite`, then why not remove `instant` also? 

But this to me is a bad idea. The point is that Companion can give more descriptive value compared to the radio. So therefore I prefer if the words that describe special settings stay.

companion/src/modeledit/logicalswitches.cpp
```diff
@@ -343,9 +334,9 @@ void LogicalSwitchesPanel::setSwitchWidgetVisibility(int i)
       populateSwitchCB(cswitchSource1[i], RawSwitch(model->logicalSw[i].val1), generalSettings, LogicalSwitchesContext);
       updateTimerParam(cswitchOffset[i], model->logicalSw[i].val2, 0.0);
       updateTimerParam(cswitchOffset2[i], model->logicalSw[i].val2+model->logicalSw[i].val3, ValToTim(TimToVal(cswitchOffset[i]->value())-1));
-      cswitchOffset2[i]->setSpecialValueText(tr("(instant)"));
+      cswitchOffset2[i]->setSpecialValueText(QString(" <<(%1)").arg(tr("instant")));
       if (model->logicalSw[i].val3 == 0) {
-        cswitchOffset2[i]->setSuffix(tr("(infinite)"));
+        cswitchOffset2[i]->setSuffix(QString(" --(%1)").arg(tr("infinite")));
```
bsongis: Aha excellent :)

projectkk2glider:
>If I remember well, there is only one "special" value in these spinboxes.

True, when the value is at the minimum a *special* value is shown instead (if set).

fraternl:
I don't want to determine the presentation. I merely want them to be as close as they can be in all 3 cases...

projectkk2glider:
Perhaps a tooltip could be used to describe special values. We could then use these representations in model edit window:
* `0.0 --`
* `<<`

 Any ideas for the tooltips? Mine:
* -- The LS will trigger on falling edge of V1
* << The LS will trigger on the rising edge of V1

fraternl:
Yes..... a tooltip is great  !!!

projectkk2glider:
Another idea is to create a custom spin box widget that will display the same values as the radio. Should not be too difficult.

projectkk2glider:
So please don't merge this yet.

bsongis:
I was about to merge it when I saw your last comment. I close it, ok?

PR #3240, opened 2016-01-18
Fixes #3233: Added ADC jitter measurement (use MEASURE_JITTER=YES to …

### LINKED ISSUE #3233: 2.1.7 - Jitter on Channel Monitor and receiver output.
Can we revisit this please. 

I understand the issue was closed (#2774) but both of my Taranis are still experiencing jitter with 2.1.7. 

I have tried the recommended fix from 2774 but it makes no difference. 

The jitter is present in the Channel Monitor, and is also visible in the receiver output.     

My radios are:  1 x Taranis B01 and 1 x Taranis Plus B02.

Both radios are jitter free on all 2.0x version firmware. (and no jitter was visible with stock factory firmware when new)

Video 1 - Jitter with 2.1.7 (Taranis Plus)

https://www.youtube.com/watch?v=7Hv4GdGJasM

Video 2 - No Jitter with 2.017 (Same Taranis Plus - same day) 

https://www.youtube.com/watch?v=l0ynqLULJE0

Video 3 - Jitter from TX is visible in Receiver output 

https://www.youtube.com/watch?v=4v9-JSiZwt4

Video 4 - Jitter still present with Internal and External RF turned off. (no external module fitted here)

https://www.youtube.com/watch?v=SQTjbXBN-gw

I can replicate these results every time, on two separate radios that both have zero jitter with 2.0X version firmware.

projectkk2glider:
…enable), Added ADC jitter filter (use JITTER_FILTER=No to disable)

So jitter filter is enabled by default, while the measurement in disabled.

Closes #3233

### DIFF
--- a/radio/src/Makefile
+++ b/radio/src/Makefile
@@ -383,6 +383,14 @@ WARNINGS_AS_ERRORS = NO
 # Values = NO, YES
 LUA_COMPILER = NO
 
+# Enable ADC jitter measurement
+# Values = NO, YES
+JITTER_MEASURE = NO
+
+# Enable ADC jitter filtering
+# Values = NO, YES
+JITTER_FILTER = YES
+
 #------- END BUILD OPTIONS ---------------------------
 
 # Define programs and commands.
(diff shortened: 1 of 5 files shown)


projectkk2glider:
Ported to next in https://github.com/opentx/opentx/commit/4aa6bd12caa541c07fed018764272bf9fb8f0154

PR #3250, opened 2016-01-24
Projectkk2glider/new jitter filter

projectkk2glider:
I am not at all happy with the jitter filter in #3240. While testing it I found out that it does not pass trough very small changes of stick postilion. This is especially true if the input ADC signal has small noise level. The problem is in the filter formula and integer arithmetics.

So, I've came up with a much better solution (in my view) that still uses the same amount of RAM, but give much more precise result in term of small stick movement. The filter can also be tuned. It's implementation is documented in the code comments.

The new method also changes what kind of input signal is used for analog calibration:
 * before raw unfiltered (noisy value) was used, this sometimes resulted in a poor center stick value.
 * now a filtered value is also used for the calibration.

Same goes for multi-pos switch, a filtered value is used. The multi-pos **was not tested** by me, since I don't have any radio with it. So somebody else should check that it still works! I also have another idea for multi-pos - to implement low pass filter to help combat glitches while slowly switching the multi-pos switch (I have read on forums that some users see occasional glitches there).

The jitter filter could be more powerful and more precise in the future (if needed), but it would require a change in `s_anaFilt[]` from 16bit to 32bit. This would take more RAM. Perhaps we should implement it for better radios (X9E and Horus).

### DIFF
--- a/radio/src/gui/9X/menu_general_calib.cpp
+++ b/radio/src/gui/9X/menu_general_calib.cpp
@@ -39,6 +39,14 @@
 #define XPOT_DELTA 10
 #define XPOT_DELAY 10 /* cycles */
 
+enum CalibrationState {
+  CALIB_START = 0,
+  CALIB_SET_MIDPOINT,
+  CALIB_MOVE_STICKS,
+  CALIB_STORE,
+  CALIB_FINISHED
+};
+
 void menuCommonCalib(uint8_t event)
 {
   for (uint8_t i=0; i<NUM_STICKS+NUM_POTS; i++) { // get low and high vals for sticks and trims
@@ -57,7 +65,7 @@ void menuCommonCalib(uint8_t event)
   switch (event)
   {
     case EVT_ENTRY:
-      reusableBuffer.calib.state = 0;
+      reusableBuffer.calib.state = CALIB_START;
       break;
 
     case EVT_KEY_BREAK(KEY_ENTER):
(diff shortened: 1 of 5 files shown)


bsongis:
Would it be possible you make a pull request on 'next' so that I can test it with the Horus? I am maily working on it right now and there is a lot of jitter. It will be interesting to see how your new filter behaves!

projectkk2glider:
@bsongis here it is https://github.com/opentx/opentx/compare/projectkk2glider/new_jitter_filter_for_next

I am looking forward to your results on Horus. Perhaps we will have to bump it to 32bit if the jitter will still be too much on Horus. But then again there might be some other problem there. Post your jitter measurements.

bsongis:
Is this one still needed?

projectkk2glider:
I am not sure what to do. I have been flying this code for a while and it works good. 

**But** going over it once more today and comparing it to the implementation in the `next` branch (where it was already merged in #3327) I noticed that one bit is still missing - the one that affects the multi-pos calibration. As I wrote before, I don't have a radio with multi-pos switch that can run 2.1. So that part was newer tested by me on this branch.

Now the actual performance of the currently used filter on the `master` branch is sort of satisfactory. I bet not many people (if any) will notice that is it is difficult to produce a very small channel output change (in order or 0.1% - 0.2%) with a very small stick movement. That is the only *downside* to the currently used jitter filter in `master`.

So perhaps it is not worth porting this change to the `master`. But it will be used (already is) in `next`.

I will port the missing change onto this branch. You then decide if it is worth merging. **But before merging** somebody has to test this on the radio:
 * verify that the multi-pos switch can be calibrated
 * check the outputs of the sticks for small and big movements in channel monitor.

bsongis:
I would prefer we don't port it on 'master', let's wait for more tests on 'next' before. If anybody notices a differences then we will port it, would it be fine with you?

projectkk2glider:
Fine with me.

PR #3269, opened 2016-01-31
Store control types and names in profile (fix #3260)

### LINKED ISSUE #3260: 2.1.7 Companion loaded setting not safe to companion profile
When is models and setting loaded to Companion and preseed, save calibration and HW setting to profile that is only partial save. 
Not saved: 
Calibration of LS and RS
S3 setting in HW tab. 

Video here 
https://youtu.be/V33OOJy7WOE?t=426

kilrah:
And fix missing slider calibration.


bsongis:
Thanks!
A port in 'next'?

kilrah:
Yep just wanted confirmation i didn't do something stupid first... ;)

PR #3276, opened 2016-02-05
Joystick now supports 24 buttons (ch9-32) (fix #3268)

### LINKED ISSUE #3268: [Suggestion] USB Joystick set to 8 Axis and 24 Buttons
Playing around with the USB joystick, I've noticed there's just not enough button outputs to handle every switch state. This greatly limits the possible uses of the transmitter in some sims, since as a joystick, there are no 3 position switches.

To fix, I suggest increasing the button count from 8 to 24, this will not only allow every switch state to be outputted, but will allow every mix to have an output while staying well within the limits of DirectInput.

kilrah:
Had basically already done it when looking at it, so...

OK on Windows, maybe someone can test on linux?


bsongis:
Thanks!

PR #3283, opened 2016-02-09
Fix it.h.txt

MicheleVilla:
Fix IT translation


MicheleVilla:
Little fix in Italian firmware translation

PR #3289, opened 2016-02-14
taranis: RX DMA for USART3 and s.bus

KIrill-ka:
Replaced s.bus fifo with DMA ring buffer to get rid of overrun errors. If you like the idea, it can be extended to other serial port functions. It seems that RX DMA for USART6 is also possible (some dma streams need to be re-assigned).
I tested it on my taranis x9d.

### DIFF
--- a/radio/src/targets/taranis/serial2_driver.cpp
+++ b/radio/src/targets/taranis/serial2_driver.cpp
@@ -39,9 +39,13 @@
 uint8_t serial2Mode = 0;
 Fifo<512> serial2TxFifo;
 extern Fifo<512> telemetryFifo;
-extern Fifo<32> sbusFifo;
 
-void uart3Setup(unsigned int baudrate)
+static uint8_t dma_buf[32];
+static uint8_t dma_read_pos;
+
+#define USART3DMAStream DMA1_Stream1
+
+void uart3Setup(unsigned int baudrate, bool dma)
 {
   USART_InitTypeDef USART_InitStructure;
   GPIO_InitTypeDef GPIO_InitStructure;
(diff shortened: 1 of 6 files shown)


radio/src/targets/taranis/serial2_driver.cpp
```diff
@@ -39,9 +39,13 @@
...
 extern Fifo<512> telemetryFifo;
-extern Fifo<32> sbusFifo;
 
-void uart3Setup(unsigned int baudrate)
+static uint8_t dma_buf[32];
+static uint8_t dma_read_pos;
+
+#define USART3DMAStream DMA1_Stream1
+
+void uart3Setup(unsigned int baudrate, bool dma)
```
bsongis: Perhaps we could always use the DMA, no?

radio/src/targets/taranis/trainer_driver.cpp
```diff
@@ -258,3 +258,15 @@ extern "C" void HEARTBEAT_USART_IRQHandler()
   }
 }
 #endif
+
+int serial_get_sbus_char(uint8_t *ch) 
+{
+  switch (currentTrainerMode) {
+          case TRAINER_MODE_MASTER_SBUS_EXTERNAL_MODULE:
```
bsongis: Indentation is wrong!

bsongis:
Excellent idea!

KIrill-ka:
Please, advise me how to solve the following build error.
```
In file included from /home/travis/build/opentx/opentx/companion/src/firmwares/opentx/simulator/opentxsimulator.cpp:238:0:

/home/travis/build/opentx/opentx/companion/src/../../radio/src/targets/taranis/serial2_driver.cpp: In function ‘void Taranis::uart3Setup(unsigned int, bool)’:

/home/travis/build/opentx/opentx/companion/src/../../radio/src/targets/taranis/serial2_driver.cpp:85:73: error: cast from ‘volatile uint16_t* {aka volatile short unsigned int*}’ to ‘uint32_t {aka unsigned int}’ loses precision [-fpermissive]
```
I guess it tries to build specific stm32 code for amd64...

radio/src/targets/taranis/serial2_driver.cpp
KIrill-ka: Probably. I didn't want to touch that much code at once.

KIrill-ka:
> Excellent idea!
Thanks! Initially I didn't think DMA implementation would be that simple.

bsongis:
There is a CONVERT_PTR_UINT macro which should do the job!

KIrill-ka:
Please, help me to resolve the next issue with simu build:
`/tmp/cchsDDAV.o:`` In function ``uart3Setup(unsigned int, bool)': /home/travis/build/opentx/opentx/radio/src/targets/taranis/serial2_driver.cpp:96: undefined reference to ``USART_DMACmd'
collect2: ld returned 1 exit status

Do you think I should use direct register access instead of StdPeriph_Driver functions?

KIrill-ka:
Do you expect anything else from me for this pull request to be merged?

bsongis:
No, thanks! I will take care of it

bsongis:
I just commited a commit inspired by yours for Taranis and Horus. Here it is:
https://github.com/opentx/opentx/commit/5cb6884ab3320dac0d83e80fd5dd871f46566611

KIrill-ka:
Great work, thank you!
There seems to be a problem with DMA1 Stream 5. It's in use by audio_driver.
That's what I meant by
> RX DMA for USART6 is also possible (some dma streams need to be re-assigned)

bsongis:
Ouch OK I just moved the audio definitions to the hal.h file to make it obvious. I will now try to find another DMA stream available ...

projectkk2glider:
Despite this, it was working perfectly for me yesterday. :smile:

bsongis:
On Taranis?

bsongis:
image

projectkk2glider:
No, on Horus.

bsongis:
Horus is safe, we don't use the DAC

bsongis:
I don't see a good solution on Taranis, we have to use USART2 and USART2 is on the same DMA Stream than the DAC. I see there are DAC1 and DAC2, but DMA1_Stream6 is also used. Do I miss anything?

kilrah:
I can't see a solution either...
Using DAC2 isn't an option anyway as it would need rewiring hardware.

bsongis:
I have reverted the part of my commit on S.PORT RX. Everything else is kept. I need to do the same on USART6 (HEARTBEAT). It's useful when used with SBUS trainer.

bsongis:
The SBUS (HEARTBEAT) port is now using DMA

bsongis:
I am not sure I want this ported to ```master```

kilrah:
IMO if it solves the SBUS trainer issue it should, otherwise no.

bsongis:
Fair enough!

KIrill-ka:
> I don't see a good solution on Taranis, we have to use USART2 and USART2
Ah, I confused trainer USART6 with telemetry USART2. I haven't  looked at USART2.
Now it looks like DMA2_Stream2 is conflicting with pulses_driver (timer8).
Current assignment for DMA2:
2s2 -> TIMER8
2s2-> USART6RX
2s1-> ADC3
2s0-> ADC1
Please, check if you can change it with:
2s2 -> TIMER8
2s1 -> USART6RX
2s4 -> ADC1
2S0 -> ADC3

kilrah:
dma

bsongis:
@KIrill-ka Would you double check? DMA on Heartbeat is only used for SBUS if I am not mistaken, and we shouldn't send pulses on external module in this case!

KIrill-ka:
I just checked model settings with 2.1.7. It's possible to set External RF=XJT and Trainer Port = Master/SBUS Module. If the configuration is not valid, why it's allowed in GUI?

bsongis:
Right. That's something we should forbid. I wait for @kilrah confirmation before doing it

kilrah:
Yep it seems invalid indeed. And UART6 never transmits.

bsongis:
Done here:
https://github.com/opentx/opentx/commit/27dc84028c868f1d6edd827e6c379f7719133cb0

bsongis:
Thanks for your help on this one, I will close it since it's implemented in 'next'

PR #3326, opened 2016-02-28
Bsongis/passthrough sport to lua

bsongis:
For discussion

### DIFF
--- a/radio/src/cli.cpp
+++ b/radio/src/cli.cpp
@@ -27,7 +27,7 @@
 
 OS_TID cliTaskId;
 TaskStack<CLI_STACK_SIZE> cliStack;
-Fifo<256> cliRxFifo;
+Fifo<uint8_t, 256> cliRxFifo;
 uint8_t cliTracesEnabled = true;
 // char cliLastLine[CLI_COMMAND_MAX_LEN+1];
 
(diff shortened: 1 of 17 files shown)


projectkk2glider:
Cool idea. 

The function `luaTelemetryRegister()` can be omitted. The fifo can be created in the `luaTelemetryPop()`.

We should also consider for Lua to be able to send some value to SPORT.

bsongis:
Yes I thought of it as well

kilrah:
> We should also consider for Lua to be able to send some value to SPORT.

Absolutely. Reversed process, luaTelemetryPush() that feeds a FIFO that is emptied in the background whenever a poll that matches a configurable sport ID (in hardware tab) is received.

floaledm:
Excellent work! Successfully tested telemetryPop() in a lua script using the bsongis/passthrough_sport_to_lua branch. Returns nil whenever the fifo is empty, which in my view is a useful feature to determine such things as when the last data was received... Fifo doesn't overflow from what I tested.

I'm interested in learning more about how luaTelemetryPush() works. Is it for the serial port in the battery bay or can it be used for two way communication with Smart Port equipment connected to the receiver? There's a Serial Port option in the hardware tab, but it's not configurable beyond selecting what the port should be used for (e.g., SPort mirror, telemetry). Is it simply that the feature is a work in progress?

kilrah:
It will be to send stuff up the smart port, but for now only the fifo and lua function to fill it are implemented, nothing actually sends its content yet.

PR #3327, opened 2016-02-28
Projectkk2glider/new jitter filter for next2

projectkk2glider:
This pull addresses two things:
 * better jitter filter (the equivalent for master is in #3250)
 * improved Horus ADC driver

Tested on my Hours.

### DIFF
--- a/radio/src/targets/horus/adc_driver.cpp
+++ b/radio/src/targets/horus/adc_driver.cpp
@@ -133,11 +133,11 @@ void adcInit()
 #define SPI_S1                         4
 #define SPI_6POS                       5
 #define SPI_S2                         6
-#define SPI_S3                         7
-#define SPI_S4                         8
+#define SPI_LS                         7
+#define SPI_RS                         8
 #define SPI_TX_VOLTAGE                 9
-#define SPI_RS                         10
-#define SPI_LS                         11
+#define SPI_L2                         10
+#define SPI_L1                         11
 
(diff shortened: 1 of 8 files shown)


radio/src/targets/horus/adc_driver.cpp
```diff
@@ -133,11 +133,11 @@ void adcInit()
 #define SPI_S1                         4
 #define SPI_6POS                       5
 #define SPI_S2                         6
-#define SPI_S3                         7
-#define SPI_S4                         8
+#define SPI_LS                         7
```
bsongis: Why? On mine they are labelled S3 / S4. Is it different on yours?

projectkk2glider: I have `L1` on the left linear slider (left of the power button) and `L2` label on the right one. I see that old photos had S3 and S4 there. I did not know that.

projectkk2glider: I see that the trim labels have also changed. I will post a picture of mine.

projectkk2glider:
dsc_9367

Here is my X12S, T1 and T4 changed places.

bsongis:
Right it's different from mine. Also T1/T4 are swapped. T2/T3 are swapped.

projectkk2glider:
Perhaps ADC channels also changed, because I corrected the ADC positions of L1/l2 and LS/RS also.

Issue #3340, opened 2016-03-04
Watchdog reset on Horus and settings restore


projectkk2glider:
This issue is about the Watchdog reset mechanism on Hours and for algorithms that enable persistence of radio and current model settings when WDT reset happens.

Current idea is (as discussed on gitter):
 * after every settings change:
  * write settings to the SD card (already implemented)
  * take care not to leave the SD card based settings in a corrupted state if a WDT reset happens in the middle of a file write
  * store a subset of radio settings and currently active model settings (only relevant parts) in battery backed SDRAM (4kB available). The current calculation shows that all settings can not be saved, so only important setting will be stored here. Also settings will need to be compressed to fit.

 * after WDT reset:
  * try to restore settings from battery backed SRAM (validity verified by checksum).
  * if settings in SRAM are bad: initialize SD card an load only radio/model settings from SD card. 
  * all other SD card based functions are disabled (bitmaps, sounds, logs, etc...)

There is also another issue to consider. When battery backup SRAM is enabled, the battey consumption goes up (when radio is powered off). http://stackoverflow.com/questions/20667754/how-to-use-backup-sram-as-eeprom-in-stm32f4 Therefore we need to:
 * enable BB SRAM on every startup
 * disable BB SRAM on every power down

kilrah:
> When battery backup SRAM is enabled, the battey consumption goes up (when radio is powered off)

...but keeping all things in perspective - it would "only" halve the battery life that currently sits at 4.5 years typically.

projectkk2glider:
Tested on pre-production radio with commit 03988a3a623d0ac5f62beaf6d1b4544e2af06c17. Works.

kilrah:
Got stuck in a reboot loop
I forgot about the PXX on external not being implemented, selected it, and bam WDT, enter emergency mode, and... re-WDT forever so somehow the backup RAM was written during those 1.5sec between menu selection and wdt reset.
Need to check the write timer, it might not always be reset on change

kilrah:
Done, but leaving it open due to the potential bad read of stick mode that was reported

projectkk2glider:
This should fix stick mode problem #3455

projectkk2glider:
Stick mode problem was resolved.

PR #3341, opened 2016-03-07
Projectkk2glider/horus reboot and external module

projectkk2glider:
Some work done for issue #3340 in first part. Second part is start of external module support on Horus (None and PPM working).

### DIFF
--- a/radio/src/CMakeLists.txt
+++ b/radio/src/CMakeLists.txt
@@ -106,6 +106,7 @@ if(PCB STREQUAL HORUS)
     pwr_driver.cpp
     sdram_driver.c
     led_driver.cpp
+    startup_stm32f40_41xxx.s
     )
   set(STM32LIB_SRC
     STM32F4xx_StdPeriph_Driver/src/stm32f4xx_sdio.c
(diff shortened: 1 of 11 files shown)


radio/src/CMakeLists.txt
```diff
@@ -18,6 +18,7 @@ option(AUTOSOURCE "Automatic source detection in menus" ON)
 option(AUTOSWITCH "Automatic switch detection in menus" ON)
 option(JITTER_MEASURE "Enable ADC jitter measurement" OFF)
 option(JITTER_FILTER "Enable ADC jitter filtering" ON)
+option(WATCHDOG_DISABLED "Disable hardware Watchdog on Horus" ON)
```
bsongis: Is this option *really* needed? If yes it should be only in the PCB=HORUS part, no?

radio/src/cli.cpp
```diff
@@ -272,7 +272,13 @@ int cliMemoryInfo(const char ** argv)
 int cliReboot(const char ** argv)
 {
 #if !defined(SIMU)
-  NVIC_SystemReset();
+  if (!strcmp(argv[1], "wdt")) {
```
bsongis: Cool!

radio/src/CMakeLists.txt
projectkk2glider: It is to enable easy turning on/off of Horus WDT, we will remove it after the firmware is working reliably. It only has effect on Horus board.

PR #3348, opened 2016-03-10
[Taranis] fixed placement of decimal point in SMLSIZE font

floaledm:
[Compiled and Tested]
Decimal point needed to be shifted one pixel to the left.
This is a correction also appearing in #2666
The next branch needs to be corrected too, and I will submit a pull request for that branch right away.


bsongis:
:+1:

PR #3349, opened 2016-03-10
[Taranis] fixed placement of decimal point in SMLSIZE font

floaledm:
[Compiled and Tested]
Decimal point needed to be shifted one pixel to the left.
This is a correction also appearing in #2666
Same as #3348 but for branch next.

### DIFF
--- a/radio/src/gui/taranis/lcd.cpp
+++ b/radio/src/gui/taranis/lcd.cpp
@@ -404,8 +404,8 @@ void lcdDrawNumber(coord_t x, coord_t y, int32_t val, LcdFlags flags, uint8_t le
         xn = x;
       }
       else if (smlsize) {
-        x -= 1;
-        lcdDrawPoint(x-1, y+5);
+        x -= 2;
+        lcdDrawPoint(x, y+5);
         if ((flags&INVERS) && ((~flags & BLINK) || BLINK_ON_PHASE)) {
           lcdDrawSolidVerticalLine(x-1, y, 7);
         }


bsongis:
:+1:

bsongis:
Perhaps a non regression test would be nice!

floaledm:
Sorry I didn't specify. I tested using lua scripts on the Taranis Plus, for both the master and next branches, that the modifications proposed (and now merged) in the pull requests #3348 and #3349 had the intended effect. Is that what you are looking for?

bsongis:
I meant an automatic non regression test, executed by Travis.
Something like:
```
TEST(Lcd, SmlsizeDecimalPoint)
{
  lcdClear();
  lcdDrawNumber(0, 0, 105, PREC1|SMLSIZE);
  lcdDrawNumber(0, 50, 105, PREC1|SMLSIZE|INVERS);
  EXPECT_TRUE(checkScreenshot("SmlsizeDecimalPoint"));
}
```

First the test will fail on your computer, you will manually check the screenshot, copy it with the others and that's it!

floaledm:
Ok. Apologies in advance if any of this sounds naive, but I've never used Travis before...

I added the code above to /radio/src/tests/lcd.cpp
For both master and next branches, how do I go about executing Travis so I can generate the screenshot(s)?

When that's done, I would submit a pull request with the code change to lcd.cpp + the screenshot in the tests folder?

Would it be better to include the decimal point test within the "Smlsize" test, or would you rather keep it separate?

bsongis:
If it's separate, it's easier to track if one day someone breaks your change ...

floaledm:
Sounds good.

How do I go about executing Travis so I can generate the screenshot(s)?

When that's done, do I submit a pull request with the code change to lcd.cpp + the screenshot in the tests folder?

bsongis:
In fact you should do locally:
```make gtests```
```./gtests```

When your new test is OK, Travis will also execute it at each commit and will prevent somebody breaks your change!

floaledm:
I managed to run ./gtests on the next branch (not the master yet) and the Lcd.SmlsizeDecimalPoint test fails as expected. I looked for but could not find an automatically generated screenshot (png, or even bmp file). Where is it?

projectkk2glider:
/tmp/

floaledm:
Thanks! I found SmlsizeDecimalPoint_212x64.png in /tmp/ and copied it to opentx/radio/src/tests then ./gtests completed successfully. Looks like I should also generate the 128x64 size png? If so, how do I do that?

projectkk2glider:
You make gtests for other radio for example 9X. I wonder if 9X even supports small fonts.

floaledm:
Thanks. There's a file named "smlsize_128x64.png" in opentx/radio/src/tests/ that seems to indicate you can get SMLSIZE in the 128x64 format. The test is surrounded by #if defined(CPUARM) and #endif

I've placed the code suggested by @bsongis in the /radio/src/tests/lcd.cpp file, right before TEST(Lcd, Smlsize) and inside the "if defined" and tried with 9X but it didn't work. I tried running this from opentx/build:
make clean
cmake
make
make gtests
./gtests

but the tests return no error. 9X is ATMEGA and not ARM I gather. Which ARM platform supported by OpenTX uses a 128x64 size display? Hopefully this is the right question to ask...

kilrah:
sky9x, 9xrpro and ar9x use a 128x64 display with ARM architecture. 

What he means is you need to make gtests with the platform and options of interest like for building the firmware, e.g.
```make gtests PCB=TARANIS OPTION=XYZ ...```

floaledm:
Thanks. I was able to generate the128x64 png file (found in /tmp/) after running this from opentx/build (on the next branch):
make clean
cmake .. -DPCB=9XRPRO
make gtests
./gtests

I've submitted pull request #3363 for the SmlsizeDecimalPoint test in the next branch, along with fixes when using SMLSIZE|INVERS on the Taranis and when using SMLSIZE on the sky9x, 9xrpro and ar9x. I haven't managed to get gtest working on the master branch yet, but I'll give it another try...

PR #3358, opened 2016-03-13
Fixes #3344: radio and voice language returned in Lua getGeneralSetti…

### LINKED ISSUE #3344: Add lua function getLanguage() to be able to make script texts match transmitter language.
There are two places for language definition:

In compile options for user interface. Covered in earlier request to make used compile options accessable from lua.

In language for sound files.

Please make those available for proper localisation of scripts.

projectkk2glider:
Closes #3344 

Test script:
```lua
local function run(event)
  local s = getGeneralSettings()
  print("lang: "..s["language"])
  print("tts: "..s["voice"])
  return 1
end

return { run=run }
```
Result:
```
lang: EN
tts: sk
Script finished with status 1
```


bsongis:
Right it's better, thanks!

PR #3364, opened 2016-03-16
SMLSIZE fixes + Gtests on SmlsizeDecimalPoint

floaledm:
Followup of #3348:

Fixed SMLSIZE display for sky9x, 9xrpro and ar9x (incl. INVERS)
Fixed SMLSIZE|INVERS display for Taranis
Added Gtest and screenshots


radio/src/gui/9X/lcd.cpp
```diff
@@ -457,9 +457,9 @@ void lcd_outdezNAtt(coord_t x, coord_t y, lcdint_t val, LcdFlags flags, uint8_t
       }
       else if (smlsize) {
```
bsongis: Interesting, so it was needed :)

bsongis:
Excellent!

radio/src/gui/9X/lcd.cpp
floaledm: I had not manually tested INVERS so the Google test helped in that sense... Glad I could learn how to use GTest. Nice work guys!

PR #3368, opened 2016-03-18
[Horus] Internal module antenna selection support

kilrah:
Don't merge yet, for discussion...

We should probably have a confirmation popup when going from internal to external for the user to make sure there is an antenna installed on the external port before actually switching.


projectkk2glider:
Do you know if IXJT has any internal `antenna missing` protection? Or does selecting external antenna while no antenna connected result in burned transmitter?

kilrah:
No idea. I intend to do some tests and measurements (I have RF test equipment) once I get my radio. I would be surprised if there is an auto shutdown.

bsongis:
Yes a confirmation would be nice!

kilrah:
Done, might still want to wait for tests to merge...

kilrah:
Tested and working. There is NO protection, selecting external antenna with none attached results in SWR about 80-90. 
By the way, the SWR warning is not functional.

PR #3373, opened 2016-03-20
Projectkk2glider/issue 3371 horus tx battery

### LINKED ISSUE #3371: Add internal battery voltage calibration

projectkk2glider:
Closes #3371 

My changes do not affect AVR platforms - no code size change there!

The `periodicTick()` system is just a start. It should be more generalized and then used for calling other functions that need to run periodically, but not on every thread loop.

Similar thing is already implemented (but well hidden) in `doMixerCalculations()` where it is used for throttle trace. These should be united probably.

### DIFF
--- a/radio/src/dmafifo.h
+++ b/radio/src/dmafifo.h
@@ -45,6 +45,9 @@ class DMAFifo
 
     bool isEmpty()
     {
+#if defined(SIMU)
+      return true;
+#endif
       return (ridx == N - stream->NDTR);
     }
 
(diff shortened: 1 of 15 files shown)


radio/src/dmafifo.h
```diff
@@ -45,6 +45,9 @@ class DMAFifo
 
     bool isEmpty()
     {
+#if defined(SIMU)
+      return true;
```
bsongis: I was on it :)

projectkk2glider: We actually need better fix in order to be able to fill telemetry fifo with telemetry simulator in Companion! My fix was just to make it compile and run.

PR #3378, opened 2016-03-22
Removed a few legacy timer limitations, fix non-working limit (fix #3301)

### LINKED ISSUE #301: Missing GVAR function switch entries
_From [bernet.a...@gmail.com](https://code.google.com/u/114409629465115275680/) on October 31, 2012 13:08:02_

Already mentioned to Bertrand, but to make it easier - "Adjust GVx" entries are missing in the function switches list.

_Original issue: http://code.google.com/p/companion9x/issues/detail?id=101_

### LINKED ISSUE #3301: Timer configuration issues
- Radio UI allows setting timer start minutes only up to 99 minutes (we have 24 bits now, for UI reasons we can go up to 99:59:59)
- Adjusting seconds allows adjusting minutes beyond the limit (bug)

Should add TIMEHOUR to the setup function (and corresponding companion field) and fix the seconds editing so it doesn't allow going beyond the minute limit

kilrah:


companion/src/modeledit/setup.cpp
```diff
@@ -42,6 +42,13 @@ TimerPanel::TimerPanel(QWidget *parent, ModelData & model, TimerData & timer, Ge
     ui->countdownBeep->addItem(tr("Voice"), TimerData::COUNTDOWN_VOICE);
     ui->countdownBeep->addItem(tr("Haptic"), TimerData::COUNTDOWN_HAPTIC);
   }
+  
+  if (IS_TARANIS(board))
+    ui->value->setMaximumTime(QTime(23, 59, 59));
+  else if (IS_ARM(board))
+    ui->value->setMaximumTime(QTime(8, 59, 59));
+  else
+    ui->value->setMaximumTime(QTime(0, 59, 59));
```
projectkk2glider: This and the same section above should go into the `firmware->getCapability()` function.

kilrah: Good idea

kilrah: In what form though? getCapability returns an int, there is no QTime constructor that takes seconds so while the value gains a bit by being defined in a single place it means more duplicated code to handle it.

projectkk2glider: Perhaps use function overloading with out parameter, see http://stackoverflow.com/questions/9568852/overloading-by-return-type

kilrah: No good t seems... Would need a dedicated, separate function in the class.

projectkk2glider: Something like this:
```c
QTime max;
firmware->getCapability(MAX_TIME_BLA_BLA, &max);
```

PR #3384, opened 2016-03-23
Kilrah/timers22

kilrah:
Same as #3378 for next.

Given the recent key layout change that doesn't allow using the keys to edit values anymore it is now impossible to set the timer start value conveniently (would need to turn the encoder hundreds of times to set a couple of hours start time). Code simplifications that have been done in the timer display function don't allow the convenient trick used on Taranis for separate seconds/minute adjustment, so that needs to be reworked. Maybe with separate hour/min/sec due to the encoder.

### DIFF
--- a/companion/src/firmwares/opentx/opentxinterface.cpp
+++ b/companion/src/firmwares/opentx/opentxinterface.cpp
@@ -813,6 +813,16 @@ int OpenTxFirmware::getCapability(const Capability capability)
   }
 }
 
+QTime OpenTxFirmware::getMaxTimerStart()
+{
+  if (IS_TARANIS(board))
+    return QTime(23, 59, 59);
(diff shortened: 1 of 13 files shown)


companion/src/firmwares/opentx/opentxinterface.cpp
```diff
@@ -813,6 +813,16 @@ int OpenTxFirmware::getCapability(const Capability capability)
   }
 }
 
+QTime OpenTxFirmware::getMaxTimerStart()
+{
+  if (IS_TARANIS(board))
```
bsongis: HORUS here?

kilrah: Oops yes

PR #3385, opened 2016-03-25
Also backup members inside unions,

schwabe:
the code is a bit more complicated since unions can have structs or simple types.


schwabe:
sorry, my mistake with last minute changes, will upload a fixed version

schwabe:
Whatever caused that build to fail it was not my commit.

kilrah:
The "Details" link is there for a reason ;) Travis failed because some server it needs to pull stuff from seems to be down so it can't configure itself.

schwabe:
I see that you reverted the pull request again, is there anything wrong with it that I can fix?

PR #3388, opened 2016-03-29
Also backup members inside unions,

schwabe:
the code is a bit more complicated since unions can have structs or simple types. Also change Telemetry sensor to be a struct, since a class with all public is identical to a struct in C++

Use the right branch for the pull request instead of the wrong one in #3385, sorry about that.


bsongis:
Good job, thanks!

PR #3391, opened 2016-03-29
Make sound work in OS X

schwabe:
This fixes sound not being included in the simulator build and sound works on the simu target. Unfortunately it currently also breaks building lib simulator with undefined symbols from SDL.


bsongis:
Perhaps this should be done in the root of the project? Because otherwise it's duplicated!

bsongis:
Excellent, thanks!

PR #3401, opened 2016-04-04
check struct sizes

schwabe:
Changing the size of a struct by accident lead to random the wrong data from the eeprom and can screw the eeprom up pretty bad. Better bail out if the eeprom size changes and let the developer make that an intentional choice. The template magic is there to print out the actual and expected size:

/Users/arne/oss/opentx/radio/src/datastructs.h:1045:3: error: static_assert failed "struct size changed"
  static_assert(expectedSize == realSize, "struct size changed");
  ^             ~~~~~~~~~~~~~~~~~~~~~~~~
/Users/arne/oss/opentx/radio/src/datastructs.h:1181:3: note: in instantiation of function template specialization 'check_size<TrainerData, 168, 16>' requested here
  CHKSIZE(TrainerData, 168);

I screwed my eeprom up badly while working on my multi module patch :(. The checks for horus might be commented out for now since its eeprom is not finalized yet.


projectkk2glider:
I can not compile this on my system. I had to do a couple of changes to get it to compile using gcc-arm 4.7 or 4.8. 

The result when I intentionally messed with one data struct is very good:
```
opentx/radio/src/datastructs.h: In instantiation of 'void check_size() [with ToCheck = MixData; 
unsigned int expectedSize = 20u; unsigned int realSize = 21u]':
```

Not sure how option `-std=gnu++11` will affect other parts of the code, but resulting binary works on Horus.

schwabe:
interesting that c++11 is  standard for me but not you

schwabe:
Found why c++ is standard for me, the main cmake file already has:

set(CMAKE_CXX_STANDARD 11)

So from the cmake documentation -std=gnu++ should be already the default. If compiling with older c++ version is need an ifdef on the c++ could be added.

projectkk2glider:
I've found the reason:

> You're using a feature (`CMAKE_CXX_STANDARD`) introduced in CMake version
3.1, so you should require a minimum version >= that.

I am using cmake version 2.8.12.2

radio/src/CMakeLists.txt
```diff
@@ -979,14 +979,15 @@ if(ARCH STREQUAL ARM)
   set(CMAKE_ASM_COMPILE_OBJECT "<CMAKE_ASM_COMPILER> -o <OBJECT> <SOURCE>")
 
   set(COMMON_FLAGS "-mcpu=${MCU} -mthumb -fomit-frame-pointer -fverbose-asm -Wa,-ahlms=firmware.lst -O${OPT} -gdwarf-2 -DHSE_VALUE=${HSE_VALUE} -fno-exceptions -fdata-sections -ffunction-sections ${WARNING_FLAGS}")
+  set(COMMON_CPP_FLAGS "${COMMON_FLAGS} -std=gnu++11")
```
bsongis: This should not be done this way, CMake has a macro for this and if I remember right I already used it somewhere, am I wrong?

bsongis:
Yes this is a nice change, just the small cosmetics I commented above and it's OK, thanks!

radio/src/CMakeLists.txt
projectkk2glider: >You're using a feature (CMAKE_CXX_STANDARD) introduced in CMake version
3.1,

schwabe:
So it boils down to 
a) force CMAKE >= 3.1
b) do an #ifdef __cplusversion in the check_struct function.

right"?

radio/src/CMakeLists.txt
bsongis: Is there a problem if we force everybody to use this CMake version?

schwabe: Seem like ubuntu has the required version only since 16.04 (not yet released iirc): http://packages.ubuntu.com/search?keywords=cmake&searchon=names&suite=all&section=all

Maybe add the c++11 flags to CMAKE_CXX_FLAGS depending on cmake version for now and later (in a year or two) force cmake 3.1?

bsongis: OK then!

bsongis: Don't forget to add a comment in the CMakeLists.txt otherwise I will forget and change it next time I modify the file!

hrenz:
Hy,
got some errors at compiling  with check struct size  ??

(I was 3 weeks out, did you  change something at the compilers?)

Error file added
[Errors at Compiling 1704a.txt](https://github.com/opentx/opentx/files/222290/Errors.at.Compiling.1704a.txt)

bsongis:
Which is the command you use to compile?

bsongis:
@kilrah Yes we should remove this line

hrenz:
normal compiling   
with cmake Taranis or Horus  

then
make
make all
make firmware

I do since month now  
now allways the same Compiling errors

schwabe:
/home/helle/githubNEU/opentx1704a/opentx/radio/src/datastructs.h:1046:3: warning: identifier ‘static_assert’ is a keyword in C++11 [-Wc++0x-compat]

That sounds like your compiler is not using C++11. Can you do a make VERBOSE=1 and post the output? (especially the call of gcc)

What platform are you using? Linux?

hrenz:
ubuntu 14.04

hrenz:
will do a make VERBOSE=1 
but its a long file

bsongis:
Perhaps we can avoid the check struct size when the c++11 is not allowed?

2016-04-16 18:45 GMT+02:00 hrenz <notifications@github.com>:

schwabe:
@hrenz the las compiler call before the failure is probably enough.

@bsongis sure if we can add an

#if __cplusplus  >= 201103L 

but the cmake file should make sure that c++11 is used.

bsongis:
Indeed!

schwabe:
I checked the Cmake files. The changes @projectkk2glider made are only C++11 in arm target mode, so simu (and probably avr) target do not have the C++11 flags set. 


Adding this to main cmake file will make the compiling work but is ugly as it throws warning on c files:
``` 
if (${CMAKE_VERSION} VERSION_LESS 3.1.0})
  add_compile_options("-std=gnu++11")
endif()
```

projectkk2glider:
Are you not satisfied with your commit a93e1538b948bf250771fea110851f4bddcd5d96

I've made the same change for `simu` in 41a5af5a6e1bd8f96e8cb86c8bd9367266827e6d

schwabe:
@projectkk2glider Looking at it again, not really. For cmake >= 3.1 everything is compile with C+11. For older cmakes we have to add to each target :/

PR #3403, opened 2016-04-06
[Horus] Backlight control added back in, driver fix (thanks Adela), OFF level configurable

kilrah:
Cleaned up this time. 
Breaks general settings due to new var.


radio/src/datastructs.h
```diff
@@ -911,6 +911,7 @@ PACK(struct TrainerData {
     uint8_t  slidersConfig:2; \
     uint32_t switchConfig; \
     uint8_t  potsConfig; /* two bits per pot */ \
+    NOBACKUP(uint8_t  blOffBright); \
```
bsongis: Would it be possible you steal 7bits from the bluetoothEnable so that we avoid changing again the EEPROM. I plan to add Json next week anyhow ...

bsongis:
Seems good to me, but I can't test it until sunday.

radio/src/datastructs.h
kilrah: Oh yeah good idea, didn't see it

kilrah:
It works fine :) Although I just noticed we have no backlight in Emergency mode anymore - before the problem didn't appear because there was never anything actually enabling backlight, it was just the pin default state.

Not sure what the best way to enable it is...

bsongis:
Perhaps such a macro:
#define BACKLIGHT_LEVEL() (unexpectedShutdown ? ... : radioData.BL_level)

PR #3405, opened 2016-04-08
[Horus] Extra trims support

kilrah:
Simu issues though, can someone have a look?


bsongis:
Does it remain simu issues? It works here (libfox simu)

kilrah:
No I think it's all good! Works here too, and the fact you found the wrong define would explain why it was intermittent

PR #3407, opened 2016-04-09
[ARM board] ADC are done 4 times inside the driver now. To be tested …

bsongis:
Not tested on the real hardware (Taranis / Horus / 9XR-PRO)

### DIFF
--- a/radio/src/opentx.cpp
+++ b/radio/src/opentx.cpp
@@ -1516,7 +1516,7 @@ tmr10ms_t jitterResetTime = 0;
 
 #if defined(VIRTUALINPUTS) && defined(JITTER_FILTER)
   #define JITTER_FILTER_STRENGTH  4         // tune this value, bigger value - more filtering (range: 1-5) (see explanation below)
-  #define ANALOG_SCALE            1         // tune this value, bigger value - more filtering (range: 0-3) (see explanation below)
+  #define ANALOG_SCALE            1         // tune this value, bigger value - more filtering (range: 0-1) (see explanation below)
 
   #define JITTER_ALPHA            (1<<JITTER_FILTER_STRENGTH)
   #define ANALOG_MULTIPLIER       (1<<ANALOG_SCALE)
--- a/radio/src/opentx.cpp
+++ b/radio/src/opentx.cpp
@@ -1503,26 +1503,16 @@ uint16_t BandGap = 225;
 // G: Note that the above would have set the ADC prescaler to 128, equating to
 // 125KHz sample rate. We now sample at 500KHz, with oversampling and other
 // filtering options to produce 11-bit results.
-uint16_t BandGap = 2040 ;
+uint16_t BandGap = 2040;
 #elif defined(PCBSTD)
-uint16_t BandGap ;
+uint16_t BandGap;
 #endif
 
 #if defined(JITTER_MEASURE)
-  JitterMeter<uint16_t> rawJitter[NUMBER_ANALOG];
-  JitterMeter<uint16_t> avgJitter[NUMBER_ANALOG];
-  tmr10ms_t jitterResetTime = 0;
-  #if defined(PCBHORUS)
-    #define JITTER_MEASURE_ACTIVE()   (menuHandlers[menuLevel] == menuStatsAnalogs)
-  #elif defined(PCBTARANIS)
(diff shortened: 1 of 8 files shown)


radio/src/opentx.cpp
```diff
@@ -1590,22 +1578,11 @@ void getADC()
...
-      temp[x] += val;
-    }
-  }
+  adcRead();
   DEBUG_TIMER_STOP(debugTimerAdcRead);
 
-  for (uint32_t x=0; x<NUMBER_ANALOG; x++) {
-    uint16_t v = temp[x] >> (3 - ANALOG_SCALE);
+  for (uint8_t x=0; x<NUMBER_ANALOG; x++) {
+    uint16_t v = getAnalogValue(x) >> (1 - ANALOG_SCALE);
```
projectkk2glider: This won't work if ANALOG_SCALE > 1. See also my later comments which are related to this.

radio/src/targets/horus/adc_driver.cpp
```diff
@@ -126,103 +124,118 @@ void adcInit()
...
+    }
+#endif
     ADC_CS_HIGH();
     delay_01us(1);
+    result += val;
   }
-  DEBUG_TIMER_STOP(debugTimerAdcLoop);
 
-  DEBUG_TIMER_START(debugTimerAdcWait);
+  return result >> 2;
```
projectkk2glider: I would prefer if `result` was not divided by 4. We have enough bits in uint16_t to store four full ADC readings (4 * 0xFFF = 3FFC). Dividing here means lost precision/resolution for the jitter filter. So

projectkk2glider:
I really need to test this, perhaps there is no need to change the division of the `result`. Then only the comment and test for  `ANALOG_SCALE` should be corrected. I will do a test on the radio later.

bsongis:
Just let me know. I did it quickly and I am ok to change anything if needed

projectkk2glider:
First the good news: the ADC time has improved `ADC read   : 0.244ms - 0.263ms` it was `ADC read   : 0.629ms - 0.653ms` before.

The bad news: From the four readings for each ADC channel at least one is from the wrong ADC channel. The problem with the ADC chip is that the current reading gives the channel value that was selected two commands before the current command!

bsongis:
Perhaps ```adcReadSPIDummy()``` must be called inside ```adcReadNextSPIChannel()```

bsongis:
Or better on the last read you send ```adcCommands[index+1]```

bsongis:
Also would you measure the on chip ADC duration? Perhaps ```if (index < 8 && !(index & 1))``` could be reduced to ```if (index < 4)```

projectkk2glider:
I already have one solution, I will also consider your suggestions.

projectkk2glider:
>Also would you measure the on chip ADC duration? Perhaps if (index < 8 && !(index & 1)) could be reduced to if (index < 4)

Is this really needed? We just have to read two values in the time it takes to read other values from ADC chip. So no need to do it faster.

bsongis:
It would only reduce the test, in case only one SPI value needs more time than all internal ADC ...

projectkk2glider:
Here are my changes.

 I have just discovered that apparently used hall sensors have some kind of hysteresis - when I move the stick to one side and return it to center I get for example value 0.0%, when I move it to the other side and back  I get 1.0%. It is totally repeatable for all four axes.

kilrah:
It would be interesting to see if that reduces the jitter on its own and would allow for removing the filter.

I still see the jitter filter as a bandaid since it doesn't explain why on Taranis the jitter became a problem in 2.1 while it was insignificant in 2.0.

projectkk2glider:
Should be easy to test, just use JITTER_FILTER=NO

kilrah:
Good news, it seems it does!

2.1:
screen-2016-04-14-184731

2.2 before:
screen-2016-04-14-184919

2.2 after:
screen-2016-04-14-185049

screen-2016-04-14-185107

By looking at the channel monitor that's pretty much back to 2.0 levels. Let's try porting that to 2.1, might be better than the filter, hopefully is since reception for it is pretty appalling right now.

bsongis:
But my change shouldn't change anything on Taranis!

kilrah:
Ugh indeed, weird...

PR #3409, opened 2016-04-09
Fixes #3374: added SD card disk cache for Horus

### LINKED ISSUE #3374: Implement disk cache
With lots of SRAM on Horus, try to implement a simple disc cache and test to see if it helps. It should improve audio stuttering for example.

Idea is simple - read cache:
* a definable number of cache blocks, each the same size (at least a couple of sectors)
* on sector(s) read:
  * if the whole read can be satisfied from the cache, do it. Mark last used time of used cache block.
  * otherwise find oldest (used first) cache block and fill it with the sectors (starting from the requested sector). Return requested sector(s)
* on sector(s) write:
 * write sectors 
 * clear all cache blocks that contain at least one written sector.

Ideas for further enhancements:
 * merge written sectors into cache blocks
 * dedicated cache block for FAT table
 * write caching

projectkk2glider:
* SIMU_DISKIO option available again
 * new option DISK_CACHE defaults to ON for Horus
 * printf("%f") fixed in CLI

Closes #3374


bsongis:
Thanks!

PR #3410, opened 2016-04-10
Fixes #3178: USB driver changed interrupt grouping which affected int…

### LINKED ISSUE #3178: Interrupt priority potential problem
Taranis/Horus:
File usb_bsp.c
Function: void USB_OTG_BSP_EnableInterrupt(USB_OTG_CORE_HANDLE *pdev)
This function calls:
  NVIC_PriorityGroupConfig(NVIC_PriorityGroup_1);
Depending on how you expect the interrupt priorities to work, this has the potential to completely upset them. The default operation is to have 16 "group" priorities thus an interrupt of priority 0 will interrupt anything else. The above call changes this to having only TWO groups (0 and 1), with 8 sub-priorities within each. This means interrupt priorities 0 to 7 are one group and 8-15 are another. An interrupt of priority 0 will no longer interrupt those of priority 1 through 7, only those of 8 through 15. Also priority 1 won't interrupt 2 through 6 etc.

Function: void USB_OTG_BSP_DisbleInterrupt(USB_OTG_CORE_HANDLE *pdev)
also calls this.
Unless you want the interrupt priority grouping changed, both functions need this call removing.

I would suggest:
void USB_OTG_BSP_EnableInterrupt(USB_OTG_CORE_HANDLE *pdev)
{
	NVIC_SetPriority( OTG_FS_IRQn, 11 ) ; // Lower priority interrupt
	NVIC_EnableIRQ(OTG_FS_IRQn) ;
}

void USB_OTG_BSP_DisableInterrupt(USB_OTG_CORE_HANDLE *pdev)
{
	NVIC_DisableIRQ(OTG_FS_IRQn) ;
}

Mike.

projectkk2glider:
…errupt priorities (fix suggested by Michael Blandford)

I recommend not merging right now for 2.1.8. It was tested on my Taranis, but should be tested some more time on the next branch (changes there done in commit 14841590b9e092d6daf7056426c449e601c15b12).

Closes #3178

### DIFF
--- a/radio/src/targets/taranis/usb_bsp.c
+++ b/radio/src/targets/taranis/usb_bsp.c
@@ -83,14 +83,8 @@ void USB_OTG_BSP_Deinit(USB_OTG_CORE_HANDLE *pdev)
 */
 void USB_OTG_BSP_EnableInterrupt(USB_OTG_CORE_HANDLE *pdev)
 {
-  NVIC_InitTypeDef NVIC_InitStructure; 
-  
-  NVIC_PriorityGroupConfig(NVIC_PriorityGroup_1);
-  NVIC_InitStructure.NVIC_IRQChannel = OTG_FS_IRQn;
-  NVIC_InitStructure.NVIC_IRQChannelPreemptionPriority = 1;
-  NVIC_InitStructure.NVIC_IRQChannelSubPriority = 3;
-  NVIC_InitStructure.NVIC_IRQChannelCmd = ENABLE;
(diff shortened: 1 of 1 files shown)


projectkk2glider:
I have been using this for a while now on my radios without any issue. I recommend merging.

projectkk2glider:
I have used this on Taranis and 9XE for at least two hours flight time and a couple of connects to computer. So it is safe to merge.

PR #3425, opened 2016-04-16
[Horus] Reboot protection reworked to also handle non-WDT events

kilrah:
Please check/comment...

### DIFF
--- a/radio/src/definitions.h
+++ b/radio/src/definitions.h
@@ -41,6 +41,12 @@
   #define __SDRAM __DMA
 #endif
 
+#if defined(PCBHORUS) && !defined(SIMU)
+  #define __NOINIT __attribute__((section(".noinit")))
+#else
+  #define __NOINIT
+#endif
+
 #if defined(SIMU) || defined(CPUARM) || GCC_VERSION < 472
 typedef int32_t int24_t;
 #else
--- a/radio/src/targets/horus/board.h
+++ b/radio/src/targets/horus/board.h
@@ -186,18 +186,30 @@ void checkRotaryEncoder(void);
 
 // WDT driver
 #define WDTO_500MS                            500
-#define wdt_disable()
-void watchdogInit(unsigned int duration);
-#if defined(WATCHDOG_DISABLED) || defined(SIMU)
-  #define wdt_enable(x)
-  #define wdt_reset()
+extern uint32_t powerupReason;
+#define SHUTDOWN_REQUEST                      0xDEADBEEF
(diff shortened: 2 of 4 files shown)


projectkk2glider:
Why three new variables. Would it not work with just one new variable in `__NOINIT` section?

kilrah:
Well you could by masking, but apart from that I haven't found how to use less than 3 separate storage units and manage to handle all cases and use the "negative logic" we discussed (one case leads to shtdown, every other to staying on). 

A normal boot looks the same as an unrequested shutdown, so we need an additional marker that survives the reset to differentiate the 2. That marker has to always be set straight away to "failure" to be ready for a reboot, so it can't be used to determine the previous reason during use.

Maybe you can find better...

projectkk2glider:
You are probably right. The logic is very complicated. I am afraid we will *forget* why it was done as it is in the future. In fact even now I was not sure why such a complicated setup is needed. Can you put in some additional comments explaining the whole setup.

bsongis:
I will also need a good night before reviewing this one. Please keep it as it is.

kilrah:
As it is:
- We're pretty damn sure to at least keep power on if anything wrong happens
- There is a slight risk we could end up booting normally with warnings instead of in emergency mode, if `shutdownReason` got corrupted. But I'd say that if we can't trust that we probably counldn't trust the watchdog flag used previously either. An additional check for the `unexpectedShutdown` var in storage (read backup ram first, check it, if 1 go to emergency mode, if not reload from SD) but it's likely overkill.

projectkk2glider:
@kilrah @bsongis why was this not merged?

bsongis:
I will try to spend some time on it later during the week-end, too tired now. I am only good for AVR things :)

bsongis:
Travis seems to report an error

kilrah:
Indeed, for the simu... no idea how to fix

bsongis:
Ok I will have a look

kilrah:
I'm being zapped all over the place this morning so I could test in real conditions if this works in case of ESD discharge... and it does, gets into emergency mode as desired. I confirmed the radio just shuts down without it too. 

Interestingly sometimes with a really good fat charge it seems to actually reboot several times until it finally manages to keep running, as suggested by multiple startup audio "pops". 

@projectkk2glider thanks for the simu fix

projectkk2glider:
That is good news, I was worried that the actual power circuit was affected by ESD and that power off was inevitable. Good work :+1:

projectkk2glider:
This fix has been integrated into #4275

PR #3428, opened 2016-04-17
Implementation of the Multi TX module logic in Companion and few minor fixes/updates for the code

schwabe:


bsongis:
Thanks!

MikeBland:
Any intention of supporting telemetry from a MULTI module in openTx?
I have it working in ersky9x, including the DSM telemetry from a re-flashed Orange module.

bsongis:
@MikeBland would you give me a reference for this orange module? I am waiting for a module for my 9XR-PRO since very long ...

schwabe:
The first problem is the inverted vs non inverted. That could be fixed by either porting the softserial from your ersky9x or but just a adding a transistor/inverter to the board. 

I will add setting the telemetry to the right baudrate of the multimodule to my todo list for the Frsky protocol. Frsky protocol and telemetry via the multi module is obviously not a focus on the Tarnis.

For the DSM a telemetry parser needs to be implemented. Since OpenTX already supports multiple telemetry formats (e.g. Crossfire, sport...) doing so should be relatively straightforward, I could do that but I have no way of testing that since I have only DSMX BNF models that do not have any telemetry (at least I would be surprised if they have).

MikeBland:
This one should be re-flashable to use MULTI: http://www.hobbyking.com/hobbyking/store/__90770__OrangeRX_DSMX_DSM2_Compatible_2_4Ghz_Transmitter_Module_V1_2_JR_Turnigy_Taranis_compatible_.html
This one also that uses the internal antenna (but has been on backorder as you say): http://www.hobbyking.com/hobbyking/store/__61213__OrangeRX_2_4Ghz_Telemetry_Module_For_Turnigy_9XR_Pro_DSMX_DSM2_Walkera_Compatible_.html
This one (also on backorder), is the same as the first one, just has the DIL switches fitted: http://www.hobbyking.com/hobbyking/store/__61212__OrangeRX_DSMX_DSM2_Devo_2_4GHz_Selectable_Transmitter_Module_JR_Turnigy_Compatible_.html
All should be able to use the MULTI firmware, see here: http://openrcforums.com/forum/viewtopic.php?f=40&t=8753#p114549

MikeBland:
The XMEGA processor on the Orange module supports inversion of the I/O pins, so it is possible to configure these to provide inverted or non-inverted output. The green PCB modules include either an 8-way DIL switch, or at least the pads where it would be mounted. One of these could be used to define whether to invert the telemetry output or not.
Adding the soft serial would be a general purpose option.

schwabe:
@MikeBland do you know a cheap dsmx receiver that works with the multi code and has telemetry support? My BNF models don't cut it.

schwabe:
@MikeBland the [https://github.com/opentx/opentx/tree/schwabe/spektrum_telemetry](spektrum_telemetry) branch, now implements Spektrum telemetry. Still needs a bit of polishing but already works. 

And BNF models actually have telemetry but it is useless.

PR #3442, opened 2016-04-24
Log viewer improvements

projectkk2glider:
Here are some improvement for log viewer that I wanted to have for a long time:
 * find and display max altitude
 * enable user to position two cursors and calculate time and altitude difference for them 

Cursors are placed with the right mouse key. Second cursor is placed with SHIFT + right button.

The implementation is open for discussion, especially the visible part. I don't particularly like the way I display the results in a simple label. Ideas how to better use the dialog space are welcome.

Here is an example screen:
screenshot


bsongis:
Thanks!

PR #3448, opened 2016-04-27
Projectkk2glider/new jitter filter with option

### LINKED ISSUE #3432: Centring precision 2.1.8
see this video https://youtu.be/ymvEtfYWABY

projectkk2glider:
This replaces #3250 and closes #3432

It implements better jitter filter than is currently available on 2.1 and also adds an option to turn it on/off. 

The multi-pos switch was no tested!


bsongis:
I merge it now so that we have tests done quickly and we can release

bsongis:
Thanks for this work!

PR #3454, opened 2016-05-01
Projectkk2glider/sensor persistent

projectkk2glider:
First a little background. I have two very similar models on the same radio (9XE with 2.1.8) but at the radio power off one model took much longer to shut down :open_mouth: . This bothered me, since I know I have no setting that would change the model EEPROM.

So I fired up the `simu` with gdb and found out that the culprit is the `calculated sensor` in the affected model. But that sensor does not have the `persistent` option checked!

So my first commit addresses this issue: there is no need to zero saved persistent value on every shutdown. There isn't even the need to zero it out when the persistent gets deselected, but I did it anyway. 

Second commit only shows `persistent` option for the sensors that actually have their value saved at the model save. This are all calculated type sensors.

### DIFF
--- a/radio/src/opentx.cpp
+++ b/radio/src/opentx.cpp
@@ -1964,10 +1964,6 @@ void opentxClose()
         sensor.persistentValue = telemetryItems[i].value;
         eeDirty(EE_MODEL);
       }
-      else if (!sensor.persistent) {
-        sensor.persistentValue = 0;
-        eeDirty(EE_MODEL);
-      }
     }
   }
 #endif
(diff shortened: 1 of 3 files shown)


radio/src/opentx.cpp
```diff
@@ -1964,10 +1964,6 @@ void opentxClose()
         sensor.persistentValue = telemetryItems[i].value;
         eeDirty(EE_MODEL);
       }
-      else if (!sensor.persistent) {
-        sensor.persistentValue = 0;
-        eeDirty(EE_MODEL);
-      }
     }
```
bsongis: Yes I can't remember why I wrote this

bsongis:
Excellent, thanks! Would you port it to next (directly)?

PR #3455, opened 2016-05-02
Stick mode now in one global variable, it should work in Horus emerge…

projectkk2glider:
…ncy mode


radio/src/opentx.h
```diff
@@ -1486,57 +1484,60 @@ extern void opentxInit();
...
-    } version;
+  struct
+  {
+    char id[27];
+  } version;
 #endif
+  struct
+  {
+    uint8_t stickMode;
+  } modelSettings;
```
projectkk2glider: The only real change to this file is new struct `modelSettings`, the rest are just cosmetics.

radio/src/gui/taranis/menu_general_setup.cpp
```diff
@@ -473,9 +477,9 @@ void menuGeneralSetup(uint8_t event)
         if (attr && s_editMode>0) {
           CHECK_INCDEC_GENVAR(event, g_eeGeneral.stickMode, 0, 3);
```
bsongis: The stick mode will change without pressing [Enter], I would prefer we change the reusable variable here

bsongis: I remember one issue we solved in the past with this code, somebody going from mode1 to mode3 with the model ON. Perhaps @romoloman will remember the details, I discussed this with him at the time

radio/src/opentx.h
bsongis: It should be called generalSettings, no?

bsongis:
Yes except the small problems I nociced, this is a better implementation than mine

bsongis:
Thanks!

PR #3460, opened 2016-05-05
Fixes #3214: Flight mode clear

### LINKED ISSUE #3214: Companion flight mode Clear does not reset all fields to defaults
v2.1.7 tested Win 10 & Ubuntu 14.4

For flight mode 0
- GVARs names are not cleared
- GVARs Popup enabled are not unchecked

For flight modes 1 thru 8
- Switch is not reset to dashes thus the flight mode is still active
- GVARs source not reset to "Flight mode 0 value"

elecpower:
Plus a minor fix for Fade In edit finished which did not flag modified.
Only compiled and tested on Ubuntu 14.04


projectkk2glider:
Thanks! I will port this to the next branch.

Issue #3279, opened 2016-02-07
DSM(X) hack module not working on 9XR-PRO


Grawp:
On 9XR-PRO the DSMX module from DX5e is not working on 2.1.x releases while on 2.0.x it does.
There seems to be no signal going to the module. I don't have a scope or logic analyzer at hand right now, however the mean voltage on the signal pin is definitely higher (3.25V) than on 2.0.x versions which may indicate real lack of signal + just having pullup on.

The problem seems to have appeared with a9e64c36036bb6d4ee0fb115c24a920c120938e8 commit. On previous commits the mentioned DSMX module works just fine.

Grawp:
I've connected to my 9XR-PRO via SWD debugger and finally I was able to trace the problem down to not setting correct `required_protocol` in `setupPulses` function in _pulses_arm.cpp_. This is caused by adding `defined(PCBTARANIS)` in a9e64c3 commit. Was there any reason for this?
I've removed the define and it seemed to be working for me so I made a pull request #3471 

Btw. If somebody is interested in knowing how to relatively easily connect SWD debugger to 9XR-PRO and use OpenOCD+GDB+QtCreator(with Bare Metal plugin) to step through the code, I could make a wiki page or a blog about it. There are some small changes necessary to be done in Makefile and the code in both _OpenRcBootloader_ and _OpenTX_.

projectkk2glider:
Your changes in #3471 look good, I will let @bsongis merge. 

>There are some small changes necessary to be done in Makefile and the code in both OpenRcBootloader and OpenTX.

Please elaborate.

projectkk2glider:
The #3471 needs to be ported to the `next` branch also.

Grawp:
> Please elaborate.

Do not switch PB7 and PB6 to GPIO via `MATRIX->CCFG_SYSIO`
and since PB7 is used as CARD DETECT you also need to debug without sdcard inserted or physically rewire the CD to some free GPIO & change the code. This must be done both for the OpenTX and the bootloader.
https://gist.github.com/Grawp/7553e529841b96c1b1c9a2511025c99c
Compiling with `-O0 -g3` is also a good idea.

projectkk2glider:
Makes sense, I did not know that SWD pins are shared. They are not shared on Taranis. 

And did you make a mistake with `-g3` option, it is not know to me?

Grawp:
> And did you make a mistake with -g3 option, it is not know to me?

`man gcc`:

>-glevel
[...]

PR #3471, opened 2016-05-07
Enable DSM protocols on all targets that define DSM2 macro.

Grawp:
Enable DSM protocols on all targets that define DSM2 macro, not just on Taranis.
#3279


bsongis:
Thanks!

projectkk2glider:
@bsongis how do you merge this kind of pull request without making two commits? I would like to do it to :smile: If I just press the merge button two commits are made:  the merge commit and actual original commit.

bsongis:
There is a button for that! It squashes all commits into one and yes that's nice!

PR #3485, opened 2016-05-12
Fix lcd.drawRectangle on Horus

### LINKED ISSUE #3486: [Horus] lcd.drawRectangle force close on Horus

3djc:
This fixes  lcd.drawRectangle force close on Horus. This changes also add a new extra mandatory field 'thickness' : doc will need to be updated accordingly

This fixes #3486

### DIFF
--- a/radio/src/lua/api_lcd.cpp
+++ b/radio/src/lua/api_lcd.cpp
@@ -367,7 +367,7 @@ Draw a rectangle from top left corner (x,y) of specified width and height
 
 @param flags (unsigned number) drawing flags
 
-@param t (number) thikness in pixels
+@param t (number) thickness in pixels
 
 @status current Introduced in 2.0.0
 */
--- a/radio/src/lua/api_lcd.cpp
+++ b/radio/src/lua/api_lcd.cpp
@@ -380,7 +380,7 @@ static int luaLcdDrawRectangle(lua_State *L)
   int h = luaL_checkinteger(L, 4);
   unsigned int flags = luaL_optunsigned(L, 5, 0);
   #if defined(PCBHORUS)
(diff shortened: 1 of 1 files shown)


radio/src/lua/api_lcd.cpp
```diff
@@ -367,6 +367,8 @@ Draw a rectangle from top left corner (x,y) of specified width and height
 
 @param flags (unsigned number) drawing flags
 
+@param t (number) thikness in pixels
```
projectkk2glider: thikness -> thickness

radio/src/lua/api_lcd.cpp
```diff
@@ -377,7 +379,12 @@ static int luaLcdDrawRectangle(lua_State *L)
   int w = luaL_checkinteger(L, 3);
   int h = luaL_checkinteger(L, 4);
   unsigned int flags = luaL_optunsigned(L, 5, 0);
+  #if defined(PCBHORUS)
```
projectkk2glider: Un-indent please, same goes for other two `#if #else #endif` lines.

radio/src/lua/api_lcd.cpp
```diff
@@ -377,7 +379,12 @@ static int luaLcdDrawRectangle(lua_State *L)
...
+  int t = luaL_optinteger(L, 6, 0);
```
projectkk2glider: The `thickness` parameter is unsigned, so I would prefer `unsigned int t =  tluaL_optunsigned()`

3djc: Indeed, changes pushed :)

bsongis:
Thanks!

PR #3487, opened 2016-05-13
Enables lcd.drawGauge on Horus

3djc:
This enables lcd.drawGauge on Horus also. No usage difference with other platforms. Unsure why it wasn't initialy included, so beware !

### DIFF
--- a/radio/src/lua/api_lcd.cpp
+++ b/radio/src/lua/api_lcd.cpp
@@ -417,7 +417,7 @@ static int luaLcdDrawFilledRectangle(lua_State *L)
 
 
 /*luadoc
-@function lcd.drawGauge(x, y, w, h, fill, maxfill)
+@function lcd.drawGauge(x, y, w, h, fill, maxfill [, flags])
 
 Draw a simple gauge that is filled based upon fill value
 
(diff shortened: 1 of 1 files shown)


radio/src/lua/api_lcd.cpp
```diff
@@ -415,7 +415,7 @@ static int luaLcdDrawFilledRectangle(lua_State *L)
   return 0;
 }
 
-#if !defined(COLORLCD)
+
 /*luadoc
 @function lcd.drawGauge(x, y, w, h, fill, maxfill)
```
projectkk2glider: Please add `flags` parameter like so: `maxfill, [flags])`

projectkk2glider:
Looks good.

radio/src/lua/api_lcd.cpp
```diff
@@ -444,13 +444,13 @@ static int luaLcdDrawGauge(lua_State *L)
   int h = luaL_checkinteger(L, 4);
   int num = luaL_checkinteger(L, 5);
   int den = luaL_checkinteger(L, 6);
-  // int flags = luaL_checkinteger(L, 7);
+  int flags = luaL_optinteger(L, 7,0);
```
bsongis: unsigned?

bsongis: cosmetics: space after the comma

radio/src/lua/api_lcd.cpp
3djc: I thought about it, but https://dsbeach.gitbooks.io/opentx-lua-reference-guide/content/lcd/drawGauge.html does not mention that options, so should I add it and we change the doc ?

bsongis:
Perfect, thanks!

PR #3489, opened 2016-05-13
Fix CUSTOM_COLOR,…

3djc:
also change to way setColor work to a more natural lcd.setColor(CUSTOM_COLOR, RED)


radio/src/lua/api_general.cpp
```diff
@@ -855,6 +855,7 @@ const luaR_value_entry opentxConstants[] = {
 #endif
 #if defined(COLORLCD)
   { "COLOR", ZoneOption::Color },
+  { "CUSTOM_COLOR_INDEX", CUSTOM_COLOR_INDEX },
```
bsongis: I think we should remove all these indexes

PR #3490, opened 2016-05-13
Do not export COLOR_INDEX to LUA…

3djc:
They are not needed anymore due to change in setColor


bsongis:
Ah yes they were a lot... Thanks!

PR #3496, opened 2016-05-14
Fixes #3492: delay_01us() and delay_ms() functions were generating wr…

### LINKED ISSUE #3492: Verfy the actual delay of delay_01us() function
Some test suggest that it is not right. Test environment in 2020807473d468fbf13db5798610d87d1c7e7c0e

projectkk2glider:
…ong delays (6 times shorter on Taranis, 8 times on X9E and Horus). Important delays adjusted, function names and code unified.

There is one problem remaining: the delay_01us(1) is now much longer than before. Now it is (I believe, I haven't measured) around 0.2us + call overhead. This is due to the fact that timer registers get updated/reset on the timer input clock.

So before delay_01us(1) was more like 1/8 of 0.1us - despite that all code that used it worked!. 

Therefore the delay_01us() for small values (let say less than 10) should be improved.

This change for example increases ADC acquisition time on Horus from 0.199ms to 0.199ms.

### DIFF
--- a/radio/src/targets/horus/adc_driver.cpp
+++ b/radio/src/targets/horus/adc_driver.cpp
@@ -165,9 +165,9 @@ uint32_t adcReadNextSPIChannel(uint8_t index)
   //
   //        delay value       A5 change     Time needed for adcRead()
   //          1               16            0.154ms - 0.156ms
-  //        300               5             0.197ms - 0.199ms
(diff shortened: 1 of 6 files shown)


bsongis:
>from 0.199ms to 0.199ms
?

projectkk2glider:
Ehh, 0.245ms

Issue #2999, opened 2015-10-27
[Taranis] Problem with SBUS Trainer mode


bsongis:
Reported by Wayne: **we had a few reports from people. We did most all of our testing with the D4R-II in CPPM mode. Only ground checks with Sbus, so we failed to fully test it. Sorry!! The issues take a little time to show up, but sure enough it seems to do it every time. 
We also wired in an X8R just to make sure it was not an RF thing, and it had the same results.
from an observation emailed to me this morning:
But the worst part was that after several minutes with the
Student flying, the plane would go crazy, and returning control
to the Instructor it started working OK.**

But it means ... who has the RSSI warning?

**I think the master radio does. I was out sick this weekend, so was not present for this round of testing.**

would you double check if you can reproduce?
I just read again the code, I really see nothing, I hope it's not just interferences between the 2 modules

**I just asked Bruce a few more details from his weekend testing. He was doing the additional testing as we had heard from a few people about issues when using Sbus with the Linker. It may be a while before I have an answer for you.**

kilrah:
Another report:
http://www.rcgroups.com/forums/showpost.php?p=33039291&postcount=14

And in French:
http://frskytaranis.forumactif.org/t2828-rx-sans-telemetrie-avec-sortir-cppm#33090

kilrah:
If we could have details on what firmware the receivers are using (EU/non-EU) it would be nice too to see if that can play a role.

bsongis:
Here is an update from Bruce:
**I could never get the bad RSSI Critical/Telemetry Lost/Recovered messages
on the ground.
Maybe the model receiver needs to be quite a ways away before it happens?

But I had some success seeing the the XR4-SB Trainer channels misbehave.

I used a D4R-II as the model receiver and used channels 1-4 for the Trainer.

With the Trainer Radio, Student Radio, and Model receiver all about 20 feet apart
I switched a few times between Student and Trainer, then kept it
with Student in control.

After a few minutes, looking at the Channels Monitor, every now and then
I could see the 4 channel values jump around.

Sometimes you had to wait several minutes for it to happen, then it would
happen for maybe 5 seconds.
**

kilrah:
> I could never get the bad RSSI Critical/Telemetry Lost/Recovered messages
on the ground.
Maybe the model receiver needs to be quite a ways away before it happens?

[...]

PR #3501, opened 2016-05-16
Re #2999: More strict SBUS frame check (only allow 25 byte frames, ch…

projectkk2glider:
…eck end byte)

This is what I had in mind when I made comment https://github.com/opentx/opentx/issues/2999#issuecomment-219238649

This code was not tested by me! I have no hardware for this at the moment.

### DIFF
--- a/radio/src/sbus.cpp
+++ b/radio/src/sbus.cpp
@@ -24,40 +24,31 @@
 #define SBUS_FRAME_GAP_DELAY   1000 // 500uS
 
 #define SBUS_START_BYTE        0x0F
+#define SBUS_END_BYTE          0x00
 #define SBUS_FLAGS_IDX         23
 #define SBUS_FRAMELOST_BIT     2
 #define SBUS_FAILSAFE_BIT      3
 
-#define SBUS_CH_BITS          11
-#define SBUS_CH_MASK          ((1<<SBUS_CH_BITS)-1)
+#define SBUS_CH_BITS           11
+#define SBUS_CH_MASK           ((1<<SBUS_CH_BITS)-1)
 
-#define SBUS_CH_CENTER        0x3E0
+#define SBUS_CH_CENTER         0x3E0
 
-uint8_t SbusFrame[SBUS_MAX_FRAME_SIZE];
-uint16_t SbusTimer ;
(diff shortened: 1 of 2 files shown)


bsongis:
I will check it is OK after work

radio/src/sbus.cpp
```diff
@@ -24,40 +24,31 @@
...
+  if (size != SBUS_FRAME_SIZE || sbus[0] != SBUS_START_BYTE || sbus[SBUS_FRAME_SIZE-1] != SBUS_END_BYTE) {
     return; // not a valid SBUS frame
   }
-
-  if (size < SBUS_MIN_FRAME_SIZE) {
-    return;
-  }
-
-  if (size > SBUS_FLAGS_IDX && ((sbus[SBUS_FLAGS_IDX] & (1<<SBUS_FAILSAFE_BIT)) || (sbus[SBUS_FLAGS_IDX] & (1<<SBUS_FRAMELOST_BIT)))) {
+  if (sbus[SBUS_FLAGS_IDX] & (1<<SBUS_FAILSAFE_BIT) || sbus[SBUS_FLAGS_IDX] & (1<<SBUS_FRAMELOST_BIT)) {
```
bsongis: I would see parenthesis here. I am never sure which one gets the higher priority between & and ||

radio/src/sbus.cpp
```diff
@@ -76,11 +67,15 @@ void processSbusInput()
...
+  static uint8_t SbusIndex = 0;
+  static uint16_t SbusTimer;
+  static uint8_t SbusFrame[SBUS_FRAME_SIZE];
+
   while (sbusGetByte(&rxchar)) {
     active = 1;
     SbusFrame[SbusIndex++] = rxchar;
-    if (SbusIndex > SBUS_MAX_FRAME_SIZE-1) {
-      SbusIndex = SBUS_MAX_FRAME_SIZE-1;
+    if (SbusIndex > SBUS_FRAME_SIZE-1) {
```
bsongis: Then the size given to processSbusFrame will be 24?

bsongis:
I just checked your changes with the fixes I did, and it seems OK

projectkk2glider:
Cool, thanks!

PR #3512, opened 2016-05-19
MEGA2560 : Voice, KS108 and "far" Splash

Ingwie:
(Add Jivarofad mods)
- New contributor
- Mega documentation
********************
Add KS108 lcd
Add voice
Add "far" splash function working for firmware > 64Kb


Ingwie:
Travis : error: ‘uint_farptr_t’ has not been declared -> I'll change to uint32_t for compatibility.

radio/src/gui/9x/lcd.cpp
```diff
@@ -1372,6 +1372,26 @@ void lcdInvertLine(int8_t y)
 }
 
 #if !defined(BOOT)
+#if defined(PCBMEGA2560)
```
bsongis: Maybe it could be placed outside the ```#if !defined(BOOT)```, it would be easier to read

radio/src/gui/9x/lcd.cpp
```diff
@@ -1372,6 +1372,26 @@ void lcdInvertLine(int8_t y)
...
+void lcd_imgfar(coord_t x, coord_t y,  uint32_t img, uint8_t idx, LcdFlags att) // progmem "far"
```
bsongis: I don't think uint32_t is the right type for img

radio/src/gui/9x/lcd.cpp
```diff
@@ -1372,6 +1372,26 @@ void lcdInvertLine(int8_t y)
...
+    uint8_t *p = &displayBuf[ (y / 8 + yb) * LCD_W + x ];
+    for (coord_t i=0; i<w; i++){
+      uint8_t b = pgm_read_byte_far(q);
+      q++;
+      ASSERT_IN_DISPLAY(p);
+      *p++ = inv ? ~b : b;
+    }
+  }
+}
+#endif
```
bsongis: It would be nice to avoid this duplicated code, perhaps with templates?

```
template <typename F>
void doOperation(F f)
{
  int temp=0;
  f(temp);
  std::cout << "Result is " << temp << std::endl;
}
```

radio/src/gui/9x/lcd.cpp
Ingwie: Yes, right type is uint_farptr_t but travis cry with

radio/src/gui/9x/lcd.h
```diff
@@ -236,6 +236,9 @@ void lcdDrawTelemetryTopBar();
   lcdDrawSolidVerticalLine(xx  ,yy-ll,ll);  \
   lcdDrawSolidVerticalLine(xx+1,yy-ll,ll)
 
+#if defined(PCBMEGA2560)
+void lcd_imgfar(coord_t x, coord_t y, uint32_t * img, uint8_t idx, LcdFlags att); // progmem "far"
```
bsongis: img doesn't have the same type here

radio/src/gui/9x/popups.cpp
```diff
@@ -82,6 +86,10 @@ void message(const pm_char *title, const pm_char *t, const char *last MESSAGE_SO
 #undef MESSAGE_LCD_OFFSET
 
   lcdRefresh();
+ #if defined(LCD_KS108)
```
bsongis: indentation problem

radio/src/gui/9x/splash.cpp
```diff
@@ -20,18 +20,46 @@
...
+#endif
+
 #if defined(SPLASH)
+
 const pm_uchar splashdata[] PROGMEM = {
   'S','P','S',0,
   #include "bitmaps/9x/splash.lbm"
   'S','P','E',0 };
+  
+#if !defined(PCBMEGA2560)
```
bsongis: Perhaps a #define would avoid this #ifdef

radio/src/gui/9x/splash.cpp
```diff
@@ -20,18 +20,46 @@
...
 const pm_uchar * const splash_lbm = splashdata+4;
+#endif
 
 void drawSplash()
 {
   lcdClear();
-  lcd_img(0, 0, splash_lbm, 0, 0);
-
+#if defined(PCBMEGA2560)
+  lcd_imgfar(0, 0, (pgm_get_far_address(splashdata)+4), 0, 0); //use progmem "far" for splash working with all other options enabled
```
bsongis: I guess you checked there is no other solution provided by the linker?

bsongis: Also you will need to define pgm_get_far_address for the simu, perhaps something like 

```c
#define pgm_get_far_address(ptr) ptr
```

radio/src/gui/9x/lcd.cpp
bsongis: The simu will need a definition for this type, such as uint8_t *

radio/src/targets/9x/lcd_driver.cpp
```diff
@@ -145,6 +192,41 @@ uint8_t lcdRefresh_ST7920(uint8_t full)
 void lcdRefresh()
 {
 #endif
+#if defined(LCD_KS108)
+static uint8_t change = 0; // toggle left or right lcd writing
```
bsongis: Does it mean you always call 2 times lcdRefresh? Why not introduce such a function:

```
void lcdRefresh()
{
  lcdRefreshLeft();
  lcdRefreshRight();
}
```

bsongis:
OK I take it in my plate

radio/src/targets/9x/lcd_driver.cpp
Ingwie: My mind was to have the fastest lcd function. They are two controllers in a KS108 256X64 and the driver toggle left and right (2.5mS by lcdrefresh call) .... But all popup needs two call for drawing correctly. The result of my mind with this LCD : More time for the mixer and less power consumption.

radio/src/gui/9x/lcd.cpp
Ingwie: Yes, it is a near duplicated code : Same as original but with far call -> slow call
This slow function is used just for Splash where we no need rapidity.
Original fast function is great for drawing the sticks and battgraph.

bsongis:
I just opened #3515 which needs tests, I hope it covers all remarks I wrote above

PR #3515, opened 2016-05-20
[MEGA2560] Add KS108 lcd + voice + "far" splash for firmware > 64Kb

bsongis:
In addition I have splitted the lcd_driver.cpp file which was impossible to read


radio/src/targets/mega2560/voice.cpp
```diff
@@ -0,0 +1,167 @@
...
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ */
+
+
+#include "../../opentx.h"
```
bsongis: This file is unchanged compared to the #3512, at least it would need a re-indentation

radio/src/targets/mega2560/voice.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+// Bracame : Driver rewrited for driving WTV20SDMini module
```
bsongis: rewritten

radio/src/targets/mega2560/voice.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+	
+#ifdef SIMU
+#define SOMOBUSY busy
+#else
+#define BUSY
+#endif
+
+
+void somo14_sendstart()
```
bsongis: is "somo" right here?

PR #3524, opened 2016-05-22
Fixes #3473: Inconsistent new model global variables

### LINKED ISSUE #3473: Inconsistent new model global variables initialisation by Companion and radio
The model created on the radio intialises the linked field for each GVar in FM 1-8 to "Flight mode 0 value" whereas Companion initialises to "Own value".

(Taranis X9D, Companion 2.1.8, Bootloader 2.1.7, Eeprom 217, Ubuntu 14.04, Windows 10)

elecpower:
Global variables default to same as radio ie Flight Mode 0 value.
Rotary encoders default to Flight Mode 0 value for general flight mode consistency however this now differs from radio. (being new contributor not confident on radio code but if someone would like to provide the diff I'll include in this pull).
Trim controls disabled when flight mode linked except for +Own trim.
Fixed Companion program crash when opening a model with rotary encoder linked flight modes.

### DIFF
--- a/companion/src/eeprominterface.cpp
+++ b/companion/src/eeprominterface.cpp
@@ -1410,8 +1410,7 @@ void ModelData::clear()
     moduleData[1].protocol = PULSES_OFF;
   }
   for (int i=0; i<C9X_MAX_FLIGHT_MODES; i++) {
-    flightModeData[i].clear();
-    flightModeData[i].setDefaultLinkedFlightModes(i);
+    flightModeData[i].clear(i);
   }
   clearInputs();
   clearMixes();
@@ -1845,8 +1844,9 @@ SimulatorInterface *GetCurrentFirmwareSimulator()
     return NULL;
 }
 
-void FlightModeData::setDefaultLinkedFlightModes(const int phase)
+void FlightModeData::clear(const int phase)
 {
+  memset(this, 0, sizeof(FlightModeData));
   if (phase != 0) {
     for (int idx=0; idx<C9X_MAX_GVARS; idx++) {
       gvars[idx] = 1025;
--- a/companion/src/eeprominterface.cpp
+++ b/companion/src/eeprominterface.cpp
@@ -1411,6 +1411,7 @@ void ModelData::clear()
   }
   for (int i=0; i<C9X_MAX_FLIGHT_MODES; i++) {
     flightModeData[i].clear();
+    flightModeData[i].setDefaultLinkedFlightModes(i);
   }
   clearInputs();
   clearMixes();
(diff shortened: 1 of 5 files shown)


companion/src/eeprominterface.cpp
```diff
@@ -1843,3 +1844,16 @@ SimulatorInterface *GetCurrentFirmwareSimulator()
   else
     return NULL;
 }
+
+void FlightModeData::setDefaultLinkedFlightModes(const int phase)
```
bsongis: Shouldn't ```clear()``` do this job? We could give it ```i``` as an argument as you did here

elecpower: Considered overloading clear() to maintain current functionality and support new.
Checked a few of the other class clear() functions and the norm was for clear() to zeroise the memory space only. Any other defaults were handled by other functions.
Also I am aware that the whole area of global variables is being reworked in 2.2 so wanted to keep this fix code somewhat isolated.
As a novice I took the safe route with the change so the main devs can make a call. Happy to rework into clear() overloaded or not. Please advise.

bsongis: Perfectly understood, you did well. But I remember I did it in some other classes, that's why I ask if it is feasible here. Not completely sure it's used somewhere else to zeroe the structure

companion/src/modeledit/flightmodes.cpp
```diff
@@ -462,11 +469,11 @@ void FlightModePanel::fmClear()
...
       if (gvCount > 0 && (firmware->getCapability(GvarsFlightModes))) {
         for (int i=0; i<gvCount; i++) {
-          gvUse[i]->setCurrentIndex(0);
+          gvUse[i]->setCurrentIndex((phase.gvars[i] > 1024 ? (phase.gvars[i] - 1024) : phase.gvars[i]));
         }
       }
       for (int i=0; i<reCount; i++) {
-        reUse[i]->setCurrentIndex(0);
+        reUse[i]->setCurrentIndex((phase.rotaryEncoders[i] > 1024 ? (phase.rotaryEncoders[i] - 1024) : phase.rotaryEncoders[i]));
       }
```
elecpower: Just spotted a mistake in the two setCurrentIndex when not linked. Thought I tested. Will fix when decision on to merge or not of the set defaults.

companion/src/eeprominterface.cpp
elecpower: As suggested I am in the process of modifying clear() to clear(const int i) and checking for calls.
Also found the radio model initialisation code section and just working thru the maze of compiler defs and macros to write the conditional initialisation of the rotary encoders.
Will update the pull request when ready for next peer review.

projectkk2glider:
@elecpower cherry pick commit 907b6337fc7fec8a0d7e2aa2e2c9ebf351e2d858 from `next` branch. It will fix Travis build.

elecpower:
@projectkk2glider thanks will do tonight

PR #3533, opened 2016-05-25
Allow companion to install in non-default location on linux

sgstreet:
The companion cmake install targets ignores CMAKE_INTALL_PREFIX when installing to a non-default location.  The current install targets always try to install images/linuxicons in /usr/share/icons and udev.rules in /lib/udev/rules.d.

This PR detects the use of non-default install locations and uses relative paths instead of absolute paths for the affected paths.


bsongis:
Thanks!

sgstreet:
No worries.  This is a great project!

PR #3534, opened 2016-05-25
[MEGA2560] KS108+ Voice+Splash "far" (Voice driver,KS108 and ST7920 tested OK)

Ingwie:
+little option (not defined) to reduce encoder resolution (please keep it)


bsongis:
Seems all good, thanks!

PR #3556, opened 2016-05-31
Allow software based volume control

sreichholf:
- enable it for AR9X (it doesn't have HW Volume Control)
- rename "HARDWARE_VOLUME" defines to "MASTER_VOLUME" to mirror it's actual meaning
- adjust build-firmware.py and fwoptions.py


radio/src/audio_arm.cpp
```diff
@@ -511,9 +515,16 @@ void audioTask(void * pdata)
 }
 #endif
 
-void mixSample(audio_data_t * result, int sample, unsigned int fade)
+void mixSample(audio_data_t * result, int sample, unsigned int fade, bool swVolumeAdjust = true)
 {
-  *result = limit(AUDIO_DATA_MIN, *result + ((sample >> fade) >> (16-AUDIO_BITS_PER_SAMPLE)), AUDIO_DATA_MAX);
+  *result = limit(AUDIO_DATA_MIN, *result + ((sample >> fade) >> (16 - AUDIO_BITS_PER_SAMPLE)), AUDIO_DATA_MAX);
+
+#if defined(SOFTWARE_VOLUME)
```
bsongis: Why do you do this here? I would see the SOFTWARE_VOLUME after all mixing is done

sreichholf: The reason are the (Sine based) Tones. 
My initial attempt was to mix loop over the fully mixed buffer and apply volume on that but that caused tones to become cracks if volume is < 100%.
At least that's true with the Speaker delivered with the AR9X (2W 8Ohm).
As this is primarily for AR9X I considered it the "most important target for a good result" and adjusted it to work better.

I haven't debugged this to it's full extent but from looking at the code I do think that tone waves are kind of becoming to steep/irregular due to the interger-related fuzzyness when mixxing before volume control. 
The solution for that was to apply tone volume before mixing which in return meant this had to be done for for each wav sample, too.

But: It's well possible I simply did something wrong, I'd be glad to fix that up once I know what it is :)

radio/src/audio_arm.cpp
```diff
@@ -511,9 +515,16 @@ void audioTask(void * pdata)
...
+  if (!swVolumeAdjust)
+    return;
+  int32_t tmpSample = (int32_t) (uint32_t) (*result);  // conversion from uint16_t
+  *((uint16_t*) result) = (int16_t) ((tmpSample * currentSpeakerVolume) / VOLUME_LEVEL_MAX);
```
projectkk2glider: The problem is here. The zero level audio signal (silence) has a value of AUDIO_DATA_SILENCE, so to convert from unsigned sample value you have to subtract the AUDIO_DATA_SILENCE. Then you do the scaling, and then convert back to unsigned value (add back AUDIO_DATA_SILENCE).

Example of the first part is in simpgmspace.cpp:
```c
    int sample = ((int32_t)(uint32_t)(buff[i]) - 0x8000);  // conversion from uint16_t
```

And like Bertrand suggested, do the sw volume at the end, after all the mixing was done.

radio/src/audio_arm.cpp
```diff
@@ -684,8 +695,12 @@ int ToneContext::mixBuffer(AudioBuffer * buffer, int volume, unsigned int fade)
     }
 
     for (int i=0; i<points; i++) {
+
       int16_t sample = sineValues[int(toneIdx)] / state.volume;
-      mixSample(&buffer->data[i], sample, fade);
+#if defined(SOFTWARE_VOLUME)
+      sample *= (float)currentSpeakerVolume / VOLUME_LEVEL_MAX ;
```
projectkk2glider: This is bad, when you correct the sw volume math as suggested above, this will not be needed.

radio/src/audio_arm.cpp
sreichholf: thx for the explanation!

radio/src/audio_arm.cpp
```diff
@@ -511,9 +515,9 @@ void audioTask(void * pdata)
 }
 #endif
 
-void mixSample(audio_data_t * result, int sample, unsigned int fade)
+void mixSample(audio_data_t * result, int sample, unsigned int fade, bool swVolumeAdjust = true)
```
bsongis: Is the new parameter still needed?

sreichholf: No I missed that one. Sorry!

bsongis:
Perfect!

PR #3558, opened 2016-06-01
Fix columns trim5 and trim6 for Horus

3djc:


3djc:
Closed while we investigates options

PR #3568, opened 2016-06-05
Fixes #3547: Flight modes linked trims display value error

### LINKED ISSUE #3547: Companion flight modes incorrect trim values displayed when cascade linked + own trim
Incorrect value and slider positions when a trim is linked to a flight mode + own trim and that trim is linked to another flight mode.
FM 0 Own Trim enter value 10
FM 1 Use Trim from Flight Mode 0 value 10 displayed
FM 8 Use Trim from Flight Mode 1 or + Own Trim as an Offset value displayed 10
Companion Simulator in FM 8 trim is 2% (correct)

FM 8 Use Trim from Flight Mode 1 + Own Trim as an Offset value displayed 10
FM 8 change the trim to 125
switch to another tab and back to FM 8 trim displays 10 (**wrong**)
Companion Simulator in FM 8 trim is 25% (correct)

Exit editing and return to model edit still displays 10 in FM 8 (**wrong**)

FM 1 Use Trim from Flight Mode 0 + Own Trim as an Offset increase to 25
FM 8 Use Trim from Flight Mode 1 + Own Trim as an Offset displays 25 change to 50
switch to another tab and back to FM 8 trim displays 50 (correct)

FM 1 Own Trim set to 40
FM 8 Use Trim from Flight Mode 1 + Own Trim as an Offset displays 40 change to 100
switch to another tab and back to FM 8 trim displays 100 (correct)

Therefore model settings appear to be stored correctly and just a display issue for specific combination.

Companion 2.1.8 Ubuntu 14.04 Taranus Plus

elecpower:
Fixes display error in Companion when a trim in a linked flight mode sequence has Own Value

### DIFF
--- a/companion/src/eeprominterface.cpp
+++ b/companion/src/eeprominterface.cpp
@@ -1523,9 +1523,7 @@ int ModelData::getTrimValue(int phaseIdx, int trimIdx)
       }
       else {
         phaseIdx = phase.trimRef[trimIdx];
-        if (phase.trimMode[trimIdx] == 0)
-          result = 0;
-        else
+        if (phase.trimMode[trimIdx] != 0)
           result += phase.trim[trimIdx];
       }
     }


projectkk2glider:
Re #3547

PR #3570, opened 2016-06-06
Fix interrupt handler not handling multiprotocol (Pulses fix)

schwabe:


radio/src/pulses/pulses_arm.cpp
```diff
@@ -65,7 +65,7 @@ void setupPulses(uint8_t port)
           required_protocol = PROTO_PXX;
           break;
 #if defined(MULTIMODULE)
-      case MODULE_TYPE_MULTIMODULE:
+        case MODULE_TYPE_MULTIMODULE:
```
bsongis: Indentation is wrong now

radio/src/pulses/pulses_arm.cpp
```diff
@@ -209,6 +209,8 @@ void setupPulses(uint8_t port)
 #endif
     case PROTO_PPM:
       setupPulsesPPM(port);
+      // 45 ms is the worst case length of the PPM frame length (800 µs between channels and 16 channels
```
bsongis: This is not the reason of the 45!
22.5ms is the default length of a PPM frame, it's stored in EEPROM as "0" to have a better compression ratio.

radio/src/targets/horus/extmodule_driver.cpp
```diff
@@ -235,7 +235,11 @@ void extmoduleSendNextFrame()
     EXTMODULE_DMA_STREAM->NDTR = modulePulsesData[EXTERNAL_MODULE].pxx.ptr - modulePulsesData[EXTERNAL_MODULE].pxx.pulses;
     EXTMODULE_DMA_STREAM->CR |= DMA_SxCR_EN | DMA_SxCR_TCIE; // Enable DMA
   }
-  else if (s_current_protocol[EXTERNAL_MODULE] >= PROTO_DSM2_LP45 && s_current_protocol[EXTERNAL_MODULE] <= PROTO_DSM2_DSMX) {
+  else if ((s_current_protocol[EXTERNAL_MODULE] >= PROTO_DSM2_LP45 && s_current_protocol[EXTERNAL_MODULE] <= PROTO_DSM2_DSMX)
```
bsongis: Is it possible to modify (and use here) the ```IS_DSM2_PROTOCOL(...)``` macro? It would avoid a #if macro inside the test

radio/src/targets/taranis/extmodule_driver.cpp
```diff
@@ -224,7 +224,11 @@ void extmoduleSendNextFrame()
     EXTMODULE_DMA_STREAM->NDTR = modulePulsesData[EXTERNAL_MODULE].pxx.ptr - modulePulsesData[EXTERNAL_MODULE].pxx.pulses;
     EXTMODULE_DMA_STREAM->CR |= DMA_SxCR_EN | DMA_SxCR_TCIE; // Enable DMA
   }
-  else if (s_current_protocol[EXTERNAL_MODULE] >= PROTO_DSM2_LP45 && s_current_protocol[EXTERNAL_MODULE] <= PROTO_DSM2_DSMX) {
```
bsongis: The same here

radio/src/pulses/pulses_arm.cpp
schwabe: But the 45 as argument of the scheduleNextMixerCalculation is really in ms. See Crossfire (4) and PXX (9)

bsongis: Yes 45/2 = 22.5ms for the default PPM frame :)

radio/src/pulses/pulses_arm.cpp
schwabe: It is now the same lavel as case MODULE_TYPE_XJT and case MODULE_TYPE_PPM. Or do I miss something like tab vs spaces?

radio/src/pulses/pulses_arm.cpp
schwabe: To me that sound still like the mixer runs only every 2nd ppm frame but maybe I just misunderstood that paramter. I will just remove the comment

radio/src/targets/horus/extmodule_driver.cpp
schwabe: I added the IS_MULTIMODULE_PROTOCOL to avoid an ifdef. The DSM is also used in the ui, where the two macros should be different.

radio/src/pulses/pulses_arm.cpp
bsongis: Sorry it was me! Your indentation is perfect.

radio/src/targets/horus/extmodule_driver.cpp
```diff
@@ -235,7 +235,7 @@ void extmoduleSendNextFrame()
     EXTMODULE_DMA_STREAM->NDTR = modulePulsesData[EXTERNAL_MODULE].pxx.ptr - modulePulsesData[EXTERNAL_MODULE].pxx.pulses;
     EXTMODULE_DMA_STREAM->CR |= DMA_SxCR_EN | DMA_SxCR_TCIE; // Enable DMA
   }
-  else if (s_current_protocol[EXTERNAL_MODULE] >= PROTO_DSM2_LP45 && s_current_protocol[EXTERNAL_MODULE] <= PROTO_DSM2_DSMX) {
+  else if (IS_DSM2_PROTOCOL(s_current_protocol[EXTERNAL_MODULE]) || IS_MULTIMODULE_PROTOCOL(s_current_protocol[EXTERNAL_MODULE])) {
```
bsongis: Even better

bsongis:
Github has a button to squash all commits into one, don't worry. Thanks for your help. Everything working now?

schwabe:
Not quite. I still running into some very obscure bug. Pulses do not work for any external module when multimode is enabled but if I remove the packed attribute of the dsm2 struct 
```
-PACK(struct Dsm2TimerPulsesData {
+struct Dsm2TimerPulsesData {
```
everything works perfect. I still want check/dig into that and see what exactly changes in the memory layout but I did not want to commit this without understanding what is happing there

schwabe:
Just, checked the pulses stop working for me if the ``ModulePulsesData``` union has an odd number of bytes. I used 

```
static inline void foo()
{
  check_size<struct Dsm2TimerPulsesData, 407>();
  check_size<union ModulePulsesData, 413>();
}
``` 

to check its size. Disabling MULTIMODULE and add an uint8_t pad; at the end of pxx pulses also makes the pulses not working. For the normal non MULTIMODULE the pxx struct happens to be bigger and aligned and thus the resulting ModulePulsesData also to be aligned.

The result is that in 
```ModulePulsesData modulePulsesData[NUM_MODULES] __DMA;```

Although the whole array is aligned (__DMA keyword) the second member of the array is not aligned, which happens to be the external pulses. I suspect the STM32 then chokes on the pointer not being 2 byte aligned but did not investigate further into that.

I suggest we either remove the PACK from the pulses struct or force padding on the ModulePulsesData or add the __DMA keword (or another align keyword) to the ModulePulsesData union. __DMA seems to be a bit overkill since it aligns for 32byte.

PR #3571, opened 2016-06-06
tell dfu-util to boot the device to boot after flashing

schwabe:
Only tested on my X9D+. It allows me to boot the device into the OpenTX firmware by using 'make flash' (on switch must be on), combined with -DUSB=SERIAL and -DCLI=YES and the reboot command I can test a new firmware on the device without removing/plugging the usb cable


bsongis:
Why not!

PR #3575, opened 2016-06-07
Fix alignment bug of pulses data array

schwabe:
The second element of the modulePulsesData was never forced to be correctly aligned. Without the new
pulses and without the Multimodule code the second element was never misaligned since the PXX pulses
union member was always the largest member and was 4 byte aligned by coincidence, also making the
second modulePulseData array member 4 byte aligned.

With the Multimodule option enabled the dsm pulses struct becomes the largest union member with 607 bytes
and the second modulePulsesData is not aligned to anything anymore which completely screws up the pulses
generation.

Forcing the alignment of the ModulePulsesData fixes this issue.

### DIFF
--- a/radio/src/pulses/pulses_arm.h
+++ b/radio/src/pulses/pulses_arm.h
@@ -115,13 +115,20 @@ union ModulePulsesData {
 #endif
   PpmPulsesData ppm;
   CrossfirePulsesData crossfire;
-};
+} __DMA;
+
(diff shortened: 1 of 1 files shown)


bsongis:
/home/bertrand/git/opentx/radio/src/pulses/pulses_arm.h:120:3: warning: 'section' attribute does not apply to types [-Wattributes]

bsongis:
I have added an ```uint8_t _alignment``` in the structure to have the alignment OK

PR #3586, opened 2016-06-11
companion 2.2 splash screen

mhotar:
Splash screen for companion 2.2
Single splash screen replaces those customized by radio type we had in 2.1


bsongis:
Excellent!

PR #3588, opened 2016-06-12
3djc/document lua bitmaps

3djc:
Create lua doc for previously added Horus bitmaps functions


bsongis:
Sorry, I forgot again the doc ...

PR #3589, opened 2016-06-12
Add scale to lcd.drawBitmap.

3djc:
For lua writer ease of use, scale is in percent (100 : no change, 50: half size, 200 double)


radio/src/lua/api_lcd.cpp
```diff
@@ -413,9 +415,10 @@ static int luaLcdDrawBitmap(lua_State *L)
   const BitmapBuffer * bitmap = checkBitmap(L, 1);
   int x = luaL_checkinteger(L, 2);
   int y = luaL_checkinteger(L, 3);
+  unsigned scale = luaL_optunsigned(L, 4, 100);
```
bsongis: I think default scale should be 0 to avoid extra maths for nothing

radio/src/lua/api_lcd.cpp
```diff
@@ -413,9 +415,10 @@ static int luaLcdDrawBitmap(lua_State *L)
...
 
   if (bitmap) {
-    lcd->drawBitmap(x, y, bitmap);
+    lcd->drawBitmap(x, y, bitmap, 0, 0, 0, 0, (float) scale/100);
```
bsongis: Here if scale input is 100 you need to give 0 as a scale to drawBitmap

PR #3590, opened 2016-06-12
Fixes 3472: circular linked global variables and rotary encoders in Companion

### LINKED ISSUE #3472: Global variables circular linked flight modes
It is possible to in Companion and radio to configure flight mode global variables to be linked to other flight modes in such a way as to create a circular reference of one or more steps. eg
FM1 GV1 -> FM2
FM2 GV1 -> FM1

FM1 GV2 -> FM3
FM3 GV2 -> FM8
FM8 GV2 -> FM1

Whilst this is a useless config however with up to 9 FMs it could easily be done in error and not noticed. Also you could have a SF which tries to adjust the circular linked GVar.

In Companion the GVar Value field displays 1024 in all the circular linked FMs. However 1024 is a valid value for a GVar. 
gvars circular linked fms

This behaviour is inconsistent with Simulator and radio. The Simulator displays and announces when played zero (0) for the Gvars. The radio announces zero (0) when played. I believe this is the correct handling.
gvars circular linked fms radio
gvars circular linked fms simulator

Companion should be fixed to be consistent with Simulator and radio.

Enhancement to Companion and radio (if can be squeezed into firmware) to warn when attempting to configure a circular reference. Less of a priority for radio as the current display shows all the values and FMs for each GVar on one line.

(Taranis X9D Windows 10 Ubuntu 14.04 Companion 2.1.8)

elecpower:
Fixes #3472 Circular linked global variables.
Fixes circular linked rotary encoders.
Warn user when creating circular links.
Move rotary encoder code to ModelData class for consistency with global variables.

### DIFF
--- a/companion/src/eeprominterface.cpp
+++ b/companion/src/eeprominterface.cpp
@@ -1536,16 +1536,57 @@ bool ModelData::isGVarLinked(int phaseIdx, int gvarIdx)
   return flightModeData[phaseIdx].gvars[gvarIdx] > 1024;
 }
 
-int ModelData::getGVarValue(int phaseIdx, int gvarIdx)
+bool ModelData::isGVarLinkedCircular(int phaseIdx, int gvarIdx)
 {
-  int idx = flightModeData[phaseIdx].gvars[gvarIdx];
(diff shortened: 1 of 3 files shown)


bsongis:
Thanks!

PR #3593, opened 2016-06-12
Implement a constant calculated telemetry source (closes #3579)

### LINKED ISSUE #3579: [Enhancement][OpenTX 2.1.x] New calculated sensor: electric fuel gauge
Hi folks,

I'm using the UniSens-E sensor. It is a great very compact sensor. One feature which I really like is the "fuel" feature. When programming the UniSens-E with its USB dongle, you can setup a LiPo capacity. Then the UniSens-E will transmit a fuel information from 100% (LiPo full) to 0% (LiPo empty).
This is very, very convenient to setup some warning/alarm (in my case a warning at 35% and an alert at 30%). The major inconvenient is: you need to used the USB dongle to change the LiPo capacity.

I realized that almost everything is present on OTX 2.1 to achieve the same feature.
There already is a "consumption sensor" (Calculated → Consumption → name of current sensor).

Would it be possible to add an "**_electric fuel/gauge**_" calculated sensor?
**Calculated → Electric fuel (or gauge) → name of current sensor → max capacity of the LiPo**
The unit will be %
The sensor will start at 100% (consumption=0) and decrease to 0% (consumption=capacity).
The basic computation will be something like:
_1 - Consumption / Capacity_

The nice thing to be able to do it from OTX, is be able to create several calculated sensors with several different LiPo capacity (for a same model).
Then, by using LS with the additional AND, we can select the appropriated capacity, but the alarm logic (and associated SF) will always be the same. It will even be better!! :smile: 

What do you think?

schwabe:
Tested on Taranis Plus. Pull request is against next since 2.1 Companion does not build for me. If the concept is okay, I can provide also a master version.


radio/src/gui/9x/menu_model_telemetry.cpp
```diff
@@ -344,6 +344,14 @@ void menuModelSensor(uint8_t event)
             }
             break;
           }
+          else if (sensor->formula == TELEM_FORMULA_CONSTANT) {
+            lcd_putsLeft(y, STR_VALUE);
+            if (attr)
+              sensor->custom.offset = checkIncDec(event, sensor->constant.value, -30000, +30000, EE_MODEL|NO_INCDEC_MARKS|INCDEC_REP10);
```
projectkk2glider: shuld be `sensor->constant.value = checkIncDec`

radio/src/gui/horus/menu_model_telemetry.cpp
```diff
@@ -259,6 +259,13 @@ bool menuModelSensor(evt_t event)
             }
             break;
           }
+          else if (sensor->formula == TELEM_FORMULA_CONSTANT) {
+            lcdDrawText(MENUS_MARGIN_LEFT, y, STR_VALUE);
+            if (attr) sensor->custom.offset = checkIncDec(event, sensor->constant.value, -30000, +30000, EE_MODEL|NO_INCDEC_MARKS|INCDEC_REP10);
```
projectkk2glider: Same as above

radio/src/gui/taranis/menu_model_telemetry.cpp
```diff
@@ -256,6 +256,14 @@ void menuModelSensor(uint8_t event)
             }
             break;
           }
+          else if (sensor->formula == TELEM_FORMULA_CONSTANT) {
+            lcd_putsLeft(y, STR_VALUE);
+            if (attr) sensor->custom.offset = checkIncDec(event, sensor->constant.value, -30000, +30000, EE_MODEL|NO_INCDEC_MARKS|INCDEC_REP10);
```
projectkk2glider: Again :smile:

bsongis:
I wonder if this PR is useful if we had a proper *ratio* handling

schwabe:
With proper Ratio handling and offset and so on you can propably caputure a lot of use cases that the constant telemetry source also handles (like subtracting/adding/multiplying with an abritrary value). 

I think the advantage of this is, that is a small change, that is easy to understand and use and might be useful for other use cases too like quick testing a script, warnings, etc.

schwabe:
Should I just close this?

bsongis:
Not now

RealTadango:
@bsongis Please re-consider this. It would be very use full to translate telemetry data into human interpret able values.

PR #3602, opened 2016-06-15
Bsongis/issue 3600 emergency shutdown

bsongis:


radio/src/tasks_arm.cpp
```diff
@@ -91,6 +91,32 @@ uint16_t stackAvailable()
 }
 #endif
 
+uint32_t timeForcePowerOffPressed = 0;
```
projectkk2glider: Should be `volatile`. It could also be `uint16_t` to save 2 bytes of RAM, still enough range (wrap around in 650 seconds).

```c
volatile uint16_t timeForcePowerOffPressed;
uint16_t delay = (int16_t)get_tmr10ms() - timeForcePowerOffPressed;
```

projectkk2glider:
Good idea, much simpler than I thought it would be.

radio/src/tasks_arm.cpp
bsongis: I agree for the uint16_t
Why the cast to int16_t?

projectkk2glider: My mistake, cast to uint16_t.

PR #3605, opened 2016-06-17
Access ChanMon from main screen using a short MODEL key push

3djc:
This enables access to the ChanMon also from main view (in addition to the existing access from MODEL pages) using a short MODEL key push.

### DIFF
--- a/radio/src/gui/horus/view_main.cpp
+++ b/radio/src/gui/horus/view_main.cpp
@@ -203,16 +203,4 @@ bool menuMainView(evt_t event)
   return true;
 }
 
-#if 0
-bool menuMainViewChannelsMonitor(evt_t event)
-{
-  switch (event) {
-    case EVT_KEY_BREAK(KEY_EXIT):
-      chainMenu(menuMainView);
-      event = 0;
-      return false;
-  }
 
-  return menuChannelsView(event);
-}
-#endif
--- a/radio/src/gui/horus/view_main.cpp
+++ b/radio/src/gui/horus/view_main.cpp
@@ -152,6 +152,11 @@ bool menuMainView(evt_t event)
       pushMenu(menuTabModel[0]);
       return false;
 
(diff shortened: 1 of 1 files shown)


3djc:
Not working as expected, removing it

kilrah:
I'd maybe rather have it in the "long enter" menu...

JimMchugh:
I agree with Kilrah. On the "long enter press" menu would be ideal...

JimMchugh:
Also, on that subject, could the (new) menu item order on that menu be:

Model Select
Channel Monitor
Reset
Setup Screens
Statistics
About

Could Statistics and About be merged into a single screen?

Only a suggestion from a user, you guys know what you're doing!

HorusChris:
It would be good to see Channel monitor as a selection in that menu. But I think I'd still access it from home screen by long press MOL, short press MOL.

mpjf01:
The operation is a little inconsistent as a long press on MOL in the HOME screen gives MODEL SETUP and another long press is required for MONITOR, whereas in the other screens just one long press is needed. Is there no other key that could be used (I assume there is a desire not to use the same key for that function as with FRSKYOS).

3djc:
I'm unsure why setup screens should be in there since the are on shortcut with long TELE, so would the following make sense ? 

Model Select
Monitors
Reset
Statistics
About

JimMchugh:
3djc: Yes I 100% agree! I assume the "Monitors" would have a series on monitor pages: Channel Monitor 1 / 2 plus your new Logical Switch Monitor(s)....?

I support this...

Again, could Statistics & About be merged into one page to keep the menu list short?

kilrah:
I think that dates back from before the long TEL shortcut... Suggestion looks good to me

3djc:
Yes Jim, I justed pushed a PR to add mixers monitor to the existing output monitors, and logical switches will follow, so definitively monitorS :)

JimMchugh:
@3djc Good work man!

PR #3612, opened 2016-06-18
Fixes #3549: disable trim processing in while shutdown is in progress…

### LINKED ISSUE #3549: Trim sound during switch off
While powering off a Taranis X9D, if a trim is moved while the power off process is being run, the trim sound is audible as if the trim for the last model could still be changed and recorded.

I consider this as a bug (because this is at best confusing the user) and would propose the following:

1/ get the confirmation that the trim of a model cannot be altered during the shutdown procedure;
2/ when this is confirmed, prevent the trim sound from beeping during the shutdown procedure.

I can assist if I am pointed out to the proper code files, I could not find any good lead in my first quick search for evidence.
Regards,
B.

projectkk2glider:
…, do no turn of the back-light before the shutdown

Closes #3549

### DIFF
--- a/radio/src/opentx.cpp
+++ b/radio/src/opentx.cpp
@@ -1944,6 +1944,11 @@ void opentxStart()
 #if defined(CPUARM) || defined(CPUM2560)
 void opentxClose()
 {
+#if defined(CPUARM)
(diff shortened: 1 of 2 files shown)


bsongis:
Thanks!

PR #3614, opened 2016-06-19
Rearange long enter menu

3djc:
New order :
Model Select
Monitors
Reset
Statistics
About


radio/src/gui/horus/view_main.cpp
```diff
@@ -29,6 +29,8 @@
 #define TRIM_LEN                       80
 #define POTS_LINE_Y                    (LCD_H-20)
 
+extern uint8_t lastMonitorPage;
```
bsongis: It should be inside menus.h then, I think there is another location it could be removed

3djc: Yeap, doing it now

3djc: It hum *was* already ion there, just forgot it was :D

bsongis:
All good, thanks!

PR #3636, opened 2016-06-30
3djc/horus improve mixer setup

3djc:
Improvement of model setup mixers view:

- real time mixer out and output view for selected mix on bottom bar (output name also displayed)
- use of icons for mix type
- optimisation of columns size and position to allow more mix settings configurations to display nicely
image


radio/src/gui/horus/menu_model_mixes.cpp
```diff
@@ -330,8 +338,32 @@ void displayMixLine(coord_t y, MixData *md)
   displayFlightModes(MIX_LINE_FM_POS, y, md->flightModes, 0);
 }
 
+void displayMixStatus(uint8_t channel)
+{
+  static const BitmapBuffer * mixer_bmp = BitmapBuffer::loadMaskOnBackground("mask_sbar_mixer.png", MENU_TITLE_COLOR, HEADER_BGCOLOR);
```
projectkk2glider: `static` here would only make sense if you loaded bitmaps only one time. The way it is now it is only consuming global RAM.

In fact I think that the bitmaps should be loaded only once when user enters this screen and released once the user leaves the screen.

radio/src/gui/horus/menu_model_mixes.cpp
```diff
@@ -330,8 +338,32 @@ void displayMixLine(coord_t y, MixData *md)
...
+  lcd->drawBitmap(MIX_STATUS_ICON_TO, MENU_FOOTER_TOP, to_bmp);
+  lcd->drawBitmap(MIX_STATUS_ICON_OUTPUT, MENU_FOOTER_TOP, output_bmp);
+
+  lcdDrawSizedText(MIX_STATUS_OUT_NAME, MENU_FOOTER_TOP, g_model.limitData[channel].name, sizeof(g_model.limitData[channel].name), MENU_TITLE_COLOR | LEFT | ZCHAR);
+  drawSingleOutputBar(MIX_STATUS_OUT_BAR, MENU_FOOTER_TOP + 4, MIX_STATUS_BAR_W, MIX_STATUS_BAR_H, channel);
+}
+
 bool menuModelMixAll(evt_t event)
 {
+  static const BitmapBuffer * mpx_mode[] = {
```
projectkk2glider: Same as previous comment.

projectkk2glider:
:+1: on bitmaps.

radio/src/gui/horus/themes/darkblue.cpp
```diff
@@ -116,7 +116,34 @@ class DarkblueTheme: public Theme
       loadMenuIcon(ICON_MONITOR_CHANNELS3, "mask_monitor_channels3.png");
       loadMenuIcon(ICON_MONITOR_CHANNELS4, "mask_monitor_channels4.png");
       loadMenuIcon(ICON_MONITOR_LOGICAL_SWITCHES, "/mask_monitor_logsw.png");
+
```
bsongis: Perhaps this should be done in a ```Theme::loadThemeBitmaps()``` function, no?

bsongis:
Excellent!

PR #3638, opened 2016-06-30
Projectkk2glider/gcc6 compilation fixes

projectkk2glider:
Tested with:
* gcc-6 (Ubuntu 6.1.1-3ubuntu11~14.04.1) 6.1.1 20160511
* gcc (Ubuntu 4.8.5-2ubuntu1~14.04.1) 4.8.5
* arm-none-eabi-gcc (GNU Tools for ARM Embedded Processors) 4.8.4 20140725 (release) [ARM/embedded-4_8-branch revision 213147]

I did not test the resulting firmware on the actual radio. The first commit should be used in any case since it is unrelated and was only exposed by the gcc 6.

### DIFF
--- a/companion/src/firmwares/opentx/simulator/opentxsimulator.cpp
+++ b/companion/src/firmwares/opentx/simulator/opentxsimulator.cpp
@@ -77,6 +77,9 @@
 #undef min
 #undef max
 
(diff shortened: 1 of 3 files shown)


bsongis:
Thanks!

PR #3646, opened 2016-07-05
update function is required

3djc:
If not, it gnerates lua errors when not displayed. it can be empty, but since we have one, let's keep it counting !

Thx Noircogi for spotting this one

### DIFF
--- a/radio/sdcard/horus/WIDGETS/Counter/main.lua
+++ b/radio/sdcard/horus/WIDGETS/Counter/main.lua
@@ -14,9 +14,13 @@ local function update(pie, options)
   pie.options = options
 end
 
+local function background(pie)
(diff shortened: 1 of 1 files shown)


bsongis:
I like this change, but the background function should be optional (as it is in telemetry scripts). I guess I missed something, what is the error?

3djc:
The error is that background gets called no matter what in interface.cpp and genrate an error in LuaWidget::background(). We could of course check for presence of the background prior trying to execute it

bsongis:
Yes I think it is done this way in telemetry scripts

bsongis:
And I did it for function scripts recently. I forgot widgets scripts for sure

PR #3647, opened 2016-07-05
Add function name to debug output

3djc:


radio/src/lua/interface.cpp
```diff
@@ -981,7 +981,7 @@ class LuaWidgetFactory: public WidgetFactory
       }
 
       if (lua_pcall(L, 2, 1, 0) != 0) {
-        TRACE("Error in widget %s: %s", getName(), lua_tostring(L, -1));
+        TRACE("Error in create() of widget %s: %s", getName(), lua_tostring(L, -1));
```
bsongis: @kilrah @projectkk2glider is it good english? we both are french :)

projectkk2glider: I am no expert :blush:, but it sounds weird. My suggestion: `Error in widget %s create() funciton: %s`

bsongis: I thought of something like this, yes. Without the typo ;)

projectkk2glider: :stuck_out_tongue:

radio/src/lua/interface.cpp
```diff
@@ -981,7 +981,7 @@ class LuaWidgetFactory: public WidgetFactory
       }
 
       if (lua_pcall(L, 2, 1, 0) != 0) {
-        TRACE("Error in widget %s: %s", getName(), lua_tostring(L, -1));
+        TRACE("Error in widget %s create() function : %s", getName(), lua_tostring(L, -1));
```
bsongis: no space before : in english (not the same than french)

bsongis:
Yes it should be OK. Just the typos

bsongis:
Thanks!

PR #3652, opened 2016-07-05
Fixes #3642: sdInit() now called in one place only (inside the opentx…

### LINKED ISSUE #3642: Model audio files not played
Model audio files (for example flight mode on/off) are not played when the radio is powered up. If the model is reloaded, then the audio files start playing. The problem is only on the radio (X9E), `simu` works just fine.

This was not a problem on 2.1.

projectkk2glider:
…Init()). Some reordering of opentxInit().

Please review carefully and test. I only tested on Taranis. 

The SD card is now initialized in `opentxInit()`. Before this change `referenceModelAudioFiles()` was being executed before SD was initialized. 

I also rearanged the flow of   `opentxInit()` to make it more simple and clear. 

I think  there was also bug - the `topbar = new Topbar(&g_model.topbarData)` was callled before the model data was read.

### DIFF
--- a/radio/src/gui/horus/topbar.cpp
+++ b/radio/src/gui/horus/topbar.cpp
@@ -21,7 +21,7 @@
 #include "opentx.h"
 
 
-static Topbar * topbar;
+static Topbar * topbar = 0;
 
 unsigned int Topbar::getZonesCount() const
 {
@@ -52,6 +52,19 @@ void drawTopbarDatetime()
 
   getTimerString(str, getValue(MIXSRC_TX_TIME));
   lcdDrawText(DATETIME_MIDDLE, DATETIME_LINE2, str, SMLSIZE|TEXT_INVERTED_COLOR|CENTERED);
+}
+
+Topbar * getTopBar()
+{
+  return topbar;
+}
+
+void loadTopBar()
+{
+  if (!topbar) {
+    topbar = new Topbar(&g_model.topbarData);  
+  }
+  topbar->load();
 }
 
 void drawTopBar()
--- a/radio/src/gui/horus/topbar.cpp
+++ b/radio/src/gui/horus/topbar.cpp
@@ -20,6 +20,9 @@
 
 #include "opentx.h"
 
+
+static Topbar * topbar;
+
 unsigned int Topbar::getZonesCount() const
 {
   return MAX_TOPBAR_ZONES;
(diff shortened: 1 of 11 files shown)


radio/src/opentx.cpp
```diff
@@ -2488,16 +2488,14 @@ void opentxInit(OPENTX_INIT_ARGS)
...
+    sdInit();
 #endif
-
-#if defined(COLORLCD)
-  topbar = new Topbar(&g_model.topbarData);
-  luaInit();
-#endif
+  }
 
 #if defined(RAMBACKUP)
```
bsongis: I don't remember what happens if lua is not initialized here, in case the screens have lua widgets inside

bsongis:
I like this PR, just need to check or test for the remark I wrote above

projectkk2glider:
I don't like the fact that `luaInit()` is called from two different places (depending on the radio). I will look into it to see if some code simplification and unification can be done.

radio/src/gui/horus/topbar.cpp
```diff
@@ -20,6 +20,9 @@
 
 #include "opentx.h"
 
+
+static Topbar * topbar;
```
bsongis: = 0 is not mandatory but good practice

radio/src/gui/horus/topbar.cpp
```diff
@@ -88,6 +91,9 @@ void drawTopBar()
     lcdDrawSolidFilledRect(LCD_W-122+4*i, 30, 2, 8, i >= bars ? MENU_TITLE_DISABLE_COLOR : MENU_TITLE_COLOR);
   }
 
+  if (!topbar) {
```
bsongis: Why did you move it there?

projectkk2glider: It is the only place where variable `topbar` is used.

bsongis: I think I would prefer a ```topbarInit()``` function, no?

projectkk2glider:
I am not sure I like what I did with the topbar. 

Lua will need more elaborate checking/refactoring. It is a bit of a mess right now with the Hours working differently than Taranis.  There also looks like permanent scripts are not loaded on Horus.

radio/src/opentx.cpp
```diff
@@ -2552,13 +2548,7 @@ void opentxInit(OPENTX_INIT_ARGS)
 
   if (g_eeGeneral.backlightMode != e_backlight_mode_off) backlightOn(); // on Tx start turn the light on
 
-  if (UNEXPECTED_SHUTDOWN()) {
-#if !defined(CPUARM)
-    // is done above on ARM
```
bsongis: FYI it was done this way (yes, difficult to understand) to save some bytes of flash on 9X. I agree on revoving this

bsongis:
Integrated into #3674

PR #3656, opened 2016-07-16
Schwabe/inofficial multi module21

schwabe:
Backport of the multi module to 2.1. The companion support is only selecting the multiprotocol option and selecting that protocol as external protocol but not the individual settings of the multi protocol. 

The code is almost identical to the one in next otherwise. The only main noteable difference to next is the definition in ``myepprom.h`` that does not move the ppm option into a name ``ppm`` substructure like the recaftoring done in next but instead keep source code compatible with master.


radio/src/pulses/pulses.h
```diff
@@ -84,6 +84,16 @@ enum ModuleFlag
   #define IS_DSM2_SERIAL_PROTOCOL(protocol)  (0)
 #endif
 
+#if defined(MULTIMODULE)
+  #define IS_MULTIMODULE_PROTOCOL(protocol)  (protocol==PROTO_MULTIMODULE)
+  #if !defined(DSM2)
+     #error You need to enable DSM2 = PPM for MULTIMODULE support
```
kilrah: Would it not be better to force it in the Makefile?
Move the 
``` ifeq ($(MULTIMODULE), YES)```
block before the DSM2 one, and add `DSM2=PPM` in it?

schwabe: But since this is Taranis only and the Taranis already has DSM2 uncoditionally enabled. This define is just as an addtional safeguard to let people out what is wrong when mixing compilation options (as it is possible in next).

kilrah: Oh yes, forgot it was always in. Forget it then :)

bsongis:
Will be merged later if 2.2 is delayed

PR #3659, opened 2016-07-20
Spektrum telemetry implementation

schwabe:
Implementation of Spektrum Telemetry for multi module. The parser should also work with other modules (OrangeRX?) but I have no idea what baud/rate/mode they use or even a module to test.

Tested with Blade nano CPX BNF model (has the frame data telemetry) and lemon rx receiver/telemetry. Most of the sensors are untested but should work.


griebd:
ok, I have a ultra micro 4-site (AR6400L receiver) and a diy dsm module using a DX6e (?) (http://openrcforums.com/wiki/index.php/How_to_modify_your_9x_to_the_DSM_protocol)...
does it support any kind of telemetry? as far as I knew the answer for BNF were always no...
yes, I can test and code if necessary ;)
how I miss knowing how my model battery is doing during fly...

schwabe:
@griebd I don't think that dsm module outputs telemetry. The BNF models I have, only send times in failsafe (holds) and frame losses but no voltage or any other useful telemetry.

kilrah:
The DIY hack module doesn't do telemetry.

bsongis:
seems good, thanks!

PR #3660, opened 2016-07-21
Add cmake configuration for Libfox.

schwabe:
CMake on OS X does not find libfox in the homebrew location otherwise


bsongis:
good job!

PR #3661, opened 2016-07-23
split up the cmakelist by target

nathantsoi:
in an effort to make this more readable, i've split most target specific stuff out into different files.

im a cmake nub, feedback is welcome. the goal here is to make it more manageable to port to / support more targets. just adding a new target was hard, so i started here.


radio/src/CMakeLists.txt
```diff
@@ -374,388 +107,21 @@ if(NOT PCB STREQUAL 9XR AND NOT PCB STREQUAL 9XRPRO)
...
-    STM32F4xx_StdPeriph_Driver/src/stm32f4xx_rcc.c
-    STM32F4xx_StdPeriph_Driver/src/stm32f4xx_spi.c
-    STM32F4xx_StdPeriph_Driver/src/stm32f4xx_i2c.c
-    STM32F4xx_StdPeriph_Driver/src/stm32f4xx_dma.c
-    STM32F4xx_StdPeriph_Driver/src/stm32f4xx_rtc.c
-    STM32F4xx_StdPeriph_Driver/src/stm32f4xx_pwr.c
-    STM32F4xx_StdPeriph_Driver/src/stm32f4xx_usart.c
-    )
-  set(FIRMWARE_TARGET_SRC ${FIRMWARE_TARGET_SRC} system_stm32f4xx.c)
+  include(cpu/stm32f4/CMakeLists.txt)
```
bsongis: I would prefer targets/common/stm32f4/CMakeLists.txt

nathantsoi:
good idea @bsongis, how about something like this?

.gitignore
```diff
@@ -10,3 +10,4 @@ radio/src/bootloader/bootloader_ramBoot*
 *.stamp
 radio/src/lua/lua_fields_*.txt
 radio/src/lua/lua_exports_*.txt
+build-*/
```
bsongis: build* will be even better

radio/src/CMakeLists.txt
```diff
@@ -107,21 +109,21 @@ if(NOT PCB STREQUAL 9XR AND NOT PCB STREQUAL 9XRPRO)
 endif()
 
 if(CPU_TYPE STREQUAL STM32F4)
-  include(cpu/stm32f4/CMakeLists.txt)
+  include(common/arm/stm32/f4/CMakeLists.txt)
```
bsongis: yes why not!

PR #3666, opened 2016-07-25
Fix AR9X build and add it to regular commit tests.

schwabe:
Note that simu still crashes on startup in the readRtc function

### DIFF
--- a/radio/src/targets/simu/opentxsimulator.cpp
+++ b/radio/src/targets/simu/opentxsimulator.cpp
@@ -95,7 +95,7 @@ void OpenTxSimulator::start(QByteArray & ee, bool tests)
 
 void OpenTxSimulator::start(const char * filename, bool tests)
 {
-#if defined(PCBSKY9X) && !defined(REVX)
+#if defined(PCBSKY9X) && !defined(REVX) && !defined(AR9X)
   g_rotenc[0] = 0;
 #elif defined(PCBGRUVIN9X)
   g_rotenc[0] = 0;
@@ -164,7 +164,7 @@ void OpenTxSimulator::wheelEvent(int steps)
     rotencValue -= 2;
   else
     rotencValue += 2;
-#elif defined(PCBSKY9X) && !defined(REVX)
+#elif defined(PCBSKY9X) && !defined(REVX) && !defined(AR9X)
   g_rotenc[0] += steps*4;
 #elif defined(PCBGRUVIN9X)
   g_rotenc[0] += steps;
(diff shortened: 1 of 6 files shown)


radio/src/targets/simu/opentxsimulator.cpp
```diff
@@ -95,7 +95,7 @@ void OpenTxSimulator::start(QByteArray & ee, bool tests)
 
 void OpenTxSimulator::start(const char * filename, bool tests)
 {
-#if defined(PCBSKY9X) && !defined(REVX)
+#if defined(PCBSKY9X) && !defined(REVX) && !defined(AR9X)
```
bsongis: perhaps && ROTARY_ENCODERS > 0 ? 
I am not sure

radio/src/targets/sky9x/i2c_driver.cpp
```diff
@@ -18,6 +18,7 @@
  * GNU General Public License for more details.
  */
 
+#include "opentx.h"
 #include "i2c_driver.h"
```
bsongis: here some includes can be removed then

radio/src/targets/sky9x/rtc_ds3231_driver.cpp
```diff
@@ -18,6 +18,7 @@
  * GNU General Public License for more details.
  */
 
+#include "opentx.h"
 #include "rtc.h"
```
bsongis: the same here 
not sure though

bsongis:
> Note that simu still crashes on startup in the readRtc function

Do you need help for this?

radio/src/targets/simu/opentxsimulator.cpp
schwabe: Neither am I

schwabe:
No the second commit fixes the crashes in the simulator

radio/src/targets/simu/opentxsimulator.cpp
bsongis: I just checked, you should use NUM_ROTARY_ENCODERS > 0 and something is missing here:
#if defined(PCBSKY9X) && (defined(REVX) || defined(AR9X)
  #define MAX_ROTARY_ENCODERS          1
  #define NUM_ROTARY_ENCODERS          0
#elif defined(PCBSKY9X)
  #define MAX_ROTARY_ENCODERS          1
  #define NUM_ROTARY_ENCODERS          1
...

PR #3675, opened 2016-08-01
Add getRAS lua function

3djc:


radio/src/lua/api_general.cpp
```diff
@@ -498,6 +498,22 @@ static int luaGetValue(lua_State * L)
...
+Return the RAS value
+
+@retval number representing RAS value. Values bellow 50 are all ok, values higher indicate an issue
+This is just a hardware pass/fail measure and does not represent the quality of the radio link
+
+@status current Introduced in 2.2.0
+*/
+static int luaGetRAS(lua_State * L)
+{
+  lua_pushinteger(L, telemetryData.swr.value);
```
bsongis: In case it's not reliable I would push nil, no?

3djc:
I have made the change, but I'm worried that it could generate a lua fault if use simply for something like drawNumber(x,y, getSWR(),DBLSIZE) in case of faulty antenna

radio/src/lua/api_general.cpp
```diff
@@ -498,6 +498,26 @@ static int luaGetValue(lua_State * L)
...
+Return the RAS value or nil in case of issue 
+
+@retval number representing RAS value or nil if if there is an hardware issue
+This is just a hardware pass/fail measure and does not represent the quality of the radio link
+
+@status current Introduced in 2.2.0
+*/
+static int luaGetRAS(lua_State * L)
+{
+  if (telemetryData.swr.value > 0x33) {
```
bsongis: I don't think the test is right, I would rather see:

```
IS_VALID_XJT_VERSION()
```

Well indeed this macro name is not very good ...

bsongis: I just pushed a change, you may use `IS_SWR_VALUE_VALID()` now

bsongis:
Sorry for changing again my mind, but I will follow @kilrah and @3djc first advice, not to export SWR/RAS to Lua. This is something which is given for DEBUG only in the statistics screen, not for a widget in the first screen of the radio ...

PR #3681, opened 2016-08-03
Fix gcc 5.4 compiler warnings

schwabe:
The order preference in flash_driver would evaluate ! before & but I think checking the bit first is what was intended


schwabe:
This also affects 2.1

Issue #3673, opened 2016-07-31
[Bug][Radio][2.1.8] Bug with Spektrum Hack Module


LapinFou:
Hi folks,
When creating a new model with the hack module, choosing "DSM2 DSMX" will coerce the channel to CH1-**8**.
It should be CH1-**6**.
I can reproduce the problem on my Taranis and also with the simulator.

LapinFou:
Will this "cosmetic" bug also be fixes in the OTX 2.1 branch?

schwabe:
@LapinFou It is already fixed in 2.1. And I called it cosmetic because the only thing that the commits changes is what is shown in the. The DSM2 code does not evaluate the channel count always sends six channels to the module

LapinFou:
Thanks for the clarification. :smile: 
Since it was tagged for the Milestone ***OpenTX 2.2.X*** I thought it will not be fixed for ***OpenTX 2.1.X***.

PR #3689, opened 2016-08-04
Fix channel count on dsm2 module, purely cosmetic (closes issue #3673)

schwabe:
Taranis menu always use 8 channels as default opposed to 9x and Horus, which use the maximum number as default channel count (which is 6 on dsm2). This commit does not change that logic.


schwabe:
For the default number of channel I think the Taranis behaviour of only 8 makes more sense than the 16 of 9x/Horus since D16 will have 11ms for 8ch and 22ms for 16

bsongis:
:+1

PR #3701, opened 2016-08-16
Bsongis/gui refactoring for x7 d

bsongis:


bsongis:
Sorry for so many diffs, but needs tests!

PR #3712, opened 2016-08-22
3djc/sound files rename

3djc:
With this change:
- situation is unchanged for 9X (but they will need to  review sound numbering for 2.2 units)
- for ARM radio with SD, units are no longer numbered files, but have the name of the english measure instead, like (they always have this mane, regardless of languages) :
meter.wav for singular meter
meter1.wav for plural meters (in most languages)
meter2.wav as needed for languages that have several plurals
meter3.wav as needed for languages that have several plurals


radio/src/audio_arm.cpp
```diff
@@ -1235,6 +1266,21 @@ void audioEvent(unsigned int index)
   }
 }
 
+#if defined(SDCARD)
+void pushUnit(uint8_t unit, uint8_t id=0)
+{
+  if (unit < 104)
```
bsongis: DIM(unitsFilenames)

radio/src/audio_arm.cpp
```diff
@@ -135,6 +135,37 @@ const int16_t sineValues[] =
 };
 
 #if defined(SDCARD)
+
+const char * unitsFilenames[] = {
```
bsongis: reduce this array to only the unit string, i.e. "volt", "amp", then add 0 / 1 / 2 / 3 in the function, using `getStringWithIndex(...)`

radio/src/audio_arm.cpp
bsongis: no newline here

radio/src/audio_arm.h
```diff
@@ -359,11 +359,13 @@ enum AutomaticPromptsEvents {
 };
 
 void pushPrompt(uint16_t prompt, uint8_t id=0);
+void pushUnit(uint8_t, uint8_t);
```
bsongis: it's easier if you name the args

PR #3716, opened 2016-08-23
Change ARM targets sound file handling for UNITS table

3djc:
Change audio generation files accordingly
New way of working :
- units files are ALWAYS named in english based on the names in audio_arm.cpp names.
- multiple form (singular, plural(s) are ammended with 0 to 4 suffix, according to languages rules.
example : meter1.wav for first form of plural of meter unit,. This filename needs to be the same for all languages (in their respective directories of course)


radio/src/audio_arm.cpp
```diff
@@ -1235,6 +1266,23 @@ void audioEvent(unsigned int index)
   }
 }
 
+#if defined(SDCARD)
+void pushUnit(uint8_t unit, uint8_t idx, uint8_t id=0)
+{
+  if (unit < DIM(unitsFilenames) && idx < 5)
```
bsongis: no newline here

bsongis: why not an assert?

radio/src/audio_arm.cpp
```diff
@@ -1235,6 +1266,23 @@ void audioEvent(unsigned int index)
...
+  {
+    char path[AUDIO_FILENAME_MAXLEN+1];
+    getSystemAudioPath(path);
+    getStringWithIndex(path, unitsFilenames[unit], (int)idx);
```
bsongis: the path is lost here, no?

bsongis: no need to cast idx

radio/src/audio_arm.cpp
```diff
@@ -1235,6 +1266,23 @@ void audioEvent(unsigned int index)
...
+    strcat(path, SOUNDS_EXT);
```
bsongis: use char * tmp = getStringWithIndex(...)
so that you can call strcpy instead of strcat

radio/src/audio_arm.cpp
```diff
@@ -1235,6 +1266,23 @@ void audioEvent(unsigned int index)
   }
 }
 
+#if defined(SDCARD)
+void pushUnit(uint8_t unit, uint8_t idx, uint8_t id=0)
```
bsongis: no default param to id here, instead you may add it in the header

radio/src/translations/tts_fr.cpp
```diff
@@ -54,6 +54,22 @@ enum FrenchPrompts {
...
+#if defined(CPUARM)
+  #define FR_PUSH_UNIT_PROMPT(u) fr_pushUnitPrompt((u), id)
+#else
+  #define FR_PUSH_UNIT_PROMPT(u) pushUnitPrompt((u))
+#endif
+
+I18N_PLAY_FUNCTION(fr, pushUnitPrompt, uint8_t unitprompt)
+{
+#if defined(CPUARM)
+    PUSH_UNIT_PROMPT(unitprompt, 0);
```
bsongis: wrong indentation

radio/src/audio_arm.cpp
bsongis: with my recent changes it could be:
strcpy(strAppendStringWithIndex(strAppendSystemAudioPath(path), unitsFilenames[unit], idx), SOUNDS_EXT);

3djc:
Another rebase failure, thanks smartgit...

PR #3717, opened 2016-08-24
Change audio generation files accordingly

3djc:
New way of working :

units files are ALWAYS named in english based on the names in audio_arm.cpp names.
multiple form (singular, plural(s) are ammended with 0 to 4 suffix, according to languages rules. example : meter1.wav for first form of plural of meter unit,. This filename needs to be the same for all languages (in their respective directories of course)


radio/util/tts_fr.py
```diff
@@ -15,34 +15,36 @@
...
+             (u"g", "g0", NO_ALTERNATE),
+             (u"degrés", "degree0", NO_ALTERNATE),
+             (u"radians", "rad0", NO_ALTERNATE),
+             (u"millilitres", "ml0", NO_ALTERNATE),
+             (u"onces", "founce0", NO_ALTERNATE),
+             (u"heure", "hour0", NO_ALTERNATE),
+             (u"minute", "minute0", NO_ALTERNATE),
+             (u"seconde", "second0", NO_ALTERNATE),
+             ]:
+    systemSounds.append((s, filename(f, a)))
```
bsongis: I don't see how the xxx.ad4 for gruvin9x are generated now

radio/util/tts_en.py
```diff
@@ -14,34 +14,34 @@
...
-                       "r p m", "r p m",
-                       "g", "g",
-                       "degree", "degrees",
-                       "radian", "radians",
-                       "milliliter", "milliliters",
-                       "fluid ounce", "fluid ounces",
-                       "hour", "hours",
-                       "minute", "minutes",
-                       "second", "seconds",
+for i, (s, f, a) in enumerate([("volt", "volt0", NO_ALTERNATE), ("volts", "volt1", NO_ALTERNATE),
```
bsongis: All those NO_ALTERNATE are not needed!

bsongis: One line dedicated to volt (as you did for amp etc.)

bsongis:
Good job, thanks a lot!

PR #3728, opened 2016-08-29
3djc/horus fix timehour

3djc:


radio/src/strhelpers.cpp
```diff
@@ -159,7 +159,7 @@ char * strAppendStringWithIndex(char * dest, const char * s, int idx)
   return strAppendUnsigned(strAppend(dest, s), abs(idx));
 }
 
-char * getTimerString(char * dest, putstime_t tme, uint8_t hours)
+char * getTimerString(char * dest, putstime_t tme, uint16_t hours)
```
bsongis: why? this should just be a boolean 0/1!

PR #3729, opened 2016-08-29
This fixes #3727

### LINKED ISSUE #3727: Global variables does not show the current Flight mode that is active
in 2.1.X when you have a flight mode active that flight mode is shown in bold at he top of the menu - this allows the operator to quickly adjust the variables pertinent to the active flight mode... even if the mode isnt on.. right now you have to guess or count where the flight modes are in that menu and it makes adjusting things a little harder.

3djc:


radio/src/gui/480x272/model_gvars.cpp
```diff
@@ -71,7 +72,8 @@ bool menuModelGVars(event_t event)
           if (v > GVAR_MAX) {
             uint8_t p = v - GVAR_MAX - 1;
             if (p >= j-1) p++;
-            drawFlightMode(x-21, y, p+1, attr);
+            if ((j-1) == curfm) drawFlightMode(x-21, y, p+1, attr|BOLD);
```
bsongis: I would do this in one line, no?

PR #3734, opened 2016-08-29
3djc/horus pcbrev

3djc:


bsongis:
Hi reused your code in my latest commits into 'next', thanks for your help on this

PR #3736, opened 2016-08-29
Horus internal GPS: settable baudrate, misc fixes, new cmake options,…

projectkk2glider:
… cosmetics, cli

This now works on my Horus. My GPS module is unconfigured and sends a lot of data, but more importantly it uses $G**N**GGA instead of normal $G**P**GGA messages. The code here supports both messages and displays correct position.

Default cmake variables are good for production radio, for my I have to use:
```
cmake -DINTERNAL_GPS_BAUDRATE=38400 -DINTERNAL_GPS_INSTALLED=YES
```


radio/src/gps.cpp
```diff
@@ -207,12 +228,14 @@ bool gpsNewFrameNMEA(char c)
           switch (gps_frame) {
             case FRAME_GGA:
               frameOK = 1;
+              gpsData.fix = gps_Msg.fix;
+              gpsData.numSat = gps_Msg.numSat;
               if (gps_Msg.fix) {
-                gpsData.fix = 1;
+                __disable_irq();    // do the atomic update of lat/lon
```
bsongis: Why is it needed for GPS position?

radio/src/targets/horus/CMakeLists.txt
```diff
@@ -49,6 +52,9 @@ if(DISK_CACHE)
   set(SRC ${SRC} disk_cache.cpp)
   add_definitions(-DDISK_CACHE)
 endif()
+if(INTERNAL_GPS_INSTALLED)
+  add_definitions(-DINTERNAL_GPS_INSTALLED)
```
bsongis: Perhaps:
if (PCBREV > 10)
  option(INTERNAL_GPS ... YES)
else()
  option(INTERNAL_GPS ... YES)
endif()

if (INTERNAL_GPS)
  add_definitions(-DINTERNAL_GPS)
endif()

radio/src/targets/horus/board.h
```diff
@@ -389,12 +389,12 @@ void hapticOff(void);
 void hapticOn(uint32_t pwmPercent);
 
 // GPS driver
-#if PCBREV >= 13
+#if PCBREV >= 13 || defined(INTERNAL_GPS_INSTALLED)
 #define INTERNAL_GPS                   1
```
bsongis: This would not be needed (because done in the CMakeLists.txt)

radio/src/gps.cpp
projectkk2glider: Currently it is not necessary, but if in future we move the gps code  in another thread it will be needed.

bsongis: Agree. Perhaps we could add a comment, but avoid to disable the interrupts?

projectkk2glider: I think we will need another thread (at least on Horus) soon:
 * the new thread will have priority lower than mixers thread, but higher than main thread
 * new thread would process all periodic things like: internal GPS, gyro/accc sensor, Bluetooth, etc. 

Basically the stuff that is now handled in the main thread, but should really be handled more often and more independently of the Lua/gui cpu usage and relatively high loop times of the main thread.

kilrah: Good idea

projectkk2glider:
My GPS module is quite chatty, these are messages, which are sent frequently:
```
$GNRMC,190415.80,,eV,,,,,,,300816,,,N*6F
$GNVTG,,,,,,,,,N*2E
$GNGGA,190415.80,,,,,0,00,99.99,,,,,,*78
$GNGSA,A,1,,,,,,,,,,,,,99.99,99.99,99.99*2E
$GNGSA,A,1,,,,,,,,,,,,,99.99,99.99,99.99*2E
$GPGSV,3,1,09,01,118,157,,06,40,306,,07,07,184,,09,57,238,*71
$GPGSV,3,2,09,17,19,245,,19,25,267,,22,37,100,,23,90,159,*70
$GPGSV,3,3,09,31,16,039,*4F
$GLGSV,3,1,09,65,66,352,,66,25,282,,72,34,066,,73,11,019,*688
```

I turned them off with following commands (they are missing checksum, which is generated by the code):
```
$PUBX,40,GSV,0,0,0,0
$PUBX,40,GSA,0,0,0,0
$PUBX,40,VTG,0,0,0,0
$PUBX,40,GLL,0,0,0,0
```
This then leaves only GGA and RMC messages which we actually parse. 

It would be interesting to get the trace from the production GPS, perhaps similar optimization could be done. I will now push my latest changes that enable one to send this command via CLI.

projectkk2glider:
CLI commands help:

* enable/disable GPS trace 
```
>gps trace
```
* send command to GPS (this actually sends `$PUBX,40,GLL,0,0,0,0*5C\r\n` to the module):
```
>gps $PUBX,40,GLL,0,0,0,0
```
  

* change GPS baudrate:
```
> gps 38400
```

projectkk2glider:
Still left to do:
* parse date and time
* add gps data (like lat/lon, time/date, number of satellites, etc...) to some GUI screen (or create a new screen if needed), most likely in Statistics, Debug and Analogs group of screens.

radio/src/gps.cpp
bsongis: Yes I do agree as well

bsongis:
Seems all good, thanks!

PR #3740, opened 2016-08-31
Haptic feedback in lua scripts

floaledm:
This is my attempt to add haptic feedback support to lua scripts. I used haptic.play() even though OpenTX usually uses haptic.event() to generate haptic feedback. Tested working using the script below. I played with the parameters of that script but it seems that only "pause" has any effect, which is how long the haptic feedback is on. That value does not seem to correspond to milliseconds or even centiseconds... Thanks!

local inputs = {{"length", VALUE, 0, 100, 0}, {"pause", VALUE, 0, 100, 0}, {"flags", VALUE, 0, 1000, 16}, {"RepeatT", VALUE,0,100,10}}

local last_reset = getTime()

local function run(length,pause,flags,RepeatT)
local timer = getTime()
	if (timer - last_reset) > RepeatT*10 then -- times 10 to convert to tenths of seconds
		last_reset = timer
        playHaptic(length,pause,flags)
    end
end

return {run=run, input=inputs}

### DIFF
--- a/radio/src/lua/api_general.cpp
+++ b/radio/src/lua/api_general.cpp
@@ -694,6 +694,32 @@ static int luaPlayTone(lua_State * L)
 }
 
 /*luadoc
+@function luaPlayHaptic(duration, pause [, flags])
+
+Generate haptic feedback
+
+@param duration (number) length of the haptic feedback in milliseconds
+
+@param pause (number) length of the silence after haptic feedback in milliseconds
+
+@param flags (number):
+ * `0 or not present` play with normal priority
+ * `PLAY_NOW` play immediately
+
+@status current Introduced in 2.2.0
+*/
+static int luaPlayHaptic(lua_State * L)
+{
+#if defined(HAPTIC)
+  int length = luaL_checkinteger(L, 1);
+  int pause = luaL_checkinteger(L, 2);
+  int flags = luaL_optinteger(L, 3, 0);
+  haptic.play(length, pause, flags);
+#endif
+  return 0;
+}
+
+/*luadoc
 @function killEvents(key)
 
 Stops key state machine.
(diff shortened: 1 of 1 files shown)


radio/src/lua/api_general.cpp
```diff
@@ -694,6 +694,30 @@ static int luaPlayTone(lua_State * L)
 }
 
 /*luadoc
```
bsongis: This block (and the function declaration) should be inside #if defined(HAPTIC), because some boards (i.e. X9D(not plus) has no HAPTIC

bsongis:
Good idea, just the X9D thing

floaledm:
Thanks for the feedback. I made the necessary changes.

For those functions with #if defined, if a script executes a function but it's not available, wouldn't that cause an error in the script and stop its execution? If that's the case, what do you suggest to prevent lua script errors? Maybe at the very least the lua doc should warn users that the lua function is compatible only with certain boards, and that they could check  radio version using getVersion()? If there's an issue there, let me know how I can help with this...

bsongis:
Excellent question which makes me think I suggested something wrong, only the contents of ```luaPlayHaptic(...)``` should be #ifdef-ed so that no error is raised in case of an X9D-nonplus

bsongis:
May I ask you to do the change? Sorry again

floaledm:
It's all good. I'll take care of it right away...
So, maybe also keep in mind my comment about #if define for the other lua functions. Seems most problematic for the ones which are expected to return something (e.g., GREY())...

bsongis:
GREY()? Would you explain what is the problem? Compatibility Horus / Taranis? Or something else?

bsongis:
Thanks for your PR!

floaledm:
If COLORLCD is defined, there is no GRAY() function, so if a script calls it, there will be an error... is that a problem?

bsongis:
I guess we will have completely different scripts on HORUS compared to TARANIS!

floaledm:
Good answer... =)

floaledm:
Thanks for moving this forward! Til next time...

PR #3741, opened 2016-08-31
3djc/companion disable horus

3djc:


companion/src/firmwareinterface.cpp
```diff
@@ -342,7 +342,7 @@ bool FirmwareInterface::isValid()
 
 unsigned int FirmwareInterface::save(QString fileName)
 {
-  uint8_t binflash[MAX_FSIZE];
+  uint8_t binflash[FSIZE_MAX];
```
bsongis: 2MBytes in stack. I guess VC++ will segfault, better to use a malloc here

bsongis:
Seems good, thanks for your help on this job which is not really funny ...

companion/src/firmwareinterface.cpp
bsongis: 500k was not very good before ...

PR #3756, opened 2016-09-03
Projectkk2glider/gps clock to rtc improvements

projectkk2glider:
Both internal and receiver GPS can now adjust RTC clock. I think we had a couple of bugs with existing GPS date/time code, most notable: time zone was not handled properly (hours roll-over, day-rollover, etc..) The easiest way to solve this was to keep the GPS clocks in UTC, so this will change display and Lua values for GPS. Still thinking what to do about that.

I also messed with function definitions (adding const) and that brought up a lot of changes for other radios. But it should be OK.

I have successfully tested:
 * RTC clock was set by internal GPS on my Horus
 * RTC clock was set by external GPS on Horus simulator

### DIFF
--- a/radio/src/gps.cpp
+++ b/radio/src/gps.cpp
@@ -134,6 +134,8 @@ typedef struct gpsDataNmea_s
   uint16_t altitude;
   uint16_t speed;
   uint16_t groundCourse;
+  uint32_t date;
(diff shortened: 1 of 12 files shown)


bsongis:
All good, thanks!

PR #3765, opened 2016-09-05
Add timer3 as default on Horus

3djc:
Not sure this  is the right place to add it


bsongis:
But no possibility to remove this extra timer! And I am afraid that it is possible to have two -DTIMERS=X on the same command line!

bsongis:
I would change the option default value in arm/CMakeLists.txt if we really want to have 3 timers by default

companion/src/firmwares/opentx/opentxinterface.cpp
```diff
@@ -597,7 +597,7 @@ int OpenTxFirmware::getCapability(const Capability capability)
     case OffsetWeight:
       return (IS_ARM(board) ? 500 : 245);
     case Timers:
-      return ((IS_ARM(board) && id.contains("timer3")) ? 3 : 2);
+      return (IS_ARM(board) ? 3 : 0);
```
bsongis: : 2

radio/src/targets/common/arm/CMakeLists.txt
```diff
@@ -12,11 +12,6 @@ option(DEBUG_INTERRUPTS "Count interrupts" OFF)
 option(DEBUG_TASKS "Task switching statistics" OFF)
 option(DEBUG_TIMERS "Time critical parts of the code" OFF)
 
-if(TIMERS EQUAL 3)
```
bsongis: I prefer you leave the option as it is, just set 3 as default
I personnally use only 2

PR #3767, opened 2016-09-05
[Horus] Give pxx its own substruct instead of using ppm (cosmetics)

schwabe:


bsongis:
Thanks a lot, this is really better

PR #3776, opened 2016-09-08
Schwabe/cpack osx

schwabe:
Overhaul the packing of OpenTX on OS X. It is not perfect but it works on my development machine and is working at all (the 2.1 stuff breaks in horrible ways on my machine).

Some of the changes could also benefit other platforms but I did test/port them.


bsongis:
@kilrah I let you test and merge, ok?

kilrah:
Yep, can't build here but the result looks fine to me! Maybe just rename the app to "OpenTX Companion 2.2" instead of just companion to avoid replacing older versions

schwabe:
@kilrah what errors are you getting? And what QT version are you using to build.

bsongis:
@kilrah would you merge this one?

kilrah:
Was just waiting for the rename. 
@schwabe I don't have a build setup at all on my mac at this point

schwabe:
@kilrah I have to figure how to the rename. Just renaming the target does not work since " " and . are not allowed. Have to figure how to rename in cmake right.

kilrah:
Hmm well if it's complicated at least have it called "companion22" like on linux... it's just so that users don't overwrite any existing 2.1 version they might have.

schwabe:
I looked into compiling for older OS X versions. This can be theorectically done by setting ``-D CMAKE_OSX_DEPLOYMENT_TARGET="10.8"``. However, this also adds the os x version flags to the firmware compilation. 

Inquering in the #cmake irc gave me reason: Cmake does not support diferent toolchains in the same source tree build and therefore that setup is broken. The /right/ way is to have two different cmake project and add the second one via "ExternalProject_Add".

For compiling for older OS X version we need to compile only OS X targets in a seperate step.

bsongis:
I trust you :)

PR #3778, opened 2016-09-08
Fix timing of Multi and DSM2 protocol

schwabe:
We send an extra stopbit and the logic for driving the transistor was swapped. Thanks Mike for noticing (http://www.rcgroups.com/forums/showpost.php?p=35671709&postcount=5151)


bsongis:
Thanks!

PR #3800, opened 2016-09-14
3djc/faster local commit tests

3djc:
Allow passing desired core usage for commit-tests (speed up local execution)
Usage:
commit-tests.sh -j=8
or
commit-tests.sh --jobs=8

Passing no arguments defaults to two cores used

On an i7@3.7Ghz : 2 cores : 13m14secs
with 8 cores : 8m11secs

### DIFF
--- a/tools/commit-tests.sh
+++ b/tools/commit-tests.sh
@@ -5,12 +5,16 @@ CORES=2
 for i in "$@"
 do
 case $i in
-    -c=*|--cores=*)
+    -j=*|--jobs=*)
       CORES="${i#*=}"
       shift
       ;;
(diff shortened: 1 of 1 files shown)


schwabe:
Looks good. Two nitpickings

I would prefer -j8 instead of -c8.
${foo} is nicer in scripts than $foo

3djc:
If you can find me a long (--) version starting with 'j' and making sense, you got it

schwabe:
--jobs? :)

3djc:
Yeah, never liked it in make hence my resistance, but you know what they say, resistance is futile !

PR #3810, opened 2016-09-15
[Horus] S.PORT to Lua passthrough

bsongis:


radio/src/targets/horus/board.cpp
```diff
@@ -89,11 +87,19 @@ void interrupt1ms()
   DEBUG_TIMER_START(debugTimerRotEnc);
   checkRotaryEncoder();
   DEBUG_TIMER_STOP(debugTimerRotEnc);
+
+#if defined(LUA)
+  if (telemetryProtocol == PROTOCOL_FRSKY_SPORT) {
+    if (telemetryFifo.last(-2) == 0x7E && telemetryFifo.last(-1) == outputTelemetryBufferTrigger) {
+      sportSendBuffer(outputTelemetryBuffer, outputTelemetryBufferSize);
```
bsongis: outputTelemetryBufferTrigger should be reset here, or the sportSendBuffer will be called every millisecond!

3djc: I have tested bith with 0x00 and 0x7E. Both work, but a new issue comes with this : if you loose telem (beeing too close for exemple), you do not get it back anymore

bsongis: Are you sure it's linked to this change? Hmm. perhaps because S.PORT doesn't return to RX mode (it stays in TX mode). I have to check this.

kilrah: Just to precise, it does NOT seem to be related to "normal" telemetry losses, I can have as many of those as I want without issues, the "crash" where you get "telemetry lost" and it won't recover without a reboot of the radio can happen at any distance (reproduced it in range check mode 10m away too).

bsongis: Are we sure we still receive telemetry? I remember having this bug, but the RX was completely silent because of a bug inside the S6R

3djc: Well,  keeping the power on S6R and just rebooting the radio does givem telemetry back

bsongis: Then ...

kilrah: I need to retry, but I think that when running a branch with these changes the same would happen also with an other receiver (X6R).

kilrah: Confirmed, not related to the S6R nor even a receiver in any way, the telemetry just hangs after a few minutes. If I power the radio on with no receiver powered, wait 5 minutes, then power up the receiver I have no telemetry.

bsongis: Do you use this branch?

DeepnineRC: I'm using OpenTX 2.2.0 RC5 (N357).

3djc:
I have tested both on the bench and inflight, with a production horus and S6R, seems to work flawlessly after the last change. It will not loose telem over time, and will recover it quickly if you loose it for too close proximity. The communcation does not get lost on a 30 min bench test with X6R as well.

DeepnineRC:
I would like to test it too, but i can't get the compiling stuff working.
I'm running opentx-horus-noheli-lua-massstorage-de

3djc:
I need to do more work on the scripts as well, so it is not ready yet for wide usage

radio/src/targets/horus/board.cpp
```diff
@@ -90,7 +90,7 @@ void interrupt1ms()
 
 #if defined(LUA)
   if (telemetryProtocol == PROTOCOL_FRSKY_SPORT && outputTelemetryBufferTrigger) {
-    if (telemetryFifo.last(-2) == 0x7E && telemetryFifo.last(-1) == outputTelemetryBufferTrigger) {
+    if (telemetryFifo.last(-2) == 0x7E && telemetryFifo.last(-1) == outputTelemetryBufferTrigger && outputTelemetryBufferSize > 0) {
```
bsongis: Right, but ```outputTelemetryBufferSize > 0``` could be tested before the fifo

bsongis:
👍

PR #3813, opened 2016-09-16
Update tts_de.py

rcbebo82:
Some more translations fixed.


radio/util/tts_de.py
```diff
@@ -11,31 +11,31 @@
...
 for i, (s, f) in enumerate([(u"Volt","volt0"),
                             (u"Ampere", "amp0"),
-                            (u"MilliAmpere", "mamp0"),
+                            (u"Milliampere", "mamp0"),
                             (u"Knoten", "knot0"),
                             (u"Meter pro sekunde", "mps0"),
                             (u"Fuss pro sekunde", "fps0"),
                             (u"Kilometer pro stunde", "kph0"),
-                            (u"Meilen pro Stunde", "mph0"),
+                            (u"Meile pro Stunde", "mph0"), (u"Meilen pro Stunde", "mph1"),
```
3djc: Adding plural values will not make them available. Currently, the code dealing with German only deals with plural for hours. If you need plural, then this need to be changes and all plural for units needs to be provided.

Also, you need to try those with PicoTTS under linux to listen to what is said, the way it sounds not always match proper spelling

bsongis:
I let others review this one, I am a noob in Deutsch

3djc:
The issue is not so much one of language than of deutch speech generation. If we want plural for some units, then it needs to be for all units

rcbebo82:
Well for Meile and Meilen we dont need this    necessarily in german. I think when anyone chooses german is would be very special if miles is choosen as unit.

PR #3816, opened 2016-09-16
Allow sdcard version change from cmake

3djc:
To increment sdcard.opentx.version, just increment SDCARD_REVISION in CMakeLists.txt (line is set(SDCARD_REVISION "0002"))


radio/src/targets/horus/CMakeLists.txt
```diff
@@ -26,6 +26,7 @@ set(RAMBACKUP YES)
 add_definitions(-DPCBHORUS -DPCBREV=${PCBREV} -DSTM32F429_439xx -DSDRAM -DCOLORLCD -DPPM_PIN_UART -DPPM_PIN_TIMER)
 add_definitions(-DEEPROM_VARIANT=0 -DAUDIO -DVOICE -DRTCLOCK)
 add_definitions(-DGPS_USART_BAUDRATE=${INTERNAL_GPS_BAUDRATE})
+add_definitions(-DREQUIRED_SDCARD_VERSION="${SDCARD_VERSION}")
```
bsongis: could be done in common/arm, no?

3djc: Indeed :)

bsongis:
Thanks!

PR #3817, opened 2016-09-16
Allow selecting a model image from the SD manager (Fix #3814)

### LINKED ISSUE #3814: Enhancement: Show small preview image on model image selector
It would be nice if there could be a small preview image of the selected model image in the model image selector.

kilrah:
To be improved with a check that we are in the IMAGES folder (also on Taranis) for cleanliness

### DIFF
--- a/radio/src/gui/480x272/radio_sdmanager.cpp
+++ b/radio/src/gui/480x272/radio_sdmanager.cpp
@@ -128,6 +128,10 @@ void onSdManagerMenu(const char * result)
     audioQueue.stopAll();
     audioQueue.playFile(lfn, 0, ID_PLAY_FROM_SD_MANAGER);
   }
+  else if (result == STR_ASSIGN_BITMAP) {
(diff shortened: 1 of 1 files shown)


radio/src/gui/480x272/radio_sdmanager.cpp
```diff
@@ -220,6 +224,11 @@ bool menuRadioSdManager(event_t _event)
           if (!strcasecmp(ext, SOUNDS_EXT)) {
             POPUP_MENU_ADD_ITEM(STR_PLAY_FILE);
           }
+          else if (!strcasecmp(ext, BMP_EXT) || !strcasecmp(ext, JPG_EXT) || !strcasecmp(ext, PNG_EXT)) {
```
bsongis: I remember this test done somewhere else, perhaps we have written a macro, no?

kilrah: It's done inside sdListFiles() function, but not quite sure how it works to take it out and make it available elsewhere

PR #3818, opened 2016-09-17
Audio fixes

projectkk2glider:
* Switches audio filenames were wrong [Horus]
* sdAvailableLogicalSwitchAudioFiles was too small (we now have 64 LSW), fixed with BitFiled class


bsongis:
Good job, thanks!

PR #3829, opened 2016-09-19
HORUS save persistant items on model change

3djc:


radio/src/gui/480x272/model_select.cpp
```diff
@@ -97,6 +97,7 @@ void onModelSelectMenu(const char * result)
 {
   if (result == STR_SELECT_MODEL) {
     // we store the latest changes if any
+    storageFlushCurrentModel();
```
bsongis: Ah I just forgot to call the function! Thanks

PR #3835, opened 2016-09-20
3djc/horus s6 r support

EricFPV:
Horus Backlight brightness adjustment using pot, 6p or p2, p1 ?
can it be done ? I am new to this forum.


kilrah:
> I am new to this forum

This is a code development platform, not a place for discussions/support/questions and you're using it very wrong by sending out a pull request that modifies code in a way that would basically break openTX.

Use the "issues" tab for reporting confirmed bugs and making well presented suggestions, for prior discussions use the chat or a forum like openrcforums.com.

PR #3838, opened 2016-09-21
Fixes 3474 Mix line drag n drop before CH1

elecpower:
Fixes Shift Drag n Drop (Move)
Only tested on Ubuntu 14.04
Will do another PR for Inputs which has same bug.


3djc:
Tested ok on Debian 8.5, waiting for a windows confirmation before merging

3djc:
Thanks a lot elecpower

PR #3844, opened 2016-09-22
Fixes Inputs Drag n Drop

elecpower:
Stops dropping inputs before Input1
Shift drag n drop (move) works
Has dependencies to Mixes drag n drop fix
Tested Ubuntu 14.04


3djc:
Thanks a lot !

PR #3848, opened 2016-09-23
Update tts_de.py

rcbebo82:
Translated crow to "Butterfly", changed filenames accordingly to the english ones so its possible to change from english to german without the need of editing the Special Functions section per model.
Removed "Minimale Trimmung", in my opinion this makes no sense a trim only can be at maximum or at center.


radio/util/tts_de.py
```diff
@@ -50,8 +50,8 @@
                 (u"Fehler", "error", NO_ALTERNATE),
                 (u"Trim zentriert", "midtrim", 495),
                 (u"Poti zentriert", "midpot", 496),
-                (u"Maximale Trimmung erreicht", "maxtrim", NO_ALTERNATE),
-                (u"Minimale Trimmung erreicht", "mintrim", NO_ALTERNATE),
```
kilrah: There need to be 2 files, mintrim and maxtrim. You can have them both say the same thing.

3djc: Also you cannot change the name of a bunch of sounds like that, many people have referenced them in special fuincitons that will suddenyl stop working

kilrah: It's a system sound, no special functions.

rcbebo82: So it would be no problem if for those two files the same is spoken. Should i correct it?

schwabe:
I think the information which limit of trimming you reached is still useful. If you don't like Mimimum and Maximum, we should go with:

"Untere Trimmungsgrenze erreicht" and
"Obere Trimmungsgrenze erreicht" 

oder something similar.

rcbebo82:
Got it now i understand. Mintrim is on the Zero side of each Input so it makes Sense. But what is with the filemames?

schwabe:
Just keep the filenames mintrim and maxtrim.

3djc:
Thanks a lot

Issue #1098, opened 2014-05-14
Add RF Module blocking mode switch to model instructions / checklist


TimGFoley:
We had discussed this in use forums so I thought I'd add it to a later version of firmware as a suggestion.

Could we look at adding a blocking mode switch/setting of some type that when the model notes/checklist is displayed, the selects RF module  is not powered up until the checklist/notes are read/cleared.

This is to remind people to connect external antennas for modules or internal if they have done the antenna mod.   Thereby preventing damage to the RF module should they forget to connect an antenna.

I'm sure there are other instances such as a reminder to plug in a specific module if the user has multiple, etc.

Thanks

projectkk2glider:
Just a reminder that in case of a Watchdog reset this should be skipped

kabturek:
i've added a simple warning on PR #3861 that reminds about plugging an antenna. not perfect but might be a good start.

craigywsm:
Great idea,
Also a great idea if you want to test failsafe without switching radio off. An option to turn RF off and on (With connected warning) would be a great feature

3djc:
The OP feature has been implemented already, the module stays off while models notes are read

PR #3861, opened 2016-09-29
Add RF antenna warning to startup warnings

kabturek:
Hey guys, 
i've really liked the feature on #1098 and decided to do some hacking while on a train - it's only tested in simulator as i don't have my taranis with me right now.
Would love some input and ideas if there are better ways to do it.


radio/src/myeeprom.h
```diff
@@ -2152,6 +2152,7 @@ PACK(typedef struct {
   ARM_FIELD(uint8_t   ignoreSensorIds:1)
   int8_t    trimInc:3;            // Trim Increments
   uint8_t   disableThrottleWarning:1;
+  uint8_t   enableRfAntennaWarning:1;
```
schwabe: Adding an int here will break compatibility with existing eeprom contents.

kilrah:
There is unfortunately a truckload of implications/dependencies that are ignored/broken in here. Please specify if you're interested in learning about them and pursuing this and would like detailed explanations.

kabturek:
thx @kilrah i'm interested in fixing this if it has any chance in being merged in future. 
I'm aware it's just simple hacking but wanted to do anything that works (small successes).

I'll rebase it on next - couldn't fetch next while i was on a train.

kabturek:
rebased this on next but have trouble building the 2.2 source on MAC OS X. will keep trying and update the wiki when i succeed.

kabturek:
It looks like it would be better to use ModuleData Struct (as this isnt model dependent) with its external_antenna and warn if external_antenna is true. 
And then add radio settings to enable/disable warning(RadioData struct or EXTRA_GENERAL_FIELDS) or use one of the spare fields  in Module struct for warning.

schwabe:
@kabturek that external antenna define is only valid if the module is XJT (see the union). The moduledata hasn't that much space left but the following should work:

```
   uint8_t channelsStart;
   int8_t  channelsCount; // 0=8 channels
-  uint8_t failsafeMode:4;  //only 3 bits used
+  uint8_t failsafeMode:3;
+  uint8_t warnAntennaEnabled:1;
   uint8_t subType:3;
   uint8_t invertedSerial:1
```

bsongis:
Please reopen when fixed, thanks!

lypse:
Any chance on fixing this? It would be very helpful. Thanks!

kabturek:
@lypse i didn't have time to look into this again (this PR was broken/naive/proof of concept) I have it in my todo list but i cant really promise any timing - sorry.

Issue #3803, opened 2016-09-14
Minute timer on Horus 


rcbebo82:
If i use the minute timer set to five minutes the system plays:
"four and zero"
"three and zero"
"two an zero"
"one an zero" 

It should play instead:
"four minutes"
"three minutes"
"two minutes"
"one minute"

projectkk2glider:
You most likely have a wrong sound pack. Opentx 2.0, 2.1 and 2.2 have different sound packs.

rcbebo82:
Well if have now used the original SD-CARD Horus nightly zip file and the error is also there!

3djc:
Appends in German it seems

rcbebo82:
I copied the english Sound files to the "de" Folder. Also used the csv inside the system folder and edited all paths and filenames to fit it. The error occurs also it seems there is something more wired up.

I now used the en folder and createt german sound files in it. Now all works as expected. 
Dont know how to get through the error. Anyone any idea?

kilrah:
> I now used the en folder and createt german sound files in it. Now all works as expected. 
Dont know how to get through the error. 

Sounds like you jsut did?

Pleas explain in detail what you tried, which language files, which language selected etc.

rcbebo82:
Yes it now works.
Voice Language: English
I took the csv file in the folder SOUNDS\EN\SYSTEM\en-US-taranis.csv and translated all english words to german. I also did some other changes to it to use with TTSAutomate 1.6.3.0. File attached here [](https://dl.dropboxusercontent.com/u/13465738/en-US-taranis%20%28aktuell%29.psv)
After i created all sound files i put the folder on the horus and overwrited all wav files.
The timer now works it Plays "vier Minuten", "drei Minuten", "zwei Minuten", "eins Minute".
So here we have the problem with the grammar it should play "eine Minute"

rcbebo82:
Mayby it has something to do with the en firmware i use on the horus?

kilrah:
That really doesn't help us to know what's needed to fix German...

rcbebo82:
I know .. i will do some further investigations with de sound files.

rcbebo82:
With the original de sound files from the sd card zip file and voice Deutsch selected the Horus plays on the minute timer:
4 und 0
3 und 0
2 und 0
1 minuten und 0
30
20
10
9
8
7
6
5
4
3
2
1

rcbebo82:
Seems the error is in tts_de.cpp

[...]

PR #3863, opened 2016-09-29
Update tts_de.cpp

rcbebo82:
Issue #3803 
German spoken numbers fixed:
eins hundert > ein hundert
eins tausend > ein tausend


rcbebo82:
Something is wrong. I will compile a new firmware and test it before i commit it.

PR #3864, opened 2016-09-29
Update tts_de.py

rcbebo82:
Sound files to match Pull Request #3863


kilrah:
You should submit all changes that go with a given topic/feature/enhancement in the same pull request. Leave it for this time, but you'll know for next.

rcbebo82:
Something is wrong. I will compile it my self and test it.

PR #3877, opened 2016-10-05
Projectkk2glider/issue 3815 audio refactoring

### LINKED ISSUE #3815: Audio stuttering on Horus
In the latest nightly 2.2.0N357. N356 was OK, the most likely culprit is 8a31c1d5551163d00b85d9576137619d73bb71dc 

That commit should not be reverted, because it fixes an important problem where interrupt were disabled for too long time.

The audio needs fixing, my proposal:
- add debug timer to the audio tak to see how often it is called and how long does one loop take
- implement DMA transfer on Horus
- other fixes that might be identified by the first step
- perhaps even have more audio buffers on Horus (we have more RAM)

projectkk2glider:
This was tested on Horus and X9D. The change also affects Sky9X but I dont' have it.

Closes #3815 

The goals:
 * make Audio code more clean and readable, more object isolation
 * use new FatFS (fixes several bugs, easier handling of long filenames) 
 * new more robust and simple audio buffer FIFO
 * try too keep the audio buffers filled (ideally keep it full) this should reduce audio stuttering
 * simplify the audio thread - audio driver interface


projectkk2glider:
I have take in account your comments. Please do not merge as it is. When you are happy I will consolidate all audio changes into one commit in order to keep the git history clean. I also want to have the FatFs commit separate so do not *squash and merge* - it will be a mess.

radio/src/audio_arm.h
```diff
@@ -254,14 +254,15 @@ class MixedContext {
       tone.clear();   // the biggest member of the uninon
     }
 
+    bool isEmpty() const { return fragment.type == FRAGMENT_EMPTY; };
     bool isTone() const { return fragment.type == FRAGMENT_TONE; };
     bool isFile() const { return fragment.type == FRAGMENT_FILE; };
     bool hasId(uint8_t id) const { return fragment.id == id; };
 
-    int mixBuffer(AudioBuffer *buffer, int volume, unsigned int fade)
+    int mixBuffer(AudioBuffer *buffer, int toneVolume, int wavVolume, unsigned int fade)
```
bsongis: Well I am wrong again, I prefer your previous solution. And I am almost sure that the compiler will optimize the double test.

bsongis:
Yes it is a good idea to merge first the latest FatFS changes, it will be also easier to review!

radio/src/audio_arm.h
projectkk2glider: Bummer, I kind of like the new solution, the code in the wakeup function is much clearer now. Please consider changing your opinion.

projectkk2glider:
I will prepare a new PR with just the new FatFS then.

projectkk2glider:
New FatFS 0.12 in #3925

projectkk2glider:
Audio changes are actually independent of the FatFS changes - so a PR for audio is in #3926

PR #3880, opened 2016-10-07
Ensure default for luaDrawNumber is RIGHT for consistency with 2.1.x

3djc:


radio/src/lua/api_lcd.cpp
```diff
@@ -226,6 +226,7 @@ static int luaLcdDrawNumber(lua_State *L)
   int y = luaL_checkinteger(L, 2);
   int val = luaL_checkinteger(L, 3);
   unsigned int att = luaL_optunsigned(L, 4, 0);
+  if (!(att & LEFT)) att |= RIGHT;
```
schwabe: I like this change. You might want to add a short comment why this is here.

// Default was RIGHT aligned in 2.1, do not change default for scripts

3djc: Even better if we specifiy it in the doc for end users, don't you think ?

PR #3881, opened 2016-10-07
Add 'Z' as key to the rotary encoder emulation

schwabe:
This helps us people with non AZERTY keyboards that have Z left to the X. Doesn't help German layouts with 'Y' next to 'X' but that key is already used as trim key.


bsongis:
Indeed!

PR #3883, opened 2016-10-07
Schwabe/spektrum dsm2 bind

schwabe:
Update OpenTX Multi code to match current MULTI code:

- Allow 63 protocols instead of 31. This unfortunately requires a change how EEPROM values are interpreted for MULT. I would like to get this change in before 2.2 is official released to minimise the impact.
- This commits fixes a lot of bugs related to selecting a custom MULTI protocol.
- Parsing DSM2/X bind repsonses is implemented and DSMX/DSM2 11ms/22ms and number channels automatically selected.
- Allow to select broken PRNG tables in DSM2 for OrangeRX receivers
- move some of the definition of MULTI to gui_common.h instead of replicating them
- Make multi companion implementation cleaner


szempruw:
This merged issue is the closest I can find to my simple question. Sorry if this is the wrong place to ask. I am a simple user, not a developer.

I am hoping that one day I can buy an RF Module and use it with my Taranis X9D or X9D+. I have a 9 Channel AR9350 AS3X Receiver. I have a current hacked RF Module that is limited to 6 channels. I see the new Jumper JP4in1 and I asked this simple question to the developers of the 4in1 module software suspecting that there must be software synergy between the RF module software and the Taranis software. (Obvious over simplification here.)

Here was my question and answer.

 "Sorry if this is the wrong place to ask. But why are the number of DSM2/DSMX servo channels limited to 6? I use module on Taranis X9D+. OpenTx 2.2.1. There are Spektrum Rx's with more, i.e. AR9350 has 9 channels and uses a DSMX Spektrum transmitter. Is this only driven by OpenTx??? Or is synergy between OpenTx and RF module needed?"

Here was his answer.

"The number of channels to be sent is selectable by software from 6 (default) to 12. 
Here are the protocol details: https://github.com/pascallanger/DIY-...ils.md#dsm---6

Some receivers don't like some number of channels , this is why you can force it. But if you bind using auto detect, the module sets the radio settings based on what the RX requires.
In short, you have to set the number of channels.
Pascal"

If I read the GitHub link correctly, 12 channel read of a Spektrum Rx is possible with this module. But will it? And by which protocol?  How am I able to set it? Can you comment?

When I currently try to choose external RF settings, DSM / DSMX I only get a choice of 6 channels. I get it that I have a 6 channel RF Module. But will the JP4in1 really solve the problem? 

So will it be possible one day? Or must I purchase a 12 channel Spektrum radio. I just sense that Spektrum's ASX3 and auto leveling features that help intermediate flyers fly better and safer (no pun intended) are further down the road than what can be done with Taranis, unless I'm really missing something. 

Aside from all this, as I tried the IRange X module, the useful flying range was limited. Poor RSSI and suggested only toy quad use was possible. Not safe for normal Sport Plane flying. I'm hoping the JP4in1 will range test better.

I apologize in advance for my ignorance.

3djc:
If you choose DSM, then it is for an official DSM module

What you want is multi, then DSM, and there you can set the number of channels you want

image

Issue #3827, opened 2016-09-18
Weird output of all channels when Flight Data reset is chosen


rcbebo82:
Today i was at our flight field and found a curious error.
I flight my T-Rex 600 Nitro landed refueled it and choose "Reset flight data" on the horus after that the horus goes in in warning mode Throttle channel not zero and all Chanels get wirde output in all directions.

A video with the error: https://youtu.be/9WwtAGJWOyA

kilrah:
Are you using the internal RF or an external module?

bsongis:
No failsafe?

rcbebo82:
Internal RF with internal antenna. Receiver XSR connected by SBUS to the V-Stabi. Failsafe set to hold.

kilrah:
Is it reproductible? What happens if you turn the radio off?

rcbebo82:
Yes it is. Every time when a warning is active eg. Throttle not zero or any switch in a alarm position and flight data reset is done this happens. 
When warning is active and this is currently Happening if i press the power button jitterings stop immediatly after the horus is of the Receiver goes in failsafe mode.

rcbebo82:
It also happens on an other model with an X8R.

projectkk2glider:
Testing on Horus with 29e618146c673f605234a19210157c244c7d9cd4 I get another problem:
 * Initial throttle, switch and SD version warning - no problems
 * Flight reset
 * if I leave any warning displayed for more than 10s or so, I get total Menus thread freeze - no more display update. Radio still works and emergency shutdown also works. This does not happen every time, but at least in half of the tries.

projectkk2glider:
I have explanation for GUI freeze in warning display. When warning is in effect the `getADC()` function gets called from two threads: main and mixer. Downstream it uses the `adcRead()` which is not reentrant! A mutex should solve this, hold on...

projectkk2glider:
@rcbebo82 can you test my fix if it solves your issue?

rcbebo82:
Sure, i will do this tomorrow.

rcbebo82:
I cant clone your branch projectkk2glider/issue_3827_adc_mutex because i dont see it. How can i clone it?

projectkk2glider:
Just checkout my branch.  You need to pull the repository before you will see the new branch.

rcbebo82:
I dont understand that. I cloned your branch opentx but there is no branch "issue_3827_adc_mutex".
[...]

Issue #3820, opened 2016-09-17
Resume after USB connection problems


projectkk2glider:
On Horus if USB=MASSSTORAGE, disconecting USB cable produces different results:
- a GUI freeze
- or a bad SD CARD message

Example trace, radio startup and USB cable connected:

```
register theme Default
register theme Darkblue
register layout Layout2x2
register layout Layout2x1
register layout Layout2x4
register layout Layout2P1
register layout Layout1x1
register widget ModelBmp
register widget Outputs
register widget Gauge
register widget Value
register widget Text
register widget Timer

Horus board started :)
>opentxInit
sdInit
sdMount
luaInit
register theme TBS
register widget Counter
register widget BattCheck
storageReadAll
loadFile(/RADIO/radio.bin)
loadFile(/MODELS/model1.bin)
disk_read bad alignment (0x100061ec)
load theme Default
disk_read bad alignment (0xd00f06bd)
disk_read bad alignment (0xd00f86b1)
opentxStart
�GC Use: 37616bytes
gps> $PUBX,40,VTG,0,0,0,0*5e
gps> $PUBX,40,GSV,0,0,0,0*59
RamBackupWrite sdsize=10227 backupsize=4531 rlcsize=294
gps> $PUBX,40,GSA,0,0,0,0*4e
gps> $PUBX,40,GLL,0,0,0,0*5c
eeprom write general
writeFile(/RADIO/radio.bin)
disk_write bad alignment (0x10003e50)
disk_write bad alignment (0x10003e50)
opentxClose
eeprom write general
writeFile(/RADIO/radio.bin)
disk_write bad alignment (0x10003e38)
disk_write bad alignment (0x10003e38)
sdDone
RamBackupWrite sdsize=10227 backupsize=4531 rlcsize=294
```

Disconnecting cable:

```
opentxResume
sdMount
storageReadAll
loadFile(/RADIO/radio.bin)
loadFile(/MODELS/model1.bin)
disk_read bad alignment (0x100061ec)
load theme Default
opentxStart
```

kilrah:
It was fine a couple of weeks ago...

projectkk2glider:
It might be just a bad USB cable that is corrupting the contents of the SD card. Will keep a close look.

projectkk2glider:
Another symptom:
* Radio apears to work normally, but subsequent USB connections do not work, I get:
```
[...]

PR #3889, opened 2016-10-07
Re #3827: added mutex to getADC(), because adcRead() is not reentrant

projectkk2glider:
This ~~might~~(https://github.com/opentx/opentx/issues/3827#issuecomment-252425167) also fix the #3827 original issue.

### DIFF
--- a/radio/src/opentx.cpp
+++ b/radio/src/opentx.cpp
@@ -1119,7 +1119,7 @@ void checkTHR()
     return;
   }
 
-  getADC();
+  getADC_ifMixerNotRunning();
 
   evalInputs(e_perout_mode_notrainer); // let do evalInputs do the job
 
@@ -1138,7 +1138,7 @@ void checkTHR()
 
   while (1) {
 
-    getADC();
+    getADC_ifMixerNotRunning();
 
     evalInputs(e_perout_mode_notrainer); // let do evalInputs do the job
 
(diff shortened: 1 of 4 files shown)


radio/src/opentx.cpp
```diff
@@ -1138,8 +1138,6 @@ void checkTHR()
 
   while (1) {
 
-    SIMU_SLEEP(1);
-
     getADC();
```
bsongis: In order to avoid the use of the mutex, we could check if the mixer is running, no?

radio/src/opentx.cpp
```diff
@@ -1414,6 +1421,8 @@ uint16_t anaIn(uint8_t chan)
 #if defined(CPUARM)
 void getADC()
 {
+  static OS_MutexID getAdcMutex;
```
bsongis: I thought there was an init method to be called, needs to be checked

bsongis:
Good catch, this could explain the freeze / reboot on model load!

projectkk2glider:
Do you like my solution without the mutex? 

I would also like you to try to reproduce the issue https://github.com/opentx/opentx/issues/3820#issuecomment-252413666

radio/src/opentx.h
```diff
@@ -847,6 +847,11 @@ void checkAll();
 
 #if !defined(SIMU)
 void getADC();
+  #if defined(CPUARM)
+    #define getADC_ifMixerNotRunning()    do { if (s_pulses_paused) getADC(); } while(0)
```
bsongis: I like this solution, yes
Instead I don't like this function name which mixes all existing styles around the world :)
May I propose GET_ADC_IF_MIXER_NOT_RUNNING()?

projectkk2glider: I knew that :grin: I will change it.

bsongis:
Thanks a lot for this problem hunting and fix!

PR #3892, opened 2016-10-09
Update de.h.txt

hrenz:
Was an error at Text , function of seriell Port
SBus  is an Input at Taranis from a SBus receiver

### DIFF
--- a/radio/src/translations/de.h.txt
+++ b/radio/src/translations/de.h.txt
@@ -80,7 +80,7 @@
 #define TR_TRNCHN              "CH1CH2CH3CH4"
 
 #define LEN_UART3MODES         "\015"
-#define TR_UART3MODES          "AUS\0         ""S-Port Mirror""Telemetrie\0  ""SBUS Schüler\0""Debug\0"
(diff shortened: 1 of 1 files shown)


radio/src/translations/de.h.txt
```diff
@@ -80,7 +80,7 @@
 #define TR_TRNCHN              "CH1CH2CH3CH4"
 
 #define LEN_UART3MODES         "\015"
-#define TR_UART3MODES          "AUS\0         ""S-Port Mirror""Telemetrie\0  ""SBUS Schüler\0""Debug\0"
+#define TR_UART3MODES          "AUS\0         ""SPort Mirror\0""Telemetrie\0  ""SBUS-Lehrer In\0""Debug\0"
```
bsongis: 15 chars in your new string, I don't think it's good

schwabe: And Lehrer in is also misleading as could input of the master or input of master data

bsongis:
Please reopen when issues fixed, thanks!

PR #3893, opened 2016-10-09
Patch 14

hrenz:
Was an error at Text , function of seriell Port
SBus  is an Input at Taranis from a SBus receiver


bsongis:
I suppose this PR was done by mistake?

PR #3894, opened 2016-10-09
3djc/add sdcard dir struct

3djc:
sdcard version change required not only for this, but also previous german changes


bsongis:
Excellent idea!

PR #3897, opened 2016-10-10
3djc/build revise handling sdcard.version

3djc:
opentx.sdcard.version will not exist on git anymore, but will be created at release time instead. This will remove git handling issues


bsongis:
Yes, good!

PR #3899, opened 2016-10-11
Registered widgets limit removed

bsongis:


3djc:
Tested ok with 12 widgets, :)

PR #3900, opened 2016-10-12
Patch 13

hrenz:


schwabe:
Lenght of the string is still changed without changing the string length define.

I find Lehrer In confusing as it should be Shüler in (you have an input of Schüler data on the Lehrer's remoe)

bsongis:
Your new sub-string is 11 bytes long instead of 13, it cannot work like that!

PR #3902, opened 2016-10-13
Fix #3837 - Fix wizard for non Horus

### LINKED ISSUE #3837: LUA Wizard no longer works in 2.2.0N358 (Taranis+)
Taranis + - opentx-x9d+-lua-mixersmon-multimodule-ppmus-en-2.2.0N358

After moving wizard folder into scripts (SD card folder structure seem to be wrong / sdcard.h might need adjustment) the wizard just return an error when create a new model.

```
Script syntax error
wizard.lua:72
attempt to call field 'lock' (a nil value)
```

I also noticed that between 2.1 and 2.2 some SD-card folder names have changes (BMP > IMAGES), (FIRMWARES > FIRMWARE), (EEPROMS > EEPROM). Maybe it would also make sense to create those empty folder in the SD-card structure (including the SCRIPTS subfolder) to reduce confusions.

lichtl:
This PR will fix the error related to the deprecated lcd.lock function and move the wizard into the correct folder. Tested via simulator and on a Taranis+ 2.2.0N358


bsongis:
Thanks!

PR #3929, opened 2016-10-18
only a string changed

hrenz:


radio/src/translations/de.h.txt
```diff
@@ -80,7 +80,7 @@
 #define TR_TRNCHN              "CH1CH2CH3CH4"
 
 #define LEN_UART3MODES         "\015"
-#define TR_UART3MODES          "AUS\0         ""S-Port Mirror""Telemetrie\0  ""SBUS Schüler\0""Debug\0"
+#define TR_UART3MODES          "AUS\0         ""S-Port Mirror""Telemetrie\0  ""SBUS Input\0  ""Debug\0"
```
schwabe: This will work from the string Length. But why not "SBUS Eingang\0"?

kilrah:
Changed to "SBUS Eingang" in other branch

Issue #3823, opened 2016-09-17
Request:  Write protect on categories or models


MyRCode:
When selecting a new category using long press, delete category is one rotation away from select category.  It is very easy to accidentally delete a category.

iflymlnm:
+1 
I've created a few models and deleting the category wiped all my models, very very easy to accidentally delete potentially lots of models.

3djc:
The solution might be to have a confirmation box for deletion

MyRCode:
I believe there is a confirmation already,  but there is also a confirmation to select a different model so I was used to just hitting the short press twice.   

I've just had to take a mental pause in between the two passes just to verify I haven't accidentally scrolled the wheel.

PR #3932, opened 2016-10-18
3djc/horus fix3924

### LINKED ISSUE #3924: Model Selection UI should not allow deleting categories if they aren't empty
Steps to reproduce on Horus with OpenTX 2.2 N358:

- Create two categories, CatA and CatB
- Create two model1 and model2.
- Put model1 in CatA and model2 in CatB
- Delete CatB
- model2 disappears

model2 still exists in the MODELS directory of the SD card and if you manually edit RADIO/models.txt you can  create a category and put model2.bin in it, which will then show up in the UI.

From chat @3djc suggested the solution would be to not allow categories to be deleted unless they were empty.

3djc:
Forbids deletion of category when not empty. Add confirmation of delete when empty
This fixes #3924 and #3823


radio/src/gui/480x272/model_select.cpp
```diff
@@ -108,6 +113,7 @@ void onModelSelectMenu(const char * result)
   else if (result == STR_DELETE_MODEL) {
     POPUP_CONFIRMATION(STR_DELETEMODEL);
     SET_WARNING_INFO(currentModel->name, LEN_MODEL_FILENAME, 0);
+    deleteMode = MODE_DELETE_MODEL;
```
bsongis: Yes no choice, but we really need to change that in the near future with a function pointer!

Issue #2998, opened 2015-10-26
Companion Curves Tab Suggestions


elecpower:
Environment - Companion 2.1.3 Taranis X9D Plus Windows 10
1. For the newbies like me suggest adding a label such as "Show" or "Overlay" above the two columns of curve checkboxes or a mouse over such as "Show | Overlay this curve on canvas with curve being edited".
2. Cosmetic layout symetry/logical - move curves 17 - 32 checkboxes to left of coloured curve label rectangle so between curve canvas and curve label.

PR #3933, opened 2016-10-18
Companion model edit curves tab enhancement

elecpower:
Addresses the requests raised in #2998.
For consideration


3djc:
Thanks a lot !

PR #3936, opened 2016-10-19
German spoken language fixed.

rcbebo82:
Spoken numbers fixed:
100 -> "ein", "hundert"
1000 -> "ein", "tausend"
999 -> "9", "hundert", "99"

Timers:
60 seconds -> "eine", "minute"
3600 seconds -> "eine", "stunde"
120 seconds -> "2", "minuten"

Sound files needed to be generated because new words are used.

### DIFF
--- a/radio/util/tts_de.py
+++ b/radio/util/tts_de.py
@@ -8,14 +8,17 @@ from tts_common import filename, NO_ALTERNATE, PROMPT_SYSTEM_BASE, PROMPT_CUSTOM
 systemSounds = []
 sounds = []
 
-for i in range(101):
+for i in range(100):
     systemSounds.append((str(i), filename(PROMPT_SYSTEM_BASE + i)))
-systemSounds.append(("tausend", filename(PROMPT_SYSTEM_BASE + 101)))
-for i, s in enumerate(["comma", "und", "minus", "uhr", "minute", "minuten", "sekunde", "sekunden"]):
-    systemSounds.append((s, filename(PROMPT_SYSTEM_BASE + 102 + i)))
-for i, (s, f) in enumerate([(u"Volt","volt0"),
-                            (u"Ampere", "amp0"),
-                            (u"MilliAmpere", "mamp0"),
+systemSounds.append(("ein", filename(PROMPT_SYSTEM_BASE + 100)))
+systemSounds.append(("eine", filename(PROMPT_SYSTEM_BASE + 101)))
+systemSounds.append(("hundert", filename(PROMPT_SYSTEM_BASE + 102)))
+systemSounds.append(("tausend", filename(PROMPT_SYSTEM_BASE + 103)))
+for i, s in enumerate(["komma", "und", "minus", "uhr", "minute", "minuten", "sekunde", "sekunden", "stunde", "stunden"]):
+    systemSounds.append((s, filename(PROMPT_SYSTEM_BASE + 104 + i)))
+for i, (s, f) in enumerate([(u"Wolt","volt0"),
+                            (u"Ampär", "amp0"),
+                            (u"MilliAmpär", "mamp0"),
                             (u"Knoten", "knot0"),
                             (u"Meter pro sekunde", "mps0"),
                             (u"Fuss pro sekunde", "fps0"),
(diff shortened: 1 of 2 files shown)


rcbebo82:
Got it with the SmartGit Client.
Now the new code file and the new python file for generating sound files!

radio/util/tts_de.py
```diff
@@ -8,11 +8,14 @@
...
+for i in range(100):
     systemSounds.append((str(i), filename(PROMPT_SYSTEM_BASE + i)))
-systemSounds.append(("tausend", filename(PROMPT_SYSTEM_BASE + 101)))
-for i, s in enumerate(["comma", "und", "minus", "uhr", "minute", "minuten", "sekunde", "sekunden"]):
-    systemSounds.append((s, filename(PROMPT_SYSTEM_BASE + 102 + i)))
+systemSounds.append(("ein", filename(PROMPT_SYSTEM_BASE + 100)))
+systemSounds.append(("eine", filename(PROMPT_SYSTEM_BASE + 101)))
+systemSounds.append(("hundert", filename(PROMPT_SYSTEM_BASE + 102)))
+systemSounds.append(("tausend", filename(PROMPT_SYSTEM_BASE + 103)))
+for i, s in enumerate(["komma", "und", "minus", "uhr", "minute", "minuten", "sekunde", "sekunden", "stunde", "stunden"]):
```
schwabe: did you check that komma really sounds better than comma? The mispelling are sometimes to get better pronounciation.

rcbebo82: It is pronounced correctly. With pico2wave there is even now difference recognizable.

bsongis:
I will let german speaking people press the button for that one, ok?

3djc:
I'll do some more test tomorrow, I need to generate a audio pack first

3djc:
I believe there is an issue when you want to play values at 0, you just have the unit and no 'zero' value

rcbebo82:
I will recheck that. Could be possible that i didnt test this!

rcbebo82:
Issue confirmed, working on the code again.

rcbebo82:
Zero works now, but im stuck at the precesion. Could anyone explain me how a telemetry source with precesion 0.00 is handled. 
Concretly i would to know how the parameter precesion on a telemtry source is set in code:
https://github.com/opentx/opentx/blob/master/radio/src/translations/tts_de.cpp#L103
So which "mode" is set by precesion:
0 -> 0?
0.0 -> 1?
0.00 -> 2?

3djc:
Let's say your value is 1024, 
PREC 0 is 1024
PREC1 is 102,4
PREC2 is 10,24

rcbebo82:
Hi, that was already clear to me. But the value "mode" controls which precesion is used. 
German language is really a nightmare. A small example:

This works already:
Precesion=0, Unit=Volt:, Value=1
"ein" "Volt"
Precesion=0, Unit=Volt:, Value=2
2 "Volt"

Here it starts to broken down:
Precesion=1, Unit=Volt:, Value=11
Should be spoken: 1 "komma" 1 "Volt"
Is spoken: "ein" "komma" "ein" "Volt"

Works:
Precesion=1, Unit=Volt:, Value=22
2 "komma" 2 "Volt"

Defect:
Precesion=2, Unit=Volt:, Value=111
"hundert" "komma" "hundert" "Volt"

Precesion=2, Unit=Volt:, Value=222
2 "komma" 2 "Volt"

rcbebo82:
TRACE("DEBUG MODE : %d", mode);
Thx will try this!

rcbebo82:
Is it possible that precesion 0.00 does not work anymore also in english. If i choose as source VFAS with a value of 4,85 Volt and telemetry sensor set to precesion 0.00 i get as number 49 instead of 485.
Or is the spoken part limited to precesion 0.0?

3djc:
Not sure what you are refereing too, since even with 0.00 precisiion (defaut for VFAS sensor), the 'play value' function uses 0.0 precision

rcbebo82:
So play value uses in this context every time 0.0 precesion? This means if i switch to precesion 0.00 in a telemetry source every time 0.0 is used?

kilrah:
Yes, voice callouts have always only been 1 decimal.
Makes little sense calling out 2 in most cases.

3djc:
Tested it (with my limited german),, sounds ok

PR #3942, opened 2016-10-21
Fixes #3937: When executed from Special Functions, Flight Reset cause…

### LINKED ISSUE #3937: SF "reset flight" warning display issues
Horus, git checkout

When I use the SF "reset flight" my radio locks up.
The power LED turns red and the radio will not be freed until I press a button at the thumbwheel.

projectkk2glider:
…d a GUI lockup

Closes #3937

### DIFF
--- a/radio/src/opentx.h
+++ b/radio/src/opentx.h
@@ -1347,8 +1347,13 @@ void clearMFP();
 extern uint8_t requiredSpeakerVolume;
 #endif
 
-#if defined(PCBTARANIS)
-extern uint8_t requestScreenshot;
+#if defined(CPUARM)
+enum MainRequest {
+  RequestScreenshot,
+  RequestFlightReset,
+};
+
+extern uint8_t mainRequestFlags;
 #endif
 
 void checkBattery();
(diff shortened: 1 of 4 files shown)


radio/src/opentx.h
```diff
@@ -1347,8 +1347,13 @@ void clearMFP();
 extern uint8_t requiredSpeakerVolume;
 #endif
 
-#if defined(PCBTARANIS)
-extern uint8_t requestScreenshot;
+#if defined(CPUARM)
+enum MainRequest {
+  RequestScreenshot,
```
bsongis: Rather name them like constants: ENUM_NAME.

projectkk2glider: You mean REQUEST_SCREENSHOT?

bsongis: Yes, I just updated the coding style wiki page, it was not written

kilrah:
All good for me, works as expected. Mixer still runs and updates servo positions while warnings are displayed, but that's probably the best way to go about it. Thanks!

PR #3947, opened 2016-10-23
Add opentx os x build vm and allow more cores to be used for fast mac…

schwabe:
…hines.

### DIFF
--- a/tools/build-companion.sh
+++ b/tools/build-companion.sh
@@ -10,9 +10,9 @@ while [ $# -gt 0 ]
 do
   case "$1" in
     --jobs=*)
-      CORES="${1#*=}";;
+      JOBS="${1#*=}";;
     -j*)
-      CORES="${1#*j}";;
+      JOBS="${1#*j}";;
     -*)
       echo >&2 "usage: $0 [-j<jobs>|--jobs=<jobs>] SRCDIR OUTDIR VERSION_SUFFIX"
       exit 1;;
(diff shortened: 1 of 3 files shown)


tools/build-companion.sh
```diff
@@ -20,42 +20,45 @@ if [ "$3" == "" ]; then
    exit 1
 fi
 
+if [ "$JOBS" == "" ]; then
+    JOBS=2
+fi
```
projectkk2glider: I would prefer if JOBS was an option to the script like in https://github.com/opentx/opentx/blob/next/tools/commit-tests.sh, but then you have to take care of the other two arguments also. Nothing critical if it stays the way it is here.

schwabe: I can also remove that from this pull request for now until we have a better option

projectkk2glider:
Man, you are to quick :stuck_out_tongue:

tools/build-companion.sh
```diff
@@ -4,11 +4,28 @@
 set -e
 set -x
 
+JOBS=2
+
+while [ $# -gt 0 ]
+do
+  case "$1" in
+    --jobs=*)
+      CORES="${1#*=}";;
```
projectkk2glider: My bad, change CORES to JOBS. Same problem two lines down.

schwabe: I should have tested it

PR #3949, opened 2016-10-23
[X9D] Allow toggling Haptic option for Taranis X9D  (closes #3946)

### LINKED ISSUE #3946: No option for Haptic on Taranis X9D Non-Plus
Ive ticked the box in companion 2.20N359 and i still can't get the haptic option to appear on my taranis I've even compiled from source and still not getting this as an option is there something I'm doing wrong or a bug somewhere?

schwabe:


projectkk2glider:
Not sure this will do anything, according to http://stackoverflow.com/questions/36358217/cmake-option-vs-set-a-cached-bool-variable the `set` and `option` can both be overridden by cmdline options. This change still leaves the HAPTIC off. 

I think the -DHAPTIC=YES should be in the cmake command somewhere in the php scripts.

schwabe:
@projectkk2glider CACHE is the important word there. See https://cmake.org/cmake/help/latest/command/set.html

We could also add CACHE but I think Option is easier to read/understand

I tested it in simu and it make a difference (Haptic shown in general option)

projectkk2glider:
OK, learned something new :+1:

schwabe:
To be honest I had to double check it too. But normal set being overridable by a user seemed strange to me

projectkk2glider:
Thanks.

PR #3955, opened 2016-10-26
Abort cmake if leftover files of 2.1 are found

schwabe:
This is becoming a FAQ. File selection might not be perfect but is good enough to trigger the error.


bsongis:
Yes I also had this problem many times...

PR #3958, opened 2016-10-26
[Companion] Fixes #3858

### LINKED ISSUE #3858: SF "increment/decrement GV" doesn't work as intended in 2.2n
Hi,
just made some migration tests to 2.2n358

oTx 2.18 was able to use GVs as a counter by define some SFs like this

sf gvar

so the stepping rate was +/-1 

after converting to 2.2, the GV value jumps quickly to its new defined max/min values (+/- 1024 as standard?) by pulling/pushing the configured switch.
incremental value seems to be completely ignored

regards
Udo

bsongis:


projectkk2glider:
Tested. All good.

PR #3968, opened 2016-10-30
Fixes #3964: Added free space check before putting new data into Lua …

### LINKED ISSUE #3964: luaSportTelemetryPop troubles when using DIY passthrough
A while ago, I had tested 75e1bead3f121c7bb6515846dff51e346ed81923 before it got merged and that was working fine: for a packet with a Data ID 0x5001, calling `id,value = telemetryPop()` from a lua script would give me id=1 and the value was good.

Now, using `physicalId, primId, dataId, value = sportTelemetryPop()` I get primId=16, dataID=0x5001 and the value is good for a while (about 10 seconds after transmitting that packet repeatedly).

But after that, I run into a problem as I permanently receive weird data that looks like it's shifted by some number of bytes: primId=1, dataId is wrong (I explain right below) and value=452984832 no matter what value I actually transmit.

So when the weird behavior happens:
If I transmit a value of 1, I get dataID=336 which in binary is 01 01010000.
If I transmit a value of 2, I get dataID=592 which in binary is 10 01010000.
If I transmit a value of 3, I get dataID=848 which in binary is 11 01010000.
So, it looks to me as if the data is shifted, but it only happens after I've been transmitting for a little while. Restarting the Taranis clears the problem, but it comes back again about 10 seconds later.

Also, when this happens, if I transmit on Data ID 0x5002, I get primId=2.

Please do not hesitate to ask followups, and I can certainly do some tests, even recompile the firmware with modifications...

projectkk2glider:
…Telemetry Fifo

Closes #3964

Completely untested!

### DIFF
--- a/radio/src/fifo.h
+++ b/radio/src/fifo.h
@@ -24,6 +24,8 @@
 template <class T, int N>
 class Fifo
 {
+  static_assert((N > 1) & !(N & (N - 1)), "Fifo size must be a power of two!");
+
   public:
     Fifo():
       widx(0),
@@ -57,7 +59,7 @@ class Fifo
       }
     }
 
-    bool isEmpty()
+    bool isEmpty() const
     {
       return (ridx == widx);
     }
@@ -73,12 +75,17 @@ class Fifo
       while (!isEmpty()) {};
     }
 
-    uint32_t size()
+    uint32_t size() const
     {
       return (N + widx - ridx) & (N-1);
     }
 
-    bool probe(T & element)
+    uint32_t hasSpace(uint32_t n) const
+    {
+      return (N > (size() + n));
+    }
+
+    bool probe(T & element) const
     {
       if (isEmpty()) {
         return false;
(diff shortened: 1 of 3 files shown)


floaledm:
Would you like me to test this PR?

projectkk2glider:
Sure, go ahead!

floaledm:
Damjan,
I tested the resulting firmware and your PR fixed the issue. I still wonder if it wouldn't be best to drop the oldest packet when the FIFO gets full, rather than dropping the new one, but this is still much better...

projectkk2glider:
Even if we drop the oldest packet, your script will sill be using the oldest packet. You should read the fifo until you get no more packets in each loop of Lua script. The last successfully read packet should be used for any processing. This way the data will be as fresh as possible. If  you allow the fifo to become full, you are working with old data.

floaledm:
Yes, I agree and that's what I'll do in my scripts. Thanks for all your work and comments!

bsongis:
Seems good

PR #3971, opened 2016-10-31
Display commit on radio

schwabe:
This makes it easier to diagnose which exact version is on a radio when dealing with build from source version


schwabe:
image
image
bildschirmfoto 2016-10-31 um 21 19 35

kilrah:
Great idea 👍

PR #3973, opened 2016-11-01
Sport: provide method to add sensors from lua

floaledm:
Useful for DIY passthrough to bring standard sensorIDs/dataIDs/values back to OpenTX for general use. For example, for using Alt, VARIO, VFAS in the OpenTX environment, this alleviates the need to transmit them separately. They can be transmitted "compressed" via the DIY passthrough and brought back from lua as standard sensors...

### DIFF
--- a/radio/src/lua/api_general.cpp
+++ b/radio/src/lua/api_general.cpp
@@ -902,6 +902,28 @@ static int luaDefaultChannel(lua_State * L)
   return 1;
 }
 
+/*luadoc
+@function registerSensor()
+
+Transmit data back to the general environment as a discoverable sensor, all RAW units
+
+@param instance (number) sensor data ID
+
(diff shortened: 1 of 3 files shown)


schwabe:
Is your intend to be able to create sensors from LUA or to update existing sensors? This is more "feed sport style data to sensor" approach. 

A cleaner approach would be to expose setTelemetryValue to lua and let it default to TELEM_PROTO_LUA.

floaledm:
I want to create sensors from what I get in lua from using the DIY passthrough.

schwabe:
I implemented an alternative approach in pull request #3977 Could you have a look and tell me what you think?

floaledm:
#3977 is superior, so I'm closing this one...

PR #3974, opened 2016-11-01
Simple cleanup of #if defined

floaledm:


schwabe:
That commit looks like it adds longer strings to AVR. Have you checke that this does not increase the size of the AVR binaries? (make firmware-size)

floaledm:
How do I check that?

projectkk2glider:
I think it is good, @floaledm just removed the fidefs that could not be reached - ie dead code.

floaledm:
You got it Damjan. Looked like there was some dead code. That's it...

kilrah:
Yup, copypasta relics...

PR #3976, opened 2016-11-01
Provide getRSSI() lua method

floaledm:
This allows the low and critical alarm levels to be accessed in lua. This also allows access to RSSI level without needing discovery.


schwabe:
looks good

bsongis:
Seems good

PR #3977, opened 2016-11-01
Implement setting sensor values from LUA

schwabe:


floaledm:
Thanks for working on this! Just tested and I just can't get the name to show up in discovery.

I tried `setTelemetryValue("test",...` or a shorter string, or even a number, but none seem to work.

radio/src/lua/api_general.cpp
```diff
@@ -875,6 +875,46 @@ static int luaDefaultStick(lua_State * L)
...
+*/
+static int luaSetTelemetryValue(lua_State * L)
+{
+  const char* name = luaL_checkstring(L, 1);
+  uint16_t id = luaL_checkinteger(L, 2);
+  uint8_t subId = luaL_checkinteger(L, 3);
+  uint8_t instance = luaL_checkinteger(L, 4);
+  int32_t value = luaL_checkinteger(L, 5);
+  uint32_t unit = luaL_checkinteger(L, 6);
+  uint32_t prec = luaL_checkinteger(L, 7);
```
floaledm: How about having  unit and prec as optional? Even the name could be optional. Typically, an unrecognized sensor takes up the id number as a name... What do you think?

radio/src/lua/api_general.cpp
```diff
@@ -875,6 +875,46 @@ static int luaDefaultStick(lua_State * L)
...
+  int index = setTelemetryValue(TELEM_PROTO_LUA, id, subId, instance, value, unit, prec);
```
floaledm: Would there be an issue in having proto selection as an option (2 for SPort, 5 for LUA, etc.)? Seems like that would expand possibilities for users since the lists of recognized sensors (including PROTO_LUA) are in the source code and non editable...

schwabe: The proto is only used to setup/initialize the sensors. E.g. TELEM_PROTO_FRSKY sets some defaults for UNIT and PREC based on sensor id.

For lua I would keep that simple and just let the user call the function with the right UNIT and PREC values instead of doing that "behind the scenes magic".

floaledm: You're right; it's more flexible that way. If the name can be straightened out, no need for other PROTOs.
Is subId useful then? I see it used to differentiate in the crossfire proto... Would it be used anywhere else than to differentiate sensors further within the potential PROTO_LUA inits? I don't see the subId shown on the discovery screen.

schwabe:
Have you started "DISCOVER Sensors" before executing the script?

I added   ```setTelemetryValue("test", 5, 3, 3, 100, 1, 0);``` before the lcd.refresh() line in the snake.lua script and that worked for me.

Sesnor name does not work yet because I need to implement the normal string to Zstring conversion.

floaledm:
Awesome. Works like a charm now!

floaledm:
What do you think about having any of unit, prec, and name as optional? Not a big deal though.. I very much appreciate your work!

schwabe:
Optional would mean that will have default values like UNIT=RAW, PREC=0 and NAME=LUA or something like that and to be honest I have no idea how to declare function parameters to be optional in lua and a quick google search also came up with nothing.

floaledm:
That's fine. I can handle it once your PR is merged... Thanks!

bsongis:
Seems all good, thanks!

PR #3978, opened 2016-11-01
Various fixes for gtests that were shown when running tests in random…

projectkk2glider:
… sequence. Enabled random gtests shuffling in commit tests.

I am almost sure this won't pass gtests...


bsongis:
Thanks for taking care of that!

PR #3980, opened 2016-11-02
Mark gtest variables cached

schwabe:
Set GTEST location variables as CACHE so they can be modified when gtest is in a different location from the cmake call (/usr/src, /usr/include is not writable on OS X)


radio/src/tests/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
-set(GTEST_INCDIR /usr/include/gtest)
-set(GTEST_SRCDIR /usr/src/gtest)
+set(GTEST_INCDIR /usr/include/gtest CACHE string "location of Googletest includes")
```
bsongis: Uppercase first letter:
"Location Google Test includes"

PR #3981, opened 2016-11-02
3djc/fix #3965

### LINKED ISSUE #3965: OpenTX N359 Trim Does not center
Development unit 5 HORUS, N359
Frsky Firmware 1.2.20
Mode 2 using Non Eu version

Seems to happen on "Course" trim settings.
Extended trims on

Repeatable:
Adjust your elevator trim to full down
Then using the same trim button, adjust back to center.

2 Symptoms show up.
Maximum trim reached only shows up on the screen as 1/3 of the way up the trim bar
When going extended, it goes to about 2/3rd up the trim bar
When centering the trim, it does not stop at "Trim Center" and it's missing the "=" which indicates trim center.

3djc:
This forces a stop at 0 in the following scenarios:
- you hit ext trims limit, then go back center with trims on coarse
- by changing trim steps, you end up in a place that does not stop at 0 (ex : fine, move to 12, set coarse, you move back to +4 and then -4 without center stop

Please someone check avr binary size , I have strange results on my compile pc

This fixes #3965

### DIFF
--- a/radio/src/opentx.cpp
+++ b/radio/src/opentx.cpp
@@ -1300,7 +1300,7 @@ uint8_t checkTrim(event_t event)
     int16_t after = (k&1) ? before + v : before - v;   // positive = k&1
     bool beepTrim = false;
 
-    if (abs(after+before) < (abs(after)+abs(before))) { //forcing a stop a centerered trim when changing sides
+    if (before != 0 && (!(after < 0) == (before < 0))) { //forcing a stop a centerered trim when changing sides
       after = 0;
     }
     if (!thro && after==0 && before!=0) {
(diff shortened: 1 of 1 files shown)


bsongis:
This is good but why don't you just change the existing test
```after==0 && before!=0```
?

File: radio/src/opentx.cpp [hunk 1]
```diff
@@ -1300,7 +1300,8 @@ uint8_t checkTrim(event_t event)
     int16_t after = (k&1) ? before + v : before - v;   // positive = k&1
     bool beepTrim = false;
 
-    if (!thro && after==0 && before!=0) {
+    if (!thro && (!(after < 0) == (before < 0)) && before!=0) { //forcing a stop at centerered trim when changing sides
```
bsongis: !(..) == (..) could be reduced to (..) != (..)

bsongis:
Perfect!

PR #3984, opened 2016-11-03
Fix column alignment between text/digits

### LINKED ISSUE #3975: [X9E] GVar overwrites channel# in mixer screen
2.2.0 RC7@X9E
If a GVar is used as weight in the first line, it overwrites the channel# .

mixx9e

3djc:
This fixes #3975


kilrah:
Changes aligmnent of the weight when numbers.... disturbing but I guess we'll get used to it :)

bsongis:
Thanks!

PR #3986, opened 2016-11-03
3djc/fix #3777

### LINKED ISSUE #3777: Open TX Horus 2.2 ON352 - Model notes not shown when switching models
When a new model selection is made from model select menu and activated, the 'text notes' are not shown on screen, however notes can be selected from long press on enter key for that newly selected model.  When transmitter is powered 'on' then notes are displayed for the last model selected.

3djc:
Delay the checks until after the main menu is launched to avoid the model notes to be overwritten even before appearing.

This fixes #3777


bsongis:
Seems all good, thanks!

PR #3988, opened 2016-11-03
Fix Special function Backlight to actually also honour its value, imp…

schwabe:
…lement LCD colour on Taranis

The range of the FUNC_BACKLIGHT is 0 to 100 which is the same as brightness, so I think this was intended to be brightness. Anyway since it never worked it is now brightness and color got its new SF.


kilrah:
No, it was meant to be color. And it was supposed to be "temporary", i.e. overwrites the color in radio settings while the SF is active, and restores when off.
The idea being to be able to signal alarms by temporarily changing color for example.

bsongis:
Not now

PR #3990, opened 2016-11-04
Add constants for LUA telemetry

schwabe:
This commit added 896 bytes to flash size which is a significant size. An alternative would to just document the values in the lua telemetry function.

Shortening some of the UNIT names (e.g. UNIT_MW instead of UNIT_milliwatts) or shortening UNIT_ to U_ could save a few bytes but the general problem is still there.

@3djc  and I disagree if spending 1k of flash is worth on this, I would just tell the user the list of units/values instead of making them available as global LUA variables.


lshems:
Just wondering. I have noticed that several different lists are used to refer to sources, switches etc, depending on the functions used.

This forces you to create a LUA library for yourself anyway if you want to use these within your code consistently and often.

Is there any reason for this, and could this be perhaps offer the solution to reducing size and memory usage?

Some examples of codes I use right now to overcome this in LUA, for OpenTX 2.0.9:

`  page.sourceSwitch=   --to find the source of a physical switch(-position)
    {
    [0]=84,
    [1]=92,[2]=92,[3]=92,
    [4]=93,[5]=93,[6]=93,
    [7]=94,[8]=94,[9]=94,
    [10]=95,[11]=95,[12]=95,
    [13]=96,[14]=96,[15]=96,
    [16]=97,[17]=97,
    [18]=98,[19]=98,[20]=98,
    [21]=99,[22]=99
    }

    
    page.mixTimer=function(source)
      if source then
        return source+61
      end
    end

    page.timerMix=function(timer)
      if timer then
        return timer -61
      end
    end
    
    
    
    page.mixLS=function(source)
      if source then
        return source-48
      end
    end

    page.LSmix=function(LS)
      if LS then
        return LS+48
      end
    end

    page.mixTrimSwitch=function(source)
      if source then
        return source-42 ---RETA LRDU
      end
    end
`

kilrah:
Because those lists are all different in content, and doing them as it is saves memory on the firmware itself (especially important for the AVR platforms, if those were dropped it could be simplified but as it is not really).

3djc:
It seems the drawback outweight the benefits, so lets not do this

bsongis:
I would keep it with an #ifdef because it could be useful on Horus
I am sure the next radios round will bring a bigger flash

schwabe:
@bsongis okay reopened it with #if 0

radio/src/lua/api_general.cpp
```diff
@@ -1145,6 +1145,8 @@ const luaR_value_entry opentxConstants[] = {
   { "PLAY_BACKGROUND", PLAY_BACKGROUND },
   { "TIMEHOUR", TIMEHOUR },
 
+#if 0
```
bsongis: Why not 
```#if defined(PCBHORUS)```

bsongis:
And X9E? I cannot remember how much flash is available

schwabe:
At least the linker script for X9E defines "only" 512K of memory. 

My Taranis+ also has really 1MB of flash instead of 512k but I am not sure later version have really a bigger chip or if STM just sells relabelled 1MB chips as lower capacity.

kilrah:
From the part no mentioned in the schematic it's 512K. But the 1M part would obviously be interchangeable, so maybe it got better availablility at some point and some radios ended up having one...

3djc:
My X9E is from the first production batch, and has a 512k flash

bsongis:
Thanks!

PR #3991, opened 2016-11-04
Add EVT_MENU_LONG on taranis

raphaelcoeffic:
Basically, that's what it says. Considering the "Page" button is not really useable in telemetry scripts, I'll start using the "Menu" button instead in my scripts. However, it is quite convenient to have the long press as well, instead of declaring it in LUA:
```
local EVT_MENU_LONG = bit32.bor(bit32.band(EVT_MENU_BREAK,0x1f),0x80)
```


bsongis:
Isn't this key already used for displaying remaining RAM?

raphaelcoeffic:
That's well possible, but then I'd say it is not functional. Does not work for me on latest next-build.

projectkk2glider:
MENU long is used to display RAM, but only for the stand-alone scripts.

raphaelcoeffic:
@projectkk2glider meaning that it is indeed useful for telemetry scripts?

projectkk2glider:
I see no harm in merging this, thanks.

PR #3992, opened 2016-11-04
Document values of UNIT in the function describtion

schwabe:
Is is a bit a less clean than pull request #3990 but has no additional memory requirement


bsongis:
This one is already merged, no?

schwabe:
This one only adds the documentation for Taranis which does not have the constants

bsongis:
I am afraid someone else included your mods in another PR, am I right?

schwabe:
Yea, it is @floaledm included that. Did not see that.

PR #3993, opened 2016-11-04
Lua setTelemetryValue with optional arguments

floaledm:
Related to #3992

setTelemetryValue() can then be called with unit, precision, or name. Defaults:
unit: 0 -> UNIT_RAW
precision: 0 -> no decimal places
name: Id as a string (same as other PROTOs that have an unrecognized Id/subID pair)

### DIFF
--- a/radio/src/lua/api_general.cpp
+++ b/radio/src/lua/api_general.cpp
@@ -908,36 +908,55 @@ static int luaDefaultStick(lua_State * L)
 }
 
 /* luadoc
-@function setTelemetryValue(name, id, subID, instance, value, unit, precision)
-
-@param name (string) Name of the sensor if it does not yet exit (4 chars)
+@function setTelemetryValue(id, subID, instance, value [, unit] [, precision [, name])
 
 @param id Id of the sensor
 
-@param subID  subID of the sensor, usually 0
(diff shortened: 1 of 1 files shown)


3djc:
Could you describe why you would be needing that ?

floaledm:
Strictly speaking not a need, since it's about convenience by giving users of the setTelemetryValue() function more flexibility on its use... Is there any drawback to having optional args?

3djc:
The only purpose I can see to that is to introduce lua to lua data, since a telemetry sensor with no name, no units, will be of no use to end user, and I'm not sure going through telem is the best meaning for lua to lua

kilrah:
The one important thing is to be able to make use of the built-in defaults. If the arguments always override them there should be another call that doesn't.

floaledm:
@3djc With this PR, a telemetry sensor with no name specified will have the Id argument as a name.

@kilrah A TELEM_PROTO_LUA case in telemetry_sensors.cpp was added by another PR for built-in defaults. Those typically use id, subId, and instance to get recognized as built-in. Having unit, precision, and name as optional arguments would not override that...

kilrah:
Ah OK, didn't follow closely lately.

PR #3994, opened 2016-11-05
Add back model bin name, currently needed if you want to use model notes

3djc:
This name is usefull for those who want to use models notes


bsongis:
If I remember right, it is displayed on something else (the last models line?)

kilrah:
Ah yeah, could be if you have many models in the list...

3djc:
image

This what we have currently, it may be displayed somewhere else

bsongis:
Perhaps @mhotar would help us to find a place for the model filename? I do agree it's something interesting to see when the model is selected

mhotar:
You know, I am not happy you are about to push user to fiddle with files :) I would prefer to have model as a simple container with icon and model name only. But on the other side I understand why you, programmers love this feature :) How long will be this filename? model99.bin is longest one?

kilrah:
This is all temporary! Once companion is ready it can be removed, it's just that form now it allows helping someone about once a week when they are lost with their settings by telling them to look at this to send the right file. 
It's not even a big problem if it overwrites something for now...

3djc:
Why I brought it back is because if you want to use model notes (checklist), you need to create a text file based on the bin name, so you have to be able to see it.
I have a nice change ready allowing you to select a file per moel not based on name, but it requires either an eeprom change or a dirty hack, none been a good choice

mhotar:
One option could be to shrink category list by one item.
modelsel3
Here is the mask for file container
mask_mlib_files

PR #4009, opened 2016-11-09
lua: allow SPORT buffer availability to be queried

raphaelcoeffic:
This simplifies the LUA script in case SPORT is used as a transport for bigger requests (like MSP/SPORT). As of now, I have to build the chunk request, attempt sending, and if the buffer was not available, roll back to send buffer state to what it was previously. This change would allow to save CPU cycles, memory, and simplify the code within the LUA script.

### DIFF
--- a/radio/src/lua/api_general.cpp
+++ b/radio/src/lua/api_general.cpp
@@ -333,7 +333,10 @@ uint8_t getDataId(uint8_t physicalId)
 
 static int luaSportTelemetryPush(lua_State * L)
 {
-  if (isSportOutputBufferAvailable()) {
+  if (lua_gettop(L) == 0) {
+    lua_pushboolean(L, isSportOutputBufferAvailable());
+  }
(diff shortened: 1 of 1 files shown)


projectkk2glider:
Instead of introducing a new function, I would just modify the existing one `luaCrossfireTelemetryPush()` to do the same if no parameters are used to call it.

raphaelcoeffic:
Ok, then I'll just do it for both luaCrossfireTelemetryPush() & luaSportTelemetryPush().

projectkk2glider:
@raphaelcoeffic  In future please do not merge your own pull requests (unless told so) it defeats the whole purpose of the pull request mechanism (peer review).

Thanks for this. If you are so kind, you could also add the Lua docs for these functions :wink:

raphaelcoeffic:
@projectkk2glider Sorry! I realised that afterwards as well... Was too impatient to get my new LUA script out the door ;-) I'll proceed now to the docs.

PR #4017, opened 2016-11-11
Projectkk2glider/companion archive support

projectkk2glider:
Do not merge yet!  This is just an example.

This is an attempt to realize the https://github.com/opentx/opentx/wiki/Companion-for-Horus-blueprint


companion/src/firmwareinterface.cpp
```diff
@@ -39,6 +39,8 @@ int getFileType(const QString &fullFileName)
     return FILE_TYPE_EEPE;
   else if (suffix == "XML")
     return FILE_TYPE_XML;
+  else if (suffix == "EEPE2")
```
bsongis: Perhaps OTX extension?

kilrah: Yes - been waiting for the occasion to get rid of the eepe extension for about 4 years :)

projectkk2glider:
I have an idea, I will create a nice class that will abstract the storage model for Horus. The class will support:
* reading from disk (archive file .otx)
* saving to disk (.otx)
* reading from radio (or any other directory containing proper structure)
* saving to radio (or any other directory)

The file will have public members:
 * contents of models.txt
 * contents of radio.bin
 * list containing the contents of all modelXX.bin files

bsongis:
First merge done, I will push the PR now to avoid another one ...

PR #4018, opened 2016-11-12
Fix analog input order for sky9x != REVA

sreichholf:
Analog inputs were completely mixed up on my AR9X (and i guess for every other sky9x except REVA)

So i'm not sure this is the proper way to fix it but least it's gives a direction about what's wrong ;).

### DIFF
--- a/radio/src/opentx.cpp
+++ b/radio/src/opentx.cpp
@@ -1401,6 +1401,10 @@ tmr10ms_t jitterResetTime = 0;
 #if !defined(SIMU)
 uint16_t anaIn(uint8_t chan)
 {
+#if defined(PCBSKY9X) && !defined(REVA)
+  static const pm_char skyAna[] PROGMEM = {1,5,7,0,4,6,2,3};
+  chan = skyAna[chan];
+#endif
 #if defined(VIRTUAL_INPUTS)
   return ANA_FILT(chan);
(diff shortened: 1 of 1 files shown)


bsongis:
I will test it on 9XR-PRO

bsongis:
WTF the bootloader doesn't like my binary!

bsongis:
It seems there is an issue here:
-rw-r--r-- 1 bertrand bertrand 0 Nov 13 19:27 radio/src/bootloader.lbm

bsongis:
Reason: if Mike's bootloader is not correctly downloaded when CMake is run there is no error!

bsongis:
OpenTX 2.2 on 9XR-PRO is currently a total mess: no audio, analogs completely broken, it reboots after 10s max etc.

bsongis:
This commit is OK (audio and ADC tested):

commit dd44da14bc84e44cb4720cd1dfc565e8c751d26a
Author: Andre Bernet <andre@andrebernet.ch>
Date:   Fri Jul 22 19:56:49 2016 +0200

bsongis:
Your fix is right, I removed this code and didn't think of readding it in the driver code. I will do this now directly on 'next', please test on AR9X, I can only test on 9XR-PRO

PR #4024, opened 2016-11-15
Fix center trim stop

3djc:
that was not working when trim was changing from positive to negative.

Fixes the issue seen by Bsongis on Sky9X (but the issue was present on all platform)


bsongis:
Thanks!

PR #4025, opened 2016-11-15
Fix ar9x compile

3djc:


sreichholf:
tested, fixes ar9x here :)

bsongis:
I will prefer a white list over the black list here. I have pushed another change meanwhile

PR #4027, opened 2016-11-16
Fix Issue 4026 - Companion 2.2 Model Setup still using SD card BMP

jtaylor2:
directory instead of IMAGES directory for model images.


projectkk2glider:
Thans!

kilrah:
This will need major refactoring anyway since the extension is currently hardcoded and that has to be changed for Horus.
BTW I believe there are a few other locations where the BMP folder is hardcoded.

jtaylor2:
I had run a grep on BMP  and didn't find any other places where it is hardcoded as a directory.  Not to say I couldn't have missed something.  There are places where it is hardcoded as an extension.

Should I leave this open as a reminder for the refactoring or close it?

projectkk2glider:
I have also checked for other places where BMP was hard-coded and did not find any more.

PR #4034, opened 2016-11-17
3djc/horus wizard

3djc:
This is a wizard handler for Horus (it does not include wizards per say, just two sample ones for test and understanding)

This is used in the following way:
1) the wizard folder need to be placed in /SCRIPTS/WIZARD
2) it needs to include a 85x130 icon (samples included) named icon.png
3) it need to include a wizard.lua
4) wizard handler is trigered by new model creation


bsongis:
👍

PR #4048, opened 2016-11-19
Add display of detailed failure message when TX module firmware updat…

projectkk2glider:
…e fails.

This could be ported to 2.1, I think it would be a good idea. Some users have difficulty updating module firmware and this will at least give more detailed reason why the update failed.

Another thing to do (not so simple to display) is to add a check for the .frk file itself. I looks like the couple hundred bytes of each valid frk files is always the same. Perhaps we could check first 4 bytes and then warn user that something might be wrong with the file. But we would still let him try the flash. 

The problem for me is how to implement a warning with a choice after the user selects the `Flash` popup menu option.

### DIFF
--- a/radio/src/io/frsky_sport.cpp
+++ b/radio/src/io/frsky_sport.cpp
@@ -194,7 +194,7 @@ const char * sportUpdatePowerOn(ModuleIndex module)
     packet[1] = PRIM_REQ_POWERUP;
     sportWritePacket(packet);
     if (sportWaitState(SPORT_POWERUP_ACK, 100))
-      return 0;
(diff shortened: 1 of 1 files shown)


radio/src/io/frsky_sport.cpp
```diff
@@ -262,21 +262,24 @@ bool sportUpdateUploadFile(const char *filename)
 
     if (count < 256) {
       f_close(&file);
-      return true;
+      return 0;
```
bsongis: NULL

bsongis:
Thanks!

PR #4049, opened 2016-11-19
Fix companion "mountlist" build under FreeBSD

tjouan:


bsongis:
Same remark for this PR
Thanks for your help!

PR #4050, opened 2016-11-19
Fix radio firmwares build under FreeBSD

tjouan:


bsongis:
I guess it's also needed in the 'next' branch (for 2.2)?

tjouan:
Yes, I'll open a PR with #4049 too against `next`.

However, for companion I need a little hack to built it on `next`,
about QT5 and QtWidgets, and I haven't really tested it much. I'm
working on porting opentx to FreeBSD, but focusing on master for now
(`radio` part, however, seems easier to build on `next` for now).

Thank you and all contributors for your work on OpenTX, it's fantastic
that we actually have an open source firmware for radio
transmitters :-)

PR #4053, opened 2016-11-19
[X7D/X9E] GVars display fixed (menus INPUTS / MIXES / OUTPUTS)

bsongis:
Needs more tests before being merged


3djc:
Simu tested x7d and x9d, and radio tested x9e, seems ok with various models including those provided by fraternl

image

bsongis:
Thanks for your test!

Issue #4057, opened 2016-11-20
splash screen resolution


hketterer:
i have a custom splash screen that was readable in 2.1.8 but unreadable in 2.2 seems like the resolution was downgraded.

kilrah:
Maybe specify what radio type?

hketterer:
sorry, taranis plus

jtaylor2:
I noticed the same Taranis splash on 2.2 didn't look as good as it did on 2.1 but hadn't had time to look into it.  It seems that Companion 2.2 does something not nice to the profile picture and also if you elect to use a different picture when you download the firmware.  Here are captures of 2.2 and 2.1 splash editing screens and also 2.2 and 2.1 download windows.  These are all the same picture.
![capture](https://cloud.githubusercontent.com/assets/6879281/20467174/8b85bbe2-af4f-11e6-8c0d-bfcf82010698.JPG)
![capture1](https://cloud.githubusercontent.com/assets/6879281/20467175/8b8a4982-af4f-11e6-800f-328a01476a8d.JPG)

bsongis:
obviously the splash is converted to 1bit

jtaylor2:
The problem is that the taranis to x9d name change missed a couple of spots and the x9d splash images are being formatted for a 9x.  I will generate a pull request to fix it.

jtaylor2:
As @kilrah  pointed out fix does not cover X9E.  I will do another pull request to correct.

bsongis:
@jtaylor2 would you have a look to this PR:
https://github.com/opentx/opentx/pull/4067

PR #4064, opened 2016-11-22
Fix issue 4957 Taranis splash screens formatted incorrectly.

jtaylor2:
Fix issue 4957 Taranis splash screens formatted incorrectly (for 9x) due to missed taranis to x9d name changes.


bsongis:
Good catch!

bsongis:
Fixes #4057

kilrah:
This might not handle the X9E case... `(g.profile[g.id()].fwType().contains("x9"))` instead?

PR #4066, opened 2016-11-22
Fix Issue 4057 for X9E also.

jtaylor2:
Fix Issue 4057 Correct fix to support X9E also as suggested by @kilrah


bsongis:
Thanks!

PR #4067, opened 2016-11-22
[Companion] Some refactoring around the splash image handling

bsongis:


jtaylor2:
Tested splashes for several radio types in Companion and on X9D radio, all looked good.  Very nice, the old way didn't scale well to more radios with different LCD sizes.

bsongis:
@jtaylor2 thanks for your tests, and yes with Horus coming it will be easier to deal with different LCD size

PR #4078, opened 2016-11-28
fixed fq777-124 and assan protocol selection

bikemike:
Fixed protocol selection for FQ777-124 and Assan protocols. These were being skipped because the code assumed FRSKYV had an ID of 23 when it actually has an ID of 25 in the multiprotocol tx module code.


schwabe:
Thanks

PR #4079, opened 2016-11-28
changed default telemetry protocol for multimodule

bikemike:
Changed default telemetry protocol from FRSKY_D to SPORT to align with the PPM default telemetry type.

### DIFF
--- a/radio/src/telemetry/frsky.h
+++ b/radio/src/telemetry/frsky.h
@@ -561,11 +561,12 @@ void frskyUpdateCells(void);
   if (g_model.moduleData[INTERNAL_MODULE].rfProtocol == RF_PROTO_OFF && g_model.moduleData[EXTERNAL_MODULE].type == MODULE_TYPE_MULTIMODULE) {
     if (g_model.moduleData[EXTERNAL_MODULE].getMultiProtocol(false) == MM_RF_PROTO_DSM2)
       return PROTOCOL_SPEKTRUM;
-    else if ((g_model.moduleData[EXTERNAL_MODULE].getMultiProtocol(false) == MM_RF_PROTO_FRSKY) && (g_model.moduleData[EXTERNAL_MODULE].subType != 1))
-      // D8
-      return PROTOCOL_FRSKY_SPORT;
-    else
+    else if (((g_model.moduleData[EXTERNAL_MODULE].getMultiProtocol(false) == MM_RF_PROTO_FRSKY) && (g_model.moduleData[EXTERNAL_MODULE].subType == 1)) ||
+            (g_model.moduleData[EXTERNAL_MODULE].getMultiProtocol(false) == MM_RF_PROTO_HUBSAN))
+            //(g_model.moduleData[EXTERNAL_MODULE].getMultiProtocol(false) == MM_RF_PROTO_AFHDS2A)) // not added yet 
       return PROTOCOL_FRSKY_D;
+    else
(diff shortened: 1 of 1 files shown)


schwabe:
The multi module uses Frsky D as default in other telemetry types and midelic and hpnuts tend to add new telemetry as Frsky D (e.g. AFHDS 2A). Also Hubsan protocol sends Frsky D protocol.

bikemike:
Looking at the code, I only see two protocols that currently use FrskyD telemetry (FrskyX CH_8 and Hubsan). AFHDS2A also does but it hasn't been added to your branch of opentx yet.

Isn't FrskyD a legacy protocol? Is there any other reason why SPORT shouldn't be the default going forward? Perhaps I will ask in the DIY Multiprotocol TX Module thread.

kilrah:
Yes it is kind of legacy, but it would be up to multi to decide to change first, not us.

schwabe:
@bikemike AFHDS 2A sends Frsky D protocol unless the highest bit of the option field in which case the telemetry are forwarded raw. OpenTX implements direct parsing of the FlySky Telemetry. If FlySky would have been translated directly into nice sport telemetry I would probably have skipped implemeting FlySky telemetry in OpenTX.

MikeBland:
Since the Multi module is also used in transmitters running er9x and ersky9x, changing the Multi module doesn't only affect it's use with openTx.
Also, adding conversion code into the Multi module (with an Atmel ATMEGA328) would reduce the number of protocols that will fit in it.
Mike.

schwabe:
FlySky IBUS Telemetry is only in next (2.2) and not backported to 2.1. https://github.com/opentx/opentx/blob/next/radio/src/telemetry/flysky_ibus.cpp

bikemike:
Sounds like this solution isn't the best so I'll close this pull request. I posted another idea in the rcgroups forum that hpnuts seemed to like:

> Another option would be to allow overriding the default telemetry protocol when multi is selected. Perhaps that is a better solution as it would allow the multimodule devs to test new protocols without needing new builds of opentx/er9x/ersky9x with the telemetry type hard coded.

hpnuts replied with the following comment:

> My vote is for this option. In er9x/ersky9x you have to choose which telemetry protocol you want to be applied on the incoming stream.
> 
> For me, OpenTX should allow per model the selection of telemetry protocol (can be set by default to the right one if known). You should also be able to disable telemetry since I think I've understood that telemetry is always active on hubsan/dsm but if the model has poor telemetry TX capability, you keep getting messages telemetry lost/ok.

schwabe:
Disabling the alarm is a diferrent issue. I also thinking about a compile option for MULTI_MODULE that says "OPENTX_TELEMETRY" where the multi module just prefixes the telemetry by one byte that indicates the type of telemetry and also reports its own status as telemetry. But I cannot give you any promises when I came to implement that.

bikemike:
That sounds great. When you say the module could report its status, could the module advertise the protocols it supports(name/id/telem type)? If so you wouldn't need this information in opentx thus wouldn't need to create new builds of opentx each time a new protocol was added to the multi module.

MikeBland:
On ersky9x I now have an optional text file "Multi.txt" you put on the root of the SD card. This has the names of all the protocols and sub-protocols. To add a new protocol/sub-protocol, you just need to edit the .txt file. This allows you to only have the protocols display that are actually compiled into your Multi module.

Mike.

PR #4082, opened 2016-11-30
3djc/horus wizard plane

3djc:
More work will be done later to further optimize image handling (zones vs full screen)


radio/src/gui/480x272/model_select.cpp
```diff
@@ -177,7 +177,7 @@ bool menuModelWizard(event_t event)
   strncpy(wizpath, WIZARD_PATH, sizeof(WIZARD_PATH));
   strcpy(&wizpath[sizeof(WIZARD_PATH)-1], "/");
   lcdDrawSolidFilledRect(0, 0, LCD_W, LCD_H, TEXT_BGCOLOR);
-  lcd->drawBitmap(0, 0, BitmapBuffer::load(getThemePath("wizard/background.png")));
+  lcd->drawBitmap(0, 0,modelselWizardBackground);
```
bsongis: space missing, but yes it's better like that!

PR #4088, opened 2016-12-02
[SKY9X] Audio fix - Tested on 9XR-PRO

bsongis:


radio/src/audio_arm.h
```diff
@@ -116,6 +118,7 @@ enum AudioBufferState
 struct AudioBuffer {
   audio_data_t data[AUDIO_BUFFER_SIZE];
   uint16_t size;
+  uint8_t  state;
```
projectkk2glider: You could put `#if defined(PCBSKY9X)` around `state`.

bsongis: indeed, I will also change the #ifdef and introduce AUDIO_DOUBLE_BUFFERING rather than a board type

bsongis: as we have LCD_DUAL_BUFFER it will be AUDIO_DUAL_BUFFER

radio/src/audio_arm.h
```diff
@@ -314,38 +317,79 @@ class AudioBufferFifo {
...
     void freeNextFilledBuffer()
     {
       audioDisableIrq();
+#if defined(PCBSKY9X)
+      if (audioBuffers[readIdx].state == AUDIO_BUFFER_PLAYING) {
+        audioBuffers[readIdx].state = AUDIO_BUFFER_FREE;
+        readIdx = nextBufferIdx(readIdx);
+        bufferFull = false;
+      }
+#else
```
bsongis: I only ported the code from 'master' as I knew it works. I will try now to remove this code which gives me headaches

PR #4097, opened 2016-12-04
Projectkk2glider/issue 4086 usb and serial

### LINKED ISSUE #4086: X9D Crashing upon plug in while turned on for simulator
I keep getting my Taranis to crash while plugging into my pc to fly in DRL. I updated from 2.1.9 to 2.2.0N359 and it worked before. I checked all the drivers as far to my knowledge but with still no luck and almost always crashes when plugged into the pc. This is a new pc but have tried installing and uninstalling various drivers without any luck.

Any Idea?

projectkk2glider:
Closes #4086

I tested this on the Taranis in all three USB modes. Tests on other radios (X9E and Horus) have to be done before the merge.

The change is big, but the only way to solve the USB=SERIAL WDT resets was to use the parts of the newer ST USB library. Then I decided to port the whole library while preserving our changes and fixes.


schwabe:
On my Taranis+ Mac OS X is works as expected.

projectkk2glider:
Test on Horus:
 * MASSSTORAGE: works most of the time, but sometimes (after power-up or multiple connection cycles) it errors in two ways:
   * no response at all to USB cable
   * host can't assign address to the device

* SERIAL: works about 50% of the time after radio start. When it works it works without any problems even with multiple connects.

* JOYSTICK: worked every time (radio started 5 times), tested multiple connections, no problems.

projectkk2glider:
Interesting observation: looking at the number of USB interrupts in one second period I get in joystick mode:
 * connected to computer: ~1000 int/s
 * disconnected without cable: 0
 * cheap USB cable connected only to the radio, other end floating: 20 - 150 int/s Looks like cable is picking up the noise which is triggering something.

projectkk2glider:
Tested again in MASSTORAGE mode, without cheap cable connected only at the radio end. 5 starts without any problem. 

That cable must really be source of some noise. It will be nice to compare how Taranis behaves under same conditions. Perhaps Horus USB circuit is more sensitive to noise...

bsongis:
USB serial on X7D works perfectly

bsongis:
USB serial has a problem, the buffer is only sent once it is full, and it's very difficult to use!

projectkk2glider:
I had this problem once in the middle of the development, but it then disappeared. I will investigate this...

projectkk2glider:
What cmake option did you use. I can not reproduce on Taranis and Linux.

PR #4105, opened 2016-12-06
Fix ROT encoder nav for wizards

### LINKED ISSUE #4102: Taranis X9E OpenTX 2.2 RC8 New Model Wizard "Hangs"
The New Model Wizard "hangs" at the airplane icon and will not allow navigation to a different model type.
Only the setup for a new airplane can be completed using the wizard.

3djc:
Thix fixes #4102

### DIFF
--- a/radio/sdcard/taranis/SCRIPTS/WIZARD/delta.lua
+++ b/radio/sdcard/taranis/SCRIPTS/WIZARD/delta.lua
@@ -36,10 +36,10 @@ end
 
 local function fieldIncDec(event, value, max, force)
   if edit or force==true then
-    if event == EVT_PLUS_BREAK then
+    if event == EVT_PLUS_BREAK or event == EVT_ROT_LEFT then
       value = (value + max)
       dirty = true
-    elseif event == EVT_MINUS_BREAK then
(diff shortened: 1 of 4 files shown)


projectkk2glider:
There is a problem, the `EVT_ROT_LEFT` and other rotary related event constants are **only** defined in Lua for the radios that have the rotary encoder. Runing such script on a radio without it will in my opinion not work. We need to define these constant to something (some non-existing event) on such radios.

3djc:
I see what you mean,  I'll look into it

kilrah:
Radios with rotary encoders are supposed to emulate the plus and minus events so this should not be needed. It's rather that that mechanism that may have been broken.

3djc:
It looks like undefined constant have a nil value, and since event cannot have a nil value, it works on X9D,.. I tested that on simu, but I do not have a real radio to fully confirm

As far as rot encoder simulation of + and -, I have mixed feelings. It is example, I have mapped right to minus in the wizard.lua so that when you rotate clockwise, your selector moves right.

But in the plane,.. lua, it is mapped to plus, so that when you rotate clockwise, you increment values

This leads to a natural usage for an X9E that would not have been possible with emulation

PR #4106, opened 2016-12-06
Also bundle avrdude in addition to dfu-util and default to the execut…

schwabe:
…ables in the bundle


bsongis:
conflicts with your other commit :(

schwabe:
I will rebase and push again.

PR #4109, opened 2016-12-08
Fix for QT53

3djc:
Change is on line 250


bsongis:
Thanks
I see my editor doesn't remove white spaces on empty lines...

PR #4115, opened 2016-12-09
fix log playback when log frequency is .1s

bikemike:
The log playback is extremely slow if the log frequency is .1s. So slow that it doesn't look like the playback is even working.

This fixes the issue.

### DIFF
--- a/companion/src/simulation/telemetrysimu.cpp
+++ b/companion/src/simulation/telemetrysimu.cpp
@@ -659,7 +659,7 @@ void TelemetrySimulator::LogPlaybackController::calcLogFrequency()
     QDateTime logTime = parseTransmittterTimestamp(csvRecords[i]);
     // ugh - no timespan in this Qt version
     double timeDiff = (logTime.toMSecsSinceEpoch() - lastTime.toMSecsSinceEpoch()) / 1000.0;
-    if ((timeDiff > 0.1) && (timeDiff < logFrequency)) {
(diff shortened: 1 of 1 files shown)


companion/src/simulation/telemetrysimu.cpp
```diff
@@ -659,7 +659,7 @@ void TelemetrySimulator::LogPlaybackController::calcLogFrequency()
     QDateTime logTime = parseTransmittterTimestamp(csvRecords[i]);
     // ugh - no timespan in this Qt version
     double timeDiff = (logTime.toMSecsSinceEpoch() - lastTime.toMSecsSinceEpoch()) / 1000.0;
-    if ((timeDiff > 0.1) && (timeDiff < logFrequency)) {
+    if ((timeDiff >= 0.1) && (timeDiff < logFrequency)) {
```
projectkk2glider: Actually I would change this to `timeDiff > 0.09` because equal operator for floating point is basically useless.

projectkk2glider:
Btw, nice catch.

This must also be ported to `next` when accepted.

bikemike:
Ok, I've changed it to .09.

bsongis:
Thanks! Needs to be ported to 'next' indeed

projectkk2glider:
Ported to next in  4f47c13efccac8f60437e7ff93e7ec63a37566ce

PR #4116, opened 2016-12-09
Fixes #4103 Companion Special Functions parameters not cleared

### LINKED ISSUE #4103: A Minor Companion Bug in Special Functions
Go to **Special Functions** and enter a **Play Sound, Play Track,** or **Play Value** function.

Change the setting in the **Enable** column to something other than **No repeat**, i.e. a number of seconds, or **Play once, not during startup**. 

Bring up that model in **Print** or **Compare Models** and you'll see the **Special Function** with those settings, properly displayed.

Now change the function setting to something other than **Play Sound, Play Track,** or **Play Value**.

Bring up that model in **Print** or **Compare Models** and you'll see the **Enable** settings that belonged to the function previously entered, inappropriately attached to the new function. Example:

`SF1    SA↓ - Override CH1(0) played once, not during startup`

Workaround: If you set that function back to any one of the **Play** functions, set **Enable** to **No repeat**, and then set the function back to the new setting, the incorrectly displayed **Enable** text will no longer show up in **Print** or **Compare Models**.

elecpower:
Companion re-initialise a Special or Global Function's parameters when the selected function is changed #4103

### DIFF
--- a/companion/src/modeledit/customfunctions.cpp
+++ b/companion/src/modeledit/customfunctions.cpp
@@ -323,8 +323,11 @@ void CustomFunctionsPanel::functionEdited()
   if (!lock) {
     lock = true;
     int index = sender()->property("index").toInt();
-    fswtchParamArmT[index]->setCurrentIndex(0);
-    refreshCustomFunction(index, true);
+    RawSwitch swtch = functions[index].swtch;
+    functions[index].clear();
+    functions[index].swtch = swtch;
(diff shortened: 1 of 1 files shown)


3djc:
Tested successfully on the next port (https://github.com/opentx/opentx/pull/4118)

projectkk2glider:
Thanks.

kilrah:
Awesome! 👍

Issue #3318, opened 2016-02-24
Adding compile lua script to companion as option.


nigelsheffield:
Seeing as the compiled lua scripts take much less memory then compiling them each time they are run , could it be added as an option so that anyone could use this option without having to build an environment?
It then allows better use of limited available memory.

dsbeach:
I agree, and favor just removing the #ifdef LUA_COMPILER preprocessor directives.

If that is not acceptable, we should discuss alternative approaches to exposing this functionality.  I did a little investigation into what it would take to expose Lua in the companion UI and decided I was not up to the task.  Our choices are limited as long as the only Lua environment is in the radio simulators.

nigelsheffield:
As I understand it the fact that you overwrite the .lua file is considered a risk and the reason it was not done already.
I dont think it a problem because normally a user won't be using this function so won't have the .lua.SRC file on their system so no scripts will be overwritten.
To make the .SRC file requires that you know about the compiler option and so should also know to back up the .lua file before hand.

And I've not tried it yet but can the .lua.SRC file just be changed back to a .lua file should the worst happen? I can still open the .lua.SRC file to edit and so could just copy this text into another lua file if needed...

dsbeach:
Yes.  You can always revert the .lua.src file back to .lua provided that enough memory exists to compile and run in a single pass.  My GPS triangle scripts are too big to run on the radio in just source form.

lshems:
I always have two scripts available on my test environment:
file.lua  and  file.luab

the luab is the backup file, and always in plain text.
If i change the .lua file, I can always revert changes by copying the .luab to .lua

The .lua can also be a compiled version..

If I want to debug in companion, i need to run the .lua all as plain text versions, since otherwise important debug info on errors is lost. Therefore I created a batch program in CMD that copies the .lua to .luac, and then .luab to .lua. I can now debug properly.

If the changes in the .lua plain text files are tested, I delete the .luac so I know I need to compile it again.

after finishing debugging, I revert the action, copy .lua to luab (being the new version) and .luac to .lua

All I need to do now is compiling the .luab files that dont have a .lua version and save them as .lua. This last step is tedious and could easily be automated in my opinion in the companion.

Doing it like this, would prevent any possible errors, since if you dont create luab files, there will never be any compiling done, and if compilation is done, you will never lose the original.
And if you dont delete .lua files that have been compiled, the will never get overwritten either.

[...]

PR #4119, opened 2016-12-10
Add runtime Lua script pre-compilation and general .luac file support (#3318)

mpaperno:
Everything requested in #3318 and a bit more.  Please see individual commit notes for details.  This really enhances the Lua capabilities by allowing larger scripts to be loaded.  I have a version of this ready for 2.1 as well.

Please note I only have one Taranis+ for hardware testing.  Other models only tested in sims (including Horus).  Most 2.2 sim testing done on Linux Mint.  GNU Arm gcc 4.7 for firmware, gcc 4.9 for sim.

Thanks,
-Max


projectkk2glider:
Wow, nicely prepared. Looks good at the first glance. This will need a more thorough review, which hopefully I can do in next couple of days.

lshems:
Would be a good idea to backport to 2.1, since that is the latest stable.

hint hint

projectkk2glider:
Sure, but my plan is first to have this in 2.2 and well tested before porting it to 2.1.

mpaperno:
I agree, it should get a good workout with all the other testing of 2.2 going on.  Then port to 2.1 once we're pretty sure of no issues.  This new code is nearly identical between the versions, so it should be a good way to test.

Somewhat off-topic, but has anyone looked into/tried [LuaJIT](http://luajit.org) before?  Seems to promise some serious performance and memory improvements over stock Lua 5.1.  I just found it and am reading up on it now.

-Max

radio/src/CMakeLists.txt
```diff
@@ -171,6 +171,12 @@ endif()
 
 if(NOT LUA STREQUAL NO)
   add_definitions(-DLUA)
+  if(LUA_COMPILER)
+    add_definitions(-DLUA_COMPILER)
+  endif()
+  if(NOT "${LUA_SCRIPT_LOAD_MODE}" STREQUAL "")
+  	add_definitions(-DLUA_SCRIPT_LOAD_MODE="${LUA_SCRIPT_LOAD_MODE}")
```
bsongis: small indentation issue

radio/src/gui/480x272/radio_sdmanager.cpp
```diff
@@ -235,15 +236,15 @@ bool menuRadioSdManager(event_t _event)
           else if (isImageFileExtension(ext)) {
             TCHAR lfn[_MAX_LFN+1];
             f_getcwd(lfn, _MAX_LFN);
-            if (!READ_ONLY() && unsigned(LEN_FILE_EXTENSION+ext-line) <= sizeof(g_model.header.bitmap) && !strcmp(lfn, BITMAPS_PATH)) {
+            if (!READ_ONLY() && strlen(line) <= sizeof(g_model.header.bitmap) && !strcmp(lfn, BITMAPS_PATH)) {
```
bsongis: why? the previous code was correct and less CPU consuming, do I forget anything?

radio/src/gui/480x272/radio_sdmanager.cpp
```diff
@@ -347,15 +348,14 @@ bool menuRadioSdManager(event_t _event)
     LcdFlags attr = (index == i ? INVERS : 0);
     if (reusableBuffer.sdmanager.lines[i][0]) {
       if (s_editMode == EDIT_MODIFY_STRING && attr) {
-        editName(MENUS_MARGIN_LEFT, y, reusableBuffer.sdmanager.lines[i], SD_SCREEN_FILE_LENGTH-4, _event, attr, 0);
+        char * ext = getFileExtension(reusableBuffer.sdmanager.originalName, sizeof(reusableBuffer.sdmanager.originalName));
+        editName(MENUS_MARGIN_LEFT, y, reusableBuffer.sdmanager.lines[i], SD_SCREEN_FILE_LENGTH - (ext ? strlen(ext) : 0), _event, attr, 0);
```
bsongis: good catch

radio/src/gui/480x272/radio_sdmanager.cpp
```diff
@@ -347,15 +348,14 @@ bool menuRadioSdManager(event_t _event)
...
         if (s_editMode == 0) {
-          unsigned int len = effectiveLen(reusableBuffer.sdmanager.lines[i], SD_SCREEN_FILE_LENGTH-LEN_FILE_EXTENSION);
-          char * ext = getFileExtension(reusableBuffer.sdmanager.originalName, sizeof(reusableBuffer.sdmanager.originalName));
           if (ext) {
-            strAppend(&reusableBuffer.sdmanager.lines[i][len], ext);
+            strAppend(&reusableBuffer.sdmanager.lines[i][effectiveLen(reusableBuffer.sdmanager.lines[i], SD_SCREEN_FILE_LENGTH-strlen(ext))], ext);
```
bsongis: I prefer the code with len calculated before the if (ext), it's easier to read

radio/src/gui/common/stdlcd/radio_sdmanager.cpp
```diff
@@ -105,13 +105,14 @@ void onSdManagerMenu(const char * result)
...
       // write spaces to allow a longer filename
-      memset(ext, ' ', SD_SCREEN_FILE_LENGTH-LEN_FILE_EXTENSION-(ext-line));
-      line[SD_SCREEN_FILE_LENGTH-LEN_FILE_EXTENSION] = '\0';
+      memset(ext, ' ', SD_SCREEN_FILE_LENGTH-len-(ext-line));
+      line[SD_SCREEN_FILE_LENGTH-len] = '\0';
     }
     else {
-      int len = strlen(line);
-      memset(line + strlen(line), ' ', SD_SCREEN_FILE_LENGTH - len);
+      size_t len = strlen(line);
```
bsongis: yes the previous code was strange

radio/src/gui/common/stdlcd/radio_sdmanager.cpp
```diff
@@ -105,13 +105,14 @@ void onSdManagerMenu(const char * result)
     memcpy(reusableBuffer.sdmanager.originalName, line, sizeof(reusableBuffer.sdmanager.originalName));
     char * ext = getFileExtension(line, SD_SCREEN_FILE_LENGTH+1);
     if (ext) {
+      size_t len = strlen(ext);
```
bsongis: is it really needed to call `strlen` here? if not I will prefer the simple substraction (but with your code on 2 lines)

radio/src/gui/common/stdlcd/radio_sdmanager.cpp
```diff
@@ -277,7 +278,7 @@ void menuRadioSdManager(event_t _event)
             POPUP_MENU_ADD_ITEM(STR_VIEW_TEXT);
           }
 #if defined(LUA)
-          else if (!strcasecmp(ext, SCRIPTS_EXT)) {
+          else if (!strcasecmp(ext, SCRIPT_EXT) || !strcasecmp(ext, SCRIPT_BIN_EXT)) {
```
bsongis: Would you add a macro for this test? I think we have done it for images already

radio/src/gui/common/stdlcd/radio_sdmanager.cpp
```diff
@@ -402,15 +403,14 @@ void menuRadioSdManager(event_t _event)
...
       if (s_editMode == EDIT_MODIFY_STRING && attr) {
-        editName(lcdNextPos, y, reusableBuffer.sdmanager.lines[i], SD_SCREEN_FILE_LENGTH-4, _event, attr, 0);
+        char * ext = getFileExtension(reusableBuffer.sdmanager.originalName, sizeof(reusableBuffer.sdmanager.originalName));
+        editName(lcdNextPos, y, reusableBuffer.sdmanager.lines[i], SD_SCREEN_FILE_LENGTH - (ext ? strlen(ext) : 0), _event, attr, 0);
         if (s_editMode == 0) {
-          unsigned int len = effectiveLen(reusableBuffer.sdmanager.lines[i], SD_SCREEN_FILE_LENGTH-LEN_FILE_EXTENSION);
-          char * ext = getFileExtension(reusableBuffer.sdmanager.originalName, sizeof(reusableBuffer.sdmanager.originalName));
           if (ext) {
-            strAppend(&reusableBuffer.sdmanager.lines[i][len], ext);
+            strAppend(&reusableBuffer.sdmanager.lines[i][effectiveLen(reusableBuffer.sdmanager.lines[i], SD_SCREEN_FILE_LENGTH-strlen(ext))], ext);
```
bsongis: here 2 lines as well and if possible no strlen

radio/src/lua/api_general.cpp
```diff
@@ -1122,6 +1122,77 @@ static int luaGetRSSI(lua_State * L)
...
+  fun, err = loadScript("/SCRIPTS/FUNCTIONS/print.lua")
+  if (fun ~= nil) then
+     fun("Hello from loadScript()")
+  else
+     print(err)
+  end
+
+@status current Introduced in 2.1.10 (?)
+*/
+static int luaLoadScript(lua_State * L) {
```
bsongis: newline before `{`

radio/src/sdcard.cpp
```diff
@@ -40,21 +40,70 @@ const char * sdCheckAndCreateDirectory(const char * path)
...
+    eg; "splash.bmp" or "splash"
+  @param pattern Optional list of one or more file extensions concatenated together, including the period(s).
+    The list is searched backwards and the first match, if any, is returned.  If null, then only the actual filename
+    passed will be searched for.
+    eg: ".gif.jpg.jpeg.bmp"
+  @param exclDir true/false whether to allow directory matches (default true, excludes directories)
+  @param match Optional container to hold the matched file extension (wide enough to hold LEN_FILE_EXTENSION_MAX + 1).
+  @retval true if a file was found, false otherwise.
+*/
+static bool isFilePatternAvailable(const char * path, const char * file, const char * pattern = NULL, bool exclDir = true, char * match = NULL)
```
bsongis: why `static`

bsongis:
Good job, only cosmetics issues, thanks!

radio/src/gui/480x272/radio_sdmanager.cpp
mpaperno: Yea... the old code seemed strange... subtracting the ext. length and then adding it back...  But now I see it's a quicker way to get the total length... clever.  How about we add just strlen(ext)?  

`unsigned(strlen(ext) + ext - line)`

Or do you see some other clever way to get the total length of `line` w/out assuming the length of file extension?

radio/src/gui/480x272/radio_sdmanager.cpp
mpaperno: Value of `len` is now different based on condition.  But I can certainly split it back into two (or rather 4) lines by putting back the temp variable if you prefer (perhaps make it a smaller uint8 this time).

radio/src/gui/common/stdlcd/radio_sdmanager.cpp
mpaperno: Ultimately we need to find the strlen of the whole `line` including ext, regardless of if there is an ext. or not.  It's just quicker if we know the location of ext. (like before with `unsigned(strlen(ext) + ext - line)`),  vs like we have to do in the other conditional case with strlen(line). Maybe I'm missing something... 

However I have a couple other thoughts on this whole part... 

A) I realized that it still won't work with file extensions > 5 chars because of the limit now set in sdcard:getFileExtension()  (the LEN_FILE_EXTENSION_MAX).  Of course we could work around that by making a new param for getFileExtension() which defines max. length to bother searching (with default of the macro value). But it would still be some arbitrary value, or perhaps simply use the whole SD_SCREEN_FILE_LENGTH (and then again later while actually editing the name).

B) I've actually found it would be useful on some occasions to edit file extensions. On "wider" LCDs (like X9D/etc) maybe we could let the user just edit the whole file name? I think SD_SCREEN_FILE_LENGTH is 64 chars in 2.2 (32 in 2.1) while most files names being used in the system (img/sound/script) are limited to 6-10 chars anyway.  With added benefit of cutting out a couple dozen lines of code... and in fact make most of this discussion completely obsolete!  :-)

radio/src/gui/common/stdlcd/radio_sdmanager.cpp
mpaperno: Do you mean behind a LUA_COMPILER macro?  I could of course...  but you can actually execute .luac files weather or not the new features of LUA_COMPILER are active (IOW it would work right now in any OTx version).  It just passes the full file name to luaExec() and as long as the file exists, it will work whether it's binary or text.  So... up to you.

radio/src/sdcard.cpp
mpaperno: Really just marking it as "local" function to make it obvious it is not (currently) being used outside the file...  certainly doesn't have to be there, more of a style thing I guess.

mpaperno:
Thanks for the review and comments Bertrand, all good stuff.  Mostly that boils down to hiding file extensions during file rename... a dubious "feature" to begin with if you ask me.  :)

-Max

radio/src/gui/common/stdlcd/radio_sdmanager.cpp
mpaperno: Another thought here...  getFileExtension() already does a strlen(filename) on the file passed to it.  Doing it again in any way is redundant... we could add an optional param to getFileExtension() into which to place the full length of the filename with ext.  Seems obvious in retrospect...  perhaps this can save some time in other places as well.

Although my points A and B above still apply.

mpaperno:
```diff
--- a/radio/src/gui/common/stdlcd/radio_sdmanager.cpp
+++ b/radio/src/gui/common/stdlcd/radio_sdmanager.cpp
@@ -103,18 +103,11 @@ void onSdManagerMenu(const char * result)
   }
   else if (result == STR_RENAME_FILE) {
     memcpy(reusableBuffer.sdmanager.originalName, line, sizeof(reusableBuffer.sdmanager.originalName));
-    char * ext = getFileExtension(line, SD_SCREEN_FILE_LENGTH+1);
-    if (ext) {
-      size_t len = strlen(ext);
-      // write spaces to allow a longer filename
-      memset(ext, ' ', SD_SCREEN_FILE_LENGTH-len-(ext-line));
-      line[SD_SCREEN_FILE_LENGTH-len] = '\0';
-    }
-    else {
-      size_t len = strlen(line);
-      memset(line + len, ' ', SD_SCREEN_FILE_LENGTH - len);
-      line[SD_SCREEN_FILE_LENGTH] = '\0';
-    }
+    uint8_t fnlen = 0, extlen = 0;
+    getFileExtension(line, 0, LEN_FILE_EXTENSION_MAX, &fnlen, &extlen);
+    // write spaces to allow extending the length of a filename
+    memset(line + fnlen - extlen, ' ', SD_SCREEN_FILE_LENGTH - fnlen + extlen);
+    line[SD_SCREEN_FILE_LENGTH /* - extlen ? */] = '\0';
     s_editMode = EDIT_MODIFY_STRING;
     editNameCursorPos = 0;
   }
```

I'm not sure why the `-len` in the original: `line[SD_SCREEN_FILE_LENGTH-len] = '\0';` ?  Seems like it should always be just SD_SCREEN_FILE_LENGTH to ensure null terminator... maximum name length is enforced later anyway in editName().

mpaperno:
Speaking of strings and file names, why `#define _MAX_LFN      255` ?  Is anyone really expecting such long filenames on the SD card (and what would one do with them anyway)?

With the change in 2.2, every time you declare a FILINFO object that's 255B just for the filename, even if you never need it (the old way in 2.1 you only allocated the LFN space if needed, which was annoying but efficient).  I count 12 times FILINFO is allocated (not a pointer) outside the FatFs code itself (also from FatFs notes: "The working buffer occupies (_MAX_LFN + 1) * 2 bytes").  Also another 15 times _MAX_LFN is used to create temporary char[] buffers in OTx firmware code itself for dealing with files.  I realize these are (mostly) temporary allocations... but still.

-Max

bsongis:
I didn't review all the answers but your remark about _MAX_LFN is right
I wonder if 31 are not a little bit too short, but 63 are good for sure

bsongis:
What do you propose to rename files then? Show files extensions and allow the user to change them (with the ```.``` as well?

radio/src/gui/common/stdlcd/radio_sdmanager.cpp
bsongis: we have the function isImageFileExtension(const char * ext) for images and we could have the same helper for scripts extensions.

3djc:
We just need to ensure we are safe with the SD card browser (and possibly lua file access), because some people WILL have insane path on their SD

radio/src/gui/common/stdlcd/radio_sdmanager.cpp
bsongis: I think you could add a new function ```getFilenameInfo``` or something like this which does what you are describing above and the ```getFileExtension``` function could call it (this function is really simple and called in many locations)

mpaperno:
> What do you propose to rename files then? Show files extensions and allow the user to change them (with the . as well)?

Yes.

> We just need to ensure we are safe with the SD card browser (and possibly lua file access), because some people WILL have insane path on their SD

This would only affect the file name itself, not the full path tree. I also verified that FatFs will handle any overflow correctly.

Plus, any files that are named > SD_SCREEN_FILE_LENGTH (including .ext) do not show up in the SD browser at all anyway. 

> I wonder if 31 are not a little bit too short, but 63 are good for sure.

Probably should be at least SD_SCREEN_FILE_LENGTH.

Speaking of which, SD_SCREEN_FILE_LENGTH is too long right now, even for Horus and definitely Taranis.  I propose:

```diff
--- a/radio/src/opentx.h
+++ b/radio/src/opentx.h
@@ -1373,7 +1373,7 @@ void opentxResume();
 #endif
 
 // Re-useable byte array to save having multiple buffers
-#define SD_SCREEN_FILE_LENGTH (64)
+#define SD_SCREEN_FILE_LENGTH (LCD_COLS)
 union ReusableBuffer
 {
```
For X9D/7D we have `#define LCD_COLS (LCD_W/FW)` and on Horus `#define LCD_COLS 40`.

mpaperno:
I think this addresses all the comments so far.  

If the decision is made to allow editing file name with .ext, that can be a separate change.  Right now renaming files with extensions > 5 chars (including the period) will not remove the extension during edit, but should otherwise work.

lshems:
Just for info: the longest path I use will be something like:

/SCRIPTS/TELEM/someappx/somesubx

This structure allows to group app scripts and app sound files for easy maintenance.

Filenames could be long, as descriptive file names help when files contain basically one function at a time.

mpaperno:
_MAX_LFN discussion split to https://github.com/opentx/opentx/issues/4133

bsongis:
All comments major comments are addressed, just a small indentation issue somewhere which can be fixed directly in 'next'. Good job, thanks a lot for the help

radio/src/sdcard.cpp
```diff
@@ -194,15 +261,33 @@ bool sdListFiles(const char * path, const char * extension, const uint8_t maxlen
...
+
+      fnExt = getFileExtension(fno.fname, 0, 0, &fnLen, &extLen);
+      fnLen -= extLen;
+
+//      TRACE_DEBUG("listSdFiles(%s, %s, %u, %s, %u): fn='%s'; fnExt='%s'; match=%d\n",
+//           path, extension, maxlen, (selection ? selection : "nul"), flags, fno.fname, (fnExt ? fnExt : "nul"), (fnExt && isExtensionMatching(fnExt, extension)));
+
+      // file validation checks
+      if (!fnLen || fnLen > maxlen || (                                              // wrong size
+            fnExt && extension && (                                                  // extension-based checks follow...
```
mpaperno: I've yet to find a clean way in any language to nest complex `if` conditions... besides just not doing it.  But I kinda liked the indentation on this one, with the comments and all... :)  I'd like to see how you'd prefer it.

mpaperno:
Thanks for the merge!  This will simplify my next PR.  ;-)

I'll help support these changes, so please @mention me on any issues if I'm not obviously aware of them already.

bsongis:
You may open branches in the project now, I just added you to the commiters list, as I am not sure you have automatic build otherwise

kilrah:
On Windows I get 
`8>C:\opentx-22\code\radio\src\targets\simu\simpgmspace.cpp(313): error C3861: '_getcwd': identifier not found`

mpaperno:
Ouch... because I moved the includes down lower.  What about this?

```diff
@@ -310,7 +311,7 @@ void StartSimu(bool tests)
 
 #if defined(SDCARD)
   if (strlen(simuSdDirectory) == 0) {
-    getcwd(simuSdDirectory, 1024);
+    f_getcwd(simuSdDirectory, 1024);
   }
 #endif
```

Also just noticed the `if defined()` clauses checking for SD card need to cleaned up... there are [1](https://github.com/opentx/opentx/blob/next/radio/src/targets/simu/simpgmspace.cpp#L76), [2](https://github.com/opentx/opentx/blob/next/radio/src/targets/simu/simpgmspace.cpp#L311), [3](https://github.com/opentx/opentx/blob/next/radio/src/targets/simu/simpgmspace.cpp#L521) different versions in that one file.

kilrah:
Yep it's happy with `f_getcwd()`

mpaperno:
OK thanks, will fix asap...

mpaperno:
Fixed (hopefully) by #4144 .

mpaperno:
@bsongis [commented](#issuecomment-267754562)
> You may open branches in the project now, I just added you to the commiters list, as I am not sure you have automatic build otherwise

Thank you for the invite!  I'll do my best to keep up  :)

I do have the Travis build set up on my fork, and usually wait for that to finish before submitting PR.  Does it make much difference if I do PRs from my repo vs. this one?  I've noticed there are a lot of branches here already.

bsongis:
If your branches are inside the repo we easily see the Travis report before merging

athertop:
Guys, I have a project where I already compiled the code manually, and have the compiled luac files renamed to .lua and .lib - my main .lua file does various loadfile calls to load in the functions from my compiled .lib files and unloads the code when it no longer needs it, saving memory. I'm wondering how these compiled lua files will get on when run using the new compile functionality in 2.2rc9 ?

What I am wondering - is it possible to have your code determine first if the .lua file being loaded is already compiled and therefore skip the new compile functionality if that is the case? This will save me (and I imagine a few others) from having to redesign projects specific to running them on 2.2. I would rather have one version of code which runs on both 2.1 and 2.2 by my suggestion above being adopted. Additionally, would it be possible to have a compile tool added to Companion - or eve a standalone command line tool would be fine.

mpaperno:
It'll work just fine.  There's no problem when dumping already compiled bytecode.  It'll just create a .luac version which will be identical to your .lua original, and use that until the .lua is updated.  I agree, otherwise it would have broken a lot of previous workarounds, mine included.

A standalone compiler would be nice, but it needs to include at least stubs for all the custom Lua API which OpenTx provides.  And it will need to be maintained to be in sync with the "real" version.  Doable for sure, but would take some time.  Might be nice for pre-compiling all those Horus scripts for the SD card build.

A "fake" OpenTx Lua library of stubs could also be used in IDEs for code hinting/linting. 

-Max

athertop:
Max, Cheers for the response. If my main.lua does a loadfile on say another compiled lua file which is called say file2.lib, will this also work in the current rc9?

I presently run my compiled code quite happily on 2.1.9 (on the Taranis+) but it will not run in the 2.2 simulator. Is this normal?

mpaperno:
Yes, loadfile() doesn't care if it's compiled or not (nor has loadfile() function changed at all, that's built-in Lua).  You may however want to check out the new loadScript() API function which could potentially automate some of what you're doing (or not).

Generated bytecode is architecture-specific, so Lua code compiled in a 64-bit simulator won't run on the radio or in a 32-bit simulator.   I'm not entirely sure what part is generating the message you're referring to.  The pre-compiler should automatically detect the bytecode incompatibility and re-compile the source file to new bytecode (assuming it has the source, of course, not just bytecode in the .lua).

HTH,
-Max

mpaperno:
@athertop Is this like the error message you were referring to?
```
luaLoadScriptFileToState(/SCRIPTS/snake.lua, bt): loading /SCRIPTS/snake.luac
-E- luaLoadScriptFileToState(/SCRIPTS/snake.lua, bt): Error loading script: /SCRIPTS/snake.luac: incompatible precompiled chunk
        Retrying with /SCRIPTS/snake.lua
luaDumpState(/SCRIPTS/snake.luac): Saved bytecode to file.
```

athertop:
@mpaperno Not exactly. I think the conditions are a little different in my case. I will try to provide the detail:
 I am running a bytecode lua telemetry script which I had compiled on a v2.0 Companion simulator under Windows (it is a version of companion that I found shared somewhere on the web - the only version I have which will do the luac compilation). I created the bytecode version by taking my original text based main.lua file, putting it inside the model folder under the sim and renaming it to telem1.lua (as per 2.0 convention), then trying to open it in the sim (which incidentally displays an error) - this however still creates telem1.luac inside the model folder. I then take this luac file, rename it back to my original filename (main.lua) and put it on the physical SD card inside TELEMETRY folder and it runs fine on the Taranis Plus under 2.1.9. I try to run the same bytecode main.lua on the 2.1.9 sim inside Apple MacBook and I get the error:


If I try to run this same bytecode file inside the sim on 2.2-rc9 on the same MacBook I just get a similar message panel displaying: Script Syntax error

So not exactly the same situation you describe. I haven't tried in on 2.1rc9 on the radio directly yet.

kilrah:
There are changes in the lua API between 2.1 and 2.2.
I imagine the script uses something that causes an error on 2.2, but since it's precompiled you get this generic error instead of a more meaningful error with the affected line.

mpaperno:
> it is a version of companion that I found shared somewhere on the web

LOL.. You mean [this one](https://www.rcgroups.com/forums/showthread.php?2180477-LUA-scripting-Technical-discussion/page93#post32706124) I posted in 2015?  :)

Your Mac using 64-bit simulator will not run bytecode built on 32-bit Windows sim or the radio itself. Nothing new here, this is a known "issue" with luac  which is why they even have a special message for when it happens (the "syntax error" you're seeing).  ;-)  Though I'm afraid you lost me a bit on what runs where or even which version we're looking at in your screenshot.  But that's what the error means.  Please try it again w/out mixing "bitness" and report back.

-Max

PR #4125, opened 2016-12-10
Projectkk2glider/usb hardening

projectkk2glider:
Tested on Taranis and Horus in all three USB modes.


radio/src/targets/horus/board.cpp
```diff
@@ -175,6 +175,7 @@ void boardInit()
   init2MhzTimer();
   init1msTimer();
   usbInit();
+  usbDeInit();
```
bsongis: This needs a small comment ;)

bsongis: Would you add a small explanation in 'next' directly? (why Init followed directly by DeInit)
I will test your change now on X7
Thanks!

projectkk2glider: I introduced two new functions `usbStart()` and `usbStop()` to make it more clear. a4d899844e6ae5ef0967cae8fc02da562e4ccc1f

PR #4141, opened 2016-12-16
Fix static initialization order problem with Horus GUI

mpaperno:
Fixes static initialization order problem with Layout, Theme, & Widget factories.  I discovered this when building simu with MinGW-w64.  The initialization depended purely on compiler "luck."

See [What’s the “static initialization order fiasco”?](https://isocpp.org/wiki/faq/ctors#static-init-order).

For example compare the registration order on Linux vs. Windows (from simu startup log):

Linux g++4.9:
```
$ ./simulator22
register theme Darkblue
register theme Default
register layout Layout1x1
register layout Layout2P1
register layout Layout2x1
register layout Layout2x2
register layout Layout2x4
register widget Gauge
register widget ModelBmp
register widget Outputs
register widget Text
register widget Timer
register widget Value
registering "opentx-horus" simulator
```
Windows MinGW-w64 4.9 & 5.3
```
> simulator.exe
register widget Value
register widget Timer
register widget Text
register widget Outputs
register widget ModelBmp
register widget Gauge
register layout Layout2x4
register layout Layout2x2
register layout Layout2x1
register layout Layout2P1
register layout Layout1x1
register theme Default
register theme Darkblue
registering "opentx-horus" simulator
```
(Yes, the opposite... from same exact CMakeLists, compiler/linker options, and source code.)

Anyway, the Horus simulator would crash hard when trying to register the first widget because the global `std::list <const WidgetFactory *> registeredWidgets` hadn't been initialized yet.  Took me a while to figure that one out!

I've built this on Linux and Windows with several gcc versions, both firmware and simu.  But I have no actual Horus to test with.

As an aside, with some minor changes everything now builds and runs great using Qt5.5 or 5.7 with MinGW toolchains (and GNU Arm Embedded for the firmware, naturally).  The whole CMake project, including fw, can be configured, built, and edited from QtCreator with a stock Qt "kit" plus SDL and dirent downloads.  Coming up in a separate PR. 

-Max

### DIFF
--- a/radio/src/gui/480x272/layout.h
+++ b/radio/src/gui/480x272/layout.h
@@ -52,7 +52,7 @@ class Layout: public WidgetsContainer<MAX_LAYOUT_ZONES, MAX_LAYOUT_OPTIONS>
     const LayoutFactory * factory;
 };
 
-void registerLayout(const LayoutFactory * factory);
+extern void registerLayout(const LayoutFactory * factory);
 
 class LayoutFactory
 {
@@ -113,11 +113,12 @@ class BaseLayoutFactory: public LayoutFactory
     const ZoneOption * options;
 };
 
-#define MAX_REGISTERED_LAYOUTS 10
-
-extern unsigned int countRegisteredLayouts;
-extern const LayoutFactory * registeredLayouts[MAX_REGISTERED_LAYOUTS];
 Layout * loadLayout(const char * name, Layout::PersistentData * persistentData);
 void loadCustomScreens();
+
+#define MAX_REGISTERED_LAYOUTS          10
+extern unsigned int countRegisteredLayouts;
+#define registeredLayouts      getRegisteredLayouts()
+extern const LayoutFactory **  getRegisteredLayouts();
 
 #endif // _LAYOUT_H_
--- a/radio/src/gui/480x272/layout.h
+++ b/radio/src/gui/480x272/layout.h
@@ -52,7 +52,7 @@ class Layout: public WidgetsContainer<MAX_LAYOUT_ZONES, MAX_LAYOUT_OPTIONS>
     const LayoutFactory * factory;
 };
 
-extern void registerLayout(const LayoutFactory * factory);
+void registerLayout(const LayoutFactory * factory);
 
 class LayoutFactory
 {
@@ -118,7 +118,6 @@ void loadCustomScreens();
 
 #define MAX_REGISTERED_LAYOUTS          10
 extern unsigned int countRegisteredLayouts;
-#define registeredLayouts      getRegisteredLayouts()
-extern const LayoutFactory **  getRegisteredLayouts();
+const LayoutFactory **  getRegisteredLayouts();
 
 #endif // _LAYOUT_H_
--- a/radio/src/gui/480x272/layout.h
+++ b/radio/src/gui/480x272/layout.h
@@ -21,6 +21,7 @@
 #ifndef _LAYOUT_H_
 #define _LAYOUT_H_
 
+#include <list>
 #include "widgets_container.h"
 
 #define MAX_LAYOUT_ZONES               10
@@ -117,7 +118,6 @@ Layout * loadLayout(const char * name, Layout::PersistentData * persistentData);
 void loadCustomScreens();
 
 #define MAX_REGISTERED_LAYOUTS          10
-extern unsigned int countRegisteredLayouts;
-const LayoutFactory **  getRegisteredLayouts();
+std::list<const LayoutFactory *> & getRegisteredLayouts();
 
 #endif // _LAYOUT_H_
--- a/radio/src/gui/480x272/lcd.cpp
+++ b/radio/src/gui/480x272/lcd.cpp
@@ -282,7 +282,8 @@ void drawTimerMode(coord_t x, coord_t y, int32_t mode, LcdFlags att)
     if (mode < TMRMODE_COUNT) {
       lcdDrawTextAtIndex(x, y, STR_VTMRMODES, mode, att);
       return;
-    } else {
+    }
+    else {
       mode -= (TMRMODE_COUNT-1);
     }
   }
--- a/radio/src/gui/480x272/lcd.cpp
+++ b/radio/src/gui/480x272/lcd.cpp
@@ -279,10 +279,12 @@ void drawCurveName(coord_t x, coord_t y, int8_t idx, LcdFlags flags)
 void drawTimerMode(coord_t x, coord_t y, int32_t mode, LcdFlags att)
 {
   if (mode >= 0) {
-    if (mode < TMRMODE_COUNT)
+    if (mode < TMRMODE_COUNT) {
       lcdDrawTextAtIndex(x, y, STR_VTMRMODES, mode, att);
-    else
+      return;
+    } else {
       mode -= (TMRMODE_COUNT-1);
+    }
   }
   drawSwitch(x, y, mode, att);
 }
--- a/radio/src/gui/480x272/lcd.cpp
+++ b/radio/src/gui/480x272/lcd.cpp
@@ -280,7 +280,7 @@ void drawTimerMode(coord_t x, coord_t y, int32_t mode, LcdFlags att)
 {
   if (mode >= 0) {
     if (mode < TMRMODE_COUNT)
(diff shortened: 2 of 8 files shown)


radio/src/gui/480x272/lcd.cpp
```diff
@@ -280,7 +280,7 @@ void drawTimerMode(coord_t x, coord_t y, int32_t mode, LcdFlags att)
 {
   if (mode >= 0) {
     if (mode < TMRMODE_COUNT)
-      return lcdDrawTextAtIndex(x, y, STR_VTMRMODES, mode, att);
+      lcdDrawTextAtIndex(x, y, STR_VTMRMODES, mode, att);
```
projectkk2glider: I think return is needed here, it was just placed a little awkward.

radio/src/gui/480x272/theme.h
```diff
@@ -26,7 +26,7 @@ class BitmapBuffer;
 #define MAX_THEME_OPTIONS              5
 
 class Theme;
-void registerTheme(Theme * theme);
+extern void registerTheme(Theme * theme);
```
projectkk2glider: `extern` is not needed in my opinion. There are several other places with the same issue. http://stackoverflow.com/questions/11712707/extern-functions-in-c-vs-c

radio/src/gui/480x272/lcd.cpp
mpaperno: Ah it goes after the lcd call...  got it.

radio/src/gui/480x272/theme.h
mpaperno: OK... I've seen it both ways in the OTx code, it seems.  I'm with the 2nd answer on SO, reason # 3. :-D

radio/src/gui/480x272/layout.h
```diff
@@ -113,11 +113,12 @@ class BaseLayoutFactory: public LayoutFactory
...
-#define MAX_REGISTERED_LAYOUTS 10
-
-extern unsigned int countRegisteredLayouts;
-extern const LayoutFactory * registeredLayouts[MAX_REGISTERED_LAYOUTS];
 Layout * loadLayout(const char * name, Layout::PersistentData * persistentData);
 void loadCustomScreens();
 
+#define MAX_REGISTERED_LAYOUTS          10
+extern unsigned int countRegisteredLayouts;
+#define registeredLayouts      getRegisteredLayouts()
```
bsongis: is it possible you remove this #define, it's a little bit confusing, I would prefer we use your new function everywhere as a singleton

radio/src/gui/480x272/layout.h
```diff
@@ -113,11 +113,12 @@ class BaseLayoutFactory: public LayoutFactory
...
+extern const LayoutFactory **  getRegisteredLayouts();
```
bsongis: no extern here, we remove all of them little after little

radio/src/gui/480x272/lcd.cpp
bsongis: return is needed yes, I have nothing against 2 lines here, but with `{ }` in the else part if they are in the if part!

radio/src/gui/480x272/theme.h
```diff
@@ -91,13 +91,13 @@ inline const char * getThemePath(const char * filename)
...
-void registerTheme(Theme * theme);
-extern Theme * registeredThemes[MAX_REGISTERED_THEMES]; // TODO dynamic
-
 Theme * getTheme(const char * name);
 void loadTheme(Theme * theme);
 void loadTheme();
 
+#define MAX_REGISTERED_THEMES          10
+extern unsigned int countRegisteredThemes;
+#define registeredThemes    getRegisteredThemes()
```
bsongis: same remark here

radio/src/gui/480x272/theme.h
mpaperno: https://github.com/mpaperno/opentx/blob/e707d9c58e6adfd8762534a9ea98998e234329b4/radio/src/gui/480x272/menus.h#L243

Not many, though I count ~ 36  ;-)   I guess from "legacy" code as Bertrand points out.

radio/src/gui/480x272/lcd.cpp
mpaperno: Bertrand, so this is a good place to ask about style....

Are you OK with this?

``` c
if (simple)
  foo();

// or

if (simple)
  foo();
else
  bar():
```

Because I personally dislike the one-line versions `if (simple) foo();` and putting in the curly braces for very simple cases seems a bit verbose at times (especially in the first example).  I agree that if braces are used in one part of the condition (eg. the `if`) then they should be used for the other part(s) (the `else`), as you say.

bsongis: all examples you proposed are OK, I also don't like the one-line version.
also I prefer the brackets when there is a need for an comment like in this example:
if (test) {
 // I explain why I return here
 return ...
}

radio/src/gui/480x272/theme.h
bsongis: Yes I remember in er9x all functions were marked as extern until someone raised an issue, I started to remove them little after little at this time

radio/src/gui/480x272/layout.h
bsongis: well I think if the compilation is OK you will be good.
if you have 5 minutes more I would even ask you if you could use std::list in themes / layouts so that there is only the singleton which is public

radio/src/gui/480x272/lcd.cpp
```diff
@@ -279,10 +279,12 @@ void drawCurveName(coord_t x, coord_t y, int8_t idx, LcdFlags flags)
 void drawTimerMode(coord_t x, coord_t y, int32_t mode, LcdFlags att)
 {
   if (mode >= 0) {
-    if (mode < TMRMODE_COUNT)
-      return lcdDrawTextAtIndex(x, y, STR_VTMRMODES, mode, att);
-    else
+    if (mode < TMRMODE_COUNT) {
+      lcdDrawTextAtIndex(x, y, STR_VTMRMODES, mode, att);
+      return;
+    } else {
```
bsongis: ah newline before the else!

mpaperno: You like those even for short blocks, eg. 2-5 lines each?  Just checking...

kilrah: Yep always

bsongis: yes, I do prefer when the else is aligned with the if

radio/src/gui/480x272/lcd.cpp
kilrah: I have seen @bsongis add the brackets even for single-line blocks multiple times (even when there's no comment) so I assume it's "always" now :)

mpaperno: Got it  :)

mpaperno:
@bsongis https://github.com/opentx/opentx/pull/4141#discussion_r92903027
> well I think if the compilation is OK you will be good.
if you have 5 minutes more I would even ask you if you could use std::list in themes / layouts so that there is only the singleton which is public

Haha, 5 minutes... right!  :)  But I did it anyway.  

I left the instance limits in place at least for now, and actually added a limit for Widget instances.  I'm not sure this is necessary or what the limits should be, but seemed like some kind of sanity check might be appropriate (esp. with ability to add via Lua).

Also the little arrows on the widget selectors now disappear when you're at the start/end of the list to indicate there are no more in that direction.  I thought about rotating back to the first/last widget, but that may be more confusing to the user.

-Max

radio/src/gui/480x272/layout.h
```diff
@@ -113,11 +114,10 @@ class BaseLayoutFactory: public LayoutFactory
...
 };
 
-#define MAX_REGISTERED_LAYOUTS 10
-
-extern unsigned int countRegisteredLayouts;
-extern const LayoutFactory * registeredLayouts[MAX_REGISTERED_LAYOUTS];
 Layout * loadLayout(const char * name, Layout::PersistentData * persistentData);
 void loadCustomScreens();
 
+#define MAX_REGISTERED_LAYOUTS          10
```
bsongis: Can't this #define be removed now?

bsongis:
All comments adressed, I merge it

bsongis:
I just removed the arbitrary limits, if any issue we will revert my commit

PR #4142, opened 2016-12-16
Minor build fixes

mpaperno:
Minor syntax errors and compiler warnings.

### DIFF
--- a/radio/src/gui/128x64/model_setup.cpp
+++ b/radio/src/gui/128x64/model_setup.cpp
@@ -1175,7 +1175,7 @@ void menuModelFailsafe(event_t event)
 #if defined(PPM_UNIT_PERCENT_PREC1)
       uint8_t wbar = SLIDER_W-6;
 #else
-      uint8_t wbar = SLIDER_W);
+      uint8_t wbar = SLIDER_W;
 #endif
 
       uint8_t xValue = x+LCD_W-4-wbar;
--- a/radio/src/gui/480x272/lcd.h
+++ b/radio/src/gui/480x272/lcd.h
@@ -133,8 +133,8 @@ void lcdDrawNumber(coord_t x, coord_t y, int32_t val, LcdFlags flags=0, uint8_t 
 
 void putsModelName(coord_t x, coord_t y, char *name, uint8_t id, LcdFlags att);
 void putsStickName(coord_t x, coord_t y, uint8_t idx, LcdFlags att=0);
-void drawSwitch(coord_t x, coord_t y, int32_t swtch, LcdFlags flags=0);
-void drawSource(coord_t x, coord_t y, uint32_t idx, LcdFlags att=0);
+void drawSwitch(coord_t x, coord_t y, swsrc_t swtch, LcdFlags flags=0);
+void drawSource(coord_t x, coord_t y, mixsrc_t idx, LcdFlags att=0);
 void drawCurveName(coord_t x, coord_t y, int8_t idx, LcdFlags att=0);
 void drawTimerMode(coord_t x, coord_t y, int32_t mode, LcdFlags att=0);
 void drawTrimMode(coord_t x, coord_t y, uint8_t phase, uint8_t idx, LcdFlags att);
--- a/radio/src/gui/480x272/lcd.h
+++ b/radio/src/gui/480x272/lcd.h
@@ -133,7 +133,7 @@ void lcdDrawNumber(coord_t x, coord_t y, int32_t val, LcdFlags flags=0, uint8_t 
 
 void putsModelName(coord_t x, coord_t y, char *name, uint8_t id, LcdFlags att);
 void putsStickName(coord_t x, coord_t y, uint8_t idx, LcdFlags att=0);
-void drawSwitch(coord_t x, coord_t y, swsrc_t swtch, LcdFlags flags=0);
+void drawSwitch(coord_t x, coord_t y, int32_t swtch, LcdFlags flags=0);
 void drawSource(coord_t x, coord_t y, uint32_t idx, LcdFlags att=0);
 void drawCurveName(coord_t x, coord_t y, int8_t idx, LcdFlags att=0);
 void drawTimerMode(coord_t x, coord_t y, int32_t mode, LcdFlags att=0);
--- a/radio/src/opentx.cpp
+++ b/radio/src/opentx.cpp
@@ -2572,7 +2572,7 @@ int main()
   // lcdSetRefVolt(25);
 #endif
 
-#if defined(PCBTARANIS) || defined(PCBHORUS)
+#if defined(SPLASH) && (defined(PCBTARANIS) || defined(PCBHORUS))
   drawSplash();
 #endif
 
--- a/radio/src/targets/taranis/diskio.cpp
+++ b/radio/src/targets/taranis/diskio.cpp
@@ -978,7 +978,7 @@ void sdPoll10ms()
 
 FATFS g_FATFS_Obj;
 #if defined(LOG_TELEMETRY)
-FIL g_telemetryFile;
+FIL g_telemetryFile = {};
 #endif
 
 #if defined(BOOT)
@@ -1005,7 +1005,6 @@ void sdInit(void)
     sdGetFreeSectors();
     
 #if defined(LOG_TELEMETRY)
-    memset(&g_telemetryFile, 0, sizeof(FIL));
     f_open(&g_telemetryFile, LOGS_PATH "/telemetry.log", FA_OPEN_ALWAYS | FA_WRITE);
     if (f_size(&g_telemetryFile) > 0) {
       f_lseek(&g_telemetryFile, f_size(&g_telemetryFile)); // append
--- a/radio/src/targets/taranis/diskio.cpp
+++ b/radio/src/targets/taranis/diskio.cpp
@@ -978,7 +978,7 @@ void sdPoll10ms()
 
 FATFS g_FATFS_Obj;
 #if defined(LOG_TELEMETRY)
-FIL g_telemetryFile = {0};
+FIL g_telemetryFile;
 #endif
 
 #if defined(BOOT)
@@ -1005,6 +1005,7 @@ void sdInit(void)
     sdGetFreeSectors();
     
 #if defined(LOG_TELEMETRY)
+    memset(&g_telemetryFile, 0, sizeof(FIL));
     f_open(&g_telemetryFile, LOGS_PATH "/telemetry.log", FA_OPEN_ALWAYS | FA_WRITE);
     if (f_size(&g_telemetryFile) > 0) {
       f_lseek(&g_telemetryFile, f_size(&g_telemetryFile)); // append
--- a/radio/src/gui/480x272/layout.h
+++ b/radio/src/gui/480x272/layout.h
@@ -22,6 +22,7 @@
 #define _LAYOUT_H_
 
 #include "widgets_container.h"
+#include "lcd.h"
 
 #define MAX_LAYOUT_ZONES               10
 #define MAX_LAYOUT_OPTIONS             10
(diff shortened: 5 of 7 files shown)


radio/src/gui/480x272/lcd.h
```diff
@@ -133,7 +133,7 @@ void lcdDrawNumber(coord_t x, coord_t y, int32_t val, LcdFlags flags=0, uint8_t
 
 void putsModelName(coord_t x, coord_t y, char *name, uint8_t id, LcdFlags att);
 void putsStickName(coord_t x, coord_t y, uint8_t idx, LcdFlags att=0);
-void drawSwitch(coord_t x, coord_t y, swsrc_t swtch, LcdFlags flags=0);
+void drawSwitch(coord_t x, coord_t y, int32_t swtch, LcdFlags flags=0);
```
projectkk2glider: I do not like this? Why is this change needed in you opinion?

radio/src/targets/taranis/diskio.cpp
```diff
@@ -1005,6 +1005,7 @@ void sdInit(void)
     sdGetFreeSectors();
     
 #if defined(LOG_TELEMETRY)
+    memset(&g_telemetryFile, 0, sizeof(FIL));
```
projectkk2glider: Same fix is needed for Horus target.

radio/src/gui/480x272/lcd.h
mpaperno: The compiler can't find the definition of swsrc_t when building this way, presumably due to some bizarre include order mixup in the amalgamated source file (? even if openth.h is included at the top of lcd.h... I don't really get it either, but it's hard to follow what depends on what due to the reliance on amalgamation in the end.  My IDE has a hard time following all the links, often can't, which I find disturbing).  I noticed it's the same way (int32) in 212x64 (but not 128x64).

radio/src/gui/128x64/model_setup.cpp
```diff
@@ -1175,7 +1175,7 @@ void menuModelFailsafe(event_t event)
 #if defined(PPM_UNIT_PERCENT_PREC1)
       uint8_t wbar = SLIDER_W-6;
 #else
-      uint8_t wbar = SLIDER_W);
```
bsongis: ouch

radio/src/gui/480x272/lcd.h
bsongis: if the `drawThumb` method goes in a layout.c does it fix the issue? it would avoid to include lcd.h in layout.h

radio/src/opentx.cpp
```diff
@@ -2572,7 +2572,7 @@ int main()
   // lcdSetRefVolt(25);
 #endif
 
-#if defined(PCBTARANIS) || defined(PCBHORUS)
+#if defined(SPLASH) && (defined(PCBTARANIS) || defined(PCBHORUS))
```
bsongis: why? I thought I disabled the possibility of removing the SPLASH on those targets because the initialization was long (the audio task searches for a lot of files), I may be wrong though it's a long time I did it

radio/src/targets/taranis/diskio.cpp
```diff
@@ -978,7 +978,7 @@ void sdPoll10ms()
 
 FATFS g_FATFS_Obj;
 #if defined(LOG_TELEMETRY)
-FIL g_telemetryFile = {0};
+FIL g_telemetryFile;
```
bsongis: why?

radio/src/opentx.cpp
mpaperno: Otherwise link error with Horus build:
```
CMakeFiles\opentx-horus-simulator.dir/objects.a(opentx.cpp.obj): In function `Z8simuMainPv':

radio/src/opentx.cpp:2576: undefined reference to `drawSplash()'
```
https://github.com/mpaperno/opentx/blob/e707d9c58e6adfd8762534a9ea98998e234329b4/radio/src/gui/480x272/splash.cpp#L24

radio/src/targets/taranis/diskio.cpp
mpaperno: > radio\src\targets\horus\diskio.cpp:313:25: warning: missing braces around initializer for '_FDID' [-Wmissing-braces]

Anyway seemed like a "strange" way to init a struct.

And yes, Horus diskio also, same warning.  Only happens when building fw. with LOG_TELEMETRY defined.

radio/src/gui/480x272/lcd.h
mpaperno: Doesn't seem like it's going to work as-is.  I followed the recommendations in [How can I avoid linker errors with my template classes?](https://isocpp.org/wiki/faq/templates#separate-template-class-defn-from-decl) (scroll up a couple of items for explanation of the base problem with templates). 

```cpp
// layout.cpp
77: template<class T>
78: void BaseLayoutFactory<T>::drawThumb(uint16_t x, uint16_t y, uint32_t flags) const
79: {
80:   lcdDrawBitmapPattern(x, y, bitmap, flags);
81: }
82: 
83: template class BaseLayoutFactory<Layout>;
```

But still get errors down the line due to the virtual functions:

```
In file included from radio\src\gui\480x272\layout.cpp:21:0:

radio\src\gui\480x272\layout.h: In instantiation of 'Layout* BaseLayoutFactory<T>::create(WidgetsContainer<10, 10>::PersistentData*) const [with T = Layout]':

radio\src\gui\480x272\layout.cpp:83:16:   required from here

radio\src\gui\480x272\layout.h:95:51: error: cannot allocate an object of abstract type 'Layout'

radio\src\gui\480x272\layout.h:31:7: note:   because the following virtual functions are pure within 'Layout':

In file included from radio\src\gui\480x272\layout.h:24:0,

                 from radio\src\gui\480x272\layout.cpp:21:

radio\src\gui\480x272\widgets_container.h:124:26: note: 	unsigned int WidgetsContainer<N, O>::getZonesCount() const [with int N = 10; int O = 10]

radio\src\gui\480x272\widgets_container.h:126:18: note: 	Zone WidgetsContainer<N, O>::getZone(unsigned int) const [with int N = 10; int O = 10]

In file included from radio\src\gui\480x272\layout.cpp:21:0:

D:\Devel\RC-Tx\OpenTx\2.2\radio\src\gui\480x272\layout.h: In instantiation of 'Layout* BaseLayoutFactory<T>::load(WidgetsContainer<10, 10>::PersistentData*) const [with T = Layout]':

D:\Devel\RC-Tx\OpenTx\2.2\radio\src\gui\480x272\layout.cpp:83:16:   required from here

D:\Devel\RC-Tx\OpenTx\2.2\radio\src\gui\480x272\layout.h:102:51: error: cannot allocate an object of abstract type 'Layout'

D:\Devel\RC-Tx\OpenTx\2.2\radio\src\gui\480x272\layout.h:31:7: note:   since type 'Layout' has pure virtual functions
```

mpaperno: I can't seem to recreate the original warning anymore (and don't remember exactly which combo of options/compilers/targets caused it to begin with).  So if you guys would prefer to go with the original extern declaration, I can just revert this commit.

Should then also maybe change, for consistency:
```diff
--- a/radio/src/gui/480x272/lcd.h  (and in 212x64/lcd.h)
+++ b/radio/src/gui/480x272/lcd.h
@@ -137,1 +137,1 @@
- void drawSource(coord_t x, coord_t y, uint32_t idx, LcdFlags att=0);
+ void drawSource(coord_t x, coord_t y, mixsrc_t idx, LcdFlags att=0);
```

radio/src/targets/taranis/diskio.cpp
mpaperno: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=53119

As I understand it, the "proper" initializer in C++ should be `= {}` (although this is fixed in g++5).  And indeed that silences the warning.  I'll change both instances.

radio/src/gui/480x272/lcd.h
bsongis: yes mixsrc_t here seems better

bsongis: and yes you are right, now I remember why I put this function as extern there :)

radio/src/opentx.cpp
bsongis: ah perhaps I did something wrong with my latest commit? I introduced SPLASH=FRSKY for X7 / X9

mpaperno: I'm almost positive that I noticed the error before I merged the latest `next` commits which included your change.  I remember noticing your commit note and wondering if perhaps that had fixed the issue.  Then again it's been a long week...  :)

radio/src/gui/480x272/lcd.h
mpaperno: And you (or someone) may remember it again if the definition changes... :)

I tried a few things... the error in lcd.h happens because layout.h is included from datastructs.h.  So one way around it might be to separate LayoutFactory into its own file and include lcd.h from there.

Reverted for now and "fixed" `mixsrc_t`.

mpaperno: Guys, I finally ran into the original problem I had with `lcdDrawBitmapPattern()` in `drawThumb()`.

Try a Linux build of `next` with 
```
cmake -DPCB=HORUS -DCMAKE_BUILD_TYPE=Release ...
make simu
```  
the key ingredient being "Release" type.  You should see some warnings (which are not present unless Release is specified), with a series of errors at the end from each layout subclass:
```make
Linking CXX executable ../../../../simu
CMakeFiles/simu.dir/__/__/gui/480x272/layouts/layout1x1.cpp.o: In function `BaseLayoutFactory<Layout1x1>::drawThumb(unsigned short, unsigned short, unsigned int) const':
layout1x1.cpp:(.text._ZNK17BaseLayoutFactoryI9Layout1x1E9drawThumbEttj[_ZNK17BaseLayoutFactoryI9Layout1x1E9drawThumbEttj]+0x15): undefined reference to `lcdDrawBitmapPattern(int, int, unsigned char const*, unsigned int, int, int)'
CMakeFiles/simu.dir/__/__/gui/480x272/layouts/layout2+1.cpp.o: In function `BaseLayoutFactory<Layout2P1>::drawThumb(unsigned short, unsigned short, unsigned int) const':
layout2+1.cpp:(.text._ZNK17BaseLayoutFactoryI9Layout2P1E9drawThumbEttj[_ZNK17BaseLayoutFactoryI9Layout2P1E9drawThumbEttj]+0x15): undefined reference to `lcdDrawBitmapPattern(int, int, unsigned char const*, unsigned int, int, int)'
...
collect2: error: ld returned 1 exit status
```

When Release is specified, CMake applies its default CMAKE_CXX_FLAGS_RELEASE which includes `-O3`.  Otherwise no optimization level is specified at all, except with Debug type when `-O0` is specifically applied in our CMakeTexts.

So with `O3` the inline `lcdDrawBitmapPattern()` is optimized out and no longer available as a reference.

In the Travis commit tests `-DCMAKE_BUILD_TYPE=Debug` is specified which is why neither the warnings nor error show up.

Interestingly, when you `make libsimulator` with a Release type, there is no linker error at the end, even though `O3` is still used.  Something else must make gcc keep the function externally linkable.  You still see the same warnings though.

mpaperno: `simu` also doesn't link with `-O2` or `-O1`, only with `-O[0|s|g]`.

projectkk2glider: Why isn't `lcdDrawBitmapPattern()` inlined into `layout1x1.cpp` like it is supposed to be, then the function `lcd->drawBitmapPattern()` would be linked!?

projectkk2glider: Found it:
```
    virtual void drawThumb(uint16_t x, uint16_t y, uint32_t flags) const
    {
      extern void lcdDrawBitmapPattern(int x, int y, const uint8_t * bitmap, uint32_t flags, int width=0, int offset=0);
      lcdDrawBitmapPattern(x, y, bitmap, flags);
    }
```
`lcdDrawBitmapPattern()` is defined extern but is really an inline function.

projectkk2glider: I have solution for this, I will submit PR shortly.

projectkk2glider: See #4188

PR #4144, opened 2016-12-17
[simu] Fix Windows builds with SD-enabled targets; 

mpaperno:
Centralize SD check condition. Also make some compiler checks more specific (eg. !minGW).


bsongis:
Don't hesitate to open branches in the opentx project so that your commits will be automatically checked, we are used to name our branches 'username/feature'

PR #4146, opened 2016-12-17
Horus outputs widget display channel name when available 4145

joaquins:


joaquins:
Not the intended branch, closing.

Issue #4145, opened 2016-12-17
Horus outputs widget: display channel name when available


joaquins:
The "outputs" widget in the Horus displays the channel number.

When a channel is assigned a name in the model output settings, this name should be displayed instead.

Rationale for this request: Specially useful when having lots of models with different channel assignment (TAER, AETR...), and labeling the channel output helps when checking the outputs widget.

NOTE: A pull request with this change will be provided by me following this issue request.

projectkk2glider:
Done

PR #4147, opened 2016-12-17
Horus outputs widget: display channel name when available (issue #4145)

joaquins:


radio/src/gui/480x272/widgets/outputs.cpp
```diff
@@ -45,7 +45,11 @@ class OutputsWidget: public Widget
         int16_t chanVal = calcRESXto100(channelOutputs[curChan-1]);
         strAppend(chanString, "CH");
         strAppendSigned(&chanString[2], curChan, 2);
-        lcdDrawText(x, y + (curChan - firstChan) * row_height + 1, chanString, SMLSIZE | TEXT_COLOR | LEFT);
+        if ( g_model.limitData[curChan - 1].name[0] != 0 )  {
```
bsongis: no space after `(` and before `)`

radio/src/gui/480x272/widgets/outputs.cpp
```diff
@@ -45,7 +45,11 @@ class OutputsWidget: public Widget
...
+          lcdDrawSizedText(x, y + (curChan - firstChan) * row_height + 1, g_model.limitData[curChan - 1].name, sizeof(g_model.limitData[curChan - 1].name), SMLSIZE | TEXT_COLOR | LEFT | ZCHAR);
+        } else {
```
bsongis: newline before `else`

3djc:
It does look weird since a different font size is used for name and CHxx, and also name overlaps with channel value at -100

joaquins:
Ok, made an update following the same approach as in the failsafe custom set screen: both channel and names are showed in  2-lines when name is set (see screenshot).

Let me know what you think.

widgets2


failsafe

3djc:
Hum, quite hard to read IMHO. Here is a proposal inspired by your latest proposal ; you keep CH1.. like the origin one, and you write the name of the chan in the side that is not used, Would that be better ?

image

joaquins:
Agree, it's more clear.

What about this other approach, to make it similar to the monitors screen?

outputs_single

monitors

3djc:
Well, if you have no names (or just 1), then it looks a bit odd. Dunno what the others think

kilrah:
The latest version could be good, maybe just drop the "CH" when there's a name, so "03 Thr"...

kilrah:
Or display the CH and name alternatively every second or 2, we have that done in a couple of places.

mpaperno:
IMHO the text inside the progress bars shouldn't switch sides... it will be distracting and hard to find quickly (esp. around center... will constantly jump back/forth? although I guess you could set min. pixels drawn before the side changes...).  Can it be outlined or on its own background color, or something, so still readable on either background?

Actually the text already gets lost when on a dark BG, like on top of the dude in the wallpaper... guess the whole widget could use a BG color setting?

> ... maybe just drop the "CH" when there's a name, so "03 Thr"...

Yea the "CH" seems redundant when a name is there.

joaquins:
I need to try Kilrah's idea, showing alternatively name/ch. That would avoid the values switching sides, if it fits in the left as previously.

mhotar:
mpaperno is right, you can consider to make simle table with bars under the text.
chbars

mhotar:
chbars

joaquins:
I liked the last proposal from @mhotar.

This is an actual screenshot of the last commit:

background

The white background color is coming from TEXT_BGCOLOR, so it's configurable in the theme.

mhotar:
Great job!
There is BARGRAPH1_COLOR and BARGRAPH_BGCOLOR you can use instead of the TEXT_BGCOLOR to be fine with themes.

radio/src/gui/480x272/widgets/outputs.cpp
```diff
@@ -36,18 +36,14 @@ class OutputsWidget: public Widget
 
     uint8_t drawChannels(uint16_t x, uint16_t y, uint16_t w, uint16_t h, uint8_t firstChan)
     {
-      char chanString[] = "CH32";
+      char chanString[] = "CH32 123456";
```
projectkk2glider: ~~Looks like we had a buffer overflow in `char chanString[]` before.~~ my bad, it was OK.

The new size is actually too big now. This line can be changed back to the original.

radio/src/gui/480x272/widgets/outputs.cpp
```diff
@@ -57,8 +53,18 @@ class OutputsWidget: public Widget
...
         lcd->drawSolidVerticalLine(x + RECT_OFFSET + RECT_WIDTH / 2, y + (curChan - firstChan) * row_height, row_height, MAINVIEW_GRAPHICS_COLOR);
+        lcdDrawNumber(x + RECT_WIDTH - 10, y + (curChan - firstChan) * row_height + 1, chanVal, SMLSIZE | TEXT_COLOR | RIGHT, 0, NULL, "%");
+        if (g_model.limitData[curChan - 1].name[0] != 0)  {
+          strAppendSigned(chanString, curChan, 2);
+          lcdDrawText(x + 2, y + (curChan - firstChan) * row_height + 1, chanString, SMLSIZE | TEXT_COLOR | LEFT);
+          lcdDrawSizedText(x + 25, y + (curChan - firstChan) * row_height + 1, g_model.limitData[curChan - 1].name, sizeof(g_model.limitData[curChan - 1].name), SMLSIZE | TEXT_COLOR | LEFT | ZCHAR);
+      } 
+        else {
+          strAppend(chanString, "CH");
+          strAppendSigned(&chanString[2], curChan, 2);
```
projectkk2glider: You are relying on first two chars of `chanString` to be `CH`, but you could already overwritten that on line 59!

radio/src/gui/480x272/widgets/outputs.cpp
joaquins: Good catch!

joaquins: Done in commit https://github.com/opentx/opentx/pull/4147/commits/e5c8cbe0cc422fee1709c93add8990e6052248f7

joaquins:
Screenshot with latest commit, using BARGRAPH_BGCOLOR

barbgcolor

3djc:
Just a small detail, there is one pixel between bars that can produce strange visual effect ,maybe you want to remove that either by increasing bar height, or drawing a background rectangle, but it is looking very nice

joaquins:
With added borders:

borders

3djc:
Nice !

mpaperno:
Looks great!

kilrah:
I quite liked the transparency but both are good :)

mhotar:
Sure, both are nice, but transparent one is better a little :)

joaquins:
Umm, the failed check seems to be unrelated to this pull request. Is there a way to force a re-check?

Also, the requested change is done. Only the original requestor, @projectkk2glider can mark it as done?

mpaperno:
It is almost impossible to please everyone when it comes to design, which is when I find options can be useful.  I'm not sure, is it difficult to add a option to draw a background?  Maybe even BG color choice (or transparent)?

I don't know of a way to force a build on Travis except by pushing another commit. :(

joaquins:
@mpaperno IMHO increasing the code complexity for just a minor cosmetic choice is not worth it.

I'm fine with both approaches (transparent/box), but in my experience, I can see people complaining more about the transparent...

Anyway, a decision should be made... how do you guys deal with these cases?

mpaperno:
Adding one or two lines to the options[] array doesn't seem overly complex... ?

```
const ZoneOption OutputsWidget::options[] = {
 ...
  { "Fill Background?", ZoneOption::Bool, OPTION_VALUE_BOOL(false) },
  { "  BG Color", ZoneOption::Color, OPTION_VALUE_UNSIGNED(LIGHTGREY) },
}
```

projectkk2glider:
You can restart the failed build on Travis if you first log-in with your github account there.

projectkk2glider:
screenshot

radio/src/gui/480x272/widgets/outputs.cpp
```diff
@@ -43,42 +43,54 @@ class OutputsWidget: public Widget
...
+      drawChannels(zone.x, zone.y, zone.w, zone.h, persistentData->options[0].unsignedValue, persistentData->options[1].boolValue, persistentData->options[2].unsignedValue);
     }
 
     static const ZoneOption options[];
 };
 
 const ZoneOption OutputsWidget::options[] = {
   { "First channel", ZoneOption::Integer, OPTION_VALUE_UNSIGNED(1), OPTION_VALUE_UNSIGNED(1), OPTION_VALUE_UNSIGNED(32) },
+  { "Fill Background?", ZoneOption::Bool, OPTION_VALUE_BOOL(false) },
+  { "BG Color", ZoneOption::Color, OPTION_VALUE_UNSIGNED(LIGHTGREY) },  { NULL, ZoneOption::Bool },
```
mpaperno: The extra NULL entry probably won't hurt anything and I assume unintentional, but it would be good to clean up.

mpaperno:
Looks good! (Except for the one minor comment.)  Thanks for the extra effort, I think it'll be worth it.

radio/src/gui/480x272/widgets/outputs.cpp
joaquins: It was originally there... it's also present in text.cpp widget... so I thought it has any purpose...
Should that be removed?

mpaperno: It just needs the one "NULL" entry, which was/is on the next line down. (I wish this view of comments within the PR would should a bit more context around the code line being discussed... save a click.)

joaquins:
Screenshot of the latest commit:

latest-config

latest

mpaperno:
Oh... I thought we were talking about the background of the unfilled bars themselves... the original version being transparent.  :pensive:

joaquins:
@mpaperno LOL actually that makes more sense ;)

joaquins:
BG color settings:

bg_color

bsongis:
Good job, thanks!

PR #4148, opened 2016-12-18
9x Mavlink fixes 

### LINKED ISSUE #4132: Mavlink not working with the 9X
The mavlink option doesn't seem to be working with the 9X.
Tried it with 57600 bps and nothing gets parsed.
One issue is that, according to the code, serial double speed is disabled which makes 57600bps greatly out of spec with the 16MHz 9X clock,
Also the serial driver seems messy and with no real difference compared to the also existing telemetry driver.
I can probably fix this even if being most used to working with ARM Cortex-M processors. If the maintainers want me to give it a shot just let me know if I should branch from main or next.

PTDreamer:
Fixes #4132 
Cleans the serial drivers by droping one of them. (negative LOC :) )
Addresses the particular issue I was having (using opentx with ulrs) by creating a new non standard 58798bps speed. This is done on a separate commit which I can drop if considered a particular use case and not worth merging.


bsongis:
This is good, but why on 'master', I would prefer you do this in 'next'

PTDreamer:
> This is good, but why on 'master', I would prefer you do this in 'next'

I couldn't find any contribution guidelines so I asked when I opened the issue and based on @schwabe answer I thought this was the way to proceed. 
I can rebase on next and force push if required.

bsongis:
It will be more complex than a simple rebase ... unless you are really lucky! I close this PR, I will review the new one.

PR #4150, opened 2016-12-18
Enable Windows build with Qt5.5+ & MinGW-w64

mpaperno:
So with these relatively minor changes (and earlier [here](https://github.com/mpaperno/opentx/commit/817c8da7b49de41d1b3c1c85e720832d942fde9a)), Companion and Simulator can be built entirely with one Qt download (which includes MinGW), even right from QtCreator (if so desired).  

QtCreator has decent support for CMake projects just by opening the main CMakeFiles.txt, and all options can be configured right there (no need to run CMake, and it sets up the right Qt environment for you automatically, based on selected Qt Kit).  So this is a pretty easy way for someone to get started (and of course QtC is great for Qt coding).

If one has/installs the Arm/AVR toolchains, then the firmware can be built from QtCreator as well.  Just add the right path to the project config in QtC, so makes it easy to switch versions also.

Of course it can all still be configured and run from a command line or some other IDE, as long as all the PATHs are correct.

This also makes all OS versions able to run on a GCC-based toolchain (using all the same options/switches), vs. relying on MSVC.   As a bonus, the special Windows pthreads and msinttypes are not needed (though dirent.h still is due to the archaic implementation in MinGW64).  Works with the same SDL libs as MSVC version (or with the MinGW versions as well).

I also tried to make the CMake configs a bit friendlier for Windows, replacing the hard-coded path, consolidating some definitions, and adding some options.  (It still relies on `grep` at one point and gives an ugly error... on to-do list.)

To make things even a bit simpler, any objections to adding the `dirent` and `msinttypes` headers to the repo?  Like in `radio/src/thirdparty/windows` or something like that.  I checked the licensing and both allow redistribution.

Oh and finally the console option for Windows is quite handy.  Helped greatly while debugging the [static init issue](https://github.com/opentx/opentx/pull/4141#issue-196041243) since it was crashing way before the usual debug console would be active.

BTW I started on this because I can't run the latest MSVC stuff on my main Win7 workstation due to a long-standing problem with installing SP1 (which isn't that uncommon if you search around).  M$ has dropped support for pre-SP1 version, even if all the other updates are installed.  Even the runtime DLL won't install, so I couldn't run the official OTx Windows nightly 2.2 builds at all (Linux ones worked after some coaxing and pointing to Qt5.5 libs).  So anyway this was my "fix."  :)

Cheers,
-Max


CMakeLists.txt
```diff
@@ -33,31 +33,57 @@ set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR})
...
-  message("SDL Include Path: " ${SDL_INCLUDE_DIR})
+if(Qt5Widgets_FOUND)
+  message(STATUS "Qt Version: ${Qt5Widgets_VERSION}")
 else()
-  message("SDL not found! Joysticks support will be disabled. Simulator audio will not work.")
+  message(WARNING "Qt not found! Companion and Simulator builds disabled.")
 endif()
 
-message(STATUS "Looking for FOX ")
-find_package(Fox)
```
schwabe: Other platforms still needs this for simu

CMakeLists.txt
```diff
@@ -33,31 +33,57 @@ set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR})
...
+endif()
+
 find_package(Qt5Widgets)
 find_package(Qt5Xml)
 find_package(Qt5LinguistTools)
 find_package(Qt5PrintSupport)
 find_package(Qt5Multimedia)
 
-message(STATUS "Looking for SDL ")
-find_package(Sdl)
```
schwabe: Why did you remove find_package(Sdl)?

CMakeLists.txt
mpaperno: Other than Windows?  They'll still get it. This change just prevents the search on Windows altogether (because QUIET isn't working for some reason so you get a bogus warning message).  Does anyone use Fox on Windows?

CMakeLists.txt
mpaperno: I did not, it's further down now, and replaced by better version included with CMake.  I also highlighted this in the commit notes.

schwabe: oh sorry missed that

CMakeLists.txt
schwabe: FindFox is a self written scrpt that might need to be modified to support QUIET

mpaperno: That  script seems to conform to the standard CMake module conventions and doesn't output any messages on its own.  I tried a few things with it but for some reason it would always report negative even with QUIET flag.  I tried a few other FindPkg modules from CMake and most honor the QUIET correctly, but some don't.  Yet I can't find a difference between them (or FindFox) which explains it.   :imp:

schwabe: argh, then lets use your approach

companion/src/CMakeLists.txt
```diff
@@ -359,9 +436,8 @@ string(TOLOWER "${CPACK_PACKAGE_NAME}" CPACK_PACKAGE_NAME_LOWERCASE)
 # very poor trace-backs from the users when they report Companion crash
 set(CPACK_STRIP_FILES FALSE)
 
-set(plugin_dest_dir bin)
-set(qtconf_dest_dir bin)
-set(APPS "\${CMAKE_INSTALL_PREFIX}/bin/${COMPANION_NAME}")
```
schwabe: I think this line got competely lost during your reorganisation

companion/src/CMakeLists.txt
```diff
@@ -373,21 +449,11 @@ IF(APPLE)
...
 
-INSTALL(TARGETS ${COMPANION_NAME}
-    BUNDLE DESTINATION . COMPONENT Runtime
-    RUNTIME DESTINATION bin COMPONENT Runtime
-    )
-
-# Qt5 + Cmake + Mac is poorly documented. A lot of this is guesswork
-# and trial and error. Do not hesitate to fix it for the better
-if(APPLE)
-  set(APPS "\${CMAKE_INSTALL_PREFIX}/${COMPANION_NAME}.app")
```
mpaperno: This line was redundant, look up [here](#diff-a1ea07d95d9868fee378f5c869d8d448L369) (lines left:369 right:445)

There were a few other redundancies in the OSX script as well.

schwabe:
OS X build still works. Other changes look good to me but I have no windows system to test.

companion/src/simulation/simulatorinterface.cpp
```diff
@@ -59,7 +59,7 @@ void registerSimulators()
   QStringList filters;
 #if defined(__APPLE__)
   filters << "*-simulator.dylib";
-#elif (!defined __GNUC__) || (defined __CYGWIN__)
```
jtaylor2: This breaks the simulator on Windows running under Cygwin (posix).  It still compiles but fails to register the dll.
Changing #elif (defined WIN32) || (defined __CYGWIN__) fixes it.

CMakeLists.txt
```diff
@@ -33,31 +33,57 @@ set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR})
...
 # Check for a file that is typically left from a OpenTX 2.1 build and abort if found
 if (EXISTS ${RADIO_SRC_DIRECTORY}/stamp.h OR EXISTS ${RADIO_SRC_DIRECTORY}/translations/en.h)
    message(FATAL_ERROR "Source directory contains files leftover from a OpenTX 2.1 build. Please run `git clean -f` in source directory (Careful: Will remove any extra files) or do a new clean git checkout")
 endif()
 
+# Windows-specific includes and libs shared by sub-projects
+if(WIN32)
+  list(APPEND WIN_INCLUDE_DIRS "${WIN_EXTRA_LIBS_PATH}/dirent/include")
+  if(MSVC)
+    set(WIN_PTHREAD_BASE_PATH "${WIN_EXTRA_LIBS_PATH}/pthreads/280-x86/Pre-built.2")
```
kilrah: Hmm why change the path (the added `280-x86/` folder)?

kilrah:
Issues on VC:
companion/simulator: some link problems
```
10>SDLmain.lib(SDL_win32_main.obj) : error LNK2005: _main already defined in companion.obj
10>MSVCRTD.lib(initializers.obj) : warning LNK4098: defaultlib 'msvcrt.lib' conflicts with use of other libs; use /NODEFAULTLIB:library
10>SDLmain.lib(SDL_win32_main.obj) : warning LNK4217: locally defined symbol _fprintf imported in function _ShowError
10>SDLmain.lib(SDL_win32_main.obj) : error LNK2019: unresolved external symbol _SDL_main referenced in function _main
10>SDLmain.lib(SDL_win32_main.obj) : error LNK2019: unresolved external symbol __imp____iob_func referenced in function _ShowError
10>C:\opentx-22\companion_build\Debug\companion.exe : fatal error LNK1120: 2 unresolved externals
```

on Horus simulator dll build:
a megaton of errors about the math functions (round, floor...)
```
14>C:\Program Files (x86)\Windows Kits\10\Include\10.0.10240.0\ucrt\math.h(521): error C2143: syntax error: missing ',' before '+' (compiling source file C:\opentx-22\code\radio\src\targets\horus\telemetry_driver.cpp)
14>C:\Program Files (x86)\Windows Kits\10\Include\10.0.10240.0\ucrt\math.h(521): error C2143: syntax error: missing ',' before '+' (compiling source file C:\opentx-22\code\radio\src\targets\horus\audio_driver.cpp)
14>C:\Program Files (x86)\Windows Kits\10\Include\10.0.10240.0\ucrt\math.h(521): error C2143: syntax error: missing ',' before '+' (compiling source file C:\opentx-22\code\radio\src\targets\horus\keys_driver.cpp)
14>C:\Program Files (x86)\Windows Kits\10\Include\10.0.10240.0\ucrt\math.h(521): error C2143: syntax error: missing ',' before '+' (compiling source file C:\opentx-22\code\radio\src\targets\horus\trainer_driver.cpp)
14>C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include\cmath(278): error C2143: syntax error: missing ',' before '+' (compiling source file C:\opentx-22\code\radio\src\targets\horus\telemetry_driver.cpp)
14>C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include\cmath(279): error C2084: function 'float floor(float) noexcept' already has a body (compiling source file C:\opentx-22\code\radio\src\targets\horus\telemetry_driver.cpp)
14>  C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include\cmath(131): note: see previous definition of 'floor' (compiling source file C:\opentx-22\code\radio\src\targets\horus\telemetry_driver.cpp)
14>C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include\cmath(278): error C2143: syntax error: missing ',' before '+' (compiling source file C:\opentx-22\code\radio\src\targets\horus\audio_driver.cpp)
```

@schwabe the VM I set up the other day is still available for download :)

companion/src/simulation/simulatorinterface.cpp
mpaperno: OK, will fix.  How do you build with Cygwin...  works with Qt?

mpaperno:
@kilrah The first error with SDL -- perhaps problem with the order of libs given to the linker? Could you try moving the order of target_link_libraries in companion/src/CMakeLists.txt, to move the Windows ones first?

```diff
--- a/companion/src/CMakeLists.txt
+++ b/companion/src/CMakeLists.txt
@@ -249,10 +249,10 @@ add_executable(${COMPANION_NAME} MACOSX_BUNDLE WIN32 ${companion_SRCS} ${compani
 add_dependencies(${COMPANION_NAME} gen_qrc)
 qt5_use_modules(${COMPANION_NAME} Core Widgets Network)
 
-target_link_libraries(${COMPANION_NAME} PRIVATE generaledit modeledit simulation common qcustomplot shared storage ${PTHREAD_LIBRARY} ${SDL_LIBRARY})
 if(WIN32)
   target_link_libraries(${COMPANION_NAME} PRIVATE "${WIN_LINK_LIBRARIES}")
 endif()
+target_link_libraries(${COMPANION_NAME} PRIVATE generaledit modeledit simulation common qcustomplot shared storage ${PTHREAD_LIBRARY} ${SDL_LIBRARY})
 
 ############# Standalone simu ###############
 
@@ -277,10 +277,10 @@ add_executable(${SIMULATOR_NAME} MACOSX_BUNDLE WIN32 ${simu_SRCS} ${companion_RC
 add_dependencies(${SIMULATOR_NAME} gen_qrc)
 
 # TODO not the same link command than companion?
-target_link_libraries(${SIMULATOR_NAME} PRIVATE simulation common storage qxtcommandoptions ${QT_LIBRARIES} ${QT_QTMAIN_LIBRARY} ${PTHREAD_LIBRARY} ${SDL_LIBRARY} ${OPENTX_SIMULATOR_LIBS})
 if(WIN32)
   target_link_libraries(${SIMULATOR_NAME} PRIVATE "${WIN_LINK_LIBRARIES}")
 endif()
+target_link_libraries(${SIMULATOR_NAME} PRIVATE simulation common storage qxtcommandoptions ${QT_LIBRARIES} ${QT_QTMAIN_LIBRARY} ${PTHREAD_LIBRARY} ${SDL_LIBRARY} ${OPENTX_SIMULATOR_LIBS})
```

mpaperno:
@kilrah I'm a bit stumped on the 2nd error with round().  It must be from [here](https://github.com/opentx/opentx/pull/4150/files#diff-bd8ad939735a868d514643e3c451fb73L76), but I'm not sure why it would be different now.  Again possibly the includes list order changed... need to investigate.

kilrah:
Change suggested above unfortunately doesn't help for the linking, exact same errors :(

CMakeLists.txt
```diff
@@ -33,31 +33,57 @@ set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR})
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR})
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${PROJECT_SOURCE_DIR}/cmake)
 
+if(WIN32)
+  set(WIN_EXTRA_LIBS_PATH "C:/Programs" CACHE PATH
```
bsongis: excellent change

CMakeLists.txt
bsongis: at the time we introduced SDL I think this find was missing, thanks for the change!

companion/src/printdialog.cpp
```diff
@@ -24,7 +24,7 @@
 #include <QPrinter>
 #include <QPrintDialog>
 
-// #if !defined WIN32 && defined __GNUC__
+// #if !defined _MSC_VER && defined __GNUC__
 // #include <unistd.h>
```
bsongis: you may completely remove those lines

mpaperno:
@jtaylor2 Ah, I see the cause of the 2nd problem you listed.  The other entries in that includes list will need the `-I` prepended.  Will fix.

mpaperno:
> source file: 'C:/opentx/git/opentx/winbuild': No such file or directory

Does anyone know what is supposed to be in `winbuild`? The folder doesn't exist in the distro.  Is this something created during auto-builds... or something?  It was in the original includes path list for Windows builds, but I don't know why.

kilrah:
I think `winbuild` is deprecated, reminds me of something but it would have been ages ago, pre-2.0. It could have been the equivalent of `C:\Programs` now. Can be removed.

I've just got a full build in a clean VM using QT Creator. Nice job, quite a bit easier to setup indeed! Only thing that doesn't seem to work is building the installer, but that's not important for personal use.

The debug console window is AWESOME! Been missing that for years, never knew that was possible!

jtaylor2:
Testing on a Windows 10 Visual Studio 2015 Update 3 system that will compile the current next.  I am getting different errors depending on what I try to compile and even what options I select.  I have spent the last several hours trying to figure out why, but I don't know enough cmake to have a clue.

For example I can compile opentx-9x-simulator without errors, but I can't compile opentx-X9D-simulator and depending on what -D options I use I get different errors.  I'm also getting the link errors that @kilrah reported with the simulator and companion targets.

Some of the errors make no sense as if it's picking up the wrong include files.  For example compiling opentx-x9d-simulator from just cmake -DPCBX9D gives errors like this:

```
5>  simpgmspace.cpp
5>C:\Program Files (x86)\Windows Kits\10\Include\10.0.14393.0\ucrt\math.h(532): error C2143: syntax error: missing ',' before '+' (compiling source file C:\opentx\git\opentx\radio\src\audio_arm.cpp)
5>C:\Program Files (x86)\Windows Kits\10\Include\10.0.14393.0\ucrt\math.h(532): error C2143: syntax error: missing ',' before '+' (compiling source file C:\opentx\git\opentx\radio\src\targets\simu\simpgmspace.cpp)
5>C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include\cmath(278): error C2143: syntax error: missing ',' before '+' (compiling source file C:\opentx\git\opentx\radio\src\targets\simu\simpgmspace.cpp)
```

But adding cmake -DPBCX9D -DLUA=YES gives those types of errors plus:

```
4>  Microsoft (R) C/C++ Optimizing Compiler Version 19.00.24215.1 for x86
4>  Copyright (C) Microsoft Corporation.  All rights reserved.
4>
4>  include
4>  msinttypes
4>c1 : fatal error C1083: Cannot open source file: 'C:/programs/pthreads/Pre-built.2/include': No such file or directory
4>  dataconstants.h
4>c1 : fatal error C1083: Cannot open source file: 'C:/programs/msinttypes': No such file or directory
6>------ Build started: Project: opentx-x9d-simulator, Configuration: Debug Win32 ------
6>  telemetry_driver.cpp
6>  audio_driver.cpp
```
Yet if I look at the command line is see those directories included:
```
.........\Qt5.6.1\5.6\msvc2015\include\QtCore" /I"C:\programs\Qt\Qt5.6.1\5.6\msvc2015\.\mkspecs\win32-msvc2015" /I"C:\programs\dirent\include" /I"C:\programs\pthreads\Pre-built.2\include" /I"C:\programs\msinttypes" /Zi /Gm- /Od /Ob0 /Fd"opentx-x9d-simulator.dir\Debug\vc140.pdb" /Zc:i.............
```

mpaperno:
Thanks for trying Jim!  I just got a VC2015 build set up and am testing now... just got the same BS about syntax errors, so I'm on the right track.  I'm almost sure it's something in the order of includes.  Will report back when figured.

mpaperno:
Does the MSVC build usually work for you guys with Pthreads 2.8?  Or did everyone update to 2.9 and forget to tell me?  :stuck_out_tongue_winking_eye:   Just spent a couple hours trying to figure that one out while trying to build `next`.  Kept getting errors about `timespec` struct getting redefined in pthread.h.

> C:\Program Files (x86)\Windows Kits\10\Include\10.0.14393.0\ucrt\math.h(532): error C2143: syntax error: missing ',' before '+' (compiling source file C:\opentx\git\opentx\radio\src\audio_arm.cpp)

Got rid of all these by removing the strange defines in definitions.h   Does anyone know/remember why they were there to begin with?  It built fine w/out them (though I haven't run it yet).

```diff
--- a/radio/src/definitions.h
@@ -73,10 +73,4 @@ typedef __int24 int24_t;
 #define CONVERT_UINT_PTR(x) ((uint32_t *)(x))
 #endif
 
-#if defined(_MSC_VER)
-#define round(x)    floor(x+0.5)
-#define strcasecmp  _stricmp
-#define strncasecmp _tcsnicmp
-#endif
-
 #endif // _DEFINITIONS_H_
```

Last thing now is the linking error in companion build.

kilrah:
I have set up a VM last week with [these instructions](https://github.com/opentx/opentx/wiki/OpenTX-Windows-Build-Instructions), updating them when necessary so they should be right - and list pthreads 2.8.

If you want the VM to try...

the `timespec` issue is handled by 
https://github.com/opentx/opentx/blob/next/radio/src/CMakeLists.txt#L345

You might have forgotten that one in your mods

mpaperno:
That's very strange indeed.  No, I hadn't forgotten or changed that line, and I was also trying to build the "stock" `next` branch when I was getting those errors.  And here's the thing... in Pthread 2.8 that define doesn't work!

```c
// pthread.h v2.8
#ifndef _TIMESPEC_DEFINED
#define HAVE_STRUCT_TIMESPEC 1
struct timespec {
        long tv_sec;
        long tv_nsec;
};
#endif /* HAVE_STRUCT_TIMESPEC */


// pthread.h v2.9
#if !defined(HAVE_STRUCT_TIMESPEC)
#define HAVE_STRUCT_TIMESPEC
#if !defined(_TIMESPEC_DEFINED)
#define _TIMESPEC_DEFINED
struct timespec {
        time_t tv_sec;
        long tv_nsec;
};
#endif /* _TIMESPEC_DEFINED */
#endif /* HAVE_STRUCT_TIMESPEC */
```

jtaylor2:
The -DLUA=YES problem is coming from radio/src/lua/CMakeLists.txt.  The macro(add_lua_export_target target) doesn't like the includes.  Changing it as below lets it compile, but I don't know enough cmake to fix it. 

```
OUTPUT lua_exports_${target}.inc
      #COMMAND ${CMAKE_C_COMPILER} -E ${ARGN} -DEXPORT ${LUA_EXPORT_INCLUDES} ${RADIO_SRC_DIRECTORY}/dataconstants.h > lua_exports_${target}.tmp
	COMMAND ${CMAKE_C_COMPILER} -E ${ARGN} -DEXPORT -IC:/Programs/msinttypes ${RADIO_SRC_DIRECTORY}/dataconstants.h > lua_exports_${target}.tmp
```

kilrah:
Are we using the same pthreads 2.8??

The link we have in the instructions installs 
`#define PTW32_VERSION 2,8,0,0`

and has 
```
#ifndef HAVE_STRUCT_TIMESPEC
#define HAVE_STRUCT_TIMESPEC 1
struct timespec {
        long tv_sec;
        long tv_nsec;
};
#endif /* HAVE_STRUCT_TIMESPEC */
```

mpaperno:
> Are we using the same pthreads 2.8??

`#define PTW32_VERSION 2,8,0,0`

Mine is labeled the same but has that fairly obvious bug in it.  Weird! But yes, their [published](ftp://sourceware.org/pub/pthreads-win32/prebuilt-dll-2-8-0-release/include/pthread.h) 2.8.0 also seems correct.  I have no idea how mine got that way... I can't imagine I'd edit it for any reason, at least not w/out leaving a note in there.  Gremlins!  :imp:

mpaperno:
OK folks, thanks for your patience.  :)   MSVC build should be OK now.

The linking issue with SDL turned out to be because the new FindSDL would also find SDLmain.lib (the old one did not).  Including SDLmain in the linker objects list was breaking MSVC (but not MinGW).

I still don't know exactly why those MSVC macros (round(), etc) were now throwing errors, but removing them didn't seem to break anything (yet!).

mpaperno:
PS. if you already tried building this version then you may need to manually clear your SDL_LIBRARY cached value in order for the linking to work. Or simply remove the "SDLmain.lib" entry from the cached value.

jtaylor2:
Using VS15 now builds and runs, however, I get no audio in the Companion simulator using your build.  I have audio using a just pulled next build.  I have audio in your build of the standalone simulator but not the Companion simulator.  Your build the audio from the SF page works, but nothing in the simulator.  And in the simulator the audio file path messages are not coming up on the console either.

Using these build options on both builds:

cmake -DPCB=X9D -DEXT=FRSKY_SPORT -DSDCARD=YES -DPPM_CENTER_ADJUSTABLE=YES -DFLIGHT_MODES=YES -DAUTOSWITCH=YES -DPPM_LIMITS_SYMETRICAL=YES -DAUTOSOURCE=YES -DDBLKEYS=YES -DHELI=YES  -DPPM_UNIT=PERCENT_PREC1 -DGVARS=YES  -DUNITS=IMPERIAL -DOVERRIDE_CHANNEL_FUNCTION=NO -DLUA=YES -DMIXERS_MONITOR=YES -DTIMERS=2 -DMULTIMODULE=YES -DSIMU_AUDIO=YES -G "Visual Studio 14 2015" ..l

mpaperno:
Jim, I can't really recreate what you're talking about... I just did a bunch of test builds (Win10, MSVC 19.00.24210, Qt 5.7-32b, SDL 1.2.15) and the audio all works.  Except I did find a bug with playing tracks directly from the SD manager (or any other file interaction for that matter).

I just updated the whole SDL detection scheme to (hopefully) be more robust.  Could you clear your CMake cache (or at least all the SDL* entries) and give it another go?

@kilrah, this should fix the installer problem with finding SDL.

I opted to go with a customized FindSDL script in the end because the original had a nasty bug of recursively appending "-lpthread" (or "-mingw32") to a cached temporary variable... after many re-configurations it would get silly-long.  Plus now it returns everything we need. Hopefully puts a nail in that :coffin:. 

-Max

jtaylor2:
@mpaperno , yes audio now works in the Companion simulator.  Everything I've tested now works.

mpaperno:
Excellent, thanks for testing!  I don't have an OS X build set up yet, so hopefully the SDL detection changes didn't break that.  Seems good on Linux.

bsongis:
@schwabe may we ask you to check that OSX build is not broken?

kilrah:
I launched a build of that branch on his VM and got a DMG in return, so it looks pretty good :)

kilrah:
Yep, contents are even working :)

Tested on VC too, all good, installer now happy as well :+1:

kilrah:
Ah, some cmake warnings in my usual firmware build:

```

C:\opentx-22\fw_build>cmake -G "MinGW Makefiles" ../code/ -DCMAKE_PREFIX_PATH=C:\Qt\Qt5.7.0\5.7\msvc2015 -DPCB=X9E -DHELI=YES -DGVARS=YES -DTRANSLATIONS=EN -DLUA=YES -DMIXERS_MONITOR=YES -DTIMERS=3 -DUSB=JOYSTICK -DJITTER_FILTER=NO -DMULTIMODULE=YES
-- The C compiler identification is GNU 4.8.1
-- The CXX compiler identification is GNU 4.8.1
-- Check for working C compiler: C:/MinGW/bin/gcc.exe
-- Check for working C compiler: C:/MinGW/bin/gcc.exe -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: C:/MinGW/bin/g++.exe
-- Check for working CXX compiler: C:/MinGW/bin/g++.exe -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- OpenTX 2.2.0
-- Qt Version: 5.7.0
-- Looking for pthread.h
-- Looking for pthread.h - not found
-- Found Threads: TRUE
-- Found SDL: C:/Programs/SDL/lib/x86/SDL.dll (found version "1.2.15")
-- SDL Lib: C:/Programs/SDL/lib/x86/SDL.dll; Headers: C:/Programs/SDL/include; Version: 1.2.15
-- Found PythonInterp: C:/Python27/python.exe (found version "2.7.11")
-- Python found, version: 2.7.11
-- The ASM compiler identification is GNU
-- Found assembler: C:/MinGW/bin/gcc.exe
CMake Warning (dev) at C:/Program Files (x86)/CMake/share/cmake-3.7/Modules/Compiler/GNU.cmake:45 (if):
  Policy CMP0054 is not set: Only interpret if() arguments as variables or
  keywords when unquoted.  Run "cmake --help-policy CMP0054" for policy
  details.  Use the cmake_policy command to set the policy and suppress this
  warning.

  Quoted variables like "0" will no longer be dereferenced when the policy is
  set to NEW.  Since the policy is not set the OLD behavior will be used.
Call Stack (most recent call first):
  C:/Program Files (x86)/CMake/share/cmake-3.7/Modules/Compiler/GNU-ASM.cmake:6 (__compiler_gnu)
  C:/Program Files (x86)/CMake/share/cmake-3.7/Modules/CMakeASMInformation.cmake:22 (include)
  radio/src/CMakeLists.txt:40 (enable_language)
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Warning (dev) at C:/Program Files (x86)/CMake/share/cmake-3.7/Modules/Compiler/GNU.cmake:45 (if):
  if given arguments:

    "NOT" "APPLE" "OR" "NOT" "CMAKE_ASM_COMPILER_VERSION" "VERSION_LESS" "4"

  An argument named "0" appears in a conditional statement.  Policy CMP0012
  is not set: if() recognizes numbers and boolean constants.  Run "cmake
  --help-policy CMP0012" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.
Call Stack (most recent call first):
  C:/Program Files (x86)/CMake/share/cmake-3.7/Modules/Compiler/GNU-ASM.cmake:6 (__compiler_gnu)
  C:/Program Files (x86)/CMake/share/cmake-3.7/Modules/CMakeASMInformation.cmake:22 (include)
  radio/src/CMakeLists.txt:40 (enable_language)
This warning is for project developers.  Use -Wno-dev to suppress it.

simu: using MinGW with  -mno-ms-bitfields -mwindows
-- C:/Qt/Qt5.7.0/5.7/msvc2015/bin/lupdate.exe
Qt lupdate: C:/Qt/Qt5.7.0/5.7/msvc2015/bin/lupdate.exe
companion: using MinGW with  -mno-ms-bitfields -mwindows
-- Partial install to C:/opentx-22/fw_build
-- Configuring done
-- Generating done
-- Build files have been written to: C:/opentx-22/fw_build
```

(builds fine though)

mpaperno:
Andre, I've seen those warning also a few times on Windows.  Seems to happen when CMake gets confused with old cache data and updated CMakefiles?  Re-running it seems to clear up the warnings.  I have absolutely no idea what they mean.  Maybe we should take their advice and suppress them  :)

Thanks for the tests!

jtaylor2:
I normally build firmware on Linux, but I did build it on Windows when I was testing your PR and didn't get those warnings.  Of course my build environment is probably different.

mpaperno:
I've never seen them on a clean/new build config.  Only after "something" changes, and CMake decides to run the whole discovery process again even though it had already found the compiler before, and all that. Maybe it happens when it finds a new FindPkg module.

kilrah:
I've cleared my whole build dir before reporting. So unless it caches stuff outside the project...
I've got a nice batch file that removes everything, it's the first port of call when something looks strange or even everytime I switch branch :)

mpaperno:
> CMake Warning (dev) at C:/Program Files (x86)/CMake/share/cmake-3.7/Modules/Compiler/GNU.cmake:45 (if):
  Policy CMP0054 is not set...

https://cmake.org/cmake/help/v3.1/policy/CMP0054.html

> This policy was introduced in CMake version 3.1. CMake version 3.1.3 warns when the policy is not set and uses OLD behavior. Use the cmake_policy() command to set it to OLD or NEW explicitly.

So _that's_ when I noticed it started happening... after i updated to a newer CMake.  You haven't noticed it before?

So CMake is throwing warnings about its own code that they forgot to update?

kilrah:
Nope, doesn't do it on next... :\

bsongis:
excellent job, just this warning from @kilrah ? I prefer pressing the button and avoid a merge if we wait more

PR #4151, opened 2016-12-18
Schwabe/multi telemetry

### LINKED ISSUE #3979: Telemetry stoped working / iXJT and MultiModule
I updated about a week ago to **N359** and subsequently updated my Multimodule too.
Telemetry worked fine using my **Multi Module with [v1.1.5](https://github.com/pascallanger/DIY-Multiprotocol-TX-Module/releases/tag/v1.1.5)** and i got RSSI feedback from my TinyWhoop (DSMX, 6ch, 22ms). Sunday i noticed that my telemetry using the internal Model was not working. (iXJT @1.2.16 with X4R-SB connected to a Motolab Tempest running Betaflight 3.0.1). i flew a few lipos without telemetry, switched quads, still no telemetry. Having a short break i turned of my **Horus** Some minutes later I turned it back on with telemetry working fine for the rest of the day. (i only flew IXJT based models that day) Today i noticed that the Telemetry is not working with my multi module too. I Checked if it is working with my XSR/X4R models, but i didn't work there either. I left a model using the internal IXJT modul selected and switched of my transmitter. After turning it on again telemetry worked with all my different quads (which have different models, i am not using one model for all of them). I tried switching over to the TinyWhoop using the Multimodule and the telemetry there was not working. If i switch back to my iXJT Models it stays broken until i reboot my transmitter. 

Telemetry worked a few days with N359 and the multi module so i don't have a direct clue what is causing the issue. Do you have any ideas/suggestions what i should check to trace the error?

schwabe:
The custom telemetry allows the module to tell the telemetry protocol to OpenTX and also allows signaling the status of the Module (e.g. invalid protocol) to OpenTX

This should also close #3979.


schwabe:
Example images how this looks:

image

kilrah:
This is cool!

lassebm:
Maybe have version and status as separate menu items for clarity.

radio/src/gui/212x64/model_setup.cpp
```diff
@@ -934,16 +935,16 @@ void menuModelSetup(event_t event)
                   modelHeaders[g_eeGeneral.currModel].modelId[moduleIdx] = g_model.header.modelId[moduleIdx];
                 }
               }
-              if (s_editMode==0 && event==EVT_KEY_BREAK(KEY_ENTER)) {
+              if (menuVerticalOffset == sub && s_editMode==0 && event==EVT_KEY_BREAK(KEY_ENTER)) {
```
schwabe: that change got in there by accident, thanks for noticing.

kilrah:
I'd rather leave it like that, takes enough space already.

radio/src/telemetry/multi.cpp
```diff
@@ -0,0 +1,274 @@
...
+      break;
+    default:
+      TRACE("[MP] Unkown multi packet type 0x%02X, len %d", type, len);
+      break;
+  }
+}
+
+// sprintf does not work AVR ARM
+// use a small helper function
+static void appendInt(char* buf, uint32_t val)
```
bsongis: Should not be done here!

bsongis:
A lot of bad indentation in this PR, but it is overall a good job and I don't want another merge, I push it now.

bsongis:
would you try to fix indentation when you have time? thanks!

schwabe:
hm could you tell me an example what is wrong and in what files, I probably made that wrong consistent.

PR #4152, opened 2016-12-19
9X Mavlink fixes

### LINKED ISSUE #4132: Mavlink not working with the 9X

PTDreamer:
Copied from #4148
Fixes #4132
Cleans the serial drivers by droping one of them. (negative LOC :) )
Addresses the particular issue I was having (using opentx with ulrs) by creating a new non standard 58798bps speed. This is done on a separate commit which I can drop if considered a particular use case and not worth merging.

PS-Although this is the same as #4148 I haven't tested this branch with real hw yet


radio/src/telemetry/frsky.cpp
```diff
@@ -48,7 +48,7 @@ enum FrSkyDataState {
 #endif
 };
 
-NOINLINE void processFrskyTelemetryData(uint8_t data)
+NOINLINE void processSerialData(uint8_t data)
```
bsongis: I don't think the name is right, what about `processTelemetryData(...)`?

radio/src/io/frsky_sport.cpp
```diff
@@ -130,7 +130,7 @@ bool sportWaitState(SportUpdateState state, int timeout)
   for (int i=timeout/2; i>=0; i--) {
     uint8_t byte ;
     while (telemetryGetByte(&byte)) {
-      processFrskyTelemetryData(byte);
+      processSerialData(byte);
```
bsongis: `processTelemetryData(...)` ?

radio/src/targets/common/avr/CMakeLists.txt
```diff
@@ -1,10 +1,10 @@
-option(TELEMETRY "Telemetry protocol" OFF)
 option(AUDIO "Audio support" OFF)
 option(VOICE "Voice support" OFF)
 option(GRAPHICS "Additional graphics" ON)
 option(BOLD "Bold font" ON)
 option(BATTGRAPH "Battery graph" OFF)
 option(HAPTIC "Haptic support" OFF)
+set(TELEMETRY "NONE" CACHE STRING "Telemetry Type (NONE/MAVLINK/ARDUPILOT/FRSKY/FRSKY_SPORT/TELEMETREZ/NMEA)")
```
bsongis: good change

radio/src/targets/common/avr/telemetry_driver.cpp
```diff
@@ -79,26 +94,177 @@ ISR(USART_RX_vect_N(TLM_USART))
     telemetryRxBufferCount = 0;
   }
   else {
-    processFrskyTelemetryData(data);
+    processSerialData(data);
```
bsongis: same comment

radio/src/targets/common/avr/telemetry_driver.cpp
```diff
@@ -79,26 +94,177 @@ ISR(USART_RX_vect_N(TLM_USART))
...
+  UBRRH_N(TLM_USART) = UBRRH_VALUE;
+  UBRRL_N(TLM_USART) = UBRRL_VALUE;
+#if defined USE_2X
+  UCSRA_N(TLM_USART) |= (1 << U2X_N(TLM_USART)); // enable double speed operation.
+#else
+  UCSRA_N(TLM_USART) &= ~(1 << U2X_N(TLM_USART)); // disable double speed operation.
+#endif
+}
+
+void telemetryPortInitFromIndex(uint8_t index) {
```
bsongis: can be simplified, here we enter in 2 switch

radio/src/telemetry/mavlink.cpp
```diff
@@ -607,114 +561,28 @@ static inline void MAVLINK_msg_request_data_stream_pack_send(uint8_t req_stream_
...
-#endif
 /*!	\brief Telemetry monitoring, calls \link MAVLINK10mspoll.
  *	\todo Reimplemnt \link MAVLINK10mspoll
  *
  */
 void telemetryWakeup() {
 	uint16_t tmr10ms = get_tmr10ms();
-	uint8_t count = tmr10ms & 0x0f; // 15*10ms
-	if (!count) {
+	static uint16_t last_time = 0;
```
bsongis: Nice catch, this bug is there since we changed the mixer frequency!
I would only keep (and use) the low byte of tmr10ms though.

radio/src/targets/common/avr/telemetry_driver.cpp
PTDreamer: sorry? can you explain better?

radio/src/telemetry/frsky.cpp
PTDreamer: Sure I can make the change although strictly speaking the driver should be usage agnostic and as such should be called something like "serial_driver" and the process function "processSerialData". Looking at the history I saw several changes going back and forth with function and variable naming based on specific usages being handled at that specific time.

radio/src/targets/common/avr/telemetry_driver.cpp
bsongis: this function is only a switch, each case calls telemetryPortInit which is another switch

radio/src/telemetry/frsky.cpp
bsongis: This function name here is too generic, when we speak serial on Taranis, we don't think directly of the Telemetry!

radio/src/targets/common/avr/telemetry_driver.cpp
PTDreamer: Sorry I still don't follow are you saying you would prefer to kill the  "void telemetryPortInit(uint32_t baudrate)"  function? That is also required by "void telemetryPortInit()" which is used throughout. This telemetryPortInitFromIndex is just a workaround because the gui widget only returns the index to the option and not the value itself.
Sorry if I'm missing the obvious.

bsongis: Yes I would completely remove the `telemetryPortInit(bauds)` function

schwabe:
@PTDreamer Could you address @bsongis comments?

PTDreamer:
> @PTDreamer Could you address @bsongis comments?

The function renaming I can easily. The removal of "telemetryPortInit" would require extensive changes throughout to make it cleanly which IMO is outside the scope of this PR.

bsongis:
Merged to 'next'

Issue #4010, opened 2016-11-09
[Simulator] main window does not show up on OSX


raphaelcoeffic:
This issue has been observed on OSX Yosemite all across the next branch, and is thus probably related to Qt5 in some way.

I looked for anything strange with GDB and could not find anything. The process is not stuck in anything, it's just not showing the main simulator window, that's it.

Any clue/pointer from someone knowledgable on QT would be highly appreciated.

bsongis:
Isn't it just a packaging problem?

schwabe:
no, unfortenately not. The simulator window even works, you can hear sounds from the simulator but the window is not visible.

raphaelcoeffic:
With gdb on, I could not find anything suspicious, but please bear with me that I have close to 0 experience with QT.

bsongis:
I have some experience with Qt but none with MacOS. Perhaps we can try to work together and fix it!
On Linux I just launch 'make' then I can test my binary
I wish it's the same on OSX, but I have an error with a Cocoa lib missing

joaquins:
I can reproduce the same issue using the latest nightly (opentx-companion-2.2.0N359.dmg) and also compiling the source code myself.

This issue is NOT happening when compiling for HORUS. Launching simulator.app with libopentx-horus-simulator.dylib works as expected (simulator window opened and responsive).

If using the X9D+ build (libopentx-x9d+-simulator.dylib), the simulator opens but no window at all is visible.

raphaelcoeffic:
@joaquins that's interesting! I might be able to find the difference between both.

raphaelcoeffic:
@bsongis I do run it right on the spot where it has been compiled. Basically, the same as you do on Linux.

joaquins:
Ok, this is weird: I tried today again compiling in a different computer, with OSX Yosemite, and the simulator works perfectly with both Taranis and Horus!

I guessed that it's related to the building environment and the QT versions used, but I just copied the binaries (simulator.app & libopentx-x9d+-simulator.dylib) into the first system (OSX Sierra), and the issue is still there: Taranis simulator window NOT showing.

So, something in the taranis UI is not working with OSX Sierra only?
@raphaelcoeffic which version of OSX are you using?

raphaelcoeffic:
@joaquins I'm still on Yosemite. I could try your binaries, just to see if it would work here as well.

joaquins:
[...]

Issue #3944, opened 2016-10-21
Companion simulator has Hold X and Hold Y, needs Hold SH. 


Miami-Mike:
I originally wrote SF but I meant SH. Thanks kilrah.

I programmed my Taranis to enable certain switch functions while I hold switch SH, but I can't test it with Companion because there's no way to hold SH like you can with the sticks. My only workaround was to temporarily change it to another switch.

kilrah:
I guess you mean SH...

But what about setting all other things as they should be first, and actuate SH last?

Miami-Mike:
What I'm working on is a special "Set Failsafe Trims" flight mode for adjusting trims to the best-flying failsafe positions while I'm actually in flight. In order to properly test this in Companion I need to secure SH↓ and then actuate all of the trims to make sure they do what they're meant to do without affecting the other flight modes. 

It  works great, by the way.

kilrah:
Can be taken into account whenever we finally do that full simulator revamp.

pindeslandes:
Please, see #1627

mpaperno:
Duplicate of #1627, closing.

PR #4154, opened 2016-12-21
Reduce width of all simulator windows

mpaperno:
And also height of Horus window. Possible fix for #3135 

One minor issue is that while this reduces the height of the Horus sim, it does increase the height of the Taranis and 9x sims by a few pixels.  

The width and height of the bottom part of the sim window are basically dictated by size the Output tabs, as shown in the screenshots.  It's a bit tight with 64 switches.  

One possible idea would be to move switches 33-64 to the 2nd outputs screen.  

Another would be to further reduce the size of the switch indicators, at a possible detriment to readability of the labels (though dropping the "L" should be OK?).

Or just leave it as is.  :)   (sorry, no OS X screens at this time...)

simu-horus-overlay-a

simu-taranis-overlay-a

simu-horus-radio-compare-linux-a

simu-horus-outputs-compare-linux-a

simu-taranis-radio-compare-linux-a

simu-taranis-outputs-compare-linux-a

simu-9xrpro-outputs-compare-linux-a

simu-horus-outputs-overlay-win-a

### DIFF
--- a/companion/src/constants.h
+++ b/companion/src/constants.h
@@ -59,10 +59,10 @@ enum BoardEnum {
 #define CPN_MAX_KEYS                   32
 #define CPN_MAX_MOUSE_ANALOGS          2
 
-const char * const ARROW_LEFT = "\xE2\x86\x90";
-const char * const ARROW_UP = "\xE2\x86\x91";
-const char * const ARROW_RIGHT = "\xE2\x86\x92";
-const char * const ARROW_DOWN = "\xE2\x86\x93";
+const char * const ARROW_LEFT = "\u2190";
+const char * const ARROW_UP = "\u2191";
+const char * const ARROW_RIGHT = "\u2192";
+const char * const ARROW_DOWN = "\u2193";
 
 #if defined(DEBUG)
 #define HORUS_READY_FOR_RELEASE()     true
--- a/companion/src/simulation/simulatordialog.cpp
+++ b/companion/src/simulation/simulatordialog.cpp
@@ -24,10 +24,7 @@
 #include <iostream>
 #include "helpers.h"
 #include "simulatorinterface.h"
-#include "sliderwidget.h"
-
-#define GBALL_SIZE  20
-#define RESX        1024
+#include "virtualjoystickwidget.h"
 
 int SimulatorDialog::screenshotIdx = 0;
 SimulatorDialog * traceCallbackInstance = 0;
@@ -85,9 +82,10 @@ SimulatorDialog::SimulatorDialog(QWidget * parent, SimulatorInterface *simulator
   TrainerSimu(0),
   DebugOut(0),
   buttonPressed(0),
-  trimPressed (TRIM_NONE),
+  trimPressed(TRIM_NONE),
   middleButtonPressed(false)
 {
+  setWindowFlags(Qt::Window);
   //shorcut for telemetry simulator
   // new QShortcut(QKeySequence(Qt::CTRL + Qt::Key_T), this, SLOT(openTelemetrySimulator()));
   new QShortcut(QKeySequence(Qt::Key_F4), this, SLOT(openTelemetrySimulator()));
@@ -108,6 +106,23 @@ void SimulatorDialog::closeEvent (QCloseEvent *)
 {
   simulator->stop();
   timer->stop();
+  //g.simuWinGeo(GetCurrentFirmware()->getId(), saveGeometry());
+}
+
+void SimulatorDialog::showEvent(QShowEvent * event)
+{
+  static bool firstShow = true;
+  if (firstShow) {
+    if (flags & SIMULATOR_FLAGS_STICK_MODE_LEFT) {
+      vJoyLeft->setStickConstraint(VirtualJoystickWidget::HOLD_Y, true);
+      vJoyLeft->setStickY(1);
+    }
+    else {
+      vJoyRight->setStickConstraint(VirtualJoystickWidget::HOLD_Y, true);
+      vJoyRight->setStickY(1);
+    }
+    firstShow = false;
+  }
 }
 
 void SimulatorDialog::mousePressEvent(QMouseEvent *event)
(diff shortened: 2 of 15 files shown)


bsongis:
Why don't we move the logical switches to their own tab?

bsongis:
And perhaps a scroll for the Outputs so that they are on only one tab...

bsongis:
... unless ... those outputs could be displayed vertically?

mpaperno:
We could scroll the outputs and make the whole dialog resizable, so on a taller screen one could see all the outputs at once if one chooses.  Even on a short screen one could still scroll to the set which one wants to see (eg. someone wants to see 10-20 at once).

The LS area could also be scrollable I suppose, or move to own tab... I'm not sure how people use those output screens, eg. if having all the LSs on the same page is important or not so much.

I wasn't sure why the sliders were chosen as channel indicators instead of, say, progress bars (which would be a bit more compact).  

Vertically would work.  I can check how many would fit that way.  But it may also be easier to scroll horizontally than vertically, if we go that route.  Plus monitors these days tend to be more wide than tall.

mpaperno:
I like the vertical layout personally, good idea!  :+1: 

Taranis outputs at default window width:
simu-taranis-outputs-win-b

After manual window resize:
simu-taranis-outputs-wide-win-b

New default Taranis window size:
simu-taranis-win-b

Horus outputs at default window width:
simu-horus-outputs-win-b

Horus outputs resized (Linux):
simu-horus-outputs-wide-linux-b

kilrah:
Nice!

Now if we could get the stick boxes to resize instead of force their current size I guess we'd solve all sizing issues!

schwabe:
if you commit that change for the vertical sliders I can check on os x

joaquins:
This may fix also https://github.com/opentx/opentx/issues/4010

mpaperno:
@schwabe If you could check this latest commit on OS X that would be great.  I'll get my hackintosh build set up soon, I promise!

schwabe:
barely fits (1218x800):

image

image

image

But you must have changed something else for the better since it does not crash with a too small window anymore:

image

elecpower:
I haven't been following the chatter in the lead up to this PR so apologies in advance if my points have been discussed.
On the overall dialogue size, just a reminder of the issues raised in the past about problems running Companion on laptops which grew with telemetry changes in 2.1. 
Like the idea of LS on same tab as Outputs. How about only displaying active LS which would require keeping the L prefix. This reduces dialogue size for the average user and has added advantage of seeing the change of state in a smaller visual area which is helpful for seeing temporary state changes. Or how about placing to the RHS of the sticks to reduce overall height.
Agree with progress bars for channel outputs if possible but would still need "indicator marks" to along the -100 to +100 range. Maybe the reason the slider was chosen. Vertical channels layout works good with resize capability.
Whilst revamping there is a request for Hold SF capability #3944.

companion/src/simulation/simulatordialog.cpp
```diff
@@ -275,8 +275,9 @@ void SimulatorDialog::initUi(T * ui)
 
   setupSticks();
 
-  resize(0, 0); // to force min height, min width
-  setFixedSize(width(), height());
+  // resize(sizeHint()); // to force min height, min width
```
bsongis: Would you completely remove those lines?

mpaperno:
And now fully resizable... these are shown at minimum size and compared to current `next` version.

simu-x9d-overlay-win-c

simu-9x-overlay-linux-c

simu-horus-overlay-win-c


But don't worry, you can still have BIG STICKS!  ;-)

simu-horus-big-linux-c


The next version will also remember and restore the size and position of your simulator windows.

Cheers,
-Max

kilrah:
Woohoo, fits easily on 720p 🥇 
Awesome job! 👍

mpaperno:
All you need now is a Tx module to attach to that little thing and a quick hack to send pulses from the simulator... :rofl:

companion/src/simulation/virtualjoystickwidget.cpp
```diff
@@ -0,0 +1,496 @@
...
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ */
+
+#include "virtualjoystickwidget.h"
+#include "../constants.h"
```
mpaperno: You're right, I had it like that before actually building the new class into the project config, so my IDE was complaining.  Definitely in the include path already.

projectkk2glider:
Valgrind seems happy (hard to inspect since there are so many Qt memory leaks) :+1:

companion/src/constants.h
```diff
@@ -59,10 +59,10 @@ enum BoardEnum {
...
 #define CPN_MAX_MOUSE_ANALOGS          2
 
-const char * const ARROW_LEFT = "\xE2\x86\x90";
-const char * const ARROW_UP = "\xE2\x86\x91";
-const char * const ARROW_RIGHT = "\xE2\x86\x92";
-const char * const ARROW_DOWN = "\xE2\x86\x93";
+const char * const ARROW_LEFT = "\u2190";
+const char * const ARROW_UP = "\u2191";
+const char * const ARROW_RIGHT = "\u2192";
+const char * const ARROW_DOWN = "\u2193";
```
kilrah: This produces a mountain of warnings on VC... French system language might have to do with it

`8>C:\opentx-22\code\companion\src\constants.h(64): warning C4566: character represented by universal-character-name '\u2192' cannot be represented in the current code page (1252) (compiling source file C:\opentx-22\code\companion\src\simulation\simulatorinterface.cpp)`

mpaperno: WTF... not very "Uni" now is it?  CP1252?  I guess we could add some define check... those arrows are much nicer than the old ones.  I'll do a search...   

Besides the warnings, does the display the arrows on the trim buttons?

mpaperno: Or wait.... never mind!  They were the same characters, just different encoding.  I'll change it back.

Once I searched it led me back to a result which I had already visited (so long ago that I'd forgotten of course)...   lol

projectkk2glider: Perhaps http://stackoverflow.com/a/12040593

mpaperno: `u8` also seems to work on MinGW & GCC Linux in US-English (didn't know about that, thanks).  I'm not sure if there's an advantage to one method over the other... if the previous one worked, I'm fine with that, just got confused by that encoding.

bsongis:
excellent job!

PR #4157, opened 2016-12-22
3djc/taranis draw number review

3djc:
Fix numerous issues  around drawNumber and drawText.

Please note that DBL numbers with a dot are 1 pixel larger, and that the dot in the font has been reduced to a single pixel.

Would appreciate some testing, including graphicaly complex lua


bsongis:
excellent job!

PR #4158, opened 2016-12-23
Protect subId

3djc:
prevent endless loop in sensor discovery if more than 7 is passed.

### DIFF
--- a/radio/src/lua/api_general.cpp
+++ b/radio/src/lua/api_general.cpp
@@ -791,7 +791,7 @@ static int luaPlayTone(lua_State * L)
 }
 
 /*luadoc
-@function luaPlayHaptic(duration, pause [, flags])
+@function playHaptic(duration, pause [, flags])
 
 Generate haptic feedback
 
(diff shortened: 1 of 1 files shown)


projectkk2glider:
This should be combined with https://github.com/opentx/opentx/tree/projectkk2glider/lua_fixes

3djc:
I have added those, Thanks a lot !

radio/src/lua/api_general.cpp
```diff
@@ -1040,23 +1040,23 @@ static int luaDefaultStick(lua_State * L)
...
 
 @status current Introduced in 2.2.0
 */
 static int luaSetTelemetryValue(lua_State * L)
 {
-  uint16_t id = luaL_checkinteger(L, 1);
-  uint8_t subId = luaL_checkinteger(L, 2);
-  uint8_t instance = luaL_checkinteger(L, 3);
+  uint16_t id = luaL_checkunsigned(L, 1);
+  uint8_t subId = luaL_checkunsigned(L, 2) & 0x7;
```
projectkk2glider: Cool, first I thought that this was not needed since compiler itself take care not to assign too big value to the `telemetrySensor.subId`, but the problem would be the call to `setTelemetryValue()` because there too big `subId` would be compared with existing sensors and of course it is never found.

radio/src/lua/api_general.cpp
```diff
@@ -1068,7 +1068,7 @@ static int luaSetTelemetryValue(lua_State * L)
     zname[2] = hex2zchar((id & 0x00f0) >> 4);
     zname[3] = hex2zchar((id & 0x000f) >> 0);
   }
-  
+
   int index = setTelemetryValue(TELEM_PROTO_LUA, id, subId, instance, value, unit, prec);
```
projectkk2glider: We need another check before the call of `setTelemetryValue()` the `id`, `subId` and `instance` must not be zero - this would then use the first empty sensor slot without actually allocating it as a new sensor!

radio/src/lua/api_general.cpp
3djc: Indeed, exactly the issue that Tadango faced

PR #4164, opened 2016-12-25
Standalone simulator startup options

mpaperno:
- Radio profile used for standalone Simulator can now be separate from Companion;
- Allow selecting a radio profile and EEPROM image file (in addition to radio type) at startup via GUI and CLI;
- All startup settings are saved between uses.
- Save and restore Simulator window sizes and positions (based on selected radio profile).
- Set startup throttle stick lock in standalone sim based on default mode in selected radio profile (until we can read that from the model settings).

simu_new_startup_linux

So the biggest change here is that you no longer have to go into Companion to switch radio profiles before starting sim with different radios.  Especially handy for developers & testers.  You still use Companion to create/edit profiles, as  before.

Though I ended up changing a lot of `simulator.cpp` in the process (hopefully for the better, but the diff isn't easy to follow).

-Max

simu-new-startup-help

### DIFF
--- a/companion/src/constants.h
+++ b/companion/src/constants.h
@@ -59,6 +59,18 @@ enum BoardEnum {
 #define CPN_MAX_KEYS                   32
 #define CPN_MAX_MOUSE_ANALOGS          2
 
+#define HEX_FILES_FILTER              "HEX files (*.hex);;"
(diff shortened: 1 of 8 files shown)


kilrah:
Really nice! 👍

bsongis:
Nice!

PR #4165, opened 2016-12-26
Do a backup when upgrading the eeprom

schwabe:
I am not sure if this has any other unwated side effects. 

This should help diagnose problems with the conversion.

### DIFF
--- a/radio/src/storage/eeprom_conversions.cpp
+++ b/radio/src/storage/eeprom_conversions.cpp
@@ -1261,7 +1261,11 @@ bool eeConvert()
   g_eeGeneral.contrast = 25;
 
   ALERT(STR_STORAGE_WARNING, msg, AU_BAD_RADIODATA);
+
+// Only do a backup if supported and if it is getting a unique filename
+#if defined(EEPROM_RLC) && defined(RTCLOCK)
   eepromBackup();
+#endif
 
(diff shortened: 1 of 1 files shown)


radio/src/storage/eeprom_conversions.cpp
```diff
@@ -1262,6 +1262,11 @@ bool eeConvert()
 
   ALERT(STR_STORAGE_WARNING, msg, AU_BAD_RADIODATA);
 
+// Only do a backup if supported and if it is getting a unique filename
+#if defined(EEPROM_RLC) && defined(RTCLOCK)
```
bsongis: I would suggest SDCARD instead of RTCLOCK

schwabe: Yes, but currently eeprombackup is only defined if EEPROM_RLC is defined. Also the 480x272 radio_version.cpp lacks the backup functionality. I think backup might not have been implemented for Horus yet.

kilrah:
Is the SD card even mounted at this point?

schwabe:
I think I tested it but I can test again to be sure.

schwabe:
okay checked the code. Seems sd card init is after this.

bsongis:
you could perhaps mount the SD card just before calling this function

schwabe:
yeah but let me test that on real hardware in order to be sure. I will reopen the pull request when I have a working version

PR #4169, opened 2016-12-28
Clean up build warnings

mpaperno:
This fixes all build warnings I could find on various platforms and different targets (especially OS X).

There is one set of 9 bogus linker warnings remaining with avr-gcc < 4.8.3 (eg. on Travis build),  which look something like:

`opentx.cpp:2102:1: warning: ‘_vector_16’ appears to be a misspelled interrupt handler [enabled by default]`

This is fixed in avr-gcc 4.8.3, see: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=59396  

BTW the firmware also builds OK with avr-gcc 4.9.2 from Arduino bundle (or presumably from Atmel toolchain itself, which Arduino uses http://www.atmel.com/tools/ATMELAVRTOOLCHAINFORLINUX.aspx)

-Max


companion/src/firmwares/er9x/er9xinterface.h
```diff
@@ -43,7 +43,7 @@ class Er9xInterface : public EEPROMInterface
 
     virtual unsigned long load(RadioData &, const uint8_t * eeprom, int size);
 
-    virtual unsigned long loadBackup(RadioData &, uint8_t * eeprom, int esize, int index);
+	 virtual unsigned long loadBackup(RadioData &, const uint8_t * eeprom, int esize, int index);
```
projectkk2glider: Use 4 spaces instead of TAB for indentation.

companion/src/firmwares/ersky9x/ersky9xinterface.h
```diff
@@ -43,7 +43,7 @@ class Ersky9xInterface : public EEPROMInterface
 
     virtual unsigned long load(RadioData &, const uint8_t * eeprom, int size);
 
-    virtual unsigned long loadBackup(RadioData &, uint8_t * eeprom, int esize, int index);
+	 virtual unsigned long loadBackup(RadioData &, const uint8_t * eeprom, int esize, int index);
```
projectkk2glider: Indentation problem

companion/src/simulation/telemetrysimu.h
```diff
@@ -121,7 +121,7 @@ class TelemetrySimulator : public QDialog
       bool stepping;
       QDateTime parseTransmittterTimestamp(QString row);
       void calcLogFrequency();
-      int32_t replayRate;
+      //int32_t replayRate;
```
projectkk2glider: Completely remove all commented out variables in this file.

companion/src/storage/mountlist.cpp
```diff
@@ -227,7 +227,7 @@
 
 #ifdef MOUNTED_GETMNTINFO
 
-# if ! HAVE_STRUCT_STATFS_F_FSTYPENAME
+# if ! HAVE_STRUCT_STATFS_F_FSTYPENAME && !defined(__clang__)
```
projectkk2glider: ~~Not your mistake, but loose space between `!` and `HAVE...`~~ Dont bother, this is thirdparty code.

radio/src/gui/128x64/model_outputs.cpp
```diff
@@ -143,7 +143,7 @@ void menuModelLimits(event_t event)
     lcdDrawChar(LIMITS_DIRECTION_POS, y, swVal);
 #endif
 
-    int8_t limit = (g_model.extendedLimits ? LIMIT_EXT_MAX : 100);
+    int limit = (g_model.extendedLimits ? LIMIT_EXT_MAX : 100);
```
projectkk2glider: Nice, this looks like it was actually a bug on X7D.

radio/src/gui/212x64/view_main.cpp
```diff
@@ -410,14 +410,14 @@ void displaySwitch(coord_t x, coord_t y, int width, unsigned int index)
...
-  { "VERSION", menuRadioVersion }
-};
+//const MenuItem MAIN_MENU[] = {
+//  { "RADIO SETTINGS", menuRadioSetup },
+//  { "MODEL SELECT", menuModelSelect },
+//  { "MODEL SETTINGS", menuModelSetup },
+//  { "CHECKLIST", menuModelNotes },
+//  { "SD MANAGER", menuRadioSdManager },
+//  { "VERSION", menuRadioVersion }
+//};
```
projectkk2glider: Remove completely.

radio/src/rtc.h
```diff
@@ -27,6 +27,9 @@
 #define SECS_PER_DAY    86400ul
 #define TM_YEAR_BASE    1900
 
+#define TIME_T_MIN      (-__LONG_MAX__)
```
projectkk2glider: What is wrong with the original definition `# define TIME_T_MIN TYPE_MINIMUM (gtime_t)`?

radio/src/storage/eeprom_rlc.h
```diff
@@ -129,7 +129,7 @@ class RlcFile: public EFile
     uint8_t  m_bRlc;      // control byte for run length decoder
     uint8_t  m_zeroes;
 
-    uint8_t m_flags;
+    //uint8_t m_flags;
```
projectkk2glider: Completely remove

companion/src/firmwares/er9x/er9xinterface.h
mpaperno: Bah, I thought I checked it all... Mac editor defaulted to tabs.  :(

radio/src/rtc.h
projectkk2glider: Ok

mpaperno:
I added a [commit ](https://github.com/opentx/opentx/commit/629e462c0deb2ac3491fd60558f2e7dd02e12c55)on a separate branch which fixes the linker warnings by updating to AVR GCC 4.9.2.  It passes all tests, but I didn't want to mix it into this PR in case sticking to avr-gcc 4.8 is important.

radio/src/functions.cpp
```diff
@@ -376,9 +376,7 @@ void evalFunctions()
           case FUNC_SET_TIMER:
             timerSet(CFN_TIMER_INDEX(cfn), CFN_PARAM(cfn));
             break;
-#endif
```
bsongis: It was better before (each case surrounded by its #ifdef if needed, no nested #ifdef), no?

radio/src/gui/128x64/model_outputs.cpp
bsongis: Yes it is indeed better for X7D, but an int is at least 16bits, this will increase flash usage on 8bits platforms!

radio/src/gui/212x64/view_main.cpp
```diff
@@ -410,15 +410,6 @@ void displaySwitch(coord_t x, coord_t y, int width, unsigned int index)
   }
 }
 
-const MenuItem MAIN_MENU[] = {
```
bsongis: interesting, I cannot even remember what was the purpose of this!

radio/src/gui/gui_common_arm.cpp
```diff
@@ -278,16 +278,16 @@ bool isSwitchAvailable(int swtch, SwitchContext context)
...
       return false;
     }
+    negative = true;
     swtch = -swtch;
   }
 
 #if defined(PCBSKY9X)
   if (swtch >= SWSRC_FIRST_SWITCH && swtch <= SWSRC_LAST_SWITCH) {
-    negative = negative;
+    (void)negative;
```
bsongis: I think we have a macro for that, perhaps `UNUSED(...)`

radio/src/functions.cpp
mpaperno: Seemed redundant to me, and it would have been a 3rd repetition to also include the RANGECHECK/BIND case (below) which was causing warnings.  But either way would work.

radio/src/gui/128x64/model_outputs.cpp
mpaperno: Do you want the var size based on `#if defined(CPUARM)` ?

radio/src/gui/gui_common_arm.cpp
mpaperno: `UNUSED()` is only defined in the third-party Lua source code.  Didn't seem relevant to include it here.  But it does the same thing.
`#define UNUSED(x)	((void)(x))`

mpaperno:
If you check the [build log](https://travis-ci.org/opentx/opentx/builds/187345730) for my "avr492" branch you'll see there is not a single compiler warning.  So yes to -Werror and let people clean up their own warnings from here!  ;-)    W/out avr-gcc 4.9.2 then only on ARM platforms.

radio/src/gui/128x64/model_outputs.cpp
bsongis: I wonder if we have already a type to avoid another #ifdef here

radio/src/functions.cpp
bsongis: Perhaps you could forbid DANGEROUS_MODULE_FUNCTIONS on !ARM boards in CMakeLists.txt, it would avoid the repetition

radio/src/gui/gui_common_arm.cpp
bsongis: no, don't worry

bsongis:
Yes, this is good. We just have to make sure that binaries produced by avr-gcc 4.9 are not bigger than those produced by our good old avr-gcc 4.7 (to avoid flames in forums / issues)

radio/src/functions.cpp
mpaperno: The CMake config is the "front end" for the user, so yes there could/should be more dummy-checking in there probably (I ran into some other situations when building different targets in random order, eg. with SDCARD or LUA being still defined on unsupported targets). I think that's a different topic.  

OTOH, if it is important to dummy-check the definitions at build-time I'd use conditional macro checks with #error/warning statements as needed, and not rely on whatever the current build system did or didn't do.  Otherwise it should just quietly build what it was told.  IMHO  :)

bsongis: We improve things little after little :)

mpaperno:
A bit larger with 4.9, yes.  Does it matter for the Travis test build? Granted ideally the same toolchain would be used, but OTOH maybe checking for new warnings is more important.  Would be nice if Travis showed a warnings report like Jenkins can.

```
> avr-size.exe firmware-9x-4.8.1.elf
   text    data     bss     dec     hex filename
  54000      20    2882   56902    de46 firmware-9x-4.8.1.elf

> avr-size.exe firmware-9x-4.9.2.elf
   text    data     bss     dec     hex filename
  56014      20    2882   58916    e624 firmware-9x-4.9.2.elf
```
```
$ avr-size firmware-9x-4.8.2.elf
   text	   data	    bss	    dec	    hex	filename
  52582	     20	   2876	  55478	   d8b6	firmware-9x-4.8.2.elf


$ avr-size firmware-9x-4.9.2.elf
   text    data     bss     dec     hex filename
  54654      20    2876   57550    e0ce firmware-9x-4.9.2.elf
```

mpaperno:
I also saw a suggestion somewhere (regarding this very problem) to omit `-Werror` in just the linker flags but keep it in the compiler flags (somehow).  That may be another option.

radio/src/gui/128x64/model_outputs.cpp
mpaperno: Uh, yeah, right where LIMIT_EXT_MAX was defined... d'oh!
`#define limit_min_max_t     int8_t`

Will fix!

bsongis: Yes it will fit perfectly

bsongis:
2kBytes? It's a huge regression!

fred99:
Sorry to bother you but I got an error during compilation under W10 using the dev env describe here: https://github.com/opentx/opentx/wiki/OpenTX-Windows-Build-Instructions
since radio/src/rtc.h modification:
```
@@ -27,6 +27,9 @@

#define SECS_PER_DAY 86400ul 

#define TM_YEAR_BASE 1900 



+#define TIME_T_MIN (-__LONG_MAX__)
+#define TIME_T_MAX (__LONG_MAX__)
```

=> Compilation :
```
7>C:\Users\devfr\Desktop\opentx\opentx-next\radio\src\rtc.cpp(190): error C2065: '__LONG_MAX__': undeclared identifier
7>C:\Users\devfr\Desktop\opentx\opentx-next\radio\src\rtc.cpp(227): error C2065: '__LONG_MAX__': undeclared identifier
7>C:\Users\devfr\Desktop\opentx\opentx-next\radio\src\rtc.cpp(236): error C2065: '__LONG_MAX__': undeclared identifier
7>C:\Users\devfr\Desktop\opentx\opentx-next\radio\src\rtc.cpp(237): error C2065: '__LONG_MAX__': undeclared identifier
7>C:\Users\devfr\Desktop\opentx\opentx-next\radio\src\rtc.cpp(238): error C2065: '__LONG_MAX__': undeclared identifier
7>C:\Users\devfr\Desktop\opentx\opentx-next\radio\src\rtc.cpp(340): error C2065: '__LONG_MAX__': undeclared identifier
7>C:\Users\devfr\Desktop\opentx\opentx-next\radio\src\rtc.cpp(377): error C2065: '__LONG_MAX__': undeclared identifier
7>C:\Users\devfr\Desktop\opentx\opentx-next\radio\src\rtc.cpp(378): error C2065: '__LONG_MAX__': undeclared identifier
```

Corrected with :
```
#include <limits.h>

#define SECS_PER_HOUR   3600ul
#define SECS_PER_DAY    86400ul
#define TM_YEAR_BASE    1900

#define TIME_T_MIN      (-LONG_MAX)
#define TIME_T_MAX      (LONG_MAX)
```

fred99:
And thanks a lot for all improvment you are adding to this 2.2

mpaperno:
Thank you Fred, fixed now and seems happy in all compilers tested... gcc, arm-gcc, avr-gcc, mingw-gcc, msvc & clang.  :)

fred99:
I am very happy to help a little bit.
I appreciate the new possibilities of interactivity of the simulator with mouse and KBD.
Works fine in standalone, with mouse from Companion.

PR #4173, opened 2016-12-30
If no WIZARD are present, do not display anything

### LINKED ISSUE #4170: X12S New Model wizard - disable
I am not a huge fan of the wizard method on opentx.   Prefer the manual creation of models.

In the past - simply renaming the 'WIZARD' folder to something like 'WIZARDX' would disable the functionality.

On the X12S, doing this only breaks the functionality.  You get a wizard prompt - but no content in the wizard.  

Is there a magic way to turn off the wizard?  or is it intentional for it to now be mandatory?

3djc:
This fixes #4170


radio/src/gui/480x272/model_select.cpp
```diff
@@ -140,6 +140,10 @@ bool menuModelWizard(event_t event)
 
   if(wizardNumber == 0) {
```
bsongis: space here

radio/src/gui/480x272/model_select.cpp
```diff
@@ -139,7 +139,11 @@ bool menuModelWizard(event_t event)
   char wizpath[MAX_WIZARD_NAME_LEN];
 
   if(wizardNumber == 0) {
-    wizardNumber = getWizardNumber();
+    wizardNumber = getWizardCount();
```
bsongis: wizardCount :)

radio/src/gui/480x272/model_select.cpp
```diff
@@ -121,25 +121,29 @@ uint8_t getWizardNumber()
...
   }
-  return wizNbr;
+  return wizCnt;
 }
 
 bool menuModelWizard(event_t event)
 {
   static uint8_t wizardSelected;
-  static uint8_t wizardNumber;
+  static uint8_t wizardCnt;
```
bsongis: why don't you initialize it here:
`wizardCnt = getWizardCount();`

radio/src/gui/480x272/model_select.cpp
```diff
@@ -121,25 +121,29 @@ uint8_t getWizardNumber()
...
       }
     }
   }
-  return wizNbr;
+  return wizCnt;
 }
 
 bool menuModelWizard(event_t event)
 {
   static uint8_t wizardSelected;
```
bsongis: it should be initialized, no?

PR #4175, opened 2016-12-30
Simulator key/mouse mapping improvements

mpaperno:
- Make key/mouse navigation more consistent between radios and ensure keyboard events are always properly detected (eg. after clicking on a control);
- Fix mouse wheel direction for rotary encoder emulation (fixes #4160);
- Fix page up/down buttons on Horus (closes #4160);
- Horus rotary encoder can now be controlled with +/- keys (also keeps X/C keys, otherwise supersedes #4087 and fixes key repeat issue);
- DEL key now another alternative for ESC/BKSP (brings all controls to number pad);
- "=" key now acts same as "+" key (+ requires SHIFT in main row on most US keyboards);
- Fix controls direction when using mouse wheel over RC switches;
- Add help text showing keyboard mappings for each radio type (activate on F1 key -- a button will be added in future changes).
- Enable mouse wheel scrolling on radios w/out a rotary encoder (acts as +/- keys on X9D, PU/DN keys on 9X).

Full help text and 9X keymap:
simu-keys-9x
X9 keymap
simu-keys-x9_crop
Horus keymap:
simu-keys-horus_crop

PR #4176, opened 2016-12-30
Fix switch diagnostics display on 9X family.

mpaperno:
Did this ever work?  Pressing "keys" made the "switch" indicators light up also.  Thought I broke something with my keymapping changes (#4175), but no. Tested on an actual stock Tgy. 9X (which I "forgot" I had!).  

The #ifdef is not strictly necessary, but clarifies the new function isn't used for X7 (sorta... X7 is "PCBTARANIS" also I assume).


bsongis:
👍

PR #4180, opened 2016-12-31
Narrow gap between previous char and up and down arrow in STDSIZE

3djc:
Reduce blank gap before up and down arrow by a couple of pixels


projectkk2glider:
I think changing `.specs` is not the correct way, this file is generated by the `font2png.py` script so it will eventually get overwritten.

3djc:
Further investigation

PR #4181, opened 2016-12-31
Add missing units

### LINKED ISSUE #4174: OTX 2.2: RSSI units & scale / ratio parameters changed when reading from Taranis X9D+ into OTX Companion
Flashed OpenTX 2.2 and Betaflight 3.1 to try out live PID tuning. I noticed that the RSSI values displayed vai FrSky telemetry have changed significantly, Taranis reads RSSI = 1 when being close.

I bumped the scaling up by 100, and now the close-by RSSI is around 30. Since Taranis also has a warning level, I'm not sure what to think about it: Have you changed the scaling for a good reason? Or is it simply a bug?

3djc:
Companion was not updated when mWatt and Rad where added

This fixes #4174

### DIFF
--- a/companion/src/eeprominterface.cpp
+++ b/companion/src/eeprominterface.cpp
@@ -193,6 +193,8 @@ QString SensorData::unitString() const
       return QObject::tr("mAh");
     case UNIT_WATTS:
       return QObject::tr("W");
+    case UNIT_MILLIWATTS:
+        return QObject::tr("mW");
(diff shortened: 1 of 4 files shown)


RipperDrone:
Wow, that was a quick one, thank you. The enumeration obviously caused the glitch btw. RC and Companion (mW unit missing in btw). Ultra-fast spotting, great work :-)

PR #4183, opened 2017-01-01
3djc/horus narrow arrows gap

3djc:
Please check thoroughly

I could not find when font2png.py was invoqued in a regular build, but just in case I modified it, and used it to generate updated font

The result of this change is that up and down arrow are two pixel more left, making switch position look better on Horus


radio/util/font2png.py
```diff
@@ -91,6 +91,12 @@ def getFontTopBottom():
             pass
         elif c == "." and fontsize <= 8:
             painter.drawPoint(width, fontsize)
+        elif c == u"↑" or c == u"↓":
+            rect = metrics.boundingRect(c)
+            if fontsize >= 24:
+                painter.drawText(width - rect.left() + 1, fontsize - 2, c)  # fontsize-bottom+1 -17 / 7
+            else:
+                painter.drawText(width - 1, fontsize, c)  # fontsize-bottom+1 -17 / 7
```
projectkk2glider: I do not like the code duplicaton here. Why not add something like this in line 105:
```
adjustment = -1 if c == u"↑" or c == u"↓" else 1
painter.drawText(width + adjustment, fontsize, c)
```

projectkk2glider: Ignore my advice, I looked only at the one changeset. Your final solution is OK.

projectkk2glider:
Just a reminder, we should really write some gtests for the big display (lines, fonts, rectangles, colors, etc..)

bsongis:
The python script is not called at each build because the result is different depending on the system you are (Windows behaves differently)

bsongis:
Perfect, thanks!

PR #4189, opened 2017-01-02
Fixes #4153: Companion Outputs alts to sub trim, min and max not prompting to save

### LINKED ISSUE #4153: Companion changes to model outputs sub trim, min and max not prompting to save changes
Changing any of the fields within the sub trim, min and max columns does not prompt to save on file close or application exit. There is no asterisk indicator in the models list dialogue title bar.

Companion 2.1.9 on Ubuntu 14.04

elecpower:
Fixes #4153 Companion Outputs - not all field changes triggering save prompt


bsongis:
Thanks!

PR #4195, opened 2017-01-04
3djc/horus checklist improvement

### LINKED ISSUE #4190: Horus calling the texts  .txt for a model does not work
Problem with OpenTx Horus

For each model, a note can be displayed at startup. This works only with the first model.
Next model no more

Raimund

3djc:
- current way of working (like Taranis stays)
- in addition to that, you will also be able to use the modelxx.txt (even if your plane is called 'MySuperPlane')
- in case both are present (you should not be doing that, but just in case), the taranis style will have priority

This fixes #4190

### DIFF
--- a/radio/src/gui/gui_common_arm.cpp
+++ b/radio/src/gui/gui_common_arm.cpp
@@ -568,18 +568,19 @@ bool modelHasNotes()
   char filename[sizeof(MODELS_PATH)+1+sizeof(g_model.header.name)+sizeof(TEXT_EXT)] = MODELS_PATH "/";
   char *buf = strcat_currentmodelname(&filename[sizeof(MODELS_PATH)]);
   strcpy(buf, TEXT_EXT);
-#if defined(PCBHORUS)
   if (isFileAvailable(filename)) {
     return true;
   }
-  else {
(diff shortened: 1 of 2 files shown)


radio/src/gui/gui_common_arm.cpp
```diff
@@ -568,7 +568,18 @@ bool modelHasNotes()
   char filename[sizeof(MODELS_PATH)+1+sizeof(g_model.header.name)+sizeof(TEXT_EXT)] = MODELS_PATH "/";
   char *buf = strcat_currentmodelname(&filename[sizeof(MODELS_PATH)]);
   strcpy(buf, TEXT_EXT);
+#if defined(PCBHORUS)
+  if (isFileAvailable(filename)) {
```
bsongis: I would keep this if for all boards

radio/src/gui/gui_common_arm.cpp
```diff
@@ -568,7 +568,18 @@ bool modelHasNotes()
...
+    return true;
+  }
+  else {
```
bsongis: ... and here this block could go in `#if !defined(EEPROM)`

bsongis:
excellent, thanks!

Issue #4179, opened 2016-12-31
translate companion to chinese, need a companion_zh_CN.ts file


Kongse:
Hi i'm trying to  to translate companion to chinese. 

At #712 I've seen some guide to do this, but seems  a companion_zh_CN.ts file should be created first. 

I'm new about github, so what should I do next as a translator, should I follow the guide in "OpenTX Git Development Workflow" exactly?

thanks a lot.

projectkk2glider:
Is it possible to display Chinese texts using western charactes? If not (which I think is true), then this will be almost impossible. Current OpenTX code only has western fonts defined. Current characters can be seen here:

https://raw.githubusercontent.com/opentx/opentx/next/radio/src/fonts/480x272/font_midsize.png

kilrah:
He seems to be interested in translating Companion only, and posted on the forum that it works (he inserted Chinese translations in the italian file and they seem to display).

projectkk2glider:
:blush: Missed that.

Kongse:
Yes the companion can display Chinese but firmware seems impossible. 

It  comes a problem that when Chinsese users use a translated companion software to edit models, they may feel unsure what items on Radio corresponding to the items on companion. 

It can be partly solved by while translating the companion to Chinese, showing English words in brackets beside the translation words. I've tested it before (as kilrah mentioned above using italian ts file) and seems working well. 

So when they edit one item on companion, they know what's the item exactly corresponding to on the radio. And by cross reference, a chinese user unknown about English would learn to user Opentx radios faster.

kilrah:
There you go:
https://github.com/opentx/opentx/tree/next/companion/src/translations

Kongse:
Thank you all too :), I'm working on the file.

kilrah:
Translation included but somehow not working, still need to fix

kilrah:
Fixed

PR #4196, opened 2017-01-04
#4179 cn translation done

Kongse:
translation finished

### DIFF
--- a/companion/src/translations/companion_cn.ts
+++ b/companion/src/translations/companion_cn.ts
@@ -1,32 +1,35 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!DOCTYPE TS>
-<TS version="2.1">
+<TS version="2.1" language="zh_CN">
 <context>
     <name>AileronsPage</name>
     <message>
         <location filename="../wizarddialog.cpp" line="432"/>
         <source>No</source>
-        <translation type="unfinished"></translation>
+        <translatorcomment>模型向导中副翼连线方式的设定</translatorcomment>
+        <translation>没有副翼</translation>
     </message>
     <message>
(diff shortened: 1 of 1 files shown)


bsongis:
Thanks!

Kongse:
LOL :)

projectkk2glider:
I checked some of the translation with Google Translate, it appears good. @Kongse thanks for this. But remember to revisit this periodically and update any changed/added strings in the future.

kilrah:
A few things:

- I see that in the file there is often some text that comes up in "translator comments". Understand this will never be seen by users, this is only for translators to make notes to refer to later and will only appear inside of QT Linguist.

- It builds, but if I select Chinese in the menus I get English language. I guess that's becasue my system has no idea about Chinese characters. @Kongse , can you test this build?
http://www.andrebernet.ch/divers/opentx/companion-windows-2.2.0-cntest.exe

Kongse:
@projectkk2glider Ok i would :)

@kilrah 

- Yes the comments is intended for examining and quick searching.

- Yes I've tested the build but got the same problem. When translating I renamed the translation file to italian ts file to do the job. I've thought it is intended, and will be corrected after the translation file commit, but now it seems because some code is missing.  I'm not familiar with coding so can not figure out where is the problem.

kilrah:
Nothing seems to be missing, and I can also get it to work when swapping files... really don't understand. Will merge and we'll try to find out later

bsongis:
swapping which files?

kilrah:
Forget it, fixed... thought the 2nd code was the significant one, but it's the first

Kongse:
@kilrah thank you :)

PR #4201, opened 2017-01-05
3djc/x7 custom scripts

3djc:
Adds CUSTOM SCRIPTS screen to X7 when selecting LUA=YES

Due to reduced screen size, the display of a detailed LUA script is now only on one column for X7

image


radio/src/gui/128x64/gui.h
```diff
@@ -333,6 +333,8 @@ void editSingleName(coord_t x, coord_t y, const pm_char * label, char * name, ui
 uint8_t editDelay(coord_t y, event_t event, uint8_t attr, const pm_char * str, uint8_t delay);
 #define EDIT_DELAY(x, y, event, attr, str, delay) editDelay(y, event, attr, str, delay)
 
+void copySelection(char * dst, const char * src, uint8_t size);
```
bsongis: why is this header needed?

3djc: It's not anymore :)

bsongis:
Thanks!

PR #4202, opened 2017-01-06
Build system updates, including -Werror on TravisCI

mpaperno:
This is a bunch of small changes grouped together, mostly to the CMake config/scripts.  A lot of consolidation and making the cmake config a bit more user-friendly.

Especially this enables treating warnings as errors on Travis build.  @bsongis , this version relies on updating the AVR compiler version, as we discussed.  Although we can easily bypass -Werror on the problematic linker instead, if you'd prefer.

With WARNINGS_AS_ERRORS now all the targets get -Werror, before this it was just the firmware builds which were affected.

Also:
 * Consolidate some C/CXX flag settings for firmware target, avoid duplication/redundancy;
 * All firmware/simu warning flags are now passed to linker stage (GCC only);
 * Add a "build report" for each main target showing compiler, flags, and defines, with verbosity controlled by new VERBOSE_CMAKELISTS option;
 * Added properties to PCB, TRANSLATIONS, and a few other options (makes option selector lists in cmake-gui/QtCreator);
 * Add macro for conditionally adding C++11 flag;
 * Silence CMake >3.0 warnings about CMP0054 policy not being set;
 * Add FIRMWARE_C[XX}_FLAGS[_DEBUG] options for user to specify additional compiler flags for firmware builds (the default CMAKE_C[XX]_FLAGS[_*] are not used, as before);
 *  Silence lots of superfluous "CRT_SECURE" series MSVC warnings (MSVC builds almost clean, I have separate patch to fix the few remaining warnings);

If you're curious, the new "build report" in CMake scripts looks somewhat like this when the verbose option is enabled (there are also reports for `simulator` and `simu/libsimulator`):
```
TARGET firmware: cpp compiler arm-none-eabi-g++ v4.7.4
	with cpp flags:  -mcpu=cortex-m4 -mthumb -fomit-frame-pointer -fverbose-asm -Wa,-ahlms=firmware.lst -Os -gdwarf-2 -DHSE_VALUE=12000000 -fno-exceptions -fdata-sections -ffunction-sections  -Wall -Wno-strict-aliasing -Wformat -Wreturn-type -Wunused -Wuninitialized -Wunknown-pragmas -Wno-switch -Wtype-limits -Werror  
	with link flags: -lm -TD:/Devel/RC-Tx/OpenTx/2.2/radio/src/targets/horus/stm32f4_flash.ld -Wl,-Map=firmware.map,--cref,--no-warn-mismatch,--gc-sections
	with defs: PCBHORUS; PCBREV=13; STM32F429_439xx; SDRAM; COLORLCD; PPM_PIN_UART; PPM_PIN_TIMER; EEPROM_VARIANT=0; AUDIO; VOICE; RTCLOCK; GPS_USART_BAUDRATE=9600; DISK_CACHE; INTERNAL_GPS; DBLKEYS; STM32F4; STM32; LUA_INPUTS; VARIO; SBUS; CROSSFIRE; USB_MASS_STORAGE; TIMERS=3; LOG_TELEMETRY; HAPTIC; CPUARM; VIRTUAL_INPUTS; TELEMETRY_FRSKY; TELEMETRY_FRSKY_SPORT; FRSKY_HUB; GPS; PXX; DSM2; BOLD_FONT; BATTGRAPH; THRTRACE; GAUGES; REQUIRED_SDCARD_VERSION="2.2V0004"; TRANSLATIONS_EN; TRANSLATIONS="EN"; FLAVOUR="horus"; RAMBACKUP; LUA; LUA_COMPILER; LUA_SCRIPT_LOAD_MODE="bt"; LUA_MODEL_SCRIPTS; HELI; FLIGHT_MODES; CURVES; GVARS; AUTOSOURCE; AUTOSWITCH; SIMU_AUDIO; SDCARD; PPM_UNIT_US; PPM_CENTER_ADJUSTABLE; PPM_LIMITS_SYMETRICAL; OVERRIDE_CHANNEL_FUNCTION; DANGEROUS_MODULE_FUNCTIONS; SPLASH; TRACE_SIMPGMSPACE; GUI; CORRECT_NEGATIVE_SHIFTS
--------------
TARGET companion: cpp compiler E:/Devel/Qt/Tools/mingw530_32/bin/g++.exe v5.3.0
	with cpp flags:  -mno-ms-bitfields  -Wall -Werror  
	with link flags: 
	with defs: ALLOW_NIGHTLY_BUILDS; WIN_USE_CONSOLE_STDIO; JOYSTICKS; SIMU; QXT_STATIC; QT_TRANSLATIONS_DIR=""
--------------
```

Thanks,
-Max


CMakeLists.txt
```diff
@@ -28,11 +31,15 @@ set(CMAKE_CXX_STANDARD 11)
...
 set(COMPANION_SRC_DIRECTORY ${PROJECT_SOURCE_DIR}/companion/src)
+set(SIMU_SRC_DIRECTORY ${COMPANION_SRC_DIRECTORY}/simulation)
 
 set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR})
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR})
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${PROJECT_SOURCE_DIR}/cmake)
 
+# options shared by all targets
+option(VERBOSE_CMAKELISTS "Show extra information while processing CMakeLists.txt files." OFF)
+option(WARNINGS_AS_ERRORS "Treat all compiler warnings as error." OFF)
```
bsongis: as error**s**?

bsongis:
While you are at it, would it be possible to reduce the logs on the Travis build?
i.e. avoid one line per file

CMakeLists.txt
mpaperno: That would probably be better.  Or `Treat any compiler warning as an error.`  It only takes one  :)

projectkk2glider:
This is so cool! I always missed something like VERBOSE_CMAKELISTS and had to resort to `make VERBOSE=1`.

mpaperno:
@bsongis Yes, we can make it be quieter by adding `-DCMAKE_RULE_MESSAGES=OFF` to the cmake command line.  This only prints one status for every target built (and some "scanning dependencies" lines).  It's actually kinda nice for a change  :)  I'll add that.

What about the tests output?  Do we need to see all the success statuses? Or just that the test is being run, and of course if error.  Actually I don't know much about gtest... is there even an option?  I can't seem to find one.  Only to redirect all the output to a file, which isn't very useful here.  :-/

@projectkk2glider Glad you like the report, I really dig seeing what the config is before building (VERBOSE=1 is a nice option but a bit much for every-day use).  May also help diagnose build/config issues for other users (ask ppl to paste the build config report in the Issue).

CMakeLists.txt
bsongis: even better yes!

bsongis: I was just afer consistency between warnings (plural) and error (singular)

bsongis:
Yes there is an option for gtests to avoid any output (perhaps only dots). I don't have it on the top of my head

bsongis:
I also used VERBOSE=1, cool there is something beween 0 and 1!

CMakeLists.txt
```diff
@@ -39,7 +39,7 @@ set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${PROJECT_SOURCE_DIR}/cmake)
 
 # options shared by all targets
 option(VERBOSE_CMAKELISTS "Show extra information while processing CMakeLists.txt files." OFF)
-option(WARNINGS_AS_ERRORS "Treat all compiler warnings as error." OFF)
+option(WARNINGS_AS_ERRORS "Treat any compiler warning as an error (adds -Werror flag)." OFF)
```
bsongis: adds or add?
:)

bsongis:
Thanks!

CMakeLists.txt
mpaperno: As in "enabling this option adds -Werror flag."  But of course it could also be "enabling this option will add -Werror flag."  So many options...   :)

PR #4199, opened 2017-01-04
Projectkk2glider/flash savings

projectkk2glider:
This is not very well tested, especially the changes in the Lua global structures. But I think it should be OK.

PR #4203, opened 2017-01-06
Fix initializer list issue with ZoneOptionValue

mpaperno:
Fix initializer list issue with ZoneOptionValue (technically not C++11 [[1](http://stackoverflow.com/questions/18731707/why-does-c11-not-support-designated-initializer-list-as-c99)]), now works x-platform as well, lowers Flash use; Spit out Zone, ZoneOption, ZoneOptionValue to own unit. Fixes MSVC build (https://github.com/opentx/opentx/pull/4199/files#r94867147)

I wasn't sure if maybe the "My Label" default value should be translated... or why it was in that "ZCHAR format" specifically.  But that can be changed separately if need be.

~~Firmware size report~~  (corrected version is below)


schwabe:
600 byte increase in .data. That is 600 byte more RAM used. That is counter productive, even if it saves 1,2 kB of flash.

kilrah:
Umm isn't that flash used for the initializer?

The RAM is allocated anyway, it's just the way it's initialized that's different and thus showing up in another section.

schwabe:
well .data is the section that copied from flash to ram on init. 

Most other strings that are read only are in .text and read from flash when they needed and do not use RAM.

The usage of .data in RAM  is not reflected by the totals :/

kilrah:
That's what I mean, the RAM was used already, it was jsut filled up later by code during program execution instead of being initialized by the C lib.

mpaperno:
That's true, my size script didn't properly take into account .data also being in CCM (I'm used to a slightly different loader script).

I'm not sure yet where the 600B are going, seems strange. Tried inlining the new setZoneOptionValue functions but that made it a tiny bit worse.  And yes the string would eventually be stored in RAM regardless since the options aren't const. 

OTOH there is a lot of SRAM which seems under-utilized  😄    

The corrected report:
Before:
```
Size report for firmware.elf

Section Details:
section         size(B)               addr      loc
.text           1329184         0x08000000      F
.ARM.extab          376         0x08144820      F
.ARM                512         0x08144998      F
.init_array         108         0x08144b98      F
.fini_array           4         0x08144c04      F
.data              2772         0x10000000      F+C
.bss              40868         0x10000ad8      C
.ram              16576         0x20000000      S1
._user_heap_stack  1024         0x200040c0      S1
.sdram           783360         0xd0000000      SD

Totals:         KB        usage of ttl  KB free
 Flash:    1301.71       63.56%   2048   746.29
   CCM:      42.62       66.59%     64    21.38
 SRAM1:      17.19       15.35%    112    94.81
 SRAM2:       0.00        0.00%     16    16.00
 SDRAM:     765.00        9.34%   8192  7427.00
```
After:
```
Section Details:
section         size(B)               addr      loc
.text           1327424         0x08000000      F
.ARM.extab          376         0x08144140      F
.ARM                512         0x081442b8      F
.init_array         108         0x081444b8      F
.fini_array           4         0x08144524      F
.data              3348         0x10000000      F+C
.bss              40868         0x10000d18      C
.ram              16576         0x20000000      S1
._user_heap_stack  1024         0x200040c0      S1
.sdram           783360         0xd0000000      SD

Totals:         KB        usage of ttl  KB free
 Flash:    1300.56       63.50%   2048   747.44
   CCM:      43.18       67.47%     64    20.82
 SRAM1:      17.19       15.35%    112    94.81
 SRAM2:       0.00        0.00%     16    16.00
 SDRAM:     765.00        9.34%   8192  7427.00
```

radio/src/gui/480x272/topbar.h
```diff
@@ -22,6 +22,7 @@
 #define _TOPBAR_H_
 
 #include "widgets_container.h"
+#include "zone.h"
```
bsongis: not needed

radio/src/gui/480x272/widgets_container.h
```diff
@@ -23,6 +23,7 @@
 
 #include <stdlib.h>
 #include "widget.h"
+#include "zone.h"
```
bsongis: not needed

radio/src/gui/480x272/zone.h
```diff
@@ -0,0 +1,54 @@
+
+#ifndef _ZONE_H_
+#define _ZONE_H_
+
+#include <inttypes.h>
+#include <string.h>
+
+#define LEN_ZONE_OPTION_STRING         8
+
+#define OPTION_VALUE_UNSIGNED(x)    setZoneOptionValue((uint32_t)x)
```
bsongis: but this will use RAM, no?

mpaperno: Yes... trying to reduce the usage... open to ideas  :)   Since that ZoneOptionValue union is used directly in some packed structs (eg. Theme data in RadioData) it kinda limits the options, seems like.

bsongis: I see no issue with the solution I found and was removed by this commit:
https://github.com/opentx/opentx/commit/82dce3ee601049cf548ff4518902ab8e8fa62a0b

projectkk2glider: Before my changes the `TextWidget::options[]` was **also** in RAM. The whole point of my change was to put it where it belongs - in flash.

projectkk2glider: Take a look at firmware.map. If you can find it there, then it means it is in a RAM (.data section to be precise.)

mpaperno:
OK... for now reverted to the initialization lists version, kept the RAM optimization, and fixed MSVC build.  Later can look into improving the string handling and saving more memory by using a `const char *` in the ZoneOptionValue union, which should also fix the "non c++11" initialization issue.

Memory usage is back to:
```
Size report for firmware.elf

Section Details:
section         size(B)               addr      loc
.text           1328256         0x08000000      F
.ARM.extab          376         0x08144480      F
.ARM                512         0x081445f8      F
.init_array         108         0x081447f8      F
.fini_array           4         0x08144864      F
.data              2644         0x10000000      F+C
.bss              40868         0x10000a58      C
.ram              16576         0x20000000      S1
._user_heap_stack  1024         0x200040c0      S1
.sdram           783360         0xd0000000      SD

Totals:         KB        usage of ttl  KB free
 Flash:    1300.68       63.51%   2048   747.32
   CCM:      42.49       66.39%     64    21.51
 SRAM1:      17.19       15.35%    112    94.81
 SRAM2:       0.00        0.00%     16    16.00
 SDRAM:     765.00        9.34%   8192  7427.00
```
(I'm not sure what the small reduction vs. the ["Before"](https://github.com/opentx/opentx/pull/4203#issuecomment-270928762) report is from.)

radio/src/gui/480x272/layout.h
```diff
@@ -24,6 +24,7 @@
 #include <list>
 #include "widgets_container.h"
 #include "lcd.h"
+#include "zone.h"
```
bsongis: not needed

radio/src/gui/480x272/theme.h
```diff
@@ -22,6 +22,7 @@
 #define _THEME_H_
 
 #include <list>
+#include "zone.h"
```
bsongis: not sure it's needed

bsongis:
Thanks!

PR #4204, opened 2017-01-06
3djc/x7: LS fix limits

3djc:
Fix an X7 bug where logical switch value where displayed wrong for some types of fields.

While I was at it,  I changed to limits values from static values to getMaximumValue() to match the larger Taranis behavior.

I tested on X7, but it is very possible that this breaks other 128x64 (don't know enough to tell)

ps : my thanks to @kilrah for his ongoing support :)

### DIFF
--- a/radio/src/gui/128x64/model_logical_switches.cpp
+++ b/radio/src/gui/128x64/model_logical_switches.cpp
@@ -52,9 +52,6 @@ void putsEdgeDelayParam(coord_t x, coord_t y, LogicalSwitchData *cs, uint8_t lat
     lcdDrawNumber(lcdLastPos+3, y, lswTimerValue(cs->v2+cs->v3), LEFT|PREC1|rattr);
   lcdDrawChar(lcdLastPos, y, ']');
 }
-#endif
-
-#if defined(CPUARM)
 
 #define CSWONE_2ND_COLUMN (11*FW)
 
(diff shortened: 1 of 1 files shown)


bsongis:
In this file there is an `#if defined(CPUARM)` duplicated test, is it possible you remove it? Thanks

bsongis:
Even 2 times!

3djc:
I really don't have a good tool for that kind of things I found 2, but maybe not the ones you where expecting !

radio/src/gui/128x64/model_logical_switches.cpp
```diff
@@ -509,6 +516,12 @@ void menuModelLogicalSwitches(event_t event)
         case LS_FIELD_V3:
           cs->v3 = CHECK_INCDEC_PARAM(event, cs->v3, v3_min, v3_max);
           break;
+        case LS_FIELD_DURATION:
```
bsongis: I don't think it's a good idea to move the case order, it's better to keep the order on the LCD

radio/src/gui/128x64/model_logical_switches.cpp
```diff
@@ -162,11 +159,7 @@ void menuModelLogicalSwitchOne(event_t event)
         else {
 #if defined(TELEMETRY_FRSKY)
           if (v1_val >= MIXSRC_FIRST_TELEM) {
-#if defined(CPUARM)
```
bsongis: yes that's what I meant, thanks!

PR #4210, opened 2017-01-08
[Horus] Fixes #4205

### LINKED ISSUE #4205: Horus 2.2 N360: SPORT packets sent by LUA are sent to random sensor IDs
I traced the SPORT signal at the back of the Horus, and it seems that SPORT packets sent by LUA scripts (here S6R.lua) are sent to random sensor IDs. However, the S6R.lua sends its READ requests to sensor 0x1B:
```
local function telemetryRead(field)
  return sportTelemetryPush(0x1B, 0x30, 0x0C30, field)
end
```

First one is sent to 0x00:
screen shot 2017-01-06 at 21 55 45

and then 0xCB:
screen shot 2017-01-06 at 21 59 00

and on and on.... each time another ID (0x00 is there a couple of times).

bsongis:
Not tested at all, would need test on both Horus and Taranis

### DIFF
--- a/radio/src/targets/horus/telemetry_driver.cpp
+++ b/radio/src/targets/horus/telemetry_driver.cpp
@@ -20,9 +20,12 @@
 
 #include "opentx.h"
 
-DMAFifo<512> telemetryFifo __DMA (TELEMETRY_DMA_Stream_RX);
+DMAFifo<TELEMETRY_FIFO_SIZE> telemetryDMAFifo __DMA (TELEMETRY_DMA_Stream_RX);
+Fifo<uint8_t, TELEMETRY_FIFO_SIZE> telemetryNoDMAFifo;
+uint8_t telemetryFifoMode;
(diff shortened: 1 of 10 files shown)


radio/src/targets/horus/telemetry_driver.cpp
```diff
@@ -20,9 +20,12 @@
 
 #include "opentx.h"
 
-DMAFifo<512> telemetryFifo __DMA (TELEMETRY_DMA_Stream_RX);
+DMAFifo<TELEMETRY_FIFO_SIZE> telemetryDMAFifo __DMA (TELEMETRY_DMA_Stream_RX);
```
bsongis: these 2 buffers could be merged to save around 512 bytes RAM , but another class would be needed

raphaelcoeffic:
Thx! I tested both platforms (Horus & Taranis X9D+) with S6R.lua and other scripts. Everything looks good to me.

PR #4215, opened 2017-01-10
Only show telemetry selection if it is sensible to do (as on Taranis/…

schwabe:
…Horus)

### DIFF
--- a/radio/src/gui/128x64/model_telemetry.cpp
+++ b/radio/src/gui/128x64/model_telemetry.cpp
@@ -163,13 +163,13 @@ enum MenuModelTelemetryFrskyItems {
   #define VARIO_RANGE_ROWS             (NAVIGATION_LINE_BY_LINE | 3)
 #endif
 
-#if defined (CPUARM)
+
 #if defined (PCBTARANIS)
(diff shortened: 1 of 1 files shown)


radio/src/gui/128x64/model_telemetry.cpp
```diff
@@ -163,10 +163,18 @@ enum MenuModelTelemetryFrskyItems {
   #define VARIO_RANGE_ROWS             (NAVIGATION_LINE_BY_LINE | 3)
 #endif
 
+#if defined (CPUARM)
+#if defined (PCBTARANIS)
+  #define TELEMETRY_TYPE_SHOW_TELEMETRY  (g_model.moduleData[INTERNAL_MODULE].rfProtocol == RF_PROTO_OFF && g_model.moduleData[EXTERNAL_MODULE].type == MODULE_TYPE_PPM) ? (uint8_t)0 : HIDDEN_ROW
+#else
```
3djc: Wouldn't 

#if defined (PCBTARANIS)
...
#elseif defined(CPUARM)
....
#enfif

be easier to read than the imbricated ones ?

bsongis:
Thanks!

PR #4220, opened 2017-01-11
Fixes #4209: Backlight was always 100% due to wrong test for the unex…

### LINKED ISSUE #4209: With the current next (08-01-2017 16:35) the backlight is not adjustable (Horus)
Title says it all. Did a clean build and the on and off backlight values have no effect. Backlight is always on....

projectkk2glider:
…pected shutdown.

Closes #4209


bsongis:
Thanks!

Issue #4224, opened 2017-01-12
Companion 2.2 problems


kilrah:
Priority issues:
- [x]  Simulate radio from within companion with Horus selected still overwrites everything in the configured SD card directory instead of using a temp folder

- [x] Horus: Switch warnings read wrong (Aup Boff Cmid Dup Eup Fdown Gup on radio gives all on, A mid, G mid, rest up in companion)

- [x] Model Settings / Horus: Module settings not read correctly

- [x] Flight modes / Horus: Trim values not read

- [x] Horus / Source list: Shift of 4 elements (SC read as SG, MAX read as CYC2, TR4 read as TR8

- [x] Horus / Switch list: Shift of 22 elements somewhere (L2 read as L24, FM1 read as ---)

- [x] All Radios / Flight modes: Trim "OFF" missing in dropdown

- [x] Simulate single Horus model causes crash on some models, need to retest once other issues are fixed

- [x] X7: Completely broken for me, stray chars in model list names, no inputs/mixes/garbage in outputs/heli

- [x] Horus: Old fourcc needs to be recognised and automatically updated

- [x] When reading from Horus the file->save command saves straight to the radio instead of creating an OTX file (same problem when opening eeprom from another radio and converting to Horus, we're still wotking on the original file instead of moving to otx)

- [x] Saving the results of a something->Horus conversion (using save as due to above problem) and reopening the default "Models" category now has an empty name. When simulating this straight away there is an "invalid model" in first position, and the selected model is not the one highlighted in the companion model list. Model select page in the simulator shows no selected model

- [x] Load an X9E EEPROM, convert to Horus, launch simulator, everything's empty

- [x] X7 export corrupts input names and telemetry sensors (also custom scripts and modules after further investigation)

- [x] X7 limits divided by 10 (setting 50% results in 95% travel)

- [x] X7 switch /pot names corrupted on import

General issues:

- [x] Selected model not showing / can't be changed

- [x] Radio settings / Horus: Haptic settings should be enabled, haptic length missing

- [x] Model settings / Horus: Antenna selection for internal module missing

- [x] Model settings / Horus: Trainer settings missing

- [x] Telemetry / Horus: Protocol selection enabled when it shouldn't

- [x] RSSI alarms have the Orange/red dropdowns instead of low/critical

- [x] Top bar settings should be hidden

- [x] Telemetry screens should be hidden

- [x] X7 simulator not really usable (switches/pots, no telemetry simulator etc)

[...]

PR #4231, opened 2017-01-14
Projectkk2glider/simu tmp sd path

projectkk2glider:
Not ready for merge yet! Just to show you if I am going in the right direction.

This addresses first item of #4224


projectkk2glider:
With this in place and running simu like this: `simu aaa /card /tmp` will:
* use `/card` path as the root of the SD card image
* any path starting with `/RADIO/...` or `/MODELS/...` will be mapped into `/tmp/RADIO/...` and `/tmp/MODELS/...`
* the SD browser will still browse `RADIO` and `MODEL` directories from `/card` path. To change this will be difficult and in my opinion unnecessary.

projectkk2glider:
Going so deep would be a problem, especially the f_readdir() would be difficult to implement. Not sure if it is needed. For simulation I think the current state is enough - the radio/model settings get read and written for the alternate path. 

Basically anything that uses absolute path will be remapped correctly. What does not work is:
 * iterating trough all files in a root of SD card (won't show RADIO and MODELS dirs from alternate path)
 * changing directories with f_cwd() using relative paths (again it wont go into the alternate dirs for RADIO and MODELS)

bsongis:
Yes it's perfect as it is now

radio/src/targets/simu/simufatfs.cpp
```diff
@@ -0,0 +1,554 @@
...
+void splitPath(const std::string & path, std::string & dir, std::string & name)
+{
+#if defined(WINDOWS_BUILD)
+  char drive[_MAX_DRIVE];
+  char dir[_MAX_DIR];
+  char fname[_MAX_FNAME];
+  char ext[_MAX_EXT];
+  _splitpath(path.c_str(), drive, dir, fname, ext);
+  name = std::string(fname) + std::string(ext);
+  dir = std::string(drive) + std::string(dir);
```
kilrah: Compilation broken here on VS:

```20>C:\opentx-22\code\radio\src\targets\simu\simufatfs.cpp(135): error C2082: redefinition of formal parameter 'dir'
20>C:\opentx-22\code\radio\src\targets\simu\simufatfs.cpp(138): error C2664: 'void _splitpath(const char *,char *,char *,char *,char *)': cannot convert argument 3 from 'std::string' to 'char *'
20>  C:\opentx-22\code\radio\src\targets\simu\simufatfs.cpp(138): note: No user-defined-conversion operator available that can perform this conversion, or the operator cannot be called
20>C:\opentx-22\code\radio\src\targets\simu\simufatfs.cpp(141): error C2065: 'dirName': undeclared identifier
20>C:\opentx-22\code\radio\src\targets\simu\simufatfs.cpp(187): error C2065: 'fileName': undeclared identifier
20>C:\opentx-22\code\radio\src\targets\simu\simufatfs.cpp(187): error C2228: left of '.c_str' must have class/struct/union
20>  C:\opentx-22\code\radio\src\targets\simu\simufatfs.cpp(187): note: type is 'unknown-type' ```

projectkk2glider: I have no Windows build environment. Can you fix this?

projectkk2glider: Hopefully fixed by 19213c8

PR #4232, opened 2017-01-14
DIY ID range changed to 0x5100 - 0x52FF 

projectkk2glider:
(according to SPORT specification V3.4.0)

### DIFF
--- a/radio/src/telemetry/frsky.h
+++ b/radio/src/telemetry/frsky.h
@@ -139,8 +139,8 @@
 #define RBOX_STATE_LAST_ID        0x0b2f
 #define RBOX_CNSP_FIRST_ID        0x0b30
 #define RBOX_CNSP_LAST_ID         0x0b3f
-#define DIY_FIRST_ID              0x5000
-#define DIY_LAST_ID               0x50ff
+#define DIY_FIRST_ID              0x5100
+#define DIY_LAST_ID               0x52ff
 #define RSSI_ID                   0xf101
 #define ADC1_ID                   0xf102
 #define ADC2_ID                   0xf103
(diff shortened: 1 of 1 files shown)


bsongis:
I asked for confirmation, this will break Craft & Theory converter for Ardupilot!

floaledm:
Thanks Bertrand. Unfortunately, this would be quite disruptive to users of Passthrough on ArduPilot. That includes our own users but also other projects using Passthrough on ArduPilot (e.g., ersky9x: http://openrcforums.com/forum/viewtopic.php?f=95&t=9164).

The FrSky forum appears to be down so I can't look at the spec. Could the range simply be changed to 0x5000 -0x52FF?

By the way, the luadoc for sportTelemetryPop() would also need to be updated to reflect the new range.

3djc:
The official range from FrSky:
image

projectkk2glider:
@floaledm try to convince FrSky to change the range. We can change the range of DIY in OpenTX no problem, the only problem would be if FrSky produces some sensor that uses range from 0x500 fo 0x5ff.

bsongis:
Email sent

floaledm:
Thanks Bertrand! I think you'll be more convincing that I ever would...

Damjan's argument is definitely valid, so it would be good to clear this out. So, you guys don't know the reason behind the change? Does FrSky sometimes do this, and then you just have to play catch-up?

bsongis:
No the error is mine, I asked for a wrong range (or I gave you a wrong ID)

bsongis:
FrSky will update the S.PORT spec, DIY range will be 0x5000-0x52ff

PR #4235, opened 2017-01-14
Port #4189 to next

elecpower:
Port #4189 to next for adding data modification for sub trim, min and max on model edit outputs


bsongis:
thanks!

PR #4243, opened 2017-01-16
3djc/enhance companion get capability

3djc:
Add the ability to query 'defaults' as per Max request


mpaperno:
As some background, this came up in discussion because I was looking for a way to get a reasonable default hardware config for a radio before any eeprom data exists (eg. as if it came from from the "factory" :) ).

I think I found that a good way to get HW defaults is to create a new [GeneralSettings](https://github.com/opentx/opentx/blob/next/companion/src/eeprominterface.cpp#L1083) instance, which also sets up some reasonable hardware defaults (number of pots/sliders).

So this is essentially duplicating what @3djc did here with with `FactoryInstalled*` `Capabilities`.  Except that a default GeneralSettings gives all 18 switches for an X9E.  Is there any particular reason for this, or just never got updated?  

When editing X9E radio settings in Companion all 18 also show up as configured by default, which doesn't seem right.  If you load that eeprom in Simulator then all 18 switches show up as configured.

So... it probably makes more sense not use this PR and instead add the proper number of [default switches](https://github.com/opentx/opentx/blob/next/companion/src/eeprominterface.cpp#L1096) for X9E in GeneralSettings() iniit.  Or perhaps just keep [FactoryInstalledSwitches](https://github.com/opentx/opentx/pull/4243/files#diff-1265a523fab26572eecdbcb6972dc370R540) ?   We could keep it all but don't really need to clutter the code if we aren't using those functions (that getCapabilities is quite a beast already).

3djc:
Hum maybe take benefits of this change by updating GeneralSettings::GeneralSettings() like that 

  for (int i=0; i<GetCurrentFirmware()->getCapability(FactoryInstalledSwitches); i++) {
    switchConfig[i] = GetCurrentFirmware()->getSwitch(i).type;
  }

or move the changes into this loop, unsure what is best

mpaperno:
That's pretty much what I was thinking.  The only "issue" is that the code in `getCapability(FactoryInstalledSwitches)` is mostly a duplicate of `getCapability(Switches)` except for X9E. 

Altho...  it could go something like this:

```diff
+    case FactoryInstalledSwitches:
+      if (IS_TARANIS_X9E(board))
+        return 8;
+      else
+        return getCapability(Switches);
+      break;
```

I kinda like that.

mpaperno:
I think this makes some more useful information available and corrects a couple previous oversights.  Goes a long way to help get the simulator UI dynamically generated.

companion/src/eeprominterface.cpp
```diff
@@ -1093,30 +1093,35 @@ GeneralSettings::GeneralSettings()
     calibSpanPos[i] = 0x180;
   }
 
-  for (int i=0; i<GetCurrentFirmware()->getCapability(Switches); i++) {
+  for (int i=0; i<GetCurrentFirmware()->getCapability(FactoryInstalledSwitches); i++) {
```
bsongis: Same comment than the other review, I think this should be part of a new Board class

companion/src/eeprominterface.h
```diff
@@ -80,16 +80,18 @@ enum Capability {
...
   VoicesMaxLength,
   MultiLangVoice,
   ModelImage,
   Pots,
   Sliders,
   Switches,
+  FactoryInstalledSwitches,
   SwitchesPositions,
+  NumTrimSwitches,
+  CustomFunctions,
```
bsongis: The 2 last ones are needed here, this is good, but the 2 first ones should move to the new Board class

bsongis:
OK we will introduce a new Board abstraction layer later

PR #4244, opened 2017-01-16
Another small multi protocol update, add  H26WH mjxq subprotocol and …

schwabe:
…telmetry as option for Bayang


bsongis:
thanks!

powercycle:
I was looking at this and wondered if the fact that the strings aren't \0 terminated is an issue or not.
The previous line did end with \0, but the new line does not.

3djc:
They have to be 0 terminated only if they are shorter than expected lenght

PR #4246, opened 2017-01-16
Allow early version of X7 firmware Eeprom to migrate to new variant

3djc:


radio/src/storage/eeprom_rlc.cpp
```diff
@@ -826,6 +826,14 @@ bool eeLoadGeneral()
     }
   }
 
+  #if defined(PCBX7)
+  if (g_eeGeneral.variant == 0) {
+    TRACE("Pre release EEPROM detected, variant %d instead of %d for X7 radio. Loading anyway", g_eeGeneral.variant, EEPROM_VARIANT);
+    g_eeGeneral.variant = EEPROM_VARIANT;
```
kilrah: `storageDirty(EE_GENERAL);`

radio/src/storage/eeprom_rlc.cpp
```diff
@@ -826,6 +826,14 @@ bool eeLoadGeneral()
     }
   }
 
+  #if defined(PCBX7)
```
bsongis: indentation problem here

bsongis:
yes you are right, it's better with your change

radio/src/storage/eeprom_rlc.cpp
bsongis: perhaps not needed as we write the unexpected shutdown bit just after. perhaps I am wrong though, an eeDirty wouldn't hurt

radio/src/storage/eeprom_rlc.cpp
```diff
@@ -827,11 +827,12 @@ bool eeLoadGeneral()
   }
 
   #if defined(PCBX7)
```
bsongis: it's worse, everything 2 spaces

bsongis:
👍

PR #4248, opened 2017-01-16
[build] Only add SIMU_AUDIO option if SDL was found, & enable it by default.

mpaperno:


bsongis:
Yes!

projectkk2glider:
@mpaperno can this be removed now? https://github.com/opentx/opentx/blob/next/companion/src/CMakeLists.txt#L53

mpaperno:
Yes, that can probably go away now.  I think there's also another CMakeLists where it does the same thing.

PR #4249, opened 2017-01-17
[Companion] Move SwitchType enum to RadioData GeneralSettings

mpaperno:
For consistency with the other enums defined there.  On the radio side those same "nnConfig" enums are also defined in one place ([myeeprom.h](https://github.com/opentx/opentx/blob/next/radio/src/myeeprom.h#L138)).  :)    This also helps with some simulator stuff I'm working on.

The `SWITCH_` prefix here is a bit awkward because it conflicts with the `Switches` enum defined in [radiodata.h](https://github.com/opentx/opentx/blob/next/companion/src/radiodata.h#L8) .  But I understand it matches with the same enum names on radio side, which I guess helps semantically.  (Also that name `Switches` in global conflicts with something else... I forget what now.)

Though it does beg the question... is some shared place these enums could be better centralized?

### DIFF
--- a/companion/src/eeprominterface.cpp
+++ b/companion/src/eeprominterface.cpp
@@ -1055,17 +1055,17 @@ bool GeneralSettings::switchPositionAllowedTaranis(int index) const
   if (index == 0)
     return true;
   SwitchInfo info = switchInfoFromSwitchPositionTaranis(abs(index));
-  if (index < 0 && switchConfig[info.index] != Firmware::SWITCH_3POS)
+  if (index < 0 && switchConfig[info.index] != GeneralSettings::SWITCH_3POS)
     return false;
   else if (info.position == 1)
-    return switchConfig[info.index] == Firmware::SWITCH_3POS;
+    return switchConfig[info.index] == GeneralSettings::SWITCH_3POS;
   else
-    return switchConfig[info.index] != Firmware::SWITCH_NONE;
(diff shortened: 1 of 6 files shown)


mpaperno:
The changes in radiodata.h are just [a few lines here](https://github.com/opentx/opentx/pull/4249/files#diff-96f2da8a25b31210b73cff78e64493d9R1138).  But Bertrand told me to keep the space trimming on... :)

bsongis:
Why this enum would be part of GeneralSettings? For me it's not even better in RadioData, nor Firmware, but a new Board class.

mpaperno:
The other enums that serve a similar purpose are in GeneralSettings already.  Seemed logical to a least group them there.  Chasing these things down all over the code is a PITA.

I did ask if there would be a better place for all these enums, so I'm open to suggestions.  Though I was thinking more like some existing shared header file, like constants.h or something.

mpaperno:
The last commit is more related to #4243 in terms of determining default hardware layout, but I don't think I can commit to that branch w/out breaking the merge.  

The way it was before this, the wrong information was returned by `GetCurrentFirmware()->getSwitch(x)`  

I verified all the places getSwitch() is used, and this doesn't break anything. The only minor side-effect is that it slightly changes the order of switches in Companion model switch warnings setup for 9X (actually may fix it too... haven't tried older version to test).

mpaperno:
Thanks Bertrand!

PR #4251, opened 2017-01-17
Bsongis/horus create radiodata from scratch

bsongis:


companion/src/modeledit/modeledit.cpp
```diff
@@ -120,9 +120,14 @@ void ModelEdit::on_pushButton_clicked()
...
-  RadioData *simuData = new RadioData();
+  RadioData * simuData = new RadioData();
   simuData->generalSettings = generalSettings;
-  simuData->models[0] = model;
+  if (simuData->models.size() == 0) {
+    simuData->models.push_back(model);
+  }
+  else {
+    simuData->models[0] = model;
+  }
```
projectkk2glider: When launched this way, the simulator will only see one model - the one being edited. All other models are not transfered into simuData. Is this intentional? Perhaps it was like this before the storage changes, but I think not.

projectkk2glider:
There is another problem, when simulation is started from `mdichild.cpp` and model list is empty, the radio simulator gets confused. It finds only `RADIO/radio.bin`, while the `RADIO/models.txt` is empty and not files are in `MODELS/`. 

This state confuses the simulator. When you for example change trim, it tries to write the model settings to the file `/MODELS/` - which is really a directory. The name `modelXX.bin` is missing - empty.

companion/src/modeledit/modeledit.cpp
kilrah: This is the single model simulator launched from the model editor. Yes it was always like that there.

The button on the model list window does the whole radio, the button in a model does just that model.

projectkk2glider: Interesting, I never noticed that :rofl:

PR #4252, opened 2017-01-17
Add dangerous special function (bind/range) for Multimodule only

schwabe:
There have been a lot of requests from people using multimodule (e.g. fraterNL) to have dangerous module SF enabled or an option for dangerous module SF to be enabled. OpenTX hid this options because we think they are too dangerous for Frsky D16 and big models. 

This commit allows the dangerous function when multimodule is enabled and only for the external module when multimodule is enabled. The multimodule already always going into bind mode by settting channel 16 to high/low so we do not add any more danger.

Also name the SF multibind and multirange when Dangerous module function are disabled to avoid confusion and make clear that they are only work with multi.


3djc:
Setting of the function to int module when only multi is available (ie no DANGEROUS) should be fixed before merge I think.
image

bsongis:
Yes I do agree, there should not be a parameter in this case

schwabe:
Closing this now. The ch16 solution is probably good enough.

PR #4253, opened 2017-01-18
Do not use color coded RSSI alarms on Taranis/Horus prints

3djc:
A "somewhat" non critical issue, but nonetheless :D


bsongis:
thanks!

PR #4254, opened 2017-01-18
Re #4224: Enable reading and writing Models and Settings from/to Horu…

projectkk2glider:
…s radio (SD card)

### DIFF
--- a/companion/src/mainwindow.cpp
+++ b/companion/src/mainwindow.cpp
@@ -740,13 +740,23 @@ void MainWindow::loadBackup()
 
 void MainWindow::readEeprom()
 {
-  if (getCurrentBoard()== BOARD_HORUS) {
-    qDebug() << "TODO readEEPROM HORUS";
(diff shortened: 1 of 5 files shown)


projectkk2glider:
Filename/window title for mdichild is kind of wrong. It should be `<something>.otx`. Not sure where to fix this.

kilrah:
Opening from radio, then clicking File->Save saves straight on radio. It should rather prompt to save to .otx somewhere to keep homogeneity, only write to radio with "write models and settings"

projectkk2glider:
>Opening from radio, then clicking File->Save saves straight on radio. It should rather prompt to save to .otx somewhere to keep homogeneity, only write to radio with "write models and settings"

Yes, and the cause for this is the same as:

>Filename/window title for mdichild is kind of wrong. It should be <something>.otx. Not sure where to fix this.

bsongis:
Thanks! I will try to take care of the .otx problem

PR #4258, opened 2017-01-19
Companion : add Horus antenna handling

3djc:


bsongis:
Seems all good, thanks!

PR #4260, opened 2017-01-20
[Simulator] Dyanamic Simulator virtual hardware generation and many other improvements

mpaperno:
Major Simulator re-structuring to reduce code redundancies and allow for fully dynamic simulated hardware setup:
 * SimulatorDialog is no longer sub-classed by individual radio UIs, instead a new class is added which handles just the "skin," LCD, and navigation controls (also makes setting up new models much quicker);
 * All radio switches/knobs/sliders are populated dynamically based on loaded hardware configuration data (eeprom/bin), including control types and names (in case of no hardware config data, reasonable "factory defaults" are loaded);
 * Starting Horus sim with alternate path for RADIO and MODEL data folders is now supported (.otx file is WIP);
 * 6-way switch now works properly and various other inconsistencies addressed (eg. some Horus analogs were backwards).
 * Other changes... TBC

TODO:
- [ ] Support loading `.otx` files (and possibly make better use of `Storage` classes).
- [ ] Momentary toggle switches do not rebound (also will have "lock" feature).
- [ ] Save/restore switch/pot positions.
- [x] Add X7 skin.
- [x] Fix Travis build.

Future:
- [ ] Allow data exchange back to Companion after using Simulator (should actually be fairly easy to add now)


kilrah:
Really really nice!
Maybe one thing, I guess the 3-pos switches should default to "Up"... kinda arbitrary, but has been so for so long...

mpaperno:
Thanks!  Actually you reminded me of the TODO item I was forgetting... save/restore the switch/pot positions (added to list).  So, do you think that will cover it?  I agree it's a change... I could really go either way.  The nice thing about middle default is that it is the same (zero) value for 2 and 3-pos switches.  Not that it would be hard to handle, but ya know... :)

kilrah:
Thing is all switch warnings default to UP, and nobody ever configures their radio to have switches in the middle by default (again UP is the most common choice).

Also noticed, a few 9X radios have broken skin:

clipboard01

mpaperno:
Yea good point on the switch warnings.  Does get annoying, though again it would only be on a fresh radio profile, after that it would remember them for you anyway.

I also noticed that with X9 skin... is it not supposed to do that?  :)  I've only ever seen it white... then found/tried the Companion bg color setting and saw what you see.  (And now there's no way to go back to white bezel.)

kilrah:
It's supposed to be like this:


clipboard01

Apparently it's ok for 9X-AVR but broken for 9X-ARM

mpaperno:
Mine was like that (white) until I changed the BG color setting in companion->simulator.  Has the color ever worked?  Is it possible I fixed it and now it looks broken?  :) 

Is your "AVR" radio profile different from your "ARM" radio profile, in Companion?

kilrah:
Mine is "Orange" but the simulator settings page is not stored in the profile

mpaperno:
Oh, right you are.  So was it ever orange before?

kilrah:
Changing to blue indeed changed the frame color - but the 2 images are still missing

clipboard01

mpaperno:
Yea it's always the same 2 side images in each color, except white.

kilrah:
Yep, 2.2RC8 follows the color and looks correct

mpaperno:
Found the bug.  Silly typo.

image

bsongis:
2600 lines of code removed, this is impressive!

mpaperno:
I was going to brag about that in my first bullet point but forgot  :-)

companion/src/simulation/widgets/buttonswidget.h
```diff
@@ -75,30 +75,30 @@ class ButtonsWidget : public QWidget
...
+
     virtual void mousePressEvent(QMouseEvent * event)
     {
       int x = event->x();
       int y = event->y();
       setFocus();
       if (event->button() == Qt::LeftButton) {
-        foreach(Area area, areas) {
-          if (area.contains(x, y)) {
+      foreach(Area area, areas) {
```
bsongis: indentation broken

mpaperno: This was the old version... I think we're looking at an individual commit here? I did the file move in one commit and then the actual changes later on (my preferred method when possible) 

LOL, man, that confused the heck out of me for about 2 minutes!  😆

kilrah:
Found another bug, when starting the simu e.g. X7 and the specified eeprom file is missing it is not created. I.e. you can get in, you get formatting, can work but nothing is saved

mpaperno:
@kilrah Thanks Andre, that was a key bug and led to other improvements.

Also `.otx` file selection for Horus is disabled for now (until I add support for loading it). So pending any more testing or bugs being found, this could technically be merged now, albeit with the few missing features in the TODO list.

mpaperno:
image


How'd I do on the LCD color?  I only found a total of 2 decent photos online with the screen on.

Something is up with the switch names/display... notice my customized hardware/labels.  Some buffer is spilling over or something...

Also, the SE switch doesn't work ("TCT" in my screenshot) but the SH switch toggles the SE slot.  Keep in mind the simulator is reading all default hardware settings from the Firmware functions now, so I think something is up... but I haven't investigated. 

-Max

kilrah:
Nice!!
Does the switch mess also happen on next? might not have tested that in the simulator...

The backlight is actually bluish white

bsongis:
Really nice, we need it in the next RC!

mpaperno:
Props to @mhotar for the nice graphics!

bsongis:
there is only one minor detail I see on this graphics: the small vertical bar at the left / right of the simulator, it would be better if we kept the simulator color, i.e. white for X7

mpaperno:
I don't like the close cropping at the top either, working on that now in fact!

kilrah:
> Something is up with the switch names/display... notice my customized hardware/labels. Some buffer is spilling over or something...

Seems companion import of all custom names is broken...
But while testing on the radio I found that we had an overlap on that main view if switch names are 3 chars. Fixed that by realigning things since we only have 2 pots

screenshot-1

mpaperno:
image

Better?  The side filler gradients aren't a perfect match, but not too bad.

I would like to discuss with @mhotar adding some padding area around the LCD screen (perhaps for all non-LED models).  Right now it is the correct size in terms of actual pixels, but the physical screens always have some extra border area around them.  It would be nicer for UI design work if this border was properly reflected, IMHO.

kilrah:
Nice! I agree about a 1-2px (i.e. LCD-size pixels!) margin around the actual LCD zone

mhotar:
Sure, I do agree once we have backlight color defined. How would you prefer to do that? The same like 9x is done? Is there transparency useable so I can do all the center part in one PNG file? Can you extend colored LCD part by few pixels and I will do transparent layer over it?

mpaperno:
@mhotar Hi Martin, thanks for chiming in.  My ideas/goals:
 * Do not want to tie the graphics to the color (like 9X is now). In fact would like to make the color completely user-configurable (per radio profile).
 * Transparency (24-bit PNG) could be used (eg. to "fix" the 9X skin now).
 * I think for the X7/9 skins we just need to add some extra space for the LCD area (make the surrounding graphics a bit wider/taller).  Then I can make the LCD simulator code properly handle the extra margins (so it keeps control over the "backlight" on the whole screen area).
* Some mockups of exactly how much extra space we want might be useful before modifying all the graphics.  I agree probably a couple LCD-size pixels would be fine, but would be good to verify first.  I'll see what I can put together in P-shop, unless you can beat me to it  :)

What do you think?

bsongis:
Thanks a lot!

mpaperno:
Continued in #4282 .

PR #4263, opened 2017-01-20
[TravisCI] Set up for parallel Travis builds and use Qt5.7

mpaperno:
Check out the build times on the parallel builds vs. one single one... ~11-16min vs. 23+. Plus I added one more board (9XRPRO, not sure it's needed).  Plus you get to see results quicker, if one set of targets fails you still get tests on the others, and the logs are much shorter/easier to view.

Note that the [build summary](https://travis-ci.org/opentx/opentx/builds) page at Travis shows the total time for all the builds together.  You have to click through to the details to see the actual time.  The key point here is that the "sub-builds" run in parallel as slots become available.

Looks like Travis gives us up to 5 parallel builds across the whole repo (so that includes any other builds in progress).  So I have 6 now and the X7 build finishes quickly to open up a slot.  It could of course be split up any other way. 

I tried getting gcc-arm 5.7 working (to sync with production build server), but failed... it couldn't find the arm-none-eabi-g++ no matter what I tried (setting PATH every which way).  The code is still in `.travis.yml` if someone wants to give it a shot (obviously copied from the Docker setup).

-Max

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,32 +11,36 @@ python:
 
 env:
   global:
-     # - QT_BASE=53
-     # - QT_BASE=54
-     # - QT_BASE=55
-     # - QT_BASE=56
-     - QT_BASE=57
-     #- GCC_ARM=/opt/gcc-arm-none-eabi/bin
+    # - QT_BASE=53
+    # - QT_BASE=54
+    # - QT_BASE=55
+    # - QT_BASE=56
+    - QT_BASE=57
+    #- GCC_ARM=/opt/gcc-arm-none-eabi/bin
+    - AVR_FLAVORS="AVR9X 9X GRUVIN9X MEGA2560"
   matrix:
-     # - FLAVOR=ALL
-     - FLAVOR=DEFAULT
-     - FLAVOR=AVR9X
-     # - FLAVOR=9X
-     # - FLAVOR=GRUVIN9X
-     # - FLAVOR=MEGA2560
-     - FLAVOR=ARM9X
-     # - FLAVOR=AR9X
-     # - FLAVOR=SKY9X
-     # - FLAVOR=9XRPRO
-     # - FLAVOR=TARANIS
-     - FLAVOR=X7
-     - FLAVOR=X9
-     # - FLAVOR=X9D
-     # - FLAVOR=X9D+
-     # - FLAVOR=X9E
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,25 +1,69 @@
...
+     - FLAVOR=DEFAULT
+     - FLAVOR=AVR9X
+     # - FLAVOR=9X
+     # - FLAVOR=GRUVIN9X
+     # - FLAVOR=MEGA2560
+     - FLAVOR=ARM9X
+     # - FLAVOR=AR9X
+     # - FLAVOR=SKY9X
+     # - FLAVOR=9XRPRO
+     # - FLAVOR=TARANIS
```
mpaperno: To build those a 2nd time?  "TARANIS" includes the X7 vs. "X9" which is just the bigger Tarani.  Is that what you mean?

.travis.yml
```diff
@@ -1,25 +1,69 @@
...
+  global:
+     # - QT_BASE=53
+     # - QT_BASE=54
+     # - QT_BASE=55
+     # - QT_BASE=56
+     - QT_BASE=57
+     #- GCC_ARM=/opt/gcc-arm-none-eabi/bin
+  matrix:
+     # - FLAVOR=ALL
+     - FLAVOR=DEFAULT
```
mpaperno: Did you see my 2nd commit?  :)  I went back and forth between the names, and then meant to add a comment but forgot.

bsongis: I see, the 1st commit seems perfect to me!

.travis.yml
bsongis: no, I meant I would delete the line completely

.travis.yml
mpaperno: OK, well I liked saving "COMPANION" for if we ever want just a Companion build... the "DEFAULT" is actually "make all" with default settings... so it seemed to make more sense in the end.  If you're OK with it, I'll just add a comment to that effect.

.travis.yml
mpaperno: Ah, and the whole TARANIS "FLAVOR" also from the .sh script?  That would be fine and clean it up a bit.  Also maybe same treatment for HORUS and just have X12 = {X12S, X12S_r10} to follow the pattern.  Yes?

PR #4272, opened 2017-01-22
Force different file name after conversion

kilrah:
If we're ok to do that, should we not also automatically change every bin and hex file to .otx when opened?

Anything that might break or is storage type well decoupled from the extension now?


kilrah:
Any comment about the proposed "convert everything to otx"?

bsongis:
Yes I vote for this change

PR #4275, opened 2017-01-23
Refactoring of startup code, mainly dealing with detection of unexpec…

projectkk2glider:
…ted restart/shutdown

None of this was tested on any radio! (I can't right now).

I will also add CLI code to test the response of failing `new()` function to see if we got unhandled exceptions or we get back NULL.


projectkk2glider:
Tested successfully on Horus and Taranis.

bsongis:
Thanks!

PR #4280, opened 2017-01-24
Shorten GVx to Gx on 128x64 radios, where screen estate is a real issue

3djc:
It is a proposal to gain some space on busy 128x64 screens. Bellow are some example on how it looks.

image

image


bsongis:
good idea

PR #4281, opened 2017-01-24
3djc/x7 mixer screen rework

3djc:
Please review thoroughly

- fix FM display that was completely off

Introduce a way to see more details on a mixer quickly without having to got to details when there is a mix name defined.

Currently : if there is a name, it is displayed
Proposed : if the is a name, it is displayed, but on active row, the name is moved to title bar, and data is displayed instead

### DIFF
--- a/radio/src/gui/common/stdlcd/model_mixes.cpp
+++ b/radio/src/gui/common/stdlcd/model_mixes.cpp
@@ -228,12 +228,23 @@ void displayMixInfos(coord_t y, MixData * md)
   lcdDrawChar(MIX_LINE_DELAY_POS, y, cs);
 }
 
-void displayMixLine(coord_t y, MixData * md)
+void displayMixLine(coord_t y, MixData * md, bool active)
 {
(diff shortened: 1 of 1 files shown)


radio/src/gui/common/stdlcd/model_mixes.cpp
```diff
@@ -195,12 +195,12 @@ void displayMixLine(coord_t y, MixData * md)
...
-#else
+#else // LCD_W >= 212
 #define MIX_LINE_WEIGHT_POS            6*FW+8
 #define MIX_LINE_SRC_POS               7*FW+3
-#define MIX_LINE_CURVE_POS             12*FW
-#define MIX_LINE_SWITCH_POS            16*FW+4
-#define MIX_LINE_FM_POS                13*FW+3
+#define MIX_LINE_CURVE_POS             12*FW+3
+#define MIX_LINE_SWITCH_POS            16*FW+5
+#define MIX_LINE_FM_POS                19*FW
```
bsongis: you are right, I fixed the same bug in the inputs screen

PR #4282, opened 2017-01-25
[Simulator] Further Simulator improvements.

mpaperno:
Continued from #4260 

Multiple changes and new features:
  *  All Horus data source options are now fully supported with data properly saved after use (SD path, .otx file, or custom data path);
  *  Loading .otx files now works for all radio types (but not yet saving data back to .otx, except on Horus);
  *  Switch and knob/slider states are now saved to the Companion radio profile (and switches are now "up" by default);
  *  Joystick configuration/calibration is now available directly from inside Simulator;
  *  **!** Due to refactoring, all previously saved per-profile simulator settings are now invalid (sorry, but the new ones are better! :)
  *  With WIN_USE_CONSOLE defined, Simulator and Companion when started from Windows command line will now use that console for all messages (instead of opening a new one). This allows full debug output to be seen at all stages of execution and after exit.

Quite a bit of code change at once... sorry for that.  I tried to break it up into some logically split commits, in case that helps.  And some stuff in AppData which is totally unrelated but I just couldn't help myself with "cleaning up" a bit.

Thanks,
-Max

### DIFF
--- a/companion/src/CMakeLists.txt
+++ b/companion/src/CMakeLists.txt
@@ -71,6 +71,9 @@ endif()
 
 if(WIN32)
   include_directories(SYSTEM ${WIN_INCLUDE_DIRS})
+  if(NOT WIN_USE_CONSOLE)
+    set(WIN_EXECUTABLE_TYPE WIN32)  # GUI (WinMain) app
+  endif()
   if(MSVC)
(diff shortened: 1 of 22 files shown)


bsongis:
Seems all good, but I didn't test and read everything carefully!

PR #4286, opened 2017-01-25
Add q303 multi protocols and fix subtype selection on 128x64 ui

schwabe:


3djc:
Functionnal test OK on x7

bsongis:
In the future, we could introduce a file for not translated / translatable strings

PR #4287, opened 2017-01-25
Update the luadoc for sportTelemetryPop()

floaledm:
Followup on #4232: DIY ID range changed to 0x5000 - 0x52FF


bsongis:
Good catch!

PR #4296, opened 2017-01-26
Fix current simulator lib not included in dmg

schwabe:


companion/src/CMakeLists.txt
```diff
@@ -300,12 +300,19 @@ endif()
 
 ############# Install ####################
 
+# Generate list of simulator plugins, used by all platforms
+file(GLOB simulator_plugins "${CMAKE_BINARY_DIR}/*opentx-*simulator${CMAKE_SHARED_LIBRARY_SUFFIX}")
+# the current flavour is not automatically added if build in the current cmake iteration, so always
+# add its library name to be sure
+string(TOLOWER ${PCB} FLAVOUR)
+get_property(current_plugin TARGET opentx-${FLAVOUR}-simulator PROPERTY LOCATION)
+list(APPEND simulator_plugins "${current_plugin}")
```
projectkk2glider: I don't like this hack. Do similar to the https://github.com/opentx/opentx/blob/c78ab55696ce5fc09726290d32d3ed913e25e579/launchpad/common/rules

 * make all simulator libs
 * run empty cmake `cmake <src dir>`  **this will detect the last simulator library**
 * then run `make package` or `make install`

companion/src/CMakeLists.txt
```diff
@@ -476,7 +483,7 @@ IF(APPLE)
   install(DIRECTORY ${PROJECT_SOURCE_DIR}/companion/targets/mac/qt_menu.nib DESTINATION "${companion_res_dir}" COMPONENT Runtime)
 
   # manually add the required plugins
-  foreach (plugin ${Qt5Multimedia_PLUGINS} ${Qt5PrintSupport_PLUGINS} ${Qt5Gui_PLUGINS})
+  foreach (plugin ${Qt5Multimedia_PLUGINS} ${Qt5PrintSupport_PLUGINS} ${Qt5Gui_PLUGINS} ${Qt5Gamepad_LIBRARIES})
```
mpaperno: Isn't the joystick stuff handled by SDL? Or is this for something besides Simulator?

schwabe: oh right, will remove that again.

companion/src/CMakeLists.txt
schwabe: running cmake a second time is even more of a hack. 

I first wrapped the file global into a INSTALL(CODE fragment but that looked even more horrible and forced more stuff inside the INSTALL(CODE fragment

schwabe: Btw. which part of the hack don't you like? Getting the filename or having always the current simulator plugin in the pluging list?

projectkk2glider: Well, that current simulator you are adding might not even been build yet. What will happen then?

schwabe: I am adding only the target filename to the list. The package target should run after all other targets. If it runs before the other targets (which I doubt since I think the install stage is after everything else) the copying of simulator libs will fail, which is probably a good thing.

companion/src/CMakeLists.txt
projectkk2glider: @schwabe remove this, then we can merge.

Issue #4159, opened 2016-12-23
Open TX 2.2 - Simulator "PF7 - Reload Lua Scripts"


rcbebo82:
Im currently programming a widget. In the simulator the option Reload LUA Scripts does not work.
I edit the script in the background click on "PF7 - Reload LUA" scripts but the script does not seem to be loaded in the new version. I have to close an start the simu again to make it working.

mpaperno:
Do you mean that specifically the Lua widgets on Horus aren't reloaded?  I can confirm this.  The MIXES scrips seem to reload OK (and telemetry scripts on Taranis).

mpaperno:
They're also not reloaded when Custom Scripts (MIXES) are updated.  I believe the widgets are only loaded at startup.  This may be more of a feature request than bug per-se.  Definitely would be nice though.

projectkk2glider:
That is because I separated the two Lua environments and forgot about this feature. But on Taranis it should be the same.

rcbebo82:
For sure only widget scripts. It costs a lot of time when you develop a widget and have to restart the sim each time. Would be nice if the button also forces the widgets to reload this would speed up development per se. If the widget contains telemetry sources you also have to startup telemetry each time and set it to your needs.

projectkk2glider:
It is clear now where the problem is. This will be fixed.

projectkk2glider:
This will be harder than I first thought. Not only has the Lua environment be reloaded, the whole system of registering themes and widgets, creating widgets, etc must be recreated. Basically it will look like a new model was selected and loaded. 77

rcbebo82:
Maybe "Reset Flight Data" could do this also?

3djc:
We have gone as far as we can on this by providing a reload radio that doesn't require to restart the simu software. Closing this as a result

PR #4303, opened 2017-01-27
[Simulator] Further improvements (reload radio, momentary switches, etc.)

### LINKED ISSUE #1627: Temporary hold SH switch down in Companion (Taranis)
In Companion, it would be nice if it were possible to temporary hold SH switch at its down position (maybe in a similar way as the Hold X and Hold Y boxes of the sticks) in order to simulate sequences like: SH-down / do something  / SH-up.

mpaperno:
* Add ability to reload all radio data w/out exiting simulator (will also adjust UI for any hardware changes) (helps with #4159);
 * Only update channel/logical switch/GVar output displays if the corresponding tab is actually visible (further CPU usage reductions);
 * Implement momentary toggle switches with "latch" function (closes #1627);
 * Using mouse wheel over multi-position knobs now properly rotates them to detent positions.
 * Fix throttle stick position/lock not being set properly upon repeated launches from Companion.

This is probably the last button that will fit up on the top toolbar... will need to switch to icons-only and/or a menu for further additions.

image


3djc:
Won't compile with the QT 5.3 of building machine:
In file included from /home/opentx/opentx/companion/src/simulation/simulatordialog.cpp:28:0:
/home/opentx/opentx/companion/src/simulation/widgets/radioswitchwidget.h: In member function ‘void RadioSwitchWidget::onValueChanged(int)’:
/home/opentx/opentx/companion/src/simulation/widgets/radioswitchwidget.h:121:77: error: no matching function for call to ‘QTimer::singleShot(int, RadioSwitchWidget*, void (RadioSwitchWidget::*)())’
         QTimer::singleShot(500, this, &RadioSwitchWidget::onMomentaryTimeout);

projectkk2glider:
Testing this, works awesome.

One wish: momentary 2 pos switch returns to OFF position even if I hold it on the ON position with the mouse. I have setups that require SH to be ON for more time that is predefined in simulator. 

Therefore I would prefer if SH stays ON as long as I hold it there with the mouse. Is that possible?

kilrah:
Last step will be an X9E skin, @mhotar? ;)

mpaperno:
Damjan, great point.  I've just been flicking the switches mostly with mouse wheel and hadn't thought of "holding it down" issue.  So now it works both ways, much better I think.

And yes, the X9E seems a bit left out...  There's even some annoying conditional code in the X9D skin to handle the (slight) differences in controls.  ;-)

projectkk2glider:
Momentary now works perfect :+1:

companion/src/simulation/widgets/lcdwidget.h
```diff
@@ -72,9 +72,7 @@ class LcdWidget : public QWidget
 
     void setBackgroundColor(QColor color)
```
bsongis: Perhap it should be `const QColor &`

bsongis:
Thanks!

companion/src/simulation/widgets/lcdwidget.h
mpaperno: Yes it should!  Fixed in next version  :)  Thanks!

PR #4308, opened 2017-01-27
Proposition : model number handling

### LINKED ISSUE #4295: Companion 2.2-RC9 Saving in .otx removes unused models and renumbers following models
Companion 2.2-RC9 Radio X9D
I don't know if this is a bug or "working as designed" but if there are any unused models (gaps), saving in .otx format removes the unused models and moves all models after that up to fill those slots (changing the model numbers accordingly).  Saving in EEPE maintains the model numbers as they are on the transmitter as it always has.
capture

3djc:
This is a proposal to fix #4295

### DIFF
--- a/companion/src/storage/categorized.cpp
+++ b/companion/src/storage/categorized.cpp
@@ -47,48 +47,70 @@ bool CategorizedStorageFormat::load(RadioData & radioData)
   QList<QByteArray> lines = modelsListBuffer.split('\n');
   int modelIndex = 0;
   int categoryIndex = -1;
-  foreach (const QByteArray & line, lines) {
-    if (!line.isEmpty()) {
-      if (line.startsWith('[') && line.endsWith(']')) {
+  foreach (const QByteArray & lineArray, lines) {
+    QString line = QString(lineArray).trimmed();
+    if (line.isEmpty()) continue;
+    qDebug() << "parsing line" << line;
+
+    if (line.startsWith('[') && line.endsWith(']')) {
+      // ignore categories for boards that do not support them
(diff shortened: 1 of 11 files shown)


projectkk2glider:
Your current proposal is too hard-coded for my liking (support max 100 models)

I have this proposal:
 * if Horus - no change
 * if not Horus - use this format:
 `<position><delimiter><model name>`
 
Where:
`<position>` is a decimal number
`<delimiter>` one or more whitespace (space, tab) characters (use one space for writing, accept one or more when reading)
`<model name>` normal model name

Use `QStringList QString::split(const QRegExp &rx, SplitBehavior behavior = KeepEmptyParts) const` to split each line into two parts and convert first part to number.

3djc:
Indeed much better. There are a fex chars  we can use as delimiters, as allowed chars for model names is quite limited, thats why I used  ^ that is not allowed for model names

companion/src/storage/categorized.cpp
```diff
@@ -54,10 +56,20 @@ bool CategorizedStorageFormat::load(RadioData & radioData)
...
+        QByteArray curline;
+        if (line.contains(FIELD_DELIMITER)) {
+          modelIndex=line.mid(0,line.indexOf(FIELD_DELIMITER)).toInt();
+          curline = line.mid(line.indexOf(FIELD_DELIMITER)+1);
+          qDebug() << "Detected" << curline;
+        }
+        else {
+          curline = line.data();
+        }
+
```
schwabe: Qstring curline = QString(curline).trimmed()  to remove all the potential extra white whitespaces

bsongis:
Excellent!

PR #4309, opened 2017-01-27
Prevent Companion crash if input data is smaller than required

projectkk2glider:
Loading a bad .otx resulted in segfault because input bits array was smaller then for example radio data array. 

Qt does assert when accessing QByteArray[n] with n >= size.
```
ASSERT: "i >= 0 && i < size()"
```


projectkk2glider:
Actually the bad .otx was due my bug (fixed in cc12601a01b1f5ff86d6c13d288352056f047005) but still we might have invalid file in input and this will at least prevent Companion from crashing.

bsongis:
👍

PR #4319, opened 2017-01-28
Rework X7 flight mode screen

3djc:
- display more info on ALL screen
- allow trim settings on FM0 One screen

NEEDS TESTING


bsongis:
This screen should perhaps be splitted, it's difficult to read, well ... later!
Thanks for this PR

Issue #4222, opened 2017-01-11
Small bug - custom 2 point curve overshoots values to select.


JapieG:
When i create a custom curve on Horus n360:

Type: Custom
Count: 2pts

When i want to select a value for the Y axis it overshoots the values when i try to select it with the rotary encoder.  I know it is unwanted to change x value on 2 point curve. But the way it works now is a bug i think :-)

Video:
https://www.youtube.com/watch?v=UVnQeREYf-0

projectkk2glider:
True, two point custom curve should not even exist. I wonder if the same bug is also on 2.1?

kellybyrd:
Is this something a new-to-the-project dev could pick up as his first bug? Or is there a bunch of history and context required here?

projectkk2glider:
Should be fairly easy. If you are luycky you just need to adjust the min value based on curve type here https://github.com/opentx/opentx/blob/next/radio/src/gui/480x272/model_curves.cpp#L135

kellybyrd:
Ok, I can repro this and I finally underatand @projectkk2glider's comment that "two point custom curve should not even exist." because you would solve that problem with a standard two point curve, like I do for a curve applied for volume control on a slider. Thanks for being patient, I'll dig in now.

projectkk2glider:
Go ahead with your idea.

kilrah:
Can be solved at editing time, when you change number of points to 2 type gets set to standard, when you change type to custom if number of points is set to 2 set it to 3. Ideally... do the same in companion :)

kellybyrd:
I've got a PR out for just skipping the X editing. This is what the current Taranis code does in the simulator, so I followed that example.

PR #4325, opened 2017-01-28
Fix for #4222: On Horus GUI, can't edit X values for < 3-point curves.

kellybyrd:
After looking at the Taranis GUI code, I decided to emulate what that
does, which is just skip over the X coord line in the menu if when
curve type is custom and points is <= 2.

TESTING DONE:
Tested Horus in the simulator. Trying a variety of combinations of 2 and 3 point custom and standad curves while using the rotory dial to scroll thru the menu positions.


bsongis:
Seems perfect to me, thanks!

projectkk2glider:
@kellybyrd thanks. Can you verify if other radios (smaller screen) need this fix also?

kellybyrd:
@projectkk2glider : I didn't, but I will. I only checked the X9D firmware, and that handles things in a different way so doesn't have this bug. I'll check the others.

PR #4327, opened 2017-01-28
Read/write backup fixes:

projectkk2glider:
* radio not found was show behind all windows (missing parent)
 * bad behavior on write when radio not found

Do not merge yet, I will do some more checking and fixing.


projectkk2glider:
Done a bunch of testing with Horus and Taranis:
 * reading/writing models works
 * backup/restore models works (on Taranis, disabled on Horus, which needs fixes and implementation)

This can be merged as far as I am concerned.

bsongis:
Seems good, but I didn't test

PR #4333, opened 2017-01-29
[Build] Filelock added to avoid to compile the same firmware many tim…

bsongis:
…es simultaneously

### DIFF
--- a/radio/util/build-firmware.py
+++ b/radio/util/build-firmware.py
@@ -4,6 +4,7 @@ import os
 import sys
 import subprocess
 import shutil
+import filelock
 from fwoptions import *
 
 # Error codes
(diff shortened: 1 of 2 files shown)


bsongis:
Completely untested
filelock module is needed:
`pip install filelock`

radio/util/build-firmware.py
```diff
@@ -219,9 +218,20 @@
...
 
+if os.path.isfile(errpath):
+    exit(COMPILATION_ERROR)
+
+lockpath = path + ".lock"
+lock = filelock.FileLock(lockpath)
+try:
+    with lock.acquire(timeout = 60*60):
+        if not os.path.isfile(path):
+            firmware_build(path)
```
3djc: You defined it L182: build_firmware(path), not firmware_build(path)
I guess it would require filelock to be installed into docker, so a dockerfile modif too

bsongis: good catch!

3djc:
tested successfully on my local build env. (missing the php front end)

3djc:
This is *well* implemented in production as of 09:00 CET

Issue #4310, opened 2017-01-27
2.2RC9 Taranis lcd.getLastPos() issue


athertop:
in lua, if you display a number using RIGHT alignment then do a getLastPos() - this does not return the right hand position of the written number but the left. So here's some code to demo this:

lcd.drawNumber(61,10,getValue("Cmin")*100,MIDSIZE+PREC2+RIGHT)
xposCons=lcd.getLastPos()
lcd.drawText(xposCons,9,"c-",SMLSIZE)
lcd.drawText(xposCons,15,"min",SMLSIZE)

in RC8 this displayed the same as in 2.1 with the c- and min aligned after the displayed number, but in RC9 it displays c- and min on top of the number left aligned with the number

mpaperno:
I see the same thing in my Lua scripts here.  @3djc  I think it started happening after your updates to the fonts/display (since RC8).  Also the `MIDSIZE` period is now quite large  :)  (larger than `DBLSIZE`)

mpaperno:
Before:
![screenshot-1](https://cloud.githubusercontent.com/assets/1366615/22383322/45d6f2f0-e497-11e6-9eb6-f51620c6cfcf.png)
After:
![screenshot_17-01-27_13-47-13](https://cloud.githubusercontent.com/assets/1366615/22383407/a4bb9d98-e497-11e6-8ad6-68fbdd1f82fb.png)

projectkk2glider:
Might be caused by fixes in #4157 and is probably intentional. @3djc will explain.

3djc:
There two different aspects to this.

YES, numbers and text drawing routine have been streamlined and will not produce the exact same results as RC8. There where bugs  and issues when unusual combination of attributes where used. All of that has been fixed, and the result is that you screens might look different, sometime by a pixel or two, but in several case, the difference can be significant (XXL+RIGHT for example).

The other one is lcd.getLastPos(). If you write RIGHT, the position you give to drawNumber is what you call your right hand position (61 in your example). It is logical that lcd.getLastPos() gives you the side you don't already know, the left one.

@Max, raise a good point about MIDSIZE dot, I'll have a look

mpaperno:
Are you saying `getLastPos()` returns different results based on how you aligned the last draw call?  Or it will now always return NOT the last position (the left/top-most one where it started drawing?) ?  😕  Seems like a major "feature" of getLastPos() was that it was consistent...

3djc:
[...]

PR #4354, opened 2017-01-31
Proposal : deprecated lcdLastPos

3djc:
- Rework ARM 128x64 drawNumber (I had to change gtest for that, but simu and radio test show no issues) to put it in line with X9 and Horus coding
- introduce lcd.getRightPos() and lcd.getLeftPos()
- deprecate lcdLastPos and lcd.getLastPos()
- lcd.getLastPost() now is an alias for lcd.getRightPos()
- lcdLastPos global is now lcdRightPos

This solve #4310


radio/src/lua/api_lcd.cpp
```diff
@@ -123,21 +123,57 @@ static int luaLcdDrawLine(lua_State *L)
...
+@notice Only available on Taranis, please use lcd.getRightPos() instead
 
-@status current Introduced in 2.0.0
+@status current Introduced in 2.0.0, deprecated in 2.2
 */
 static int luaLcdGetLastPos(lua_State *L)
 {
-  lua_pushinteger(L, lcdLastPos);
+  lua_pushinteger(L, lcdRightPos);
+  TRACE("WARNING : lcd.getLastPos() is deprecated, please move to lcd.getLeftPos() or lcd.getRightPos()");
```
mpaperno: As previously discussed, this completely spams the console, very nasty.  Warn once if you need to, but don't make the console otherwise useless (also potentially crash on smaller/slower computers).  It is not going to be a "simple replace" of getLastPos() for all users due to numerous backward compatibility issues.

mpaperno:
IMHO this seems like a lot of trouble to rename a function that people are already used to just because the name isn't technically/entirely accurate.  I think "last" for most people already means the right side since the coordinate system goes left-to-right.  For that matter to take Y into account you need to name them even more specifically (getLastRightX(), getLastLeftX(), etc).  (And then add getLastBottomY(), getLastTopY().   :) )

Perhaps by adding some more constants like `FW` and `FH` to `opentxConstants[]` we could help people do their own calculations. Maybe also some more for the different sized fonts as well (eg. `FW_DBLSIZE` etc), so script authors don't have to calculate/assume that on their own.  For my own scripts I always add FW and FH definitions first thing, and base all my calculations on those, adding/subtracting for different sizes as needed (duplicating what is often done in the C code now).  If a user is drawing graphics, then they already know all the coordinates anyway, so the whole "lastPos()" thing is really for just for fonts anyway.

And I see getLastPos() isn't even available for color LCD, is that right?  So the change is mostly for already-existing radios (besides the X7)?

3djc:
Closed

athertop:
Trying to understand - did nothing come of this?

mpaperno:
Yea I hope my comments were taken as constructive and not just critical.  I was intending a conversation on the subject.  Sorry @3djc if that came across otherwise.

There's still the issue of drawNumber now defaulting to LEFT alignment which is a big change in behavior from past versions IMHO.  It could be handled in each individual script which is affected, but I wonder if that's the "right" way.

athertop:
Yes, I have worked around this issue in my scripts but feel that it does still need to be addressed. getLastRightX(), getLastLeftX() are great ideas and I for one would be happy to see these implemented. Should this be re-opened?

PR #4355, opened 2017-01-31
Move antenna switch to module switches

3djc:
Fix display when module is OFF

This fixes "Horus: Internal/external antenna choice showing up even if internal module is OFF (should be in modulePanel too)" in #4224

### DIFF
--- a/companion/src/modeledit/setup.cpp
+++ b/companion/src/modeledit/setup.cpp
@@ -193,15 +193,6 @@ ModulePanel::ModulePanel(QWidget * parent, ModelData & model, ModuleData & modul
   }
   ui->label_module->setText(label);
 
-  // Antena selection for Horus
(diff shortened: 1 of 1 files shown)


companion/src/modeledit/setup.cpp
```diff
@@ -193,15 +193,6 @@ ModulePanel::ModulePanel(QWidget * parent, ModelData & model, ModuleData & modul
   }
   ui->label_module->setText(label);
 
-  // Antena selection for Horus
-  if (IS_HORUS(firmware->getBoard()) && moduleIdx == 0) {
-    ui->antennaMode->setCurrentIndex(module.ppm.pulsePol);
```
kilrah: Initial dropdown position is now missing!

kilrah:
👍

PR #4358, opened 2017-01-31
Use C++11 clock implementation instead implementation platform specific code ourselves

schwabe:
The old implementation broke on OS < 10.12. Let the C++ stdlib handle the mess.


bsongis:
Excellent, thanks for the tip!

projectkk2glider:
Works on Linux. I learned something new here, I was not aware of std chrono.

kilrah:
Broken on Windows/VC++...

schwabe:
Damn, what happens on Windows?

kilrah:
Edited the message above, and at least temporarily reverted to prior working solution so I can build...

Just including <chrono> is enough to break it, seems there are missing defines in the suite's implementation...

PR #4360, opened 2017-02-01
Re #4224 Add "rename" item to Category context menu in companion.

kellybyrd:
- Adds "rename" to the context menu for categories, it uses the
same code path as the default dblClick handler for the base class.

- Added a separator so the context menu looks more like the one
for models

- Some whitespace changes in mdichild.h, my editor truncates
whitespace on blank lines. Let me know if you don't like this.


bsongis:
Perfect, thanks!

bsongis:
Yes it's good to do some boyscout work while editing a file (as far as it is possible to do the review)

PR #4364, opened 2017-02-01
Check X7 epprom sizes

schwabe:
Since X7 is released now, we should not modify its data struct sizes anymore

X7 is too different from Taranis/sky9x to merge it into one of the other blocks


bsongis:
Excellent, thanks!

PR #4367, opened 2017-02-01
Issue #4224: Implement "duplicate model" in Companion

kellybyrd:
Just what it says in the title, implement the "duplicate model" action in
Companion's context menu. This isn't actually listed in #4224, but it
feels like something that would be and I couldn't find a separate issue
for this.

Also, fixed "modelAdd" so that it only sets the default model if it just
added the first one. Previously, the newly added model would become the
default, which seemed like the wrong thing.


3djc:
I'm not sure I understand, the way to duplicate something is copy paste, or am I missing something ?

kellybyrd:
There was an existing "duplicate" context menu action for categories. It didn't do anything. I agree you can get the equivalent by copy/pasting a model to a blank slot (or to a category in the case of Horus). I just saw duplicate, tried to use it and when it didn't work I fixed it.

kellybyrd:
What should I do with this PR? IMO the code should either have the "duplicate" action wired up properly or it should be removed entirely. Also, I think the "don't change the default model when doing "Add Model" fix belongs in, but I can break that into a separate PR if the work for duplicate needs more discussion.

projectkk2glider:
I had to add small correction 2b83160284f2af4c97e33c4ae6299c60bc44e092. Thanks.

Issue #4284, opened 2017-01-25
Prevent Lua from allocating all heap memory


projectkk2glider:
We need to limit the amount of heap memory that Lua allocates on Horus. 

Currently there is no limit other than the amount of the actual physical memory size. So when bogus Lua script uses up all memory, the rest of the firmware that also uses heap (like model images) will have problems working. The user interface will be affected.

The Taranis is not affected, because it doesn't use heap in other parts of firmware.

Proposed solution:
 * keep track of all heap memory used up by Lua and limit that number to something less (lets call it **system reserve**) than the actual max heap size. 
 * do not forget to count all memory allocations done by Lua scripts that are not counted by the Lua, like bitmaps.

Open questions:
 * how big should system reserve be?
 * what to do to Lua when our calculation shows excessive memory usage. We have two independent Lua states one for widgets/themes other for mix/function/standalone scripts. How to decide which script to kill or even which Lua state. Tracking of Lua memory usage per script is not possible (I think), we can only get the memory usage of the whole Lua state.

kellybyrd:
Which to kill? Fun/Hard problem. The OOM-Killer on Linux has a similar problem. It tries to kill things in order of freeing up the most memory and which are least important to the system, to do this it scores each process. If OpenTX doesn't have per-process metrics, can we make some guesses which state is least least important or which state is more likely to use more RAM? 

Just guessing, but widget/themes seem like they would fit both of these criteria. They're supposed to be just drawing things on the screen, and it feels to me like they're more likely to contain bitmaps.

PR #4369, opened 2017-02-01
Several Lua improvements for stability

### LINKED ISSUE #3885: An error in Lua widget should not disable Lua globally
Such erorr should only disable the offending widget.

Test case:

``` lua
local options = {
    "kill Lua",
}

function create(zone, options)
    local context = { zone=zone, options=options }
    return context
end

function update(context, options)
    context.options = options
end

function refresh(context)
    lcd.drawText(context.zone.x, context.zone.y, "Val: " .. context.options.Val)
end

return { name="Test", options=options, create=create, update=update, refresh=refresh }
```

Result of above script is:

```
PANIC: unprotected error in call to Lua API (bad argument #-1 (table expected, got string))
```

resulting from the failed `luaL_checktype()`

projectkk2glider:
Lua: Use incremental GC and also call it for Widgets
Fixes #3885: Error in Lua Widget options handled better (does not disable entire Lua state)
Disable Lua Widget if any of its functions has error.

This is a rough concept:
 * if a widget has Lua erorr, remember error and disable offending widget (it is still availabe in factory, should it be removed from there also?)
 * show widget error to user in widget settings screen   (very rough, should be improved visually)
 * I thought GC was running in Lua automagically, looks like it is not, lua_gc() call is needed, this commit also fixes this. The Scripts are also not GC so aggressively now (was always full GC). Needs testing and tuning (amount of work for incremental GC can be tuned).
 * Added a variable that keeps the total value of memory used by bitmaps that were loaded from withing Lua. This will enable Lua script memory usage limiting. See #4284 

Further ideas for this PR:
 * add Lua debug buffer (Horus only) which would store last N lua outputs (errors from interpreter and print()), Create a new screen that will display this buffer and allow user to scroll trough it. Options: save to file, clear.
 * Implement #4284

### DIFF
--- a/radio/src/gui/480x272/screens_setup.cpp
+++ b/radio/src/gui/480x272/screens_setup.cpp
@@ -176,6 +176,23 @@ int getOptionsCount(const ZoneOption * options)
 template <class T>
 bool menuSettings(const char * title, const T * object, uint32_t i_flags, event_t event)
 {
+
+  if (object->getErrorMessage()) {
+    // display error instead of widget settings
+    // TODO nicer display (proper word-wrap)
+    SIMPLE_SUBMENU("Widget Error", ICON_MODEL_LUA_SCRIPTS, 1);
+    int len = strlen(object->getErrorMessage());
+    int y = 3*FH;
+    const char * p = object->getErrorMessage();
+    while (len > 0) {
+      lcdDrawSizedText(MENUS_MARGIN_LEFT, y, p, 30);
+      p += 30;
+      y += FH;
+      len -= 30;
+    }
+    return true;
+  }
+
   const ZoneOption * options = object->getOptions();
   linesCount = getOptionsCount(options);
   uint8_t mstate_tab[MAX_WIDGET_OPTIONS];
--- a/radio/src/lua/widgets.cpp
+++ b/radio/src/lua/widgets.cpp
@@ -22,7 +22,7 @@
 #include <stdio.h>
 #include "opentx.h"
 #include "bin_allocator.h"
-#include "lua/lua_api.h"
+#include "lua_api.h"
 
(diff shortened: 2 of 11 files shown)


projectkk2glider:
None of this was tested on actual radio :stuck_out_tongue_closed_eyes: only in `simu`

kellybyrd:
This seems reasonable at first glance.

radio/src/gui/480x272/screens_setup.cpp
```diff
@@ -176,6 +176,23 @@ int getOptionsCount(const ZoneOption * options)
 template <class T>
 bool menuSettings(const char * title, const T * object, uint32_t i_flags, event_t event)
 {
+
+  if (object->getErrorMessage()) {
+    // display error instead of widget settings
+    // TODO nicer display (proper word-wrap)
```
bsongis: Right we could add another flag in LcdFlags for this

radio/src/lua/widgets.cpp
```diff
@@ -385,8 +427,8 @@ void luaLoadFile(const char * filename, void (*callback)())
     }
   }
   else {
-    // luaDisable();
-    lsWidgets = 0;
+    // error while loading Lua widget/theme,
+    // do not disable whole Lua state, just ingnore bad widget/theme
```
bsongis: typo

bsongis:
Yes it seems good. I am only afraid that we will have problems with Taranis where the GC is very sensible!

projectkk2glider:
>I am only afraid that we will have problems with Taranis where the GC is very sensible!

Should not be a problem in my opinion. Lua will do an emergency full GC every time it fails to allocate a new chunk of memory and fails, then it tries to allocate second time - this should take care of your concerns.

bsongis:
I didn't know this, thanks for the info!

Issue #4368, opened 2017-02-01
Companion 2.2 save/load otx file duplicates categories


kellybyrd:
To repro:
Setup a new otx file with:
- Category1
  - Model1
- Category2
  - Model2

Now highlight "Category1", add a new model. You'll have:
- Category1
  - Model1
  - New Model
- Category2
  - Model2

Now save, close and reopen the otx file. You'll have:
- Category1
  - Model1
- Category2
  - Model2
- Category1
  - New Model

The logic in Companion's CategorizedStorageFormat::write() writes the file as it iterates the vector, so if a model at the end of the vector belongs to a category found earlier in the vector, it is written as a new category.

kellybyrd:
I'm looking at this now, I'll end up either sorting the model vector based on category, brute force iterating n*m times, or something more clever.

PR #4370, opened 2017-02-01
Issue #4368. Don't duplicate categories in OTX's models.txt

kellybyrd:
The previous code  didn't account for this situations where the models vector
contained models from category 1, followed by models from category 2,
followed by models from category 1. In this situaion, models.txt would have
two categories named "category 1". 

The overall problem is that we need to output models to models.txt sorted first by category index than by model index.  I'm not really happy with the approach here, but I this is what came to mind:

- Create a map: of category indices to a vector of model indices 
- While running thru the models vector writing .bin files, store indices to models and categories in sortedModels. For category-less models, store them with map key '-1' 
- When done, iterate over known category indices in the map, outputting the category line followed by the model lines 

I tested this in Companion on OSX with both the Horus and Taranis radio profiles, converting back and forth, testing edge cases like empty categories etc. It all seems work. But, the code doesn't feel very readable to me, it's a bit too opaque. So, I'm hoping for suggestions. Some approaches I thought about but discarded:

- Leave existing code, but do std:stable_sort on radioData.models before iterating over it. Use a comparison function that sorted by category index. This feels like the best solution, but given how new I am to the codebase, I was worried sorting the models vector or adding move constructors would had side effects I couldn't see. 

- Forget the "map of indices" data structure and just brute force it. This would be something like and outer loop over categories, then an inner loop over models looking for the current category index. It's m*n time, but maybe we don't care? It would probably leave cleaner looking code.

- Store something more straightforward in the map, like the actual category and model strings to be outputted to models.txt. This doesn't quite work because it should be  have two different categories with the same name, so this would end up actually a little more complicated than what I had

Hopefully someone has a better idea.

### DIFF
--- a/companion/src/storage/categorized.cpp
+++ b/companion/src/storage/categorized.cpp
@@ -125,9 +125,16 @@ bool CategorizedStorageFormat::write(const RadioData & radioData)
     return false;
   }
 
-  // all models
-  for (unsigned int i=0; i<radioData.models.size(); i++) {
-    const ModelData & model = radioData.models[i];
+  // Iterate through all models, writing to their .bin files. Because radioData
+  // isn't sorted w.r.t categories, we have to build up sortedModels so ww can
+  // write models.txt in the correct order. We use category index '-1' to handle
+  // all the models that have no category.
+  unsigned int numModels = radioData.models.size();
+  unsigned int numCategories = radioData.categories.size();
+  std::map<int, std::vector<int>> sortedModels;
+
+  for (unsigned int m=0; m<numModels; m++) {
+    const ModelData & model = radioData.models[m];
     if (model.isEmpty()) continue;
 
     QString modelFilename = QString("MODELS/%1").arg(model.filename);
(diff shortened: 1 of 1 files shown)


companion/src/storage/categorized.cpp
```diff
@@ -136,20 +143,33 @@ bool CategorizedStorageFormat::write(const RadioData & radioData)
...
+      const ModelData & model = radioData.models[sortedModels[c][m]];
+      QString line;
+      if (IS_HORUS(getCurrentBoard())) {
+        line = QString("%1\n").arg(model.filename);
+      }
+      else {
+        // use format with model number and file name
+        // this is needed, because this kind of radios can have unused model slots
+        // NOTE: The model number will only be unique within a category
+        line = QString("%1\n").arg(model.filename);
```
projectkk2glider: This won't work on radios that have no categories. Such radios need a different line format with `<model number> <model name>` because of #4295.

So for such radios, be sure to keep the old code (in this case there is no need for map<> because you have no categories.

For radios that have categories your code looks good.

companion/src/storage/categorized.cpp
```diff
@@ -136,20 +143,33 @@ bool CategorizedStorageFormat::write(const RadioData & radioData)
...
     }
-    if (getCurrentFirmware()->getCapability(HasModelCategories)) {
-      int categoryIndex = model.category;
-      if (currentCategoryIndex != categoryIndex) {
-        modelsList.append(QString().sprintf("[%s]\n", radioData.categories[model.category].name));
-        currentCategoryIndex = categoryIndex;
-      }
-    }
-    if (IS_HORUS(getCurrentBoard())) {
-      modelsList.append(QString("%1\n").arg(model.filename));
```
projectkk2glider: While you at it, the original code here basically uses these two conditions:
 * `getCurrentFirmware()->getCapability(HasModelCategories)`
 * `IS_HORUS(getCurrentBoard())`

In this context the two actually mean the same. So I would loose the second condition and base all decisions in this functions on a `getCurrentFirmware()->getCapability(HasModelCategories)` This condition determines if output has either:
 * categories, models, sorted with your map<> code
 * no categories, but each model has model number.

projectkk2glider:
Closes #4368

companion/src/storage/categorized.cpp
kellybyrd: Ok, so I think I'll just have an explicit if/else at the top and two distinct code paths, either the "hasModelCategories" or doesn't. The "doesn't" path will be a simpler version of the old code. That should be overall more readable because it won't have to try and be tricky to handle both at the same time.

kellybyrd:
Ok, re-worked the code, should be cleaner and actually correct this time.

projectkk2glider:
Look good now, you have one warning which prevents compile, look at Travis output.

kellybyrd:
Glad CI caught that, I should have had WARNINGS_AS_ERRORS set locally, I've done it now. Fixup pushed.

Would it be useful to have that as part of cmake line for all the build instructions? It would let people catch things that before pushing.

projectkk2glider:
Excellent :+1:

PR #4371, opened 2017-02-02
Issue #4224: Paste into and across categories in Companion

kellybyrd:
Issue #4224. Allow pasting models into and across categories. This one is actually on the list for #4224. 

   To do this, we need both the current category and model index
    passed into doPaste(), which then decides if the user is pasting
    into a category based on whether or not modelIdx == -1.

    Doing a category paste just means we stub out a ModelData object,
    insert it into the models vector then use the existing code to
    paste over it.

    Also, the code now allows users to copy/paste models across
    categories.

Testing Done:
- With Horus radio profile
  - I tried various combinations of pasting single models and multiple models onto existing models, on to blank new categories and existing categories. Saved, reopened, it all looked fine.
- With Taranis radio profile
  - I tried various combinations of pasting single models and multiple models onto existing models and on to blank slots. Saved, reooned, it all looked fine.

### DIFF
--- a/companion/src/mdichild.cpp
+++ b/companion/src/mdichild.cpp
@@ -235,11 +235,13 @@ void MdiChild::doCopy(QByteArray * gmData)
 // gmData->append((char *) &radioData.generalSettings, sizeof(GeneralSettings));
 }
 
-void MdiChild::doPaste(QByteArray * gmData, int index)
+void MdiChild::doPaste(QByteArray * gmData, int modelIdx, int categoryIdx)
 {
   char * gData = gmData->data();
   bool modified = false;
   int size = 0;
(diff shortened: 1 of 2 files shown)


projectkk2glider:
Looks nice, thanks! 

I will merge and then do a test.

projectkk2glider:
One small issue remains:
 * copy two or more models
 * paste onto the last model
 * only first model gets pasted (actually overwrites the last model - this is OK)
 * other models do not get created, but the question dialog box is displayed for each of them regardless.

I would create a new models in such case and not ask user anything. So
* first model overwrites the model on which paste was called
* all subsequent models get added to the end of appropriate category.

kellybyrd:
Thanks for pointing that out, I'll have a look.

Issue #4375, opened 2017-02-02
Companion copy/pasting multiple models at the end doesn't work


kellybyrd:
From https://github.com/opentx/opentx/pull/4371
 @projectkk2glider said:

One small issue remains:
 - copy two or more models
- paste onto the last model
- only first model gets pasted (actually overwrites the last model - this is OK)
- other models do not get created, but the question dialog box is displayed for each of them regardless.

I would create a new models in such case and not ask user anything. So
- first model overwrites the model on which paste was called
- all subsequent models get added to the end of appropriate category.

This should be assigned to me, I'm familiar with this code :-)

kellybyrd:
Hmm, I just realized the right behavior is probably:
- When pasting multiple models on the last few models inside a category, the pasted models should all end up inside the same category. 

This means we need a way to know how many models are in a given category. I don't see an obvious way to get this from the radioData without walking the list models. Maybe I get figure it out from TreeModel .

PR #4381, opened 2017-02-03
Issue #4375, fix pasting multiple items in Companion.

kellybyrd:
Fixes issue #4375.

With category-capable radio profiles, finding the next row to paste into mean following the UI rows rather than just getting the next model in radioData. Also, we have to account for possibly pasting "off-the-end" of the category. In that case, the extra models in the paste data will end added to the same category. 

For non-category capable models, the same code works (I tested on Taranis profile!)  because the UI layout appears sort of like one big category.


projectkk2glider:
:+1:

Issue #3685, opened 2016-08-03
Lua debugging requests – part 1


pau33:
I worked a lot the last time on writing complex mix scripts in lua for the taranis. Debugging of these scripts in OpenTx is not the easiest thing. So I have 3 request for making debugging easier. The first is the possibility to apply a filter to the debug console output (PF6) in companion. For example in a way, that only lines beginning with a defined string (for ex. „lua“) are shown. So I have partly the ability to look at my print commands (starting with „lua...“) or at the lua script error lines. Another solution would be to print the lua statements additionally into a second window. Also the ability to delete all content in the debug console output would help.

projectkk2glider:
Nice ideas.

mpaperno:
Filter is a great idea and should be pretty easy to add.  The file I/O logging can get pretty verbose indeed.  Also the GC messages can flood the screen on more complex scripts.

I wonder if filtering messages at a higher level would be useful in addition or instead (perhaps at simulator program interface level).  I realize there are some build-time options to control that now, but those aren't helpful for typical use.

Also on my list is getting the debug info window to remember its size and position instead of obscuring the sim each time.

-Max

mpaperno:
Closed via #4385.

Issue #3284, opened 2016-02-09
Seperate debug output from PRINT command output or increase debug window buffer


lshems:
when debugging, you get a lot of info, for instance about files being loaded.

This blocks viewing the content of the variables you want to check by using the PRINT command. It has already scrolled away, so you cannot analyse it.

2 options perhaps:
make the buffer on the debug screen bigger (easy one perhaps)
make 2 screens, on for debug info, and the other for results of the PRINT command

Thanks

lshems:
Found the CLI option. Will check that out.
#3000

projectkk2glider:
>make the buffer on the debug screen bigger (easy one perhaps)

The scroll back buffer should already be unlimited. Are you sure you have missing lines?

lshems:
In 2.0 sure. I'll check 2.1
Op 11-feb.-2016 09:36 schreef "Damjan Adamic" <notifications@github.com>
het volgende:

lshems:
2.1 as well.

My main problem is file loading being reported. That completely fills my debug window.

mpaperno:
A filter as suggested in #3685 would solve it?

mpaperno:
Closed via #4385 (and lack of further input).

PR #4385, opened 2017-02-03
[Simulator] New main UI window, debug console, separate outputs window, more.

mpaperno:
Replaced the "dialog" paradigm with a full "Main Window" interface.  This brings full, persistent state management of all child windows, menus, toolbar, dockable/floatable windows, and more.

 * Use a MainWindow interface with dockable/floatable windows, proper toolbar & menu;
 * Many aspects of UI can be customized, including many layout options, show/hide elements, etc;
 * Persistent size/layout state of all windows/docks/etc between uses, per radio profile;
 * Completely separate radio Outputs window with configurable layout (also persistent between uses);
 * More compact layout for Telemetry simulator, reduces min. width and adds scrolling columns to allow smaller window sizes.
 * Trainer simulator gets hold/lock buttons and resize support;
 * Translations now enabled in standalone Simulator, follows main Companion setting (they need a lot of updates);
 * New custom icons throughout, toolbar icon sizes adhere to user preference in Companion.

Debug Output console updates:
 * New filter feature with optional full RegEx support, can work in inclusive or exclusive modes (full help text is included in UI);
 * Up to 50 user-defined filters are saved to permanent settings, also a few predefined filters are provided;
 * Saves/restores the last filter used and if it was en/disabled;
 * Added word wrap and clear screen features;
 * Configurable scroll-back buffer size;
 * Increased efficiency & performance of the buffering/printing process.

Note this update requires adding **SVG** support for Qt (basically including Qt5Svg module/libs).  I've updated the Windows packaging script (`companion-vs.nsi.in`) [here](https://github.com/opentx/opentx/commit/83d7cc043e593f6d9d41058e0ad43967b042a511) as well as `commit-tests.sh` and also took a stab at updating Docker (but not tested) [here](https://github.com/opentx/opentx/commit/70b06c05353e2bbefc795f940c51d8fa183b5489).  I'm not sure where else I may have missed, but let me know.  It's a bit of trouble up front, but SVG is superior to raster graphics in many ways.

References: #3685, #3284

win1

win2

osx2

linux1

osx1

win4

linux3

filter


mpaperno:
Only problem with SVG graphics is that they count as +lines of code!   😛

kilrah:
Wowwwwwww.... 😳

mpaperno:
As a note, sometimes trying to dock a window just doesn't seem to work, or it doesn't want to snap into the right spot.  You can pretty much always get what you want by first making the main window big(ger), so the new widget has plenty of room to fit in.  Then after it's docked you can shrink the size of the whole window back down if you want.   A floating widget will always snap back to it's old spot when re-docked using dbl-click or icon on title bar (so no need to re-position it manually again).

dev-fred:
Very impressed.
Minor problem under W10/MSVC, I have no icon on the left column.
capture

kilrah:
> As a note, sometimes trying to dock a window just doesn't seem to work, or it doesn't want to snap into the right spot. 

Yup, noticed that and found the workaround - no problem making the window bigger on my 4K screen, but others might have it harder 😜
Also once everythingis docked you don't seem to be able to slide the separators to redistribute the sizes. The cursor does change to the appropriate type, but it doens't work.

> Minor problem under W10/MSVC, I have no icon on the left column.

Put a copy of Qt5Svg.dll in the program's folder.

mpaperno:
@dev-fred Thanks!  The missing icons are most likely a missing Qt image plugin or DLL.  Do you have Qt5Svg.dll and (possibly needed) `/imageformats/qsvg.dll` ?    (Andre is quick!)

mpaperno:
> Also once everythingis docked you don't seem to be able to slide the separators to redistribute the sizes. The cursor does change to the appropriate type, but it doens't work.

What happens is the radio widget is fixed width (when docked into the main window).  So you can/should be able to resize any widgets on the right/left of the radio and have them grow instead of stretching the radio widget wider (which would be pointless).  The height should remain adjustable but of course you need extra height available otherwise you can't move the splitters.

mpaperno:
Another note.  The decision to have the widgets start out un-docked was deliberate, partly because I wanted to keep the existing behavior as much as reasonable (where new windows open around the radio window).  But it could easily be done the other way, that they show up docked by default (though it's not always in the "best" spot so user will likely want to move it anyway).  So, that's an option.

kilrah:
> What happens is the radio widget is fixed width (when docked into the main window)

Yeah makes sense. Other weird thing then, if I dock something below the radio there is no separator, above is OK. 

>The decision to have the widgets start out un-docked was deliberate

About that, on Win the first time I opened them they would be tucked in the top left corner of my desktop, and had no title bar to move them. Resizing them made the title bar appear, and only then could I move them.

clipboard01

jtaylor2:
Windows 10 VS15 QT5.7.0 - Very nice indeed.  Is it possible to get the switches on the X9D simulator laid out the same way they are on the transmitter and  2.1 simulator -  SF SE SA SB    SC SD SG SH - instead of alphabetical left to right, or is that asking too much?  Regardless great work!

mpaperno:
> Other weird thing then, if I dock something below the radio there is no separator, above is OK.

I see that too.... strange, and not sure when that started.  The order of how the widgets are added and to which area can make differences and is a bit of a PITA.

> on Win the first time I opened them they would be tucked in the top left corner of my desktop,

I've seen that also, and other times they show up right by the main window like they should.  I haven't found a pattern (or SO answer :) ) yet.  Maybe have to brute-force move them over to the main window on first show, but I'd really like to avoid that.

mpaperno:
>  SF SE SA SB SC SD SG SH

Possibly... but maybe not on this rev.  You can always change the names in the radio hardware config...  :)   j/k    There's a balance here between having it all dynamically generated from actual radio settings (like it is now) vs. closely matching the physical layout of a stock radio but being fixed (like it was).

kilrah:
Seems to run fine on mac. No "widgets in the corner" problem either.

mpaperno:
Andre,
> on Win the first time I opened them they would be tucked in the top left corner of my desktop,

Do you by any chance have the console window enabled with WIN_USE_CONSOLE?   I just tried w/out it and can't recreate the issue (new windows open around main win).   Enabled console and issue is back (though again not consistent).  It kinda does make sense... that top left 0,0 corner is also where the console window opens by default (if not launching from CLI).  So I wonder if having an "extra" window attached to the process is screwing Qt up with window placement.

On my Linux VM the new windows always open in top left corner, but they're not hidden at all and easy to  move.  On OSX VM the new windows open around the main window.  Both those VMs use only one monitor.

kilrah:
No, console wasn't enabled.

bsongis:
Really nice result!

mpaperno:
> Other weird thing then, if I dock something below the radio there is no separator, above is OK.

Figured out this is a result of making the radio widget fixed-width.  I'm not sure why, tried a bunch of stuff but it's always consistent, also across platforms.  Perhaps a Qt bug.

So at this point I'm not sure what would be best... I do like how it behaves when fixed-width in terms of resizing the other docked windows w/out stretching the radio UI.  OTOH there's this strange bug/issue, and also this fixed-width behavior may not be entirely intuitive.  Eg. needing to resize the radio widget after stretching out the debug output isn't the worst thing in the world.  My first reaction to such situations is usually to add a user-controlled option (fixed-size radio widget on/off)...  but too  many options can also spoil the soup.

PR #4386, opened 2017-02-03
Fixes telemetry stream confusion

RealTadango:
Fixes telemetry stream confusion. 
0x5100 to 0x51ff is DIY stream data, 
0x5200 to 0x52ff is diy sensor data and is treated the same as any other telemetry packet.

### DIFF
--- a/radio/src/telemetry/frsky.h
+++ b/radio/src/telemetry/frsky.h
@@ -139,6 +139,8 @@
 #define RBOX_STATE_LAST_ID        0x0b2f
 #define RBOX_CNSP_FIRST_ID        0x0b30
 #define RBOX_CNSP_LAST_ID         0x0b3f
+#define DIY_FIRST_ID              0x5100
+#define DIY_LAST_ID               0x52ff
 #define DIY_STREAM_FIRST_ID       0x5100
 #define DIY_STREAM_LAST_ID        0x51ff
 #define RSSI_ID                   0xf101
--- a/radio/src/telemetry/frsky.h
+++ b/radio/src/telemetry/frsky.h
@@ -139,8 +139,8 @@
 #define RBOX_STATE_LAST_ID        0x0b2f
 #define RBOX_CNSP_FIRST_ID        0x0b30
 #define RBOX_CNSP_LAST_ID         0x0b3f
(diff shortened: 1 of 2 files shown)


radio/src/telemetry/frsky.h
```diff
@@ -139,8 +139,8 @@
 #define RBOX_STATE_LAST_ID        0x0b2f
 #define RBOX_CNSP_FIRST_ID        0x0b30
 #define RBOX_CNSP_LAST_ID         0x0b3f
-#define DIY_FIRST_ID              0x5000
-#define DIY_LAST_ID               0x52ff
```
kilrah: Maybe leave the DIY defines (corrected) for reference

RealTadango: Ah so you want to have a complete DIY range (not used in code) and a Streaming part

radio/src/telemetry/frsky.h
```diff
@@ -139,8 +139,10 @@
 #define RBOX_STATE_LAST_ID        0x0b2f
 #define RBOX_CNSP_FIRST_ID        0x0b30
 #define RBOX_CNSP_LAST_ID         0x0b3f
-#define DIY_FIRST_ID              0x5000
+#define DIY_FIRST_ID              0x5100
```
projectkk2glider: I think this can stay at 0x5000

RealTadango: But how about the FrSky specs? DIY_FIRST_ID is not used in code so why not simply maintain the specs?

projectkk2glider: We had email confirmation from FrSky that the new reserved range in their spec will be https://github.com/opentx/opentx/pull/4232#issuecomment-274943497

kilrah: Yes this means that 5000-52ff are free to use, BUT we don't want all of that to go to lua exclusively. 
We want 5000-50ff to go to lua, and the rest to be available for standard sensors.

RealTadango: @kilrah That would work great, and just as it always has....... i'l adjust the range for streaming

bsongis:
Good

PR #4388, opened 2017-02-04
Chinese translation update for better alignment

Kongse:
Chineses fonts width of ubuntu is differents from windows, layout ( words align ) re-modified for windows version, considering most people use windows version


bsongis:
Seems perfect chinese :)

PR #4397, opened 2017-02-04
Removed const modifer to Widget::update in the 480x272 GUI.

kellybyrd:
When compiling on OSX with px4/px4/gcc-arm-none-eabi: stable 20160928, I was getting const errors with a recent gcc. 

So I removed the const modifer to Widget::update in the 480x272 GUI. This caused menu_settings.cpp menuSettings() to complain, because it was being passed a param called
'object' of type pointer to a const T, and then called object->update(),
which is not a const function.

However, the one call site I found to menuSettings() (menu_settings.cpp:224)
is passing currentWidget, which is of type Widget*. 

'currentWidget' is a static though, and I don't have the full picture on this code yet, so maybe there's a problem with multithreaded use that the const-ness was trying to solve?

### DIFF
--- a/radio/src/gui/480x272/widget.h
+++ b/radio/src/gui/480x272/widget.h
@@ -45,7 +45,7 @@ class Widget
     {
     }
 
-    virtual void update() const
+    virtual void update()
     {
     }
 
--- a/radio/src/gui/480x272/screens_setup.cpp
+++ b/radio/src/gui/480x272/screens_setup.cpp
@@ -174,7 +174,7 @@ int getOptionsCount(const ZoneOption * options)
 }
 
 template <class T>
-bool menuSettings(const char * title, const T * object, uint32_t i_flags, event_t event)
+bool menuSettings(const char * title, T * object, uint32_t i_flags, event_t event)
 {
 
   if (object->getErrorMessage()) {


kellybyrd:
Marking this as "undecided" until some people familiar with this area can comment.

bsongis:
no multithread issue there, we only update widgets in the "menus" thread

kellybyrd:
Ok, so then the LuaWidges::update() definition is correctly non-const (I think, it based on its name), and the code calling it in the 480x272 GUI is passing in an object which started life without the const modifier, so this change should be the right thing to do.

kellybyrd:
After some discussion, we realized this is a result of -DWARNINGS_AS_ERRORS in cmake and specific to the simu target. Not sure if these are warnings for other compilers.

mpaperno:
`-DWARNINGS_AS_ERRORS` just turns on -Werror.  It should not cause new warnings to appear.  The compiler choice certainly will make a difference though.  Also, for everything except the `firmware` target we're passing on all compiler/linker flags as per CMake config (which can be different between CMake versions)... so things like setting the type to Release can affect the outcome.

When I first tried building the 480x272 gui stuff on a new compiler it [barfed](https://github.com/opentx/opentx/pull/4141) all over the place.

radio/src/gui/480x272/widget.h
```diff
@@ -45,7 +45,7 @@ class Widget
     {
     }
 
-    virtual void update() const
+    virtual void update()
```
projectkk2glider: This is good change. I had to remove const here because Lua widget will modify itself on update() if it encounters any error. I guess my compiler did not catch this.

bsongis:
OK your compiler is right. Did it clash with the simu? Then it's not the arm compiler which is involved!

PR #4406, opened 2017-02-05
Fixes #4315

### LINKED ISSUE #4315: 2.2 RC9 Horus. Old model (2.1 eepe) converted LS to L1
The conversion of a model from OpenTX 2.1 format to to 2.2 format for the Horus seems to have a minor glitch. I was using LS as input to a mixer. In the converted file it became L1 as input to the mixer. 

**Edit by projectkk2glider**: see also description of another similar issue https://github.com/opentx/opentx/issues/4315#issuecomment-275899641

bsongis:
A lot of conversions missing, but at least this one will be ok ...


companion/src/radiodata.h
```diff
@@ -1213,7 +1217,7 @@ class GeneralSettings {
     unsigned int backlightColor;
     CustomFunctionData customFn[CPN_MAX_CUSTOM_FUNCTIONS];
     char switchName[18][3+1];
-    unsigned int switchConfig[18];
+    unsigned int switchConfig[CPN_MAX_SWITCHES];
```
mpaperno: Nice, I had forgotten about that FIXME until now.  Also, good move to `getBoardCapability()`, I think.  (Though I'm still not in love with the globals floating around.)

bsongis: which global?

mpaperno: `getBoardCapability()`, `getCurrentFirmware()`, `g`, `current_firmware_variant`, etc...  Quite a few in eeprominterface.h, it seems like.

companion/src/eeprominterface.cpp
```diff
@@ -582,14 +582,14 @@ bool RawSource::isPot() const
 {
   return (type == SOURCE_TYPE_STICK &&
           index >= CPN_MAX_STICKS &&
-          index < CPN_MAX_STICKS+getCurrentFirmware()->getCapability(Pots));
+          index < CPN_MAX_STICKS + getBoardCapability(getCurrentBoard(), Board::Pots));
```
projectkk2glider: Maybe add new shortcut function `getCurrentBoardCapability(Board::Capability capability)` :wink:

companion/src/radiodata.h
bsongis: not wrong, still a lot to do :)

companion/src/eeprominterface.cpp
bsongis: with a new class Board it would be `getCurrentBoard()->getCapability(Pots)` which is easy to read, would it be ok?

projectkk2glider: Sure.

PR #4409, opened 2017-02-05
Multiprotocol working on sky9x

kilrah:
Telemetry not working yet though, still something to debug

### DIFF
--- a/radio/src/pulses/multi_arm.cpp
+++ b/radio/src/pulses/multi_arm.cpp
@@ -46,7 +46,7 @@ static void sendByteMulti(uint8_t b)
     parity = parity ^ (b & 1);
     b >>= 1;
   }
-  putDsm2SerialBit(parity);      // Even Parity bit
+  putDsm2SerialBit(!parity);     // Even Parity bit
 
   putDsm2SerialBit(1);           // Stop bit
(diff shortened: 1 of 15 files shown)


radio/src/pulses/multi_arm.cpp
```diff
@@ -46,7 +46,7 @@ static void sendByteMulti(uint8_t b)
     parity = parity ^ (b & 1);
     b >>= 1;
   }
-  putDsm2SerialBit(parity);      // Even Parity bit
+  putDsm2SerialBit(!parity);     // Even Parity bit
```
schwabe: I am not sure about this change. Wouldn't that change the parity on the other platforms as well?

kilrah: It's only for PPM_PIN_SERIAL, which I believe isn't used on any other platforms, which explains it was never noticed that it was wrong.

schwabe: Oh then did not see enough context. Thought horus was that too, never mind then.

bsongis:
Perhaps you could test with my latest PR?

kilrah:
Nope doesn't help unfortunately

PR #4411, opened 2017-02-06
No longer ignore output files

schwabe:
If these are present in a 2.2 OpenTX repo, they are remains from a
OpenTX 2.1 build


bsongis:
good idea

PR #4417, opened 2017-02-06
Do not reset telem on ENT push while in telem view

### LINKED ISSUE #4412: [QX7/LUA] Pressing rotary dial button disables RSSI (and possibly other telemetry values)
When executing this very simple script:
```local function run_ui(event)
   lcd.clear()
      -- do we have valid telemetry data?
   if getRSSI() == 0 then
      -- No!
      lcd.drawText(36,25,"No Telemetry",INVERS)
   end
end

return {run=run_ui}
```

As a telemetry script on a QX7, you will notice that strangely, the message "No Telemetry" gets printed. Isn't it a bit odd?

3djc:
on ARM 128x64. Instead gracefully display a popup instead

Fix #4412

### DIFF
--- a/radio/src/gui/128x64/view_telemetry.cpp
+++ b/radio/src/gui/128x64/view_telemetry.cpp
@@ -540,13 +540,14 @@ void menuViewTelemetryFrsky(event_t event)
       incrTelemetryScreen();
       break;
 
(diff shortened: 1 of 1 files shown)


3djc:
I'll improve it tomorow morning

PR #4422, opened 2017-02-07
[Simulator] Add ability to update .otx files.

mpaperno:
This could use some testing to make sure all the various startup permutations are covered and data is properly updated.

- [ ] Non-Horus with existing old-school .bin|.eepe files.
- [ ] Non-Horus with existing .otx files.
- [ ] Non-Horus with new .bin file.
- [ ] Non-Horus with new .otx file.
- [ ] Non-Horus from within Companion (whole radio and model-only).
- [ ] Horus with existing .otx files.
- [ ] Horus with new .otx file.
- [ ] Horus with SD card path.
- [ ] Horus with existing data in separate folder from SD path.
- [ ] Horus with new data in separate folder.
- [ ] Horus from within Companion (whole radio and model-only).

(Did I miss any?)

Radio & model settings should be saved back to the file/data path in all cases except when launched from within Companion.

Also, I could not find an existing way to get radio data back from SimulatorInterface, so I created the new `readEepromData()` function.  But if I missed an existing method, or there's a better way to go about this, I'm all ears.

All `libsimulator` flavors need to be rebuilt after this change, the old versions WILL crash when launched with this Simulator/Companion.

Thanks,
-Max

### DIFF
--- a/companion/src/helpers.cpp
+++ b/companion/src/helpers.cpp
@@ -824,22 +824,23 @@ void startSimulation(QWidget * parent, RadioData & radioData, int modelIdx)
     }
 
     SimulatorMainWindow * dialog = new SimulatorMainWindow(parent, simulator, flags);
-    if (IS_HORUS(getCurrentBoard()) && !dialog->useTempDataPath(true)) {
-      QMessageBox::critical(NULL, QObject::tr("Data Load Error"), QObject::tr("Error: Could not create temporary directory in '%1'").arg(QDir::tempPath()));
+    if (!dialog->setRadioData(simuData)) {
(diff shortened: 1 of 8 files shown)


mpaperno:
Oh yes, forgot some permutations a'rite...   Added above:

- [ ] Non-Horus with **new** .bin file.
- [ ] Non-Horus with **new** .otx file.
- [ ] Horus with **new** .otx file.
- [ ] Horus with **new** data in separate folder.

"New" meaning a non-existing file/data path entered in the startup dialog.

You can probably see why I can use some help testing....  :-o

bsongis:
Nice feature, users asked for it since long!

PR #4425, opened 2017-02-07
Bsongis/issue 4400

### LINKED ISSUE #4400: Companion Mixer Copy problem, n362
In Companion I tried to copy a mix from Ch16 to Channel 10.
QX7 selected as transmitter, Operating system Ubuntu.
This is repeatable.

compcopy

bsongis:
It should fix #4400 
Please review
Discussion needed about all the TODO I introduced

### DIFF
--- a/companion/src/eepromimportexport.h
+++ b/companion/src/eepromimportexport.h
@@ -26,15 +26,24 @@
 
 class DataField {
   public:
-    DataField(const char *name=""):
+    DataField(DataField * parent, const char * name=""):
+      parent(parent),
       name(name)
     {
     }
-    virtual const char *getName() { return name; }
-    virtual ~DataField() { }
+
+    virtual ~DataField()
+    {
+    }
+
+    virtual const char * getName()
+    {
+      return name;
+    }
+
+    virtual unsigned int size() = 0;
     virtual void ExportBits(QBitArray & output) = 0;
     virtual void ImportBits(const QBitArray & input) = 0;
-    virtual unsigned int size() = 0;
 
     QBitArray bytesToBits(QByteArray bytes)
     {
@@ -99,13 +108,24 @@ class DataField {
     }
 
   protected:
-    const char *name;
+    virtual void setError(const QString & error)
+    {
+      if (parent) {
+        parent->setError(error);
+      }
+      else {
+        qWarning() << qPrintable(error);
+      }
+    }
+
+    DataField * parent;
+    const char * name;
 };
 
 class ProxyField: public DataField {
   public:
-    ProxyField():
-      DataField("Proxy")
+    explicit ProxyField(DataField * parent):
+      DataField(parent, "Proxy")
     {
     }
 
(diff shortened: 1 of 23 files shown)


companion/src/simulation/simulatordialog.cpp
```diff
@@ -255,12 +255,14 @@ bool SimulatorDialog::setRadioData(RadioData * radioData)
   bool ret = false;
   if (radioDataPath.isEmpty()) {
     QByteArray eeprom(getEEpromSize(m_board), 0);
-    if (firmware->getEEpromInterface()->save((uint8_t *)eeprom.data(), *radioData, 0, firmware->getCapability(SimulatorVariant)) > 0)
+    if (firmware->getEEpromInterface()->save((uint8_t *)eeprom.data(), *radioData, 0, firmware->getCapability(SimulatorVariant)) > 0) {
```
mpaperno: Definite merge issue here with #4422 ... maybe we do this cleanup later?

companion/src/eepromimportexport.h
```diff
@@ -698,8 +698,11 @@ class ConversionField: public TransformedField {
       if (table) {
         if (table->exportValue(_field, _field))
           return;
-        if (!error.isEmpty())
+        if (!error.isEmpty()) {
+          // TODO how could we avoid this global?
```
mpaperno: Not sure there's a quick fix here... who "owns" `EEPROMWarnings`?  

Basically to avoid globals we need everything to belong to a class, even if that class is a singleton.  Then use singletons as, eg. (totally made up):
```
EepromInterface myEepe = EepromInterface::instance();
QStringList warns = myEeepe.warnings();  // or whatever we want to do with the instance
```

companion/src/simulation/simulatordialog.cpp
bsongis: yes that's what I just did

companion/src/eepromimportexport.h
bsongis: well, I am not convinced :) I would like to add a parent to each field so that `setError(...)` calls `parent->setError(...)` but I am afraid of having far too many changes ... well perhaps not an issue

mpaperno: What is the ultimate goal in this particular case?  I only see `EEPROMWarnings` displayed from inside `OpenTxEeepromInterface::save()`.

The following idea does a bit of assuming but I'll toss it out there anyway.

If each field is a subtype of `DataField` then it would make a lot of sense to have an error/warning handler in the parent class (or perhaps in `TransformedField` if `DataField` is too broad).  I'm not sure why you suggest the parent/child relationship, but if you need to track a hierarchy then that could be built into the handler (you could pass the parent and child object names for example, along with the message and e.g. error code). Of course the parent class already has some of that info, like `name`, so that could easily be added to the warning data already.  E.g. just something as simple as `DataField::addWarning(const QString &txt)` would do it, no?

As to how to actually get the error "collector" in there (`EEPROMWarnings`), I can think of a few ways... but here's one.  Since `EEPROMInterface` already essentially exists as a singleton, I'd add the "collector" as a member of that (eg. it could be a public struct definition with more data than just message).  Then in the `DataField` (or `TransformedField`) initializer I'd get a reference to the current `EEPROMInterface` instance (`EEPROMInterface * eeInstance = getCurrentEEpromInterface();`).  I'd then use whatever function(s) `EEPROMInterface` provides for adding warning messages (I'd suggest a function rather than direct access to the "collector" variable).  Or in other words inside our theoretical `DataField::addWarning(txt)` function we'd just call something like `eeInstance->addWarning(txt);`.

Actually I'd probably make `EEPROMInterface` and `DataField` inherit from QObject and then connect `DataField::warningAdded()` to `EEPROMInterface::onWarningAdded()` (or something like that).

Of course `EEPROMWarnings` could also be its own self-contained independent singleton/static class/struct, but I'm not sure it would add much benefit.  Or a part of new class for that matter, eg. `EepromImportExport`.  At that point it becomes a question of if that same warnings container would be useful for other aspects of dealing with `EEPROMInterface`, or if it's purely for import/export.

PR #4427, opened 2017-02-08
[MinGW] make install takes the pthread dll from Qt directory

bsongis:


mpaperno:
I don't really like this.  The runtime libs should be taken from the toolchain just used to build the application.  Not whatever Qt happens to be built with.  Just because they may be binary compatible doesn't mean they're the same versions.  Also you're using a different MinGW to begin with, it should be `mingw-w64`.  (BTW the former has `libpthread-2.dll`)

mpaperno:
Though really since both Qt libs and the built app itself link to the pthread dll they should ideally be the same version.  The recommended toolchain is the one each Qt distribution comes with, eg. for 5.7 it is mingw-w64 5.3.

bsongis:
OK then :)

PR #4428, opened 2017-02-08
X7 switch warning, take 2

3djc:
Using arrow a bit like x9, while keeping X7 menu behaviour.

ENT on a switch enables or disable that switch check (when anble, and arrow or - is played, when disabled, a blank space)

Long ENT while anywhere in that line stores the positions


radio/src/gui/128x64/model_setup.cpp
```diff
@@ -499,7 +504,7 @@ void menuModelSetup(event_t event)
         }
 
         LcdFlags line = attr;
-
+#if !defined(PCBX7)
```
bsongis: Perhaps avoid the negative tests: here PCBX7 code could go first!

bsongis:
Just cosmetics, thanks!

PR #4432, opened 2017-02-09
Horus : limit model bitmaps filemane to 6 chars

### LINKED ISSUE #4415: RC10 N362: It does not accept any file with more than 6 letters in the name.
I put my custom WAV files. If larger than 6 letter name than it does not appear in the track list.

With the JPG's if the file is with larger than 6 letters in the name, it is visible you can select it as model icon but then it does not load and the model stays with no image. 

Once I renamed the files to 6 letter names then no problem. Sounds load, jpgs are visible.

3djc:
Limit filename to 6 as sdcardfilelist controls size without extension (and size with ext was sent).

This closes #4415

### DIFF
--- a/radio/src/sdcard.h
+++ b/radio/src/sdcard.h
@@ -69,7 +69,7 @@ const char RADIO_SETTINGS_PATH[] = RADIO_PATH "/radio.bin";
 
 #if defined(PCBHORUS)
 #define BITMAPS_EXT         BMP_EXT JPG_EXT PNG_EXT
-#define BITMAPS_EXT_LEN     4
+#define LEN_BITMAPS_EXT     4
 #else
 #define BITMAPS_EXT         BMP_EXT
 #endif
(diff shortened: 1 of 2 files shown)


radio/src/sdcard.h
```diff
@@ -69,6 +69,7 @@ const char RADIO_SETTINGS_PATH[] = RADIO_PATH "/radio.bin";
 
 #if defined(PCBHORUS)
 #define BITMAPS_EXT         BMP_EXT JPG_EXT PNG_EXT
+#define BITMAPS_EXT_LEN     4
```
kilrah: Maybe LEN_BITMAPS_EXT for consistency with other similar defs

kilrah:
Interesting, was sure that was OK but obviously not!

tubeme:
So what is the problem with longer file names??

We are back in the era of MSDOS with this... Even there we had 8 letters...

PR #4435, opened 2017-02-10
Make build setup a little more Windows-friendly

mpaperno:
This adds friendlier warnings in case `grep` or `git` are missing and replaces using `rm` on Windows with `del`.  (git is not actually required, just makes the version stamp more specific, and grep is "only" required for building Lua-enabled firmware/simu).

I also propose including `dirent` and `msinttypes` headers in the source tree -- their licenses allow it, they're not likely to change at this point (and if they do the changes should be reviewed first anyway), it eliminates 2 extra setup steps, plus leaves less room for error/support requests.  

I'm not sure about also including pthreads, but maybe it's not a bad idea (eg. the latest official 2.9 version doesn't seem to work so we're relying on the old 2.8 version to still be available). License is GPL so shouldn't be a problem there.

Thanks,
-Max


cmake/Macros.cmake
```diff
@@ -25,17 +25,20 @@ macro(now RESULT)
...
+      execute_process(COMMAND "${GIT_EXECUTABLE}" "--git-dir=${PROJECT_SOURCE_DIR}/.git" "rev-parse" "--short=8" "HEAD" OUTPUT_VARIABLE ${RESULT})
+    endif()
     string(REGEX REPLACE "(.*)\n" "\\1" ${RESULT} ${${RESULT}})
-  else(WIN32)
-    message(SEND_ERROR "Git ID implemented")
+  else()
+    message(WARNING "Git executable not found, revision number not available.")
     set(${RESULT} 0)
-  endif(WIN32)
+  endif()
```
projectkk2glider: I like this change, I did a more cumbersome solution in https://github.com/opentx/opentx/commit/84ebeb01c34122c3959f2b55d3592c440d93a96d.

Could you add one more enhancement: If we build from a source that is not git repository (ie from downloaded zip archive), detect that and also use 0 for git_id. But first we have to test how this behaves in such case (no git files).

radio/src/lua/CMakeLists.txt
```diff
@@ -1,14 +1,23 @@
...
-      COMMAND ${CMAKE_C_COMPILER} -E ${ARGN} -DEXPORT ${RADIO_SRC_DIRECTORY}/dataconstants.h > lua_exports_${target}.tmp
-      COMMAND grep LEXP lua_exports_${target}.tmp > lua_exports_${target}.txt
-      COMMAND rm lua_exports_${target}.tmp
-      COMMAND ${PYTHON_EXECUTABLE} ${RADIO_DIRECTORY}/util/luaexport.py ${VERSION} lua_exports_${target}.txt lua_exports_${target}.inc lua_fields_${target}.txt
+  add_custom_command(
+    OUTPUT lua_exports_${target}.inc
+    COMMAND ${CMAKE_C_COMPILER} -E ${ARGN} -DEXPORT ${RADIO_SRC_DIRECTORY}/dataconstants.h > lua_exports_${target}.tmp
+    COMMAND ${GREP} LEXP lua_exports_${target}.tmp > lua_exports_${target}.txt
+    COMMAND ${RM} lua_exports_${target}.tmp
+    COMMAND ${PYTHON_EXECUTABLE} ${RADIO_DIRECTORY}/util/luaexport.py ${VERSION} lua_exports_${target}.txt lua_exports_${target}.inc lua_fields_${target}.txt
```
projectkk2glider: We could even remove the grep from the equation, the script `luaexport.py` would then have to do the filtering itself(should be very simple to implement).

cmake/Macros.cmake
mpaperno: Ah yes, good point, I'm sure git will throw some error but not clear what happens then.  We could check for `.git` folder in source root?  I'm not sure that's 100% reliable but should help.  I'll also see what happens when it just errors.

radio/src/lua/CMakeLists.txt
mpaperno: 👍  Have to admit I didn't even look very closely at why grep was there to begin with.

cmake/Macros.cmake
projectkk2glider: Git will return 128 on a dir without git data
```
git --git-dir=../ rev-parse --short=8 HEAD
fatal: Not a git repository: '../'
echo $?
128
```

you could test for this with code like this:
```
execute_process(COMMAND process param1 RESULT_VARIABLE res_var)
if(NOT "${res_var}" STREQUAL "0")
  # do something here about the failed "process" call...
  message(FATAL_ERROR "process failed res_var='${res_var}'")
endif()
```

mpaperno:
Nice, you just beat me to it on the python script (had to brew some fresh tea).  👍 
I think we can rid of the rm/del detection as well.  Woot!

Git next.  Also found some raw `python` calls hiding in the Cmake files.

bsongis:
Thanks!

PR #4440, opened 2017-02-11
Prevent Lua from allocating all heap memory (Horus)

### LINKED ISSUE #4284: Prevent Lua from allocating all heap memory

projectkk2glider:
Lua bitmap functions cleanup and docs update
Added x or y < 0 check to drawBitmap()

Closes #4284

### DIFF
--- a/radio/src/gui/480x272/bitmapbuffer.h
+++ b/radio/src/gui/480x272/bitmapbuffer.h
@@ -255,9 +255,9 @@ class BitmapBuffer: public BitmapBufferBase<uint16_t>
     template<class T>
     void drawBitmap(coord_t x, coord_t y, const T * bmp, coord_t srcx=0, coord_t srcy=0, coord_t w=0, coord_t h=0, float scale=0)
     {
-      if (!data || !bmp || x >= width || y >= height)
+      if (!data || !bmp || x < 0 || x >= width || y < 0 || y >= height)
         return;
-  
+
       coord_t srcw = bmp->getWidth();
       coord_t srch = bmp->getHeight();
 
--- a/radio/src/lua/api_lcd.cpp
+++ b/radio/src/lua/api_lcd.cpp
@@ -468,7 +468,7 @@ static int luaLcdDrawBitmap(lua_State *L)
     unsigned int y = luaL_checkunsigned(L, 3);
     unsigned int scale = luaL_optunsigned(L, 4, 0);
     if (scale) {
-      lcd->drawBitmap(x, y, b, 0, 0, 0, 0, scale/100.0);
+      lcd->drawBitmap(x, y, b, 0, 0, 0, 0, scale/100.0f);
     }
     else {
       lcd->drawBitmap(x, y, b);
--- a/radio/src/lua/api_lcd.cpp
+++ b/radio/src/lua/api_lcd.cpp
@@ -325,9 +325,18 @@ static int luaLcdDrawSource(lua_State *L)
 /*luadoc
 @function Bitmap.open(name)
 
-Loads a bitmap in memory, for later use with lcd.drawBitmap()
+Loads a bitmap in memory, for later use with lcd.drawBitmap(). Bitmaps should be loaded only
+once, returned object should be stored and used for drawing. If loading fails for whatever
+reason the resulting bitmap object will have width and height set to zero.
+
+Bitmap loading can fail if:
+ * File is not found or contains invalid image
+ * System is low on memory
+ * Combined memory usage of all Lua script bitmaps exceeds certain value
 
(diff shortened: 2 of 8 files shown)


radio/src/gui/480x272/bitmapbuffer.h
```diff
@@ -255,9 +255,9 @@ class BitmapBuffer: public BitmapBufferBase<uint16_t>
     template<class T>
     void drawBitmap(coord_t x, coord_t y, const T * bmp, coord_t srcx=0, coord_t srcy=0, coord_t w=0, coord_t h=0, float scale=0)
     {
-      if (!data || !bmp || x >= width || y >= height)
+      if (!data || !bmp || x < 0 || x >= width || y < 0 || y >= height)
```
projectkk2glider: `coord_t` is signed!

radio/src/lua/api_lcd.cpp
```diff
@@ -392,11 +411,18 @@ static int luaGetBitmapSize(lua_State * L)
 
 static int luaDestroyBitmap(lua_State * L)
 {
-  BitmapBuffer * ptr = checkBitmap(L, 1);
-  uint32_t size = ptr->getDataSize();
-  TRACE("luaDestroyBitmap: %p (%u)", ptr, size);
-  if (luaExtraMemoryUsage > size) luaExtraMemoryUsage -= size;
-  delete ptr;
+  BitmapBuffer * b = checkBitmap(L, 1);
+  if (b) {
```
projectkk2glider: `b` can be NULL

radio/src/lua/api_lcd.cpp
```diff
@@ -434,15 +461,19 @@ Displays a bitmap at (x,y)
...
-  unsigned scale = luaL_optunsigned(L, 4, 0);
+  const BitmapBuffer * b = checkBitmap(L, 1);
 
-  if (bitmap) {
-    lcd->drawBitmap(x, y, bitmap, 0, 0, 0, 0, (float) scale/100);
+  if (b) {
+    unsigned int x = luaL_checkunsigned(L, 2);
+    unsigned int y = luaL_checkunsigned(L, 3);
+    unsigned int scale = luaL_optunsigned(L, 4, 0);
+    if (scale) {
```
projectkk2glider: Small optimization if `scale == 0`, one less floating point division

radio/src/lua/interface.cpp
```diff
@@ -954,6 +954,27 @@ bool luaTask(event_t evt, uint8_t scriptType, bool allowLcdUsage)
...
 
+void checkLuaMemoryUsage()
+{
+#if (LUA_MEM_MAX > 0)
+  uint32_t totalMemUsed = luaGetMemUsed(lsScripts);
+#if defined(COLORLCD)
+  totalMemUsed += luaGetMemUsed(lsWidgets);
+  totalMemUsed += luaExtraMemoryUsage;
+#endif
+  if (totalMemUsed > LUA_MEM_MAX) {
```
projectkk2glider: Did not come up with smarter solution, so in this case both Lua states are disabled.

radio/src/lua/api_lcd.cpp
```diff
@@ -434,15 +461,19 @@ Displays a bitmap at (x,y)
...
+      lcd->drawBitmap(x, y, b, 0, 0, 0, 0, scale/100.0);
```
mpaperno: `100.0f` maybe?  Don't mean to pick on small stuff, PR looks great.

projectkk2glider: No, no, pick away. This is why we have these reviews. Good catch.

radio/src/lua/api_lcd.cpp
mpaperno: 👍   Assuming valid pointers seems to be a somewhat common issue in the code.

bsongis:
Thanks!

PR #4446, opened 2017-02-12
Bsongis/horus flash external device

bsongis:


bsongis:
Tested on X10 and X9D+

kilrah:
Fine on Horus too

bsongis:
It remains a problem with simulators
I commit directly to next

PR #4448, opened 2017-02-12
[X12S] No additional duration for the splash

bsongis:


bsongis:
And the prompt `hello.wav` is not missing its 1st second

kilrah:
Awesome 👍

PR #4452, opened 2017-02-14
Menu updates handling cat/models for Horus

### LINKED ISSUE #4349: n362 Companion: shortcut Alt-S malfunction
Hello,

shortcut "ALT-S" seems to have some problems even in n362 / Windows.
Directly after starting companion, open an eepe (or .otx) and highlighting a model, ALT-S is without functionality.

Tested with Win10 & Vista, Horus and X9e Profile.

After some editing and simulating by right mouseclick, ALT-S starts operating.
Unfortunately, couldn't find any systematic "root cause" which brings this shortcut into live.

btw
"RELOAD LUA" not suported in horus Sim but in X9e?

regards
Udo

3djc:
Fixes Alt-S not always activated when needed
Fixes edit menu for Horus (activated for models, not category)

This fixes #4349

### DIFF
--- a/companion/src/mdichild.cpp
+++ b/companion/src/mdichild.cpp
@@ -340,8 +340,7 @@ bool MdiChild::hasPasteData() const
 
 bool MdiChild::hasModelSelected()
 {
-  QModelIndex modelIndex  = ui->modelsList->currentIndex();
-  return isModel(modelIndex);
+  return isModel(ui->modelsList->currentIndex());
 }
 
 void MdiChild::updateTitle()
--- a/companion/src/mdichild.cpp
+++ b/companion/src/mdichild.cpp
@@ -764,11 +764,6 @@ void MdiChild::print(int model, const QString & filename)
   }
 }
 
-void MdiChild::viableModelSelected(bool viable)
-{
-  emit copyAvailable(viable);
-}
-
 int MdiChild::getCurrentModel() const
 {
   return modelsListModel->getModelIndex(ui->modelsList->currentIndex());
--- a/companion/src/mdichild.cpp
+++ b/companion/src/mdichild.cpp
@@ -56,9 +56,9 @@ MdiChild::MdiChild(MainWindow * parent):
   initModelsList();
   connect(parent, SIGNAL(FirmwareChanged()), this, SLOT(onFirmwareChanged()));
   connect(ui->modelsList, SIGNAL(doubleClicked(QModelIndex)), this, SLOT(openModelEditWindow()));
+  connect(ui->modelsList, SIGNAL(activated(QModelIndex)), this, SLOT(openModelEditWindow()));
   connect(ui->modelsList, SIGNAL(pressed(QModelIndex)), this, SLOT(onItemSelected(QModelIndex)));
   connect(ui->modelsList, SIGNAL(customContextMenuRequested(const QPoint &)), this, SLOT(showModelsListContextMenu(const QPoint &)));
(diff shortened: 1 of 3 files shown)


companion/src/mdichild.cpp
```diff
@@ -87,6 +89,22 @@ void MdiChild::refresh(bool expand)
   updateTitle();
 }
 
+void MdiChild::onItemSelected(QModelIndex idx)
+{
+  emit copyAvailable(isModel(idx) ? true : false);
```
mpaperno: Well if you're not going to use `viableModelSelected()` it can probably be removed or commented since it's not used otherwise.  The old usage was in modelslist.cpp and is now behind both an `#if 0` and also comments.  But I'd vote for using it.

companion/src/mdichild.cpp
```diff
@@ -87,6 +89,22 @@ void MdiChild::refresh(bool expand)
...
+}
+
+bool MdiChild::isModel(QModelIndex idx)
+{
+  return (modelsListModel->getModelIndex(idx) >= 0 ? true : false);
+}
+
+bool MdiChild::isCategory(QModelIndex idx)
+{
+  if (isModel(idx)) return false;
```
mpaperno: Line break please.  Always, please.  :)

companion/src/mdichild.cpp
```diff
@@ -87,6 +89,22 @@ void MdiChild::refresh(bool expand)
...
+  else return (modelsListModel->getCategoryIndex(idx) >= 0 ? true : false);
```
mpaperno: You could actually just `return (modelsListModel->getCategoryIndex(idx) >= 0);`  here if you'd like.

companion/src/mdichild.cpp
```diff
@@ -320,9 +338,9 @@ bool MdiChild::hasPasteData() const
   return mimeData->hasFormat("application/x-companion");
 }
 
-bool MdiChild::hasSelection() const
+bool MdiChild::hasModelSelected()
 {
-  return ui->modelsList->selectionModel()->hasSelection();
+  return isModel(ui->modelsList->currentIndex());
```
mpaperno: Hmm, it's undocumented what returns if the `currentIndex()` is invalid (that is, no current selection).  Can you verify?  It's probably -1 or something similarly harmless, but...

companion/src/mdichild.cpp
```diff
@@ -56,7 +56,9 @@ MdiChild::MdiChild(MainWindow * parent):
   initModelsList();
   connect(parent, SIGNAL(FirmwareChanged()), this, SLOT(onFirmwareChanged()));
   connect(ui->modelsList, SIGNAL(doubleClicked(QModelIndex)), this, SLOT(openModelEditWindow()));
+  connect(ui->modelsList, SIGNAL(pressed(QModelIndex)), this, SLOT(onItemSelected(QModelIndex)));
   connect(ui->modelsList, SIGNAL(customContextMenuRequested(const QPoint &)), this, SLOT(showModelsListContextMenu(const QPoint &)));
+  connect(ui->modelsList, SIGNAL(activated(QModelIndex)), this, SLOT(openModelEditWindow()));
```
mpaperno: This could move up, just below/above the `doubleClicked` connect.  Much easier to see that they connect to same slot that way.

mpaperno:
Looks good to me  👍

bsongis:
Looks good, thanks!

PR #4453, opened 2017-02-14
[firmware][gui] Extended limits fixes, issue #4447

### LINKED ISSUE #4447: Failsafe model settings don't allow full range with extended limits.
As mentioned in #4443, in model_setup.cpp we have `g_model.extendedLimits ? 640*2 : 512*2;` whereas I believe `g_model.extendedLimits ? 768*2 : 512*2` is correct.  This is the case for all 3 GUI sizes.  Please correct me if I'm wrong.  Fixes in progress.

This seems to be going back to at least 2.0... extended limits used to be 125%, right?  And still are for AVR?

mpaperno:
* Adjust custom failsafe (and some channel display) ranges for proper extended limits (fixes #4447).  
 * Fix some display issues in x64 GUIs when adjusting failsafe settings, and cleans up code in those functions.
 * Fix 480x272 channel output progress bars to show full extended limits.
 * Fix alignment issues on 212x & 480x gui mixes screens with long custom channel names.
 * Streamline some channel outputs display code.

### DIFF
--- a/radio/src/gui/128x64/model_setup.cpp
+++ b/radio/src/gui/128x64/model_setup.cpp
@@ -232,8 +232,8 @@ void menuModelSetup(event_t event)
     for (int j=0; j<=k; j++) {
       if (mstate_tab[j+HEADER_LINE] == HIDDEN_ROW) {
         if (++k >= (int)DIM(mstate_tab)) {
-    	  return;
-    	}
+          return;
+        }
       }
     }
 #endif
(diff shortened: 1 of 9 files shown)


bsongis:
Seems all good, thanks!

mpaperno:
Thanks Bertrand.

@ All, is there any reason to keep the MIXERS_MONITOR macro/option?  The only remaining difference is whether the ENT key does anything on the channels view screen or not (and only on 212x64).   Just seems like a needless complication and yet another fw. choice for users to understand/navigate.

kilrah:
I agree this could be always enabled. Was again one of those "some people will be confused by it" things.

PR #4458, opened 2017-02-14
Projectkk2glider/lua float

projectkk2glider:
This PR has mixed in other double to float conversion. The main purpose of this is review and testing. 

The actual full conversion from double to float still needs some consideration, one obvious thing is GPS coordinates for which float will perhaps not be precise enough.

### DIFF
--- a/radio/src/thirdparty/Lua/src/lrotable.c
+++ b/radio/src/thirdparty/Lua/src/lrotable.c
@@ -34,6 +34,25 @@ static luaR_result luaR_findkey(const void * where, const char * key, int type, 
   return 0;
 }
 
+
+char luaR_cachedName[LUA_MAX_ROTABLE_NAME + 1] = {0};
+TValue luaR_cachedValue;
+
+static inline void luaR_saveLastFoundGlobal(const char * name, const TValue * val)
+{
+  strcpy(luaR_cachedName, name);
+  luaR_cachedValue = *val;
+}
+
+static inline luaR_result luaR_checkLastFoundGlobal(const char * name, TValue * val)
+{
+  if (!strcmp(luaR_cachedName, name)) {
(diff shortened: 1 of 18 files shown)


radio/src/thirdparty/Lua/src/lrotable.c
```diff
@@ -10,77 +10,67 @@
...
-  if (strlen(name) > LUA_MAX_ROTABLE_NAME)
+  if (strlen(name) > LUA_MAX_ROTABLE_NAME) {
+    TRACE_LUA_INTERNALS("luaR_findglobal() '%s' = NAME TOO LONG", name);
     return 0;
+  }
   for (i=0; lua_rotable[i].name; i++) {
     if (!strcmp(lua_rotable[i].name, name)) {
-      *ptype = LUA_TROTABLE;
-      return i+1;
+      setrvalue(val, (void *)(size_t)(i+1));
```
projectkk2glider: This still uses ugly cast, but does not involve double of float. Not sure if we can improve it.

radio/src/thirdparty/Lua/src/lrotable.c
```diff
@@ -60,17 +61,17 @@ luaR_result luaR_findglobal(const char * name, TValue * val) {
 
 
 luaR_result luaR_findentry(void *data, const char * key, TValue * val) {
-  unsigned idx = (unsigned)(size_t)data - 1;
+  luaR_table * table = (luaR_table *)data;
```
projectkk2glider: I think this is much nicer.

mpaperno: Definitely!

projectkk2glider:
Replaced by #4502

PR #4459, opened 2017-02-14
[MULTI] Implement option Value adjustment in Companion

schwabe:
Also refactor ui code to use the same multi table as the radio to make updates easier and consistent


mpaperno:
Personally I'm not a fan of `auto` type, I think it's much clearer to use the actual type, even if "just" for code readability or IDE assistance.  

I'd also vote for using QPair and QVector (or QStringList?) types instead of the std:: versions since this is all Qt code (eg. QString is already used) and it avoids a conversion layer.

companion/src/CMakeLists.txt
```diff
@@ -166,6 +167,7 @@ set(companion_SRCS
   radionotfound.cpp
   wizarddata.cpp
   wizarddialog.cpp
+  multi.cpp
```
mpaperno: 2nd time?

companion/src/multi.h
```diff
@@ -0,0 +1,40 @@
...
+ * GNU General Public License for more details.
+ */
+
+#ifndef OPENTX_MULTI_H
+#define OPENTX_MULTI_H
+
+#include <vector>
+#include <QString>
+
+struct mm_protocol_definition {
```
mpaperno: Couldn't/shouldn't this have all the access functions and other const/static declarations as members, instead of making new globals?  Is this C or C++?

schwabe: Good point.

companion/src/CMakeLists.txt
schwabe: Fixed

mpaperno:
Yes, seems much better encapsulated now.  I'm not sure if some of the other static strings/lists or maybe struct definitions could be class/struct members also.  But anyway this seems like an improvement.  👍

schwabe:
The string list is already only a initilisation list. Not sure how to encapsulate that even more :)
The strings themselves are just static strings inside the cpp files so they will only be available inside that class. 

I am open for other better solutions.

schwabe:
(I rebased the commit to next since there were conflicts)

mpaperno:
My comments below are purely semantics, since you asked... :)  I don't mean to block a merge or be critical.  I also have a "narrow view" since I'm just looking at the diffs in this PR.

Could the two structs `MultiProtocolDefinition` and `radio_mm_definition` be part of the `Multiprotocols` class?  Both objects (`MultiProtocolDefinition` & `Multiprotocols` ) should still be able to use `radio_mm_definition` in this case.

Global/extern `multiProtocols` variable could perhaps be a static member of `Multiprotocols`, returned from a static function (eg. `static const Multiprotocols & getProtocols();` used as `Multiprotocols::getProtocols();`)  `Multiprotocols::protocols` variable would also need to be static in this case.

A bit curious also what kind of errors you were getting with QVector.  But not important.

You're probably right about the const static strings and string lists... You could qualify them further or put them inside a namespace but it would really just be more verbose with no actual benefit.

I might rename the files `multi` to `multiprotocols` to fully reflect the class name and avoid possible ambiguity, esp. since it's in the src root.

-Max

schwabe:
```
/usr/local/opt/qt@5.7/lib/QtCore.framework/Headers/qvector.h:319:26: error: no
      matching constructor for initialization of 'MultiProtocolDefinition'
            new (from++) T();
```
QVector seems to require to instantiate its members with an empty constructor.

bsongis:
Thanks!

PR #4463, opened 2017-02-15
Bsongis/9x fixes

bsongis:


radio/src/CMakeLists.txt
```diff
@@ -492,12 +492,12 @@ if(NOT MSVC)
...
-    file(WRITE ${ALLSRC} "${ALLSRC_CONTENT}")
+    add_custom_command(
+      OUTPUT ${ALLSRC}
+      COMMAND sh -c 'rm -f ${ALLSRC}\; for f in ${SRC}\; do echo "\\# 1 \"$$f\"" >> allsrc.cpp\; cat "$$f" >> ${ALLSRC}\; done'
+      WORKING_DIRECTORY ${RADIO_SRC_DIRECTORY}
+      DEPENDS ${SRC}
+    )
 
     add_executable(firmware ${ALLSRC})
     add_dependencies(firmware ${FIRMWARE_DEPENDENCIES})
```
mpaperno: Can't we just add ${SRC} here?

radio/src/CMakeLists.txt
```diff
@@ -492,12 +492,12 @@ if(NOT MSVC)
...
     set(ALLSRC ${PROJECT_BINARY_DIR}/allsrc.cpp)
-    foreach(FILE ${SRC})
-      file(READ ${FILE} FILE_CONTENT)
-      set(ALLSRC_CONTENT "${ALLSRC_CONTENT}# 1 \"${FILE}\"\n${FILE_CONTENT}")
-    endforeach()
-
-    file(WRITE ${ALLSRC} "${ALLSRC_CONTENT}")
+    add_custom_command(
+      OUTPUT ${ALLSRC}
+      COMMAND sh -c 'rm -f ${ALLSRC}\; for f in ${SRC}\; do echo "\\# 1 \"$$f\"" >> allsrc.cpp\; cat "$$f" >> ${ALLSRC}\; done'
```
mpaperno: We just got rid of the last `sh` call so that it all works on Windows better.

radio/src/CMakeLists.txt
mpaperno: Hmm, no, that doesn't work....   :-/

bsongis: It would take far more flash!

mpaperno:
Simulator now starts properly from within Companion if `libsimulator` is built with `GVARS + (FRSKY || FRSKY_SPORT || TELEMETREZ)`

This combination (variant=3) seems to be hard-coded in Companion, not affected by user-selected options in radio profile.

mpaperno:
There's still an overflow here, I'm not sure what the actual min/max values should be.  The vars are INT8.

https://github.com/opentx/opentx/blob/bsongis/9x_fixes/radio/src/gui/128x64/model_logical_switches.cpp#L415

PR #4474, opened 2017-02-17
Enable print (even for model idx 0)

### LINKED ISSUE #4345: Companion 2.2 v362 Print function does not work
Print option under File pull down menu is greyed out.  Right click on individual models brings up the document (except the first model in first category where this does not work) showing all settings.  But there is no way to print the document within Companion (info can be selected then copied to to another program for printing).

3djc:
You only get there if a valid model is selected

This also prevent blanking model name.

This fixes  #4345

### DIFF
--- a/companion/src/mdichild.cpp
+++ b/companion/src/mdichild.cpp
@@ -91,23 +91,12 @@ void MdiChild::refresh(bool expand)
 
 void MdiChild::onItemSelected(QModelIndex idx)
 {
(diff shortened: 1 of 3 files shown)


projectkk2glider:
This now allows printing of empty models for Taranis.
screenshot

3djc:
Apparently this is OK,  if you want to print empty,  well you can (but please think about the trees before printing on paper)

PR #4483, opened 2017-02-19
Re #4477: All usage (except Lua) of double converted to float. printf…

projectkk2glider:
…("%f") still uses promotion to double, (floating point va_args are always promoted to double)

Without this the hardware FPU does not get much usage.


bsongis:
Thanks!

Issue #2645, opened 2015-08-06
Companion sometimes import the install directory from 2.0 settings


rguptaca:
I uninstalled both 2.0 and 2.1 using the uninstall program provided.
Downloaded 2.1.1N407
It wanted to create/install into 2.0
See pic
![21 into 20](https://cloud.githubusercontent.com/assets/9390066/9103000/3303ac10-3bc7-11e5-9489-8ecf6e63567a.GIF)

projectkk2glider:
I saw the same issue. It seem that installer program somehow (I bet it is in the registry) remembers the last installation path for Companion and offers it in this dialog despite the fact that default values in cmake are configured differently.

kilrah:
Yes and it's desired so that if the user chooses another directory it's remembered on subsequent updates. The problem is when settings are imported from a previous version, this particular key should be skipped.

projectkk2glider:
When installing on a fresh Windows machine, first installed 2.0.17 and then 2.1.1 nightly `#407` I don't have this problem. It seems that some other version of Companion messed up that setting in the registry. 

@rguptaca what do you have in the registry key `HKEY_CURRENT_USER\Software\OpenTX\Companion 2.1`?

kilrah:
It might only happen if the user did not go through 2.0.17 and updated straight from <2.0.17 to 2.1.

2.0.17 moved the settings to the "2.0" key to allow for concurrent versions, before that it wasn't expected to have multiple versions.

projectkk2glider:
@kilrah I understand that, but I can't see where the 2.1 is picking up the directory of 2.0. It is looking in completely different registry key which should not even exist!?

kilrah:
Will have to look again in detail

rguptaca:
Attached are the Registry entries for Comopanion 2.1 

Key Name:          HKEY_CURRENT_USER\Software\OpenTX\Companion 2.1
Class Name:        <NO CLASS>
Last Write Time:   2015-08-06 - 1:31 AM
Value 0
  Name:            <NO NAME>
  Type:            REG_SZ
  Data:            C:\Program Files (x86)\OpenTX\Companion 2.1

Value 1
  Name:            Start Menu Folder
  Type:            REG_SZ
  Data:            OpenTX Companion 2.1

Value 2
  Name:            arm_mcu
  Type:            REG_SZ
  Data:            at91sam3s4-9x

Value 3
  Name:            avr_arguments
  Type:            REG_SZ
  Data:            

[...]

PR #4485, opened 2017-02-20
[Companion] Smarter app settings import from old versions...

### LINKED ISSUE #4330: Installing Companion 2.2 side by side with 2.1 retains the DFU-Util location
Installing Companion 2.2 side by side with 2.1 retains the 2.1 DFU-Util location in
Settings->Configure Communications...

Presumably it should always update to point to the new install. Like this...?
C:/Program Files (x86)/OpenTX/Companion 2.2/dfu-util.exe

Note: I have Companion 2.0, 2.1, 9X, and 2.2 RC9 installed side by side which I installed in that order over time.  Possibly it was wrong when I installed Companion 2.1 and I fixed the path up.  The current problem may be that the installer is seeing it as user modified and bringing it forward.  There was a similar bug in 2.1 where it retained the 2.0 install path when installed side by side which I worked around by resetting various paths manually.   I'm using Windows 7 Pro SP1

The only reason I can see to retain the existing path is if it is not pointing to the the prior install's location and even then it may be better to reset it to the new install.  I'm not clear why it is user settable so I could be misunderstanding why it should be retained.

Probably not an important bug but I have come across various people using 2.1 that have the DFU-Util location be not set to or possibly set incorrectly which seemed to cause problems for them and this may be the reason.

mpaperno:
Do not import flash tool locations on Windows which are already included with Companion (closes #4330).

Tested, even with companion9x settings which I had in my registry.

The old code had this (what I thought was) strange condition/exclusion:  
`settings21.value(*i) != QString("Start Menu Folder")`  
This sounded more like a setting name (not value) but I couldn't find any instance of either throughout all my old and new Companion settings (5 versions!).  I removed this condition, but maybe someone knows why it was there to begin with?

### DIFF
--- a/companion/src/storage/appdata.cpp
+++ b/companion/src/storage/appdata.cpp
@@ -547,133 +547,17 @@ void AppData::init()
 {
     qRegisterMetaTypeStreamOperators<SimulatorOptions>("SimulatorOptions");
 
-    //Initialize the profiles
+    QSettings settings(COMPANY, PRODUCT);
+    if (!settings.contains("settings_version"))
+      importSettings(settings);
+
+    // Initialize the profiles
     for (int i=0; i<MAX_PROFILES; i++)
         profile[i].init( i );
 
-    //Initialize the joysticks
+    // Initialize the joysticks
     for (int i=0; i<MAX_JOYSTICKS; i++)
         joystick[i].init( i );
-
-    QSettings settings(COMPANY, PRODUCT);
-
-    // Copy existing 2.1 settings if present
-    if (profile[0].name().isEmpty())
-    {
-        QSettings settings21("OpenTX", "Companion 2.1");
-
-        QStringList keys = settings21.allKeys();
-        for (QStringList::iterator i=keys.begin(); i!=keys.end(); i++)
-        {
-            if (settings21.value(*i) != QString("") && settings21.value(*i) != QString("Start Menu Folder"))
-            {
-                settings.setValue(*i, settings21.value(*i));
-            }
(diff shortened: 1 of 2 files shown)


projectkk2glider:
See https://github.com/opentx/opentx/commit/08756af9f597d8cd4f26a61544134b049c00cb66 for your question.

mpaperno:
I see... so "Start Menu Folder" was in fact the key name, not value, and the other "blank value" check was supposed to prevent the "(Default)" key (with install path) from being copied.  Did it work?  :-)   I'm not sure because my "Companion 2.2" default key points to the 2.1 install path.  The only place I have a Start Menu Folder is in the OpenTX root key, but I probably chose not to create shortcuts on most installs.

projectkk2glider:
Then the issue #2645 remains. Perhaps @kilrah can chime in. I prefer to stay away from Windows stuff.

kilrah:
Looks good, I'll check when I get a moment

mpaperno:
My earlier "Did it work?" question was rhetorical.  There is no way it could have worked, for a number of reasons.  I've tested this new code with a variety of situations and old Companion versions.  It is fundamentally same as old version, just less code duplication (plus the fixed issues).

The "better" way to do this import would be to have a list of all setting key names currently being used and only import those (or migrate values from old key names as needed).  But with the current architecture this would require yet a 3rd place where the key names are defined.  I'd rather refactor the system so the key names and default values are only specified once, but this would take more time and testing, and probably not a high priority right now.

PR #4486, opened 2017-02-20
Fix to many failsafe channels editable when less then "1 column (8)" …

RealTadango:
…of channels is used


mpaperno:
OK I see the issue on Horus screen with extra channel showing up in the right column.  But not on 212x64.  Can you post a screenshot?  Thanks.

RealTadango:
I only checked the code for all guis and it looked like the same issue (also a double loop). I have not tested it on other devices.

3djc:
Same here, cannot reproduce on x9. A LOT of wo work as gone into making the simu available and working, please make sure at least some level of testing is done, even if you do not own the physical radio

mpaperno:
On the smaller screen it properly set the number of columns already.  Fixed, thanks for reporting.

PR #4487, opened 2017-02-20
Fix lcdLastPos

### LINKED ISSUE #4310: 2.2RC9 Taranis lcd.getLastPos() issue

3djc:
Introduce lcdLastLeftPos to native code
Native lcdLastPos renamed to lcdLastRightPos

Introduce to lua:
lcd.getLastRightPos() as an alias to lcd.getLastPos()
lcd.getLastLeftPos()

Please note that on 128x64 ARM, it did require moving the code to X9/Horus style. Similar to what did happen on those platform, I had to do minor adjustment to test screens, but I do not anticipate any "omg my screen as changed !" (beside drawnumber alignement, see bellow)

The dot on MIDSIZE mentioned in #4310 as already been fixed in another PR.

Changes to alignment of lua drawnumber are NOT in the scope of this PR

This fixes #4310


radio/src/gui/128x64/lcd.cpp
```diff
@@ -534,8 +559,7 @@ void lcdDrawNumber(coord_t x, coord_t y, lcdint_t val, LcdFlags flags, uint8_t l
   lcdLastPos = x;
   x -= fw;
   if (dblsize) x++;
-
-  for (uint8_t i=1; i<=len; i++) {
+    for (uint8_t i=1; i<=len; i++) {
```
mpaperno: Indent level.  Good example of why at least a linebreak after any `if` is really useful.

radio/src/gui/128x64/lcd.cpp
```diff
@@ -29,6 +29,7 @@ void lcdClear()
 
 coord_t lcdLastPos;
 coord_t lcdNextPos;
+coord_t lcdLeftPos;
```
bsongis: also on !ARM platforms?

radio/src/gui/128x64/lcd.cpp
```diff
@@ -458,24 +464,43 @@ void lcdDrawNumber(coord_t x, coord_t y, lcdint_t val, LcdFlags flags)
...
-#if defined(CPUARM)
-  uint32_t fontsize = FONTSIZE(flags);
-  bool dblsize = (fontsize == DBLSIZE);
-  bool xxlsize = (fontsize == XXLSIZE);
-  bool midsize = (fontsize == MIDSIZE);
-  bool smlsize = (fontsize == SMLSIZE);
-  bool tinsize = (fontsize == TINSIZE);
-#else
   bool dblsize = flags & DBLSIZE;
   #define xxlsize 0
```
bsongis: If we have dedicated code, all these `#define` have no more sense!

radio/src/gui/128x64/lcd.cpp
```diff
@@ -588,29 +620,18 @@ void lcdDrawNumber(coord_t x, coord_t y, lcdint_t val, LcdFlags flags, uint8_t l
...
-      if ((flags&INVERS) && ((~flags & BLINK) || BLINK_ON_PHASE)) {
-        lcdDrawSolidVerticalLine(xn, y, 12);
-        lcdDrawSolidVerticalLine(xn+1, y, 12);
-      }
-      lcdDrawSolidHorizontalLine(xn, y+9, 2);
-      lcdDrawSolidHorizontalLine(xn, y+10, 2);
-    }
-    else {
-      // TODO needed on CPUAVR? y &= ~0x07;
-      lcdDrawSolidFilledRect(xn, y+2*FH-3, ln, 2);
```
bsongis: Are you sure this `lcdDrawSolidFilledRect` not needed anymore for PREC1 / PREC2 with DBLSIZE fonts?

radio/src/gui/128x64/lcd.cpp
3djc: Hum yes, but that goes far beyond just those define ! Major cuts incoming :)

bsongis:
The tweaks for saving 2 pixels on 128x64 screens are not anymore applied now, I hope there is no problem on telemetry screens (ARM 128x64)

Unless ... we could use the MIDSIZE font on those screens as we do on 212x64? If it's a good solution, it could even be done on AVR platforms and we would save a nice bunch of flash

3djc:
On X7:
image

But maybe you where not concerned with X7 but non x7 ARM ?

bsongis:
What do other think, I find MIDSIZE nice. About font size, we use this one on X9D, it should be good on X7 as well

3djc:
I like midsize better too

3djc:
image

3djc:
image

mpaperno:
👍  midsize looks much nicer.  

Unrelated, but the dblsize colon is kinda massive... couldn't that be same size as the period/dot?

Question, also unrelated to this PR: I see a lot of repeated code which calculates pixel widths/heights depending on font size (also often needed in Lua scripts).  There are `FH/FW/FWNUM` constants, but those are only for the "normal" size.  Would it make sense to add more constants for the different named font sizes?  Or is there some reason that wouldn't work?

bsongis:
@3djc I like it a lot

3djc:
It is screenshot with this PR in it's current status ;)

radio/src/gui/128x64/lcd.cpp
```diff
@@ -29,8 +29,9 @@ void lcdClear()
 
 coord_t lcdLastPos;
```
bsongis: I would propose to rename it to `lcdLastRightPos`

radio/src/gui/128x64/lcd.cpp
```diff
@@ -29,8 +29,9 @@ void lcdClear()
...
 coord_t lcdNextPos;
-
 #if defined(CPUARM)
+coord_t lcdLeftPos;
```
bsongis: ... and `lcdLastLeftPos`

radio/src/lua/api_lcd.cpp
```diff
@@ -128,20 +128,55 @@ static int luaLcdDrawLine(lua_State *L)
...
 
 @notice Only available on Taranis
 
+@notice For added clarity, it is recommended to use lcd.getRightPos()
+
 @status current Introduced in 2.0.0
 */
+
+/*luadoc
+@function lcd.getRightPos()
```
bsongis: `getLastRightPos()` ?

3djc:
I have pushed the proposed renaming, in case we want to go ahead with it as proposed

bsongis:
Thanks!

PR #4475, opened 2017-02-17
[Companion][Simulator] Add custom qDebug message handler/formatter

mpaperno:
This uses a Qt hook to install a custom handler for qDebug() (and family) output.  

It also allows other classes to be notified of debug/error events/messages, for example to show a message to the user.  

In the future this would also make it easier to add category-based filtering  (sort of like for the various TRACE_* messages now).  qCDebug, qCInfo, etc, all support categories already.

On Linux (at least) there are a few lines of backtrace included in the output in case of a hard fault (`qFatal()`).

Default output now looks like this:
```
[D] registerStorageFactory():54 - register storage "otx"
[D] registerStorageFactory():54 - register storage "sdcard"
[D] registerSimulatorFactory():37 - registering "opentx-9x" simulator
[D] registerSimulatorFactory():37 - registering "opentx-9xrpro" simulator
....
[I] LcdWidget::makeScreenshot():99 - Screenshot saved to: "D:/Devel/RC-Tx/OpenTx/build/2.2-gui/screenshot_x7_17-02-17_05-50-44.png"
[D] OtxFormat::write():54 - Saving to archive "C:/Users/god-like/Documents/OpenTX/test-x7.otx"
[D] OtxFormat::write():68 - Archive size 578
[I] SimulatorWidget::saveTempData():384 - Saved radio data to file "C:/Users/max/Documents/OpenTX/test-x7.otx"
[D] finish():276 - SIMULATOR EXIT 0
```

Or with source paths enabled (from OSX test) :
```
[D] firmwares/opentx/opentxinterface.cpp::OpenTxEepromInterface::loadFromByteArray():198 - "OpenTX for FrSky Horus: OK"
[D] storage/categorized.cpp::CategorizedStorageFormat::load():97 - currModelIndex = 0
[D] storage/otx.cpp::OtxFormat::write():54 - Saving to archive "/Users/max/Documents/OpenTX/eeprom-horus.otx"
[D] storage/otx.cpp::OtxFormat::write():68 - Archive size 670
[I] simulation/simulatorwidget.cpp::SimulatorWidget::saveTempData():384 - Saved radio data to file "/Users/max/Documents/OpenTX/eeprom-horus.otx"
[D] simulation/simulatorwidget.cpp::SimulatorWidget::deleteTempData():401 - Deleting temporary settings directory "/var/folders/62/304qwlnx469fdw2_4njpb63m0000gn/T/otx-u6oHFS"
[...]

PR #4488, opened 2017-02-20
[Simulator] Add asynchronous FIFO buffer for handling debug output/display more efficiently.

mpaperno:
FIFO device runs in own thread(s) w/out blocking GUI thread.

Merging in #4484 makes the whole thing more efficient due to half the number of callbacks required to handle the output (currently each CRLF is sent in a separate event from main text body).

There will probably be a minor conflict with #4475 so depending on which is merged first (or at all), I'll fix the other.


bsongis:
Seems cool :)

PR #4492, opened 2017-02-21
[Companion] Fix/improve setting radio module custom failsafe values.

### LINKED ISSUE #4443: Companion Model Edit Setup Failsafe Positions
Most channel failsafe position values rounded down 0.1 when saved after changing.
Open and close edit dialog without changing a failsafe value preserves the values as read.
Only rounds the channel being changed. Changing from a starting value and then back to the same value triggers the rounding issue,

eg 20.0 saved as 19.9 and 20.1 saved as 20.0
but 10.0 saved as 10.0 and 10.1 saved as 10.1

Taranis Plus model with Internal D16 Protocol
Companion 2.1.9
Tested on Win10 and Ubuntu 14.04

mpaperno:
* Properly set limits based on model's Extended Limits setting;
  * Fix rounding issues between percentile display and underlying value in μs;
  * Add ability to display values in μs or percent.
  * Channels numbering & count is responsive to module's defined start/end channels.

Fixes #4443

Please give this a good test first... I'm a bit rushed on final testing, and may be not be available for a few days for quick fixes.  I did give it a good workout, but you never know.  Thanks!

### DIFF
--- a/companion/src/modeledit/setup.cpp
+++ b/companion/src/modeledit/setup.cpp
@@ -148,6 +148,18 @@ void TimerPanel::on_name_editingFinished()
 #define FAILSAFE_CHANNEL_HOLD    2000
 #define FAILSAFE_CHANNEL_NOPULSE 2001
 
+#define MASK_PROTOCOL       1
+#define MASK_CHANNELS_COUNT 2
+#define MASK_RX_NUMBER      4
+#define MASK_CHANNELS_RANGE 8
+#define MASK_PPM_FIELDS     16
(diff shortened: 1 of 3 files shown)


mpaperno:
I should add that the μs display does not take the center value into account, it's purely +/- from center, exactly as stored on the actual radio.  It could probably show actual pulse width, if there's a way to get the correct center value per channel.  I ran out of time to look for that though.  Could be added later of course.

bsongis:
Looks good, but I didn't test it as well

PR #4494, opened 2017-02-22
Fixes #3305 Companion model edit add clear, set and invert all flight modes context menu

### LINKED ISSUE #3305: [Enhancement, Usage][Companion ,OpenTx] Inputs and Mixer, add Clear/Set all Flightmodes
I use Flightmodes in all my models 
And when i make new input or mixer, all flightmodes are selected as default and that is good.
But often i only need an input or mix to be active in 1-3 flightmodes
In Companion it would bee nice to have a Clear/Set-all flightmode Button,
then if i want to have a mix active in FM1 only, I click  the "Clear flightmode Button" and
click FM1 checkbox Done. 
Look at pictures below
companion-input-mixer-clear fm

And on Radio when input or mixer flightmode is selected, Long press enter could clear/Set-all flightmodes.
Then i select the FMx i want as usual.
radio-input-mixer-clear fm

If this is a useless idea just close this issue

elecpower:
Companion part of #3305. Any votes for adding to radios or close the issue with this PR?

Apologies for the messy diff caused by the indent changes.

Suggestions for shortening the fm checkbox pointer array initialisation code welcome. New at c++ and tried a number of syntax variants based on web searches but kept hitting compiler errors.

Tested on Ubuntu 16.04

### DIFF
--- a/companion/src/modeledit/expodialog.cpp
+++ b/companion/src/modeledit/expodialog.cpp
@@ -22,7 +22,7 @@
 #include "ui_expodialog.h"
 #include "helpers.h"
 
-ExpoDialog::ExpoDialog(QWidget *parent, ModelData & model, ExpoData *expoData, GeneralSettings & generalSettings, 
+ExpoDialog::ExpoDialog(QWidget *parent, ModelData & model, ExpoData *expoData, GeneralSettings & generalSettings,
                           Firmware * firmware, QString & inputName) :
   QDialog(parent),
   ui(new Ui::ExpoDialog),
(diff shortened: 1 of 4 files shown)


bsongis:
There is a way, using the `findChildren` function:
http://doc.qt.io/qt-5/qobject.html#findChildren
There is one example in modeledit/setup.cpp

mpaperno:
I'd use `QList`s instead of fixed-size arrays, eg. 
```cpp
QList<QLabel *> fpLabels;

fpLabels = findChildren<QLabel *>(QRegExp("lb_FP\\d"));

foreach (QLabel * label, fpLabels) {
  label->doStuff();
}
```
Besides ease of use, this also "guarantees" that you have the correct object type in the list and that it actually exists (or at least that it did at init time 😉 ).

If keeping an index value tied to the object is important (though I don't think it is in this case), then a QMap, e.g.  `QMap<int, QLabel *>` can be one good way to handle it.  There is also `QVector` type which acts more like an array than QList and is sometimes faster, but in the above case QList is convenient since findChildren() returns a QList already.

elecpower:
Thanks for the coding suggestions @bsongis and @mpaperno. I appreciate your time for a new coder.
After considering use of findChildren with RegExp whilst it would work for the labels there is, I would expect, no guarantee it would return the checkboxes in perfect sequence 0-8. Should that occur there would be a mismatch to the flight modes field bit pattern.
Use of QMap may in this case not shorten the code. 
So probably best to leave well alone for the time being but will keep your use case suggestions for consideration in any future PRs.

PR #4502, opened 2017-02-24
Lua: All usage of double converted to float

projectkk2glider:
This is the same as #4458 just squashed and rebased on latest next.


bsongis:
All good, thanks!

PR #4503, opened 2017-02-24
LUA getCoordDistance()

3djc:
- introduce a new LUA function to compute distance in meters between two coord
- on F4 targets, use float computation (allow earth wide distance computation)
- move this computation to math.cpp (for LUA and for calculated sensor dist)

### DIFF
--- a/radio/src/maths.cpp
+++ b/radio/src/maths.cpp
@@ -183,9 +183,9 @@ uint32_t getCoordDistance(float  y1, float x1, float y2, float x2){
   y1 *= rad;
   y2 *= rad;
 
-  float nA =   pow( sin(nDLat/2), 2 ) + cos(y1) * cos(y2) * pow( sin(nDLon/2), 2 );
-  float nC = 2 * atan2( sqrt(nA), sqrt( 1 - nA ));
-  return round(nRadius * nC);
+  float nA =   powf( sinf(nDLat/2), 2 ) + cosf(y1) * cosf(y2) * powf( sin(nDLon/2), 2 );
+  float nC = 2 * atan2f( sqrtf(nA), sqrtf( 1 - nA ));
+  return roundf(nRadius * nC);
 }
 #endif // STM32F4
 #if defined(CPUARM)
--- a/radio/src/maths.cpp
+++ b/radio/src/maths.cpp
@@ -2,7 +2,7 @@
  * Copyright (C) OpenTX
  *
  * Based on code named
- *   th9x - http://code.google.com/p/th9x 
+ *   th9x - http://code.google.com/p/th9x
  *   er9x - http://code.google.com/p/er9x
  *   gruvin9x - http://code.google.com/p/gruvin9x
  *
@@ -19,6 +19,7 @@
  */
 
 #include "opentx.h"
+extern uint32_t getDistFromEarthAxis(int32_t);
 
 #if !defined(CPUARM)
 // #define CORRECT_NEGATIVE_SHIFTS
@@ -120,12 +121,12 @@ uint16_t isqrt32(uint32_t n)
 /*
   Division by 10 and rounding or fixed point arithmetic values
 
-  Examples: 
+  Examples:
     value -> result
     105 ->  11
     104 ->  10
    -205 -> -21
-   -204 -> -20 
+   -204 -> -20
 */
 
 #if defined(FRSKY_HUB) && !defined(CPUARM)
@@ -170,5 +171,31 @@ void getGpsDistance()
   telemetryData.hub.gpsDistance = isqrt32(result);
   if (telemetryData.hub.gpsDistance > telemetryData.hub.maxGpsDistance)
     telemetryData.hub.maxGpsDistance = telemetryData.hub.gpsDistance;
+}
+#endif
+
+#if defined(STM32F4)
+uint32_t getCoordDistance(float  y1, float x1, float y2, float x2){
+  float nRadius = 6370997.0; // Earth’s radius in meters
+  float rad = 0.0174532925199433;
+  float nDLat = (y2 - y1) * rad;
+  float nDLon = (x2 - x1) * rad;
+  y1 *= rad;
+  y2 *= rad;
+
+  float nA =   pow( sin(nDLat/2), 2 ) + cos(y1) * cos(y2) * pow( sin(nDLon/2), 2 );
+  float nC = 2 * atan2( sqrt(nA), sqrt( 1 - nA ));
+  return round(nRadius * nC);
+}
+#elif defined(CPUARM)
+uint32_t getCoordDistance(int32_t  y1, int32_t x1, int32_t y2, int32_t x2){
+  uint32_t angle = abs(y1 - y2);
+  uint32_t dist = EARTH_RADIUS * angle / 1000000;
+  uint32_t result = dist*dist;
+
+  angle = abs(x1 - x2);
+  dist = getDistFromEarthAxis(y2) * angle / 1000000;
+  result += dist*dist;
+  return (isqrt32(result));
 }
 #endif
--- a/radio/src/maths.cpp
+++ b/radio/src/maths.cpp
@@ -181,8 +181,8 @@ inline float degToRad(float x)
     return x * 0.0174532925f; // Pi / 180
 }
 
-uint32_t getCoordDistance(int32_t  y1, int32_t x1, int32_t y2, int32_t x2) {
-  return(getCoordDistance( (float) y1/1000000, (float) x1/1000000, (float)y2/1000000, (float) x2/1000000));
+  uint32_t getCoordDistance(int32_t  y1, int32_t x1, int32_t y2, int32_t x2, uint16_t alt) {
+  return(getCoordDistance( (float) y1/1000000, (float) x1/1000000, (float)y2/1000000, (float) x2/1000000, alt));
 }
 
 uint32_t getCoordDistance(float  y1, float x1, float y2, float x2, uint16_t alt){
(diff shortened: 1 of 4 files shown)


radio/src/maths.cpp
```diff
@@ -172,3 +173,30 @@ void getGpsDistance()
...
+  float nRadius = 6370997.0; // Earth’s radius in meters
+  float rad = 0.0174532925199433;
+  float nDLat = (y2 - y1) * rad;
+  float nDLon = (x2 - x1) * rad;
+  y1 *= rad;
+  y2 *= rad;
+
+  float nA =   pow( sin(nDLat/2), 2 ) + cos(y1) * cos(y2) * pow( sin(nDLon/2), 2 );
+  float nC = 2 * atan2( sqrt(nA), sqrt( 1 - nA ));
+  return round(nRadius * nC);
```
projectkk2glider: roundf

projectkk2glider: In fat all math functions here miss `f` - the ones you used are for `double`

radio/src/maths.cpp
```diff
@@ -172,3 +173,30 @@ void getGpsDistance()
...
+  float nDLat = (y2 - y1) * rad;
+  float nDLon = (x2 - x1) * rad;
+  y1 *= rad;
+  y2 *= rad;
+
+  float nA =   powf( sinf(nDLat/2), 2 ) + cosf(y1) * cosf(y2) * powf( sin(nDLon/2), 2 );
+  float nC = 2 * atan2f( sqrtf(nA), sqrtf( 1 - nA ));
+  return roundf(nRadius * nC);
+}
+#endif // STM32F4
```
bsongis: `elif defined(CPUARM)`

radio/src/opentx.h
```diff
@@ -1561,6 +1561,14 @@ extern uint8_t s_frsky_view;
 
 void getGpsPilotPosition();
 void getGpsDistance();
+
+#if defined(STM32F4)
+uint32_t getCoordDistance(float, float, float, float );
+#endif
```
bsongis: `elif defined(CPUARM)`

radio/src/telemetry/telemetry_sensors.cpp
```diff
@@ -296,20 +296,12 @@ void TelemetryItem::eval(const TelemetrySensor & sensor)
...
-
-        angle = abs(gpsItem.gps.longitude - gpsItem.pilotLongitude);
-        dist = gpsItem.distFromEarthAxis * angle / 1000000;
-        result += dist*dist;
-
+        uint32_t result = getCoordDistance(gpsItem.gps.latitude, gpsItem.gps.longitude, gpsItem.pilotLatitude, gpsItem.pilotLongitude);
         if (altItem) {
-          dist = abs(altItem->value) / g_model.telemetrySensors[sensor.dist.alt-1].getPrecDivisor();
-          result += dist*dist;
+          result += abs(altItem->value) / g_model.telemetrySensors[sensor.dist.alt-1].getPrecDivisor();
```
bsongis: Aie, Pythagore will wake up now ...

radio/src/maths.cpp
3djc: Don't we also need the int32 version on F4 because the gps data are in int32 and therefore it is needed for sensors computation ? Or do you rather call by dividing the int32 by 100000 on F4 ?

3djc:
I'll add altitude computing to the function

radio/src/telemetry/telemetry_sensors.cpp
```diff
@@ -297,8 +297,10 @@ void TelemetryItem::eval(const TelemetrySensor & sensor)
           }
         }
         uint32_t result = getCoordDistance(gpsItem.gps.latitude, gpsItem.gps.longitude, gpsItem.pilotLatitude, gpsItem.pilotLongitude);
+
         if (altItem) {
-          result += abs(altItem->value) / g_model.telemetrySensors[sensor.dist.alt-1].getPrecDivisor();
+          unint32_t alt = abs(altItem->value) / g_model.telemetrySensors[sensor.dist.alt-1].getPrecDivisor();
```
bsongis: Yes Travis is upset almost everyday :)

radio/src/maths.cpp
bsongis: But the new version is really better, no? i.e. no `sin(x) ~= x` approximation

bsongis: Would you write a small google test for this function? It would allow us to easily see the difference

mpaperno:
Floats for lat/lon?  Somewhat pointless (no pun intended).   Since we can't use doubles from Lua anymore, accept values in int32 then convert to double precision with **7** decimal places (int32 * 1e-7), then do the math, then convert back to int32 (double * 1e7) for return result.  

Which method are you using to calculate distance?

radio/src/maths.cpp
```diff
@@ -172,3 +173,30 @@ void getGpsDistance()
     telemetryData.hub.maxGpsDistance = telemetryData.hub.gpsDistance;
 }
 #endif
+
+#if defined(STM32F4)
+uint32_t getCoordDistance(float  y1, float x1, float y2, float x2){
+  float nRadius = 6370997.0; // Earth’s radius in meters
```
mpaperno: How was this radius value arrived at?  Seems close to "mean radius" but not exactly...  Anyway a fixed radius like this will limit precision unless you happen to be exactly at the matching latitude.  

Also, when using floats, always add an `f` at the end of the number.  This should be default behavior (i.e. good habit) unless you specifically need a double.

radio/src/maths.cpp
```diff
@@ -172,3 +173,30 @@ void getGpsDistance()
...
+  float rad = 0.0174532925199433;
+  float nDLat = (y2 - y1) * rad;
+  float nDLon = (x2 - x1) * rad;
+  y1 *= rad;
+  y2 *= rad;
+
+  float nA =   powf( sinf(nDLat/2), 2 ) + cosf(y1) * cosf(y2) * powf( sin(nDLon/2), 2 );
+  float nC = 2 * atan2f( sqrtf(nA), sqrtf( 1 - nA ));
```
mpaperno: Careful!  atan2f() can return NaN or Inf if both operands are **close** to zero!

radio/src/maths.cpp
```diff
@@ -172,3 +173,30 @@ void getGpsDistance()
...
+#if defined(STM32F4)
+uint32_t getCoordDistance(float  y1, float x1, float y2, float x2){
+  float nRadius = 6370997.0; // Earth’s radius in meters
+  float rad = 0.0174532925199433;
+  float nDLat = (y2 - y1) * rad;
+  float nDLon = (x2 - x1) * rad;
+  y1 *= rad;
+  y2 *= rad;
+
+  float nA =   powf( sinf(nDLat/2), 2 ) + cosf(y1) * cosf(y2) * powf( sin(nDLon/2), 2 );
```
mpaperno: I'd get rid of the `powf` calls here.  Calculate the sin values and store in a variable, then multiply those together, it should save a few CPU cycles.  Also you probably meant to use `sinf` in the last instance.

radio/src/maths.cpp
mpaperno: Why the F4 exception specifically?  F2 is capable of floating point math also, even if it is marginally slower w/out FPU.  I assume the same is true for the AT ARM chips though I'm not familiar with them specifically.

lshems:
Just out of interest: Why should this be a LUA call? Couldn't this be done within LUA itself? Or is there some calculation precision missing inside LUA?

Thanks for any clarifications

3djc:
Since we need it for dist calculated sensor, we might as well expose it to Lua, as simple as that

mpaperno:
Lua operates on floats only now (it used to use doubles for all numeric calculations, even integers, which was a huge waste of CPU cycles).  With float precision on degrees of lat/lon you get "only" ~11cm accuracy (though it depends greatly on the actual lat/lon being used... some values can be expressed as floats more accurately than others).  And of course the calculations are faster in compiled C code than Lua (these are pretty "expensive" operations due to the trigonometry involved).

3djc:
We tested it to be wrong by less than a metter in worse case scenario (0, 120) which is really not an issue since we return an integer anyway . Current computing for distance sensor overflow around 19 km distance, hence my desire to review it

lshems:
Ok, good thinking if the code was there already. Thanks for the reply.

radio/src/maths.cpp
```diff
@@ -172,3 +173,50 @@ void getGpsDistance()
...
+  //for greek adverse   a = sin²(delta_phi/2) + cos phi1 * cos phi2 * sin²(delta_lambda/2)
+  //                    c = 2 * atan2( √a, √(1−a) )
+  //                    d = R * c
+  // where	φ is latitude, λ is longitude, R is earth’s radius (mean radius = 6,371km); note that angles need to be in radians to pass to trig functions!
+  float R = 6371000.0f; // Earth’s mean, radius in meters
+
+  float phi1 = degToRad(lat1);
+  float phi2 = degToRad(lat2);
+  float delta_phi = degToRad(lat2-lat1);
+  float delta_lambda = degToRad(long2-long1);
```
mpaperno: Here you're storing the value of a "cheap" math function (though BTW delta_phi could just be phi2-phi1 to avoid more math converting to radians a 2nd time)...

radio/src/maths.cpp
```diff
@@ -172,3 +173,50 @@ void getGpsDistance()
...
+
+  float a =   sinf(delta_phi/2) * sinf(delta_phi/2) + cosf(phi1) * cosf(phi2) * sinf(delta_lambda/2) * sinf(delta_lambda/2);
```
mpaperno: ... then here you're repeating expensive trig (sinf()) functions twice!

radio/src/maths.cpp
```diff
@@ -172,3 +173,50 @@ void getGpsDistance()
...
+  float c = 2 * atan2f( sqrtf(a), sqrtf( 1 - a));
```
mpaperno: And I guess you don't believe me when I say atan2 can return NaN/Inf.  From [Wikipedia](https://en.wikipedia.org/wiki/Atan2) (emphasis mine):

"It also avoids the problem of division by zero, as atan2(y, 0) will return a valid answer **as long as y is non-zero**."

Due to float rounding issues, this also applies to values **near** zero.

radio/src/maths.cpp
```diff
@@ -172,3 +173,50 @@ void getGpsDistance()
...
+
+#if defined(STM32F4)
+
+inline float degToRad(float x)
+{
+    return x * 0.0174532925f; // Pi / 180
+}
+
+  uint32_t getCoordDistance(int32_t  y1, int32_t x1, int32_t y2, int32_t x2, uint16_t alt) {
+  return(getCoordDistance( (float) y1/1000000, (float) x1/1000000, (float)y2/1000000, (float) x2/1000000, alt));
```
mpaperno: `(float) y1 * 0.000001f`  
(or, what I find more readable `(float)y1 * 1e-6f`

3djc:
Scheduled for 2.2.1

PR #4504, opened 2017-02-24
Make lcd.clear() color aware on color screens

3djc:
Allows to specify a blanking color when using it on color screens


bsongis:
👍

PR #4517, opened 2017-02-26
Fixes #4136 Reset logical switch fields when changing function type

### LINKED ISSUE #4136: Companion logical switch parameters not resetting when function changed
Some parameter values are reset but not all and it depends on what the function is changed from and to.
Even when the function is changed to "----" all parameters are not reset to defaults.
Some old values are hidden depending on the active function.

Companion 2.1.9 Ubuntu 14.04

elecpower:
Fixes #4136 for Companion and radios where not all fields were being reset.
For a logical switch on change of function family clear all fields and set family defaults.
OFF function is unable to act as an enable/disable switch for most functions as it is a member of a function family.

If and when this PR is merged I will port to next and create another PR.

### DIFF
--- a/companion/src/modeledit/logicalswitches.cpp
+++ b/companion/src/modeledit/logicalswitches.cpp
@@ -11,7 +11,7 @@ LogicalSwitchesPanel::LogicalSwitchesPanel(QWidget * parent, ModelData & model, 
   ModelPanel(parent, model, generalSettings, firmware),
   selectedSwitch(0)
 {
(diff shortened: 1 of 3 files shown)


elecpower:
Refer PR #4536 for port to next

bsongis:
Not sure I want to see it in 'master'

elecpower:
@bsongis if you do not want the whole PR for 'master' would you accept a one line fix to only zero 'enabled' (reported bug) when switching to Edge?

PR #4530, opened 2017-03-01
remove src/radio/stamp.h from the gitignore (make cmake idempotent)

nathantsoi:
without this, i get:

```
CMake Error at CMakeLists.txt:88 (message):
  Source directory contains files leftover from a OpenTX 2.1 build.  Please
  run `git clean -f` in source directory (Careful: Will remove any extra
  files) or do a new clean git checkout
```

but `git clean -f` does not remove `stamp.h`

### DIFF
--- a/radio/src/.gitignore
+++ b/radio/src/.gitignore
@@ -1,4 +1,3 @@
-/stamp.h
 /opentx.elf
 /simu
 /opentx.map


projectkk2glider:
I think all of the files in this gitignore should be removed.

3djc:
But then git will want to up it to the repo, no ?

projectkk2glider:
You should not have those files in 2.2 in radio/src - they were built there in 2.1.  In fact there are probably other gitignore files in src tree that should be cleaned similarly.

kilrah:
Indeed, looks like none of that stuff would ever be in there anymore in 2.2.

bsongis:
Indeed more files could be removed from there, can be done later.

PR #4531, opened 2017-03-01
Fix a regression that omitted filenames from allsrc.cpp

alteman:
Commit ab6f910 added filename markers to allsrc.cpp, those subsequently got broken at some point.
This PR fixed the regression.

Filename markers are helpful when trying to figure out which file contains compilation errors.

### DIFF
--- a/radio/src/CMakeLists.txt
+++ b/radio/src/CMakeLists.txt
@@ -507,7 +507,7 @@ if(NOT MSVC)
     if (BASH AND NOT WIN32)
       add_custom_command(
         OUTPUT ${ALLSRC}
-        COMMAND ${BASH} -kc 'rm -f ${ALLSRC}\; for f in ${SRC}\; do echo "\\# 1 \"$$f\"" >> allsrc.cpp\; cat "$$f" >> ${ALLSRC}\; done'
+        COMMAND ${BASH} -kc 'rm -f ${ALLSRC}\; for f in ${SRC}\; do echo "\\# 1 \\\"$$f\\\"" >> ${ALLSRC}\; cat "$$f" >> ${ALLSRC}\; done'
         WORKING_DIRECTORY ${RADIO_SRC_DIRECTORY}
         DEPENDS ${SRC}
         )


mpaperno:
Ah, this probably explains an error I was getting but thought it was a "Windows problem" with Cygwin bash.  (https://github.com/opentx/opentx/commit/ac9b781647d02ed4505ee34c16b332738f214095#diff-d6d3457411b1de39b080f1117a560f6bR502)  But I'm surprised because I thought this worked for @bsongis.   Were you getting a compiler error (badly formatted markers), bash error, or... ?  Linux or OS X?

alteman:
Building on Linux (Ubuntu 16.04)
Repro here (run from fresh git clone):
$ # force a syntax error somewhere..
$ echo 'MAVLINK BUILD ERROR HERE;' >> ./radio/src/telemetry/mavlink.cpp 
$ mkdir build-9x-debug/
$ cd build-9x-debug/
$ cmake -DPCB=9X -DDEBUG=YES -DTELEMETRY=MAVLINK -DCMAKE_BUILD_TYPE=Debug ../
$ make firmware
...
[ 88%] Building CXX object radio/src/CMakeFiles/firmware.dir/__/__/allsrc.cpp.o
/home/user/opentx/build-9x-debug/allsrc.cpp:1170:1: error: ‘MAVLINK’ does not name a type
 MAVLINK BUILD ERROR HERE;
 ^
make[3]: *** [radio/src/CMakeFiles/firmware.dir/__/__/allsrc.cpp.o] Error 1
...

With this PR:

[ 88%] Building CXX object radio/src/CMakeFiles/firmware.dir/__/__/allsrc.cpp.o
telemetry/mavlink.cpp:590:1: error: ‘MAVLINK’ does not name a type
make[3]: *** [radio/src/CMakeFiles/firmware.dir/__/__/allsrc.cpp.o] Error 1

mpaperno:
Sorry, I wasn't clear.  I know why the markers are useful.  I was wondering what was wrong with the code/script which generates the allsrc.cpp file, the part you are suggesting to change.  Are there simply no markers at all in the generated .cpp, or are they badly formatted somehow?  Or perhaps it was creating the markers in `${RADIO_SRC_DIRECTORY}/allsrc.cpp` instead of `${PROJECT_BINARY_DIR}/allsrc.cpp`?

alteman:
There were two issues. First one is insufficient escaping of quotes in the filename causing the filename to be unquoted in the source file 
`# 1 targets/9x/board.cpp`
instead of 
`# 1 "targets/9x/board.cpp"`
, which in turn causes a syntax error during compilation. The second issue was with the path of the output file not being set correctly. The second issue actually prevented the markers from being inserted altogether, that's why the build wasn't failing with syntax errors due to the first issue :-)

bsongis:
You are right, I messed up things and there were no compile errors / warnings, that's why I didn't see it ...

PR #4534, opened 2017-03-02
Allow datestamp for ARM 128x64 model backups

### LINKED ISSUE #4511: X7 backups to model name only, no warning for overwrite last backup
On X9, backups have a date to the name. Multiple backups can be made, without the risk of loosing a backup.

On X7, no date is added, and existing backup is overwritten without warning.

3djc:
To allow restore, I've had to increase MENU_LINE_LENGTH, could not find adverse effect, but want to warn about it

This fixes #4511


schwabe:
I think this should rather check for RTC clock than displaywidth and PCB. 

I think the reason is that not all 9x boards have a rtc. I think the majority has none.

PR #4536, opened 2017-03-02
Fixes #4136 Reset all logical switch fields on change of function type

### LINKED ISSUE #4136: Companion logical switch parameters not resetting when function changed

elecpower:
Port to next of #4136 to reset all logical switch fields on change of function family

### DIFF
--- a/radio/src/gui/128x64/model_logical_switches.cpp
+++ b/radio/src/gui/128x64/model_logical_switches.cpp
@@ -445,7 +445,12 @@ void menuModelLogicalSwitches(event_t event)
           CHECK_INCDEC_MODELVAR_ZERO(event, cs->func, LS_FUNC_MAX);
           uint8_t new_cstate = lswFamily(cs->func);
           if (cstate != new_cstate) {
-            cs->v1 = cs->v2 = (new_cstate==LS_FAMILY_TIMER ? -119/*1.0*/ : 0);
+            unsigned int save_func = cs->func;
+            memset(cs, 0, sizeof(LogicalSwitchData));
+            cs->func = save_func;
+            if (new_cstate == LS_FAMILY_TIMER) {
+              cs->v1 = cs->v2 = -119;
+            }
           }
           break;
         }
--- a/companion/src/modeledit/logicalswitches.cpp
+++ b/companion/src/modeledit/logicalswitches.cpp
@@ -25,7 +25,7 @@ LogicalSwitchesPanel::LogicalSwitchesPanel(QWidget * parent, ModelData & model, 
   ModelPanel(parent, model, generalSettings, firmware),
   selectedSwitch(0)
 {
-  Stopwatch s1("LogicalSwitchesPanel"); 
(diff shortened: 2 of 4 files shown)


elecpower:
Refer #4517 for merge to master

radio/src/gui/128x64/model_logical_switches.cpp
```diff
@@ -445,7 +445,12 @@ void menuModelLogicalSwitches(event_t event)
           CHECK_INCDEC_MODELVAR_ZERO(event, cs->func, LS_FUNC_MAX);
           uint8_t new_cstate = lswFamily(cs->func);
           if (cstate != new_cstate) {
-            cs->v1 = cs->v2 = (new_cstate==LS_FAMILY_TIMER ? -119/*1.0*/ : 0);
+            unsigned int save_func = cs->func;
+            memset(cs, 0, sizeof(LogicalSwitchData));
+            cs->func = save_func;
+            if (new_cstate == LS_FAMILY_TIMER) {
+              cs->v1 = cs->v2 = -119;
```
bsongis: You removed the only comment I wrote in this project!

radio/src/gui/128x64/model_logical_switches.cpp
```diff
@@ -445,7 +445,12 @@ void menuModelLogicalSwitches(event_t event)
           CHECK_INCDEC_MODELVAR_ZERO(event, cs->func, LS_FUNC_MAX);
           uint8_t new_cstate = lswFamily(cs->func);
           if (cstate != new_cstate) {
-            cs->v1 = cs->v2 = (new_cstate==LS_FAMILY_TIMER ? -119/*1.0*/ : 0);
+            unsigned int save_func = cs->func;
+            memset(cs, 0, sizeof(LogicalSwitchData));
```
bsongis: `memclear(...)` saves flash on AVR platforms and does the same job

bsongis: I would prefer this memclear is called only on ARM and CPUM2560 platforms, for 9X / 9XR I would prefer we cast to uint16_t the structure (or add a new union type) and apply a mask. I will give it a try

bsongis: This solution costs only 4bytes on 9X:
```
          if (cstate != new_cstate) {
#if defined(CPUARM)
            unsigned int save_func = cs->func;
            memset(cs, 0, sizeof(LogicalSwitchData));
            cs->func = save_func;
#else
            cs->v1 = cs->v2 = 0;
            cs->andsw = 0;
#endif
            if (new_cstate == LS_FAMILY_TIMER) {
              cs->v1 = cs->v2 = -119;
            }
          }
```

bsongis:
I just pushed my proposal above in the branch 'next'

PR #4538, opened 2017-03-02
Remove Heli choice from model wiz as we do not have one

### LINKED ISSUE #4537: Setting new model heli
Running opentx 2.2rc10 opentx-x9d+-lua-massstorage-multimodule-en-2.2.0N362.bin on a Taranis Plus

I wanted to setup a new model (Helicopter all others are multicopters) and on the first model create page on the transmitter where you can select Plane, Heli, Fixed wing and Multi I can select all but Heli.

Nothing happens when pressing enter or page, I had noheli selected first so went back to companion and disabled noheli and still same problem.

Any suggestions? it's my first Heli so might be user error.

3djc:
This (kinda) fixes #4537


bsongis:
Nice fix :)

Dutchy-79:
Lol, easiest fix ever

PR #4543, opened 2017-03-02
[Horus] Fixes #4407

### LINKED ISSUE #4407: Companion, Does not accept any settings internal HF module from the Horus
Companion 2.20N362
Does not accept any settings internal HF module from the Horus
No data exchange values in the data fields
The data of the models were taken over for the first time by Horus

Raimund

bsongis:


companion/src/firmwares/opentx/opentxeeprom.h
```diff
@@ -99,7 +99,7 @@ class ProtocolsConversionTable: public ConversionTable
         addConversion(PULSES_DSMX, val++);
       }
       if (IS_HORUS_OR_TARANIS(board)) {
-    	  addConversion(PULSES_CROSSFIRE, val++);
+    	addConversion(PULSES_CROSSFIRE, val++);
```
mpaperno: OMG, a TAB!   :-)

companion/src/modeledit/setup.cpp
```diff
@@ -238,9 +238,9 @@ ModulePanel::ModulePanel(QWidget * parent, ModelData & model, ModuleData & modul
   lock = false;
 
   // a change will not register if only one item was added
-  if (ui->protocol->count() == 1)
+  if (ui->protocol->count() == 1) {
```
mpaperno: Does this mean we should always use brackets even for single-line ifs?  Our previous discussion on the matter seemed to suggest otherwise (in certain circumstances).  I'm fine with it either way, just looking for consistency.

bsongis: No idea what is the best choice

mpaperno: Doesn't matter.  I just didn't want you or anyone to feel like they should be cleaning up.  I used no-brackets syntax a bunch in those changes, so if it bothers anyone just let me know.  If some instances are less readable than others and can use brackets, that's fine with me also.

projectkk2glider: I prefer to use braces in such cases http://stackoverflow.com/questions/2125066/is-it-bad-practice-to-use-an-if-statement-without-brackets

I also like to use single line in some cases like so, it makes it more obvious:
```
if (something) return;
```

kilrah: I personally like the 
```C
if (statement)
    do this;
else
    do this;
```
But NOT the single-line construction.

We'll never find something that wveryone's happy with ;)

mpaperno: I'm with Andre.  :)   Especially:
> But NOT the single-line construction.

I will take the "extra" brackets over that any day.

Here's where we left it last time... :)  
https://github.com/opentx/opentx/pull/4141#discussion_r92862868

3djc: Can I suggest that we document it (a wiki page ?) when we agree ? There is just too much for my little brain to remember :)

bsongis: Aha yes, but we agree that we don't agree :)
I also don't like the single line
Otherwise I have to admit that I change my mind everyday

mpaperno: You mean like document it [here](https://github.com/opentx/opentx/wiki/OpenTX-Code-Style-Guide) for example?  :)

PR #4544, opened 2017-03-02
[Horus] Fixes #3775

### LINKED ISSUE #3775: OpenTX 2.2.0N356 Changing Themes
After changing themes the background on the transparent area for the model images stays like it was on the last theme, It only changes after a power cycle (Horus)
bug

bsongis:
Quick fix ... comments welcome :)

### DIFF
--- a/radio/src/gui/480x272/widgets/modelbmp.cpp
+++ b/radio/src/gui/480x272/widgets/modelbmp.cpp
@@ -25,9 +25,9 @@ class ModelBitmapWidget: public Widget
   public:
     ModelBitmapWidget(const WidgetFactory * factory, const Zone & zone, Widget::PersistentData * persistentData):
       Widget(factory, zone, persistentData),
-      buffer(NULL)
+      buffer(NULL),
+      hash(255)
     {
-      memset(bitmapFilename, 255, sizeof(bitmapFilename));
     }
 
     virtual ~ModelBitmapWidget()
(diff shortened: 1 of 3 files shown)


mpaperno:
Tested, and fixes the issue, at least in Simulator.

Comments follow...  :)

There is no way right now for the widget itself or the layout manager (or theme, or...?) to signal that something has changed?  E.g. a `modified` bool flag that is quick to check?  The hash or memcmp loops are relatively fast, but still... loops.  Or is `refresh()` not time-sensitive?

I see, none of the other widgets seem to need this right now...   But could such a mechanism be useful? 

Eg. add such a flag at the base Widget class level with a public setter, and some way to trigger that in a cascading fashion from different events, eg. when Theme (or Layout?) settings change, or simply when the Widget settings are modified.  I suppose the other components involved may also need an `updated` flag/mechanism of their own, but perhaps they may also benefit from it.

radio/src/gui/480x272/widgets/modelbmp.cpp
```diff
@@ -63,13 +63,24 @@ class ModelBitmapWidget: public Widget
       }
     }
 
+    uint8_t evalHash(const void * data, uint32_t size)
```
mpaperno: I actually compared generated assy code (gcc-arm 4.7) and this looks faster, though mostly because `memcmp` doesn't seem to be very efficient.  If you wrote your own "memcmp" which just simply compared bytes by value it may be marginally faster, but we're talking 1-2 cycle(s) difference per loop iteration.

BTW the compiler inlined `evalHash()`.  The resulting `refresh()` is only 5 extra lines of assy over the old version, which also jumped to `memcmp` twice (`memcpy` gets inlined to 6 lines of assy each).  `memcmp` is 82 lines!

Just in case anyone was wondering.    :-)

radio/src/gui/480x272/widgets/modelbmp.cpp
```diff
@@ -78,9 +89,8 @@ class ModelBitmapWidget: public Widget
     }
 
   protected:
-    char bitmapFilename[sizeof(g_model.header.bitmap)];
-    char modelName[sizeof(g_model.header.name)];
     BitmapBuffer * buffer;
+    uint32_t hash;
```
bsongis: uint8_t

bsongis:
I was rather afraid of collisions (because of this hash lazy function) than code size :)

mpaperno:
Code size was an interesting aside.  I was more curious about the execution speed differences myself.

projectkk2glider:
This could all be avoided if the widget kept in its buffer a proper transparent image of itself (RGB4444). Then whatever it is behind it would be visible without the need for refresh.

Issue #4480, opened 2017-02-19
T5 / T6 Horus => Companion transfers that sources as "Lua1a" 


Casichispea:
if you have created a model in the Horus using  T5 as source, when you copy it with companion 2.2 into the computer, the process changes the source "T5" and appears  "LUA1a".

Red-Devil74:
As T5 and T6 are not present in Companion, I assume it defaults to the first entry in the list "LUA1a". I have the same issue. After transfer from Horus to Companion it shows 3 question marks in the mixes tab. Only after opening said mix is appears as "LUA1a". (hope this info helps a bit...)
![image](https://cloud.githubusercontent.com/assets/23432540/23231635/cc56e4dc-f948-11e6-8d31-548809394068.png)
![image](https://cloud.githubusercontent.com/assets/23432540/23231654/e446cd5a-f948-11e6-8de6-72c98044d84b.png)

Red-Devil74:
Is this not fixed as per remark in buglist "Horus misc issues #3393"?

kilrah:
No that referred to radio firmware only, still missing in companion

Issue #4336, opened 2017-01-29
 Joystick values ​​invisible in companion, joystick immovable in simulator HORUS


1Sven:
The nice future, a Joystick, usefull for special functions like a cam or  switches and so on is not supported. In mixers the values mean: " ??? " .  No values, are assignable, no inputs is definable.  The knob for Joystick is not movable.
I think a importand funktion for many users. How about, to aktivate this futures of the HORUS in companion and simulator?  
![image](https://cloud.githubusercontent.com/assets/25420478/22407359/5a5de4e8-e665-11e6-893c-54392337acf9.png)

projectkk2glider:
The `???` is a bug, while the movable joystick in  simulator  is an enhancement.

PR #4550, opened 2017-03-04
[Companion][Horus] Add joystick axes (JSx & JSy) and trims T5 & T6 as usable sources.

mpaperno:
Fixes bug in #4336.  JSx/JSy are available in the same lists which now show pots/sliders.

horus-js-source

### DIFF
--- a/companion/src/boards.cpp
+++ b/companion/src/boards.cpp
@@ -123,6 +123,11 @@ int getBoardCapability(Board::Type board, Board::Capability capability)
         return 2;
       else
         return 0;
+    case Joysticks:
+      if (IS_HORUS(board))
+        return 2;
+      else
+        return 0;
     case FactoryInstalledSwitches:
       if (IS_TARANIS_X9E(board))
         return 8;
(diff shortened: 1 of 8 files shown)


kilrah:
Cool 👍 Any chance you could add T5/T6 too while you're there?

mpaperno:
Yes... I think I've located all the pieces...  :)

Also the trim switches (5u/d & 6u/d) need to be added, right?

kilrah:
Correct, and the trim choices in Inputs

mpaperno:
I guess this now fixes #4480 as well.

kilrah:
Looks good!

mpaperno:
Andre, thanks for the tests!

@bsongis LMK if you're [not] happy with the enum naming, placement, etc.  Otherwise can be merged I think.

(Notice how I cleverly [avoided](https://github.com/opentx/opentx/pull/4550/commits/ea8852b1516d93911cc75daa4b6f3d7bff79c378#diff-14171a0cc43b987f2dd399c3f25a390fL696) the `if` bracket "controversy..." 😛  )

bsongis:
I prefer this way as well, it allows to be renamed without more indentation work

PR #4552, opened 2017-03-04
Update third party mavlink header files to git revision 8e864d2

alteman:
Current snapshot of mavlink headers is pretty old (~3 years) and doesn't have a definition of RC_CHANNELS message used by PX4 FC to transmit radio RSSI. 

First commit: results of running radio/src/thirdparty/GCS_MAVLink/sync script
Updated headers should correspond to mavlink repo git revision 8e864d2 (mavlink repo hasn't used tags since 2014, so I'm specifying a git revision for clarity).

Second commit: mavlink_system_t structure got rid of previously unused fields, changing initialization in telemetry/mavlink.cpp prevents a build error.

I'm submitting a PR adding support for PX4 RSSI next.


bsongis:
I only checked you modified things in thirdparties. If it's possible to only have a ling to another git repo it would be excellent (git clone --recursive to put in the docs)

PR #4554, opened 2017-03-04
[simu][horus] Replace 2nd knob/pot with functional 6-pos switch

mpaperno:
As with Simulator, the pot must be "calibrated."  This is because the current [radio code](https://github.com/opentx/opentx/blob/mpaperno/simu_6pos/radio/src/switches.cpp#L253) won't even try to read the switch/pot positions w/out at least some valid calibration data ( 👍 ).

This is just hard-coding the 6-pos switch with yet another `#ifdef`... open to better ideas or if anyone thinks the 6-pos should be added to other radio(s) also.

simu-horus-6pos

### DIFF
--- a/radio/src/simu.cpp
+++ b/radio/src/simu.cpp
@@ -117,8 +117,17 @@ Open9xSim::Open9xSim(FXApp* a):
 
   for(int i=0; i<NUM_POTS+NUM_SLIDERS; i++){
     knobs[i]= new FXKnob(hf11,NULL,0,KNOB_TICKS|LAYOUT_LEFT);
-    knobs[i]->setRange(-1024, 1024);
     knobs[i]->setValue(0);
+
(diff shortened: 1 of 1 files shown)


bsongis:
Good!

radio/src/simu.cpp
```diff
@@ -117,8 +117,17 @@ Open9xSim::Open9xSim(FXApp* a):
 
   for(int i=0; i<NUM_POTS+NUM_SLIDERS; i++){
     knobs[i]= new FXKnob(hf11,NULL,0,KNOB_TICKS|LAYOUT_LEFT);
-    knobs[i]->setRange(-1024, 1024);
     knobs[i]->setValue(0);
+
+#if defined(PCBHORUS)
```
bsongis: Only problem if the knob (pot1) is configured as a pot ...

PR #4555, opened 2017-03-05
[Companion] Add context-menu actions for moving a model between categories.

mpaperno:
The new option is only shown if > 1 categories exist.
Ref. #4224 : Horus models cannot be moved from one category to another.

model_move_cat


bsongis:
👍

PR #4556, opened 2017-03-05
Replace thirdparty/GCS_MAVLink with a git submodule

alteman:
Needs git clone --recursive now. Hopefully recursive clone is on by default in Travis CI.
I'll update build instructions in wiki.


mpaperno:
I'm not against submodules (I use them for Mavlink includes in my own projects also), but there's at least one caveat which may not be obvious:
 
When downloading the repo from GitHub (as zip), submodule content is NOT included.  One has to do this song-and-dance by clicking through to the [submodule folder](https://github.com/alteman/opentx/tree/mavlink-subrepo/radio/src/thirdparty/GCS_MAVLink) and downloading it separately from the origin repo which you get to by following the link there (and making sure to place the contents in the right folder of your local copy!). 

I'm not sure how many people d-load vs. clone, but it's something to be aware of.

bsongis:
Thanks!

alteman:
@mpaperno Good point. I've added some recovery logic to CMake to try and clone the submodule (covers users who updated their repo and didn't clone with --recursive). There's also an error message suggesting users who downloaded zip should git clone instead ;).
Note that it only affects 9x build with DTELEMETRY=MAVLINK.

PR #4566, opened 2017-03-06
Also define ana_mppings if horus sticks are used (closes #4565)

### LINKED ISSUE #4565: OpenTX 2.2.0 RC11 (N363) -- Compilation error when downloading Taranis X9E Firmware.
Unable to install that latest build on to my X9E because it keeps giving me a "Compilation error".

The problem occurs when i have the "horussticks" option selected, if i deselect that option then it will download the firmware but will take about a minute to do so.

However i have just installed Horus sticks into my X9E, so i do need to have that option ticked.

Thanks...

compilation error

schwabe:


radio/src/targets/common/arm/stm32/adc_driver.cpp
```diff
@@ -24,10 +24,12 @@
   // not needed
 #elif defined(PCBX10)
   const int8_t ana_direction[NUM_ANALOGS] = {1,-1,1,-1,  -1,-1,-1, -1,1,1,1,  -1};
-#elif defined(PCBX9E) && defined(HORUS_STICKS)
-  const int8_t ana_direction[NUM_ANALOGS] = {1,-1,1,-1,  -1,-1,-1,1, -1,1,1,1,  -1};
 #elif defined(PCBX9E)
-  const int8_t ana_direction[NUM_ANALOGS] = {1,1,-1,-1,  -1,-1,-1,1, -1,1,1,1,  -1};
+  #if defined(HORUS_STICKS)
```
3djc: I think we do not ident #if and the likes

mpaperno: Not too late to start though... 😉   It is done in a few places already, but not consistently.  I find it helps a lot in some cases, very awkward-looking in others.  The common problem with indenting them around C code (like this) is which level to indent the C code to... the # directives or the other C code.

kilrah: We typically don't indent them so that they're obvious when reading Found ourselves follwoing code too often that happened not to apply to the desired config, but the #ifdef wasn't noticed.

But again it's hard to be consistent - for example in a file like hal.h where it's nothing but defines it's nice and helpful to indent the #ifdefs - when mixed with code I don't like it for the above reason

mpaperno: Yes when the code is all `#` preproc stuff then indenting works nicely I think.  When mixed with C code I end up not using indents.  But when the block spans more than a few lines or has nested blocks I find it important to comment all the `#else`s and `#endif`s copiously.   

Of course the best way to view it in any case is when the IDE does proper highlighting of "dead" code.

schwabe: We indent ifdef in a few places. I felt here it improved readability but I can change it back to no indentation. I have no strong feelings either way.

bsongis: I do agree we have not a consistent rule for `#ifdef` indentation, sometimes it's good, someone less. It would be nice to avoid nested `#ifdef` as much as possible anywhay

PR #4569, opened 2017-03-06
Projectkk2glider/issue 4479 slow joystick

### LINKED ISSUE #4479: USB joystick lag since upgrading to 2.2 firmware on Taranis X9Dplus
Iv been using a lot of emulators and had no issues all input was smooth.
Since updating to 2.2 firmware iv found that USB joystick input is not smooth.
when moving throttle quickly you can hear input it sharp stepped and not smooth.
Smooth if you throttle slowley but sharp movements as its as its skipped steps and throttle is not smooth because of not getting the data quick enough to the pc?

is it a known issue or can someone test see if they have the same?

projectkk2glider:
~~Wait for feedback in https://github.com/opentx/opentx/issues/4479#issuecomment-284531024 before merging.~~

Closes #4479

### DIFF
--- a/radio/src/targets/common/arm/stm32/usb_driver.cpp
+++ b/radio/src/targets/common/arm/stm32/usb_driver.cpp
@@ -30,6 +30,8 @@
 #include "opentx.h"
 #include "debug.h"
 
+static bool usbDriverStarted = false;
+
 int usbPlugged(void)
 {
   // debounce
@@ -63,6 +65,7 @@ void usbInit(void)
 {
   // Initialize hardware
   USB_OTG_BSP_Init(&USB_OTG_dev);
+  usbDriverStarted = false;
 }
 
 void usbStart(void)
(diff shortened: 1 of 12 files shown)


radio/src/targets/common/arm/stm32/usb_driver.cpp
```diff
@@ -46,7 +55,7 @@ int usbPlugged(void)
 
 USB_OTG_CORE_HANDLE USB_OTG_dev;
 
-void OTG_FS_IRQHandler(void)
+extern "C" void OTG_FS_IRQHandler(void)
```
schwabe: This extern "C" should not be necessary if the header already has a extern "C" definition. You can also replace (void)  with () argument in C++.

radio/src/targets/common/arm/stm32/usb_driver.cpp
```diff
@@ -70,9 +80,60 @@ void usbStart(void)
   // initialize USB as MSC device
   USBD_Init(&USB_OTG_dev, USB_OTG_FS_CORE_ID, &USR_desc, &USBD_MSC_cb, &USR_cb);
 #endif
+  usbDriverStarted = true;
 }
 
 void usbStop(void)
```
schwabe: more void that can be removed. I will not comment the other ones :)

radio/src/targets/common/arm/stm32/usb_driver.cpp
projectkk2glider: It is necessary here.

bsongis:
Good catch!

kilrah:
I've read about potential issues/disconnections with X9 in masstorage mode (in the main firmware with the option), might need to check that too

PR #4572, opened 2017-03-07
Casting fix

### LINKED ISSUE #4568: RC11 N363 on Taranis Q X7
Nice effort with the RC11 N323, I like it very much!

I found a minor issue:

I can set a Logical Switch with e.g. Cells 3.25V:

logicalswitch001

But in the Logical Switches overview it looks like this:

logicalswitch002

I think it should be:
Cels 3,25V

Tested with real Q X7 and Companion.

Bye
JR

3djc:
Thix fixes #4568


bsongis:
Good catch!

PR #4573, opened 2017-03-07
Remove mixermon option (make it always true)

### LINKED ISSUE #4557: Remove MIXERS_MONITOR option, make MM always available.
Hy,
Profil X9D+

No Mixermonitor at Simulator with ENTER
At Option Mixermonitor is selceted

Same with  X9E

3djc:
This fixes #4557


mpaperno:
👍  

I hope you get as much satisfaction from removing code as I do.  :)

kilrah:
Maybe FW build script too?

bsongis:
@mpaperno there are potential satisfaction for a couple of months in opentx :)

lthole:
Does this mean that the mixermon screen will be implemented on the X7?

schwabe:
@lthole unrelated to that, this commit only affects radios that already have a mixermon.

radio/src/gui/128x64/view_channels.cpp
```diff
@@ -39,23 +38,20 @@ void menuChannelsView(event_t event)
...
   else
     ch = 0;
 
-#ifdef MIXERS_MONITOR
   if (mixersView)
   lcdDrawTextAlignedCenter(0*FH, MIXERS_MONITOR);
   else
-#endif
+
   lcdDrawTextAlignedCenter(0*FH, CHANNELS_MONITOR);
```
mpaperno: These could use some indents I think.  Also you'd get another -1 LoC (if you don't use braces).  :-)
Or even `lcdDrawTextAlignedCenter(0, mixersView ? MIXERS_MONITOR : CHANNELS_MONITOR);`  🤓

mpaperno: Sorry, I didn't realize `lcdDrawTextAlignedCenter` was a macro.  If you put the conditional in parenthesis, it works (tested on AVR compiler).  I guess that's a good reason to always use brackets in such cases (note to self).  

`lcdDrawTextAlignedCenter(0, (mixersView ? MIXERS_MONITOR : CHANNELS_MONITOR));`

(Not saying you need to change this, but just FYI).

mpaperno: Never mind, sorry.... still doesn't work as one line.  Tested with wrong compiler!

bsongis:
👍

PR #4597, opened 2017-03-13
Re-organize nightly sequence

3djc:
A possibly more logical sequence of event


gasguru:
..a wee shell tweak: While ln 57: _+rm -f $output/companion/windows/companion-windows.stamp_ is clear about the stamp file removal, the ">>" in:... 
ln 58 _+echo "#define VERSION  "'"2.2.0'$suffix'"' >> $output/companion/windows/companion-windows.stamp_ 

...allows appending to an old stubborn not-yet-stomped-out stamp file, where an " > " would forcefully create-or-overwrite it "either way", e.g. like in: ...
'ln 58 +echo "#define VERSION  "'"2.2.0'$suffix'"' > $output/companion/windows/companion-windows.stamp'

...which would allow the removal of ln 57 and making my ln 58, the new ln 57.

PR #4605, opened 2017-03-13
Fix #4603

### LINKED ISSUE #4603: Q-X7 Does not remember Date if only date is changed.
Q-X7 2.2.0N363

If you change just the Date field on the Radio Setup page and switch off the radio, the change will not be remembered.  I tried changing the year and the day separately and navigating away from the Radio Setup pages.  The date is remembered until the radio is switched off.

The Date change is only remembered if you change first the Date field and then the Time field.  

If you change the Date after the Time, the date change will not be remembered

A quick check on my X9D+ with the same version did not show the same problem.  Editing just the Date is remembered after switching off the radio.

3djc:
IO think this fixes the issue (tested on my X7), however I'm unsure why those +1 where in there (apparently added way before x7 (2 years), so I might brake other radio doing that, please check thoroughly

This fixes #4603

### DIFF
--- a/radio/src/gui/128x64/radio_setup.cpp
+++ b/radio/src/gui/128x64/radio_setup.cpp
@@ -124,7 +124,7 @@ void menuRadioSetup(event_t event)
   struct gtm t;
   gettime(&t);
 
-  if ((menuVerticalPosition==ITEM_SETUP_DATE || menuVerticalPosition==ITEM_SETUP_TIME) &&
+  if ((menuVerticalPosition==(ITEM_SETUP_DATE + HEADER_LINE) || menuVerticalPosition==(ITEM_SETUP_TIME + HEADER_LINE) &&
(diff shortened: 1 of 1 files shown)


radio/src/gui/128x64/radio_setup.cpp
```diff
@@ -124,7 +124,7 @@ void menuRadioSetup(event_t event)
   struct gtm t;
   gettime(&t);
 
-  if ((menuVerticalPosition==ITEM_SETUP_DATE+1 || menuVerticalPosition==ITEM_SETUP_TIME+1) &&
```
bsongis: `HEADER_LINE + ITEM_SETUP_DATE` or you break the 9XR-PRO

bsongis:
Missing parenthesis?

bsongis:
👍

Issue #4589, opened 2017-03-12
Horus crashing on model image selection


joaquins:
Hi,

Just upgraded my Horus to firmware:
opentx-x12s-ppmus-multimodule-noheli-lua-massstorage-en-2.2.0N363.bin

When selecting a new image model, the radio crashed under these conditions:
- Case one: Quick scroll backwards of model images, radio shows a red block covering the lower end of the screen, and becomes non-responsive. Needed to shutdown with long-press on power button.
- Case two: Quick scroll forward of model images. Radio reboots by itself into "emergency mode". Needed to shutdown with long-press on power button

- Case three: Slow scroll of image models, after scrolling last image, radio stop responding, and shutdown is needed.

Attaching my images folder in case it helps.
[IMAGES.zip](https://github.com/opentx/opentx/files/836215/IMAGES.zip)

kilrah:
Scrolling where? On the model setup page or in the SD card browser?

Can't reproduce on either even with your folder. Coudl be bad SD card...

marcogpw:
I've had the same happen to me: on the Horus trying to change the model picture via the telemetry menu, it crashed on me as well, showing the emergency mode screen

3djc:
Changing model picture using telemetry menu ????

marcogpw:
long press telemetry on the right selector, then you get the screen where you can change the widgets for telemetry etc, one of the screens holds the model picture selected in the main model setup and you can change it there as well, during the picture change it froze on my Horus

3djc:
hum, how exactly to you change the picture there ?? It is a display only as fas as I know

joaquins:
Update: I found some hidden files under my IMAGES folder. Files starting with "._", created by OSX.

After deleting those files, I can't reproduce the crash any longer.

bsongis:
are you able to reproduce this issue? would you give us an example of a file which does this?

marcogpw:
OK, update on my crash event: I checked the IMAGES folder, there are no strange files in there. I tried some more things and I found the culprit: the telenowa script. As soon as I am setting widgets and the telenowa widget was in the selection box the Horus crashed and Emergency mode was on the display. Most probably this is what happened last time as well. Deleted this widget, so I expect no more issues. Thanks for all your thinking guys !

projectkk2glider:
@marcogpw the telenowa script issue should be handled in RC12, which version are you using?

marcogpw:
[...]

PR #4609, opened 2017-03-14
Filter out hidden and system files

3djc:
This should help #4589


projectkk2glider:
I agree with this change. But I am wondering what is your rationale why would this help/prevent a crash?

3djc:
https://github.com/opentx/opentx/issues/4589#issuecomment-286313662

projectkk2glider:
Sure, but **why** did those files cause a crash and what needs to be fixed so another file does not produce the same issue.

3djc:
It is more that given the wide range of OS and image utilities that may create all kinds of thumbnails, it sounded to me like a good thing to do no matter what.

bsongis:
It doesn't fix the issue, but it won't hurt!

Issue #4099, opened 2016-12-05
Horus OpenTX 2.2.360 RC8 Problem with Screen again


Githubbachelor:
Horus OpenTX 2.2.360 RC8 Problem with screen again after laoding some models. Maybe same problem as known? But I dont know. Because new Version 2.2.360
After I must switch off the Horus and at closing there were also these pics during closing.
![20161205_002709](https://cloud.githubusercontent.com/assets/22581690/20870488/46a51a78-ba89-11e6-9d47-5c8ad76d5635.jpg)

bsongis:
Is it 100% reproductible? Would you attach the model pictures?

3djc:
There is an issue with cat lenght at least, when I testes I could not put more than 9, en know it's 15, which goes way beyong thr colunm

3djc:
![image](https://cloud.githubusercontent.com/assets/5167938/20875481/6f13d652-babb-11e6-8508-34375efad5b8.png)

3djc:
@Githubbachelor : please include a full copy of your sd content

Githubbachelor:
OK I make a zip for uploading now

Githubbachelor:
Cannot upload is too big the zip

Githubbachelor:
where can I upload

kilrah:
It's up here: http://www.filedropper.com/sdkarteanlagehorusaktuelleversion360_1

3djc:
The zp provided contains 92 models, the screen above show 93 ...

Githubbachelor:
Yes thats right I killed the wrong new model.5 which was model125.bin

kilrah:
We still don't know what sequence of things we shoudl do to reproduce this. 
If this happened while deleting the model then we can't reproduce as it's not there anymore, we'd have needed a copy of the card BEFORE the problem.

3djc:
Or at least in the very same state on when the problem happened. When things liek that happen, you can remove the sd card, make a copy on your pc, then put it back and delete the model causing you problems. Like this, nothing to investigate

Githubbachelor:
Again this has noting to do with the deleted Model.5. Two different things!

The screen problem happen when I opened peace for peace each model to control your new NB360 in each model. And at the end of doing this happen that screen problem again. After switching off the TX and switch on everything again OK. But it happen everytime during opening any model and only 1 time a week maybe.

Githubbachelor:
The SD Card you get from me have had that problem.

[...]

PR #4617, opened 2017-03-14
Projectkk2glider/issue 4099 leaks

projectkk2glider:
This fixes memory leaks, but we still need to solve the issue of the `new()` failure, either by enabling C++ exceptions or using `std:nothrow`.

 I can't find issue where we discussed this already, but second solution still leaves with `std::list` problem. I have a CLI test for that already written: https://github.com/opentx/opentx/blob/a1a58ad8f8cbcadf838e766a96046034607bec33/radio/src/cli.cpp#L307

These leaks were found with valgrind:
``` 
valgrind --leak-check=full --log-file="logfile.out" -v  ./simu
```

Might solve (at least partially) the  #4099.


bsongis:
👍

PR #4623, opened 2017-03-15
Handle neg gvars precision

### LINKED ISSUE #4582: using gVar as offset wrong scaling (x7 radio)
When using a Gvar as offset on a mixer line, the actual effect is only 10 % of the Gvar setting.

So on a mixer line, source max, weight 0, offset gVar1 with gVar set to -100, the output is -10.

3djc:
This fixes #4582

### DIFF
--- a/radio/src/gvars.cpp
+++ b/radio/src/gvars.cpp
@@ -74,7 +74,7 @@ int16_t getGVarValue(int8_t gv, int8_t fm)
 int32_t getGVarValuePrec1(int8_t gv, int8_t fm)
 {
   int8_t mul;
-  uint8_t prec = g_model.gvars[gv].prec;
+  uint8_t prec = g_model.gvars[abs(gv)].prec;
   if (prec == 0)
     mul = 10;
   else


bsongis:
good catch!

projectkk2glider:
On Ubuntu Yakkety (gcc 6.2.0):
```
/<<PKGBUILDDIR>>/radio/src/gvars.cpp: In function ‘int32_t getGVarValuePrec1(int8_t, int8_t)’:
/<<PKGBUILDDIR>>/radio/src/gvars.cpp:77:39: error: invalid types ‘GVarData [9][__gnu_cxx::__enable_if<true, double>::__type {aka double}]’ for array subscript
   uint8_t prec = g_model.gvars[abs(gv)].prec;
                                       ^
radio/src/targets/simu/CMakeFiles/opentx-x7-simulator.dir/build.make:2078: recipe for target 'radio/src/targets/simu/CMakeFiles/opentx-x7-simulator.dir/__/__/gvars.cpp.o' failed
```

Solution http://stackoverflow.com/a/8227799

projectkk2glider:
Fixed in fab3b96815794b9731d776437f026baad59fe85e

PR #4581, opened 2017-03-09
LUA drawline fix

3djc:
This fixes : #4567
Fix and simplify lcd.drawlines. Executing:
```
lcd.clear()
-- outward star
xc = LCD_W / 4
yc = LCD_H / 2
lcd.drawPoint(xc, yc)
lcd.drawLine(xc, yc, xc+10, yc, SOLID, 0)
lcd.drawLine(xc, yc, xc-10, yc, SOLID, 0)
lcd.drawLine(xc, yc, xc, yc+10, SOLID, 0)
lcd.drawLine(xc, yc, xc, yc-10, SOLID, 0)
lcd.drawLine(xc, yc, xc-10, yc - 10, SOLID, 0)
lcd.drawLine(xc, yc, xc+10, yc+10, SOLID, 0)
lcd.drawLine(xc, yc, xc+10, yc-10, SOLID, 0)
lcd.drawLine(xc, yc, xc-10, yc+10, SOLID, 0)

-- inward star
xc = xc + (LCD_W / 2)
lcd.drawPoint(xc, yc)
lcd.drawLine(xc-10, yc, xc, yc, SOLID, 0)
lcd.drawLine(xc+10, yc, xc, yc, SOLID, 0)
lcd.drawLine(xc, yc-10, xc, yc, SOLID, 0)
lcd.drawLine(xc, yc+10, xc, yc, SOLID, 0)
lcd.drawLine(xc-10, yc-10, xc, yc, SOLID, 0)
lcd.drawLine(xc+10, yc+10, xc, yc, SOLID, 0)
lcd.drawLine(xc-10, yc+10, xc, yc, SOLID, 0)
lcd.drawLine(xc+10, yc-10, xc, yc, SOLID, 0)
```
Produces:
image

image

image

PR #4630, opened 2017-03-17
LUA horizontal and vertical lcd.drawLine(...) fix

### LINKED ISSUE #4608: lcd.drawLine(...) gives wrong result
There's still an issue with lcd.drawLine(...) on both the X7 and X9.

I didn't notice before I had a chance to try RC12, but in #4581 the stars are missing one pixel each on the bottom vertical line and the right horizontal line.

Other example of the same problem, these lines should be two pixels long, but are only one pixel long:
`lcd.drawLine(3, 1, 3, 2, SOLID, 0)`
`lcd.drawLine(1, 3, 2, 3, SOLID, 0)`

floaledm:
Fixes #4608 

Compiled and tested the #4581 script on an X9D+

### DIFF
--- a/radio/src/lua/api_lcd.cpp
+++ b/radio/src/lua/api_lcd.cpp
@@ -119,11 +119,11 @@ static int luaLcdDrawLine(lua_State *L)
 
   if (pat == SOLID) {
     if (x1 == x2) {
-      lcdDrawSolidVerticalLine(x1, y1<y2 ? y1 : y2,  y1<y2 ? y2-y1 : y1 - y2, flags);
+      lcdDrawSolidVerticalLine(x1, y1<y2 ? y1 : y2,  y1<y2 ? (y2-y1)+1 : (y1-y2)+1, flags);
       return 0;
     }
     else if (y1 == y2) {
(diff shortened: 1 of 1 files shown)


radio/src/lua/api_lcd.cpp
```diff
@@ -119,11 +119,11 @@ static int luaLcdDrawLine(lua_State *L)
 
   if (pat == SOLID) {
     if (x1 == x2) {
-      lcdDrawSolidVerticalLine(x1, y1<y2 ? y1 : y2,  y1<y2 ? y2-y1 : y1 - y2, flags);
+      lcdDrawSolidVerticalLine(x1, y1<y2 ? y1 : y2,  y1<y2 ? y2-y1+1 : y1-y2+1, flags);
```
3djc: Maybe explicitly set priority in all 4 instances, like  (y1-y2)+1

floaledm: Thanks for reviewing this. Just made your suggested change.

3djc:
Thanks a lot!

PR #4634, opened 2017-03-19
[simulation] Fix some potential crashes & 26 memory leaks, improve LCD redraws.

mpaperno:
A bunch "small" fixes all bunched together... pre-squashed, if you will (I cold break them up but if they'll be squashed anyway... this was easier).  These should probably be included in next 2.2 RC.

  * [simpgmspace] Init trims; Make sure `REa` is really defined (to match board files).
  * [simpgmspace][opentxsimulator] Verify current running state before start/stop; Move rotary enc. init.
  * [simpgmspace][LcdWidget] Improve performance by moving LCD content change check to lcdRefresh() & limiting LcdWidget refresh time to 60 fps max; Ensure thread safety with shared LCD buffer.
  * [simueeprom] Ensure thread could be started, set default running state to false.
  * [simufatfs] Fix paths report trace.
  * [eepromimportexport] Fix memory leaks resulting from import debugging scheme being used.
  * [customdebug] Introduce new scheme for custom debug output in compliance with Qt recommendations (see [docs for QLoggingCategory](http://doc.qt.io/qt-5/qloggingcategory.html)).
  * [opentxeeprom] Fix extra conversion table cache elements being created and also not properly deleted (and hence leaking).
  * [opentxinterface] Unregister EEpromInterfaces in unregisterOpenTxFirmwares();
  * [storage] Unregister storage factories on exit (fixes leak); create virtual StorageFormat/StorageFactory destructors (prevents warnings).
  * [helpers] GVarGroup now emits own signal, no need to pass ModelPanel pointer (removes dependency on modeledit.h)
  * [DebugOutput] Clear simulator trace hook before exiting (prevent possible issues); Fix leak and possible bad QString allocations when reading from buffer; Fix leak with combo box event filter.
  * [TelemetrySimulator] Fix leak by deleting LogPlaybackController object on exit; Only set up data fields once; convert timers to static.
  * [build] Consolidate all Companion/Simulator shared items in `common` library to reduce build time/etc (node & edge still remain awkward).

The leak fixes are result of running standalone Simulator (not Companion) through valgrind & gdb on Linux gcc 4.9.  Moral of this story... don't just automatically blame Qt for memory leaks.  :)

### DIFF
--- a/companion/src/storage/storage.h
+++ b/companion/src/storage/storage.h
@@ -48,14 +48,14 @@ class StorageFormat
       board(Board::BOARD_UNKNOWN)
     {
     }
-    virtual ~StorageFormat() { }
+    virtual ~StorageFormat() {}
     virtual bool load(RadioData & radioData) = 0;
     virtual bool write(const RadioData & radioData) = 0;
 
     QString error() {
       return _error;
     }
-    
+
     QString warning() {
       return _warning;
     }
@@ -84,13 +84,13 @@ class StorageFormat
       qDebug() << qPrintable(QString("[%1] error: %2").arg(name()).arg(error));
       _error = error;
     }
-    
+
     void setWarning(const QString & warning)
     {
       qDebug() << qPrintable(QString("[%1] warning: %2").arg(name()).arg(warning));
       _warning = warning;
     }
-    
+
     QString filename;
     uint8_t version;
     QString _error;
@@ -119,22 +119,22 @@ class DefaultStorageFactory : public StorageFactory
       _name(name)
     {
     }
-    
+
     virtual QString name()
     {
       return _name;
     }
-    
+
     virtual bool probe(const QString & filename)
     {
       return filename.toLower().endsWith("." + _name);
     }
-    
+
     virtual StorageFormat * instance(const QString & filename)
     {
       return new T(filename);
     }
-    
+
     QString _name;
 };
 
@@ -145,19 +145,19 @@ class Storage : public StorageFormat
       StorageFormat(filename)
     {
     }
-    
+
     virtual QString name() { return "storage"; }
-    
+
     void setError(const QString & error)
     {
       _error = error;
     }
-    
+
     void setWarning(const QString & warning)
     {
       _warning = warning;
     }
-    
+
     virtual bool load(RadioData & radioData);
     virtual bool write(const RadioData & radioData);
 };
--- a/companion/src/storage/storage.h
+++ b/companion/src/storage/storage.h
@@ -48,7 +48,7 @@ class StorageFormat
       board(Board::BOARD_UNKNOWN)
     {
     }
-    
+    virtual ~StorageFormat() { }
     virtual bool load(RadioData & radioData) = 0;
     virtual bool write(const RadioData & radioData) = 0;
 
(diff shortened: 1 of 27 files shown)


3djc:
Max, I think it's great. I would like to suggest something tho. If we have issues that drive us to do another RC anyway, let's put that in, but if not, would you be ok to get those after 2,2,0 ?

companion/src/storage/storage.h
```diff
@@ -48,7 +48,7 @@ class StorageFormat
       board(Board::BOARD_UNKNOWN)
     {
     }
-    
+    virtual ~StorageFormat() { }
```
projectkk2glider: This is the only problem I could found :+1:  (space between `{ }`)

mpaperno:
@3djc Sure, whatever works best.   This started by trying to address some of the Simulator crash-on-startup issues, though I'm not sure how much of a practical difference it will make.  

I separately also implemented re-loading of radio libraries (dll/so) for each Simulator run, which works very nicely and I think will have a bigger effect on startup consistency (as discussed on RC).

I did forget one small leak fix which I'll push now, and I think there's also something up with the simulator debug window repeating some trailing characters, so I'm investigating that.

kilrah:
May or may not be related and/or fixed by this but today I got a repeatable crash, if the companion document contains an export error and we try to simulate, since the data was not exported the simu starts with "Data load error" and once the messagebox is dismissed companion crashes,

mpaperno:
> May or may not be related and/or fixed by this but today I got a repeatable crash, if the companion document contains an export error...

Indeed it seems to be related and fixed.  I wondered why exactly, so I looked into it (comparing RC13 tag to this branch).  The behavior was also different with MSVC vs. MinGW (and I imagine it would be yet different on various other toolchains).

> ... and we try to simulate, since the data was not exported the simu starts with "Data load error" and once the messagebox is dismissed companion crashes,

I also got this with RC13 also, much more often with MSVC than MinGW.  When it happens, it's consistently deep inside pthread (both versions) and with no discernible backtrace... almost guaranteed to be memory corruption somewhere.  Judging from the states of the other threads, etc, I'm pretty sure this happens while the simu interface/firmware was being told to stop. 

Mostly on MinGW but also a couple times with MSVC, Companion crashed right away while trying to open a 9xrpro.bin file I had been using to test (this didn't even have a Multi model).  The crash (SIGSEGV) was always in `CurvesField::afterImport()` with what again looked like memory corruption.   Once in maybe a dozen tries it would open fine in MinGW version, and the opposite on MSVC (one crash in ~ dozen tries).

With MinGW I could create a new file with a Multi model and run the simulator -- sometimes it did NOT crash after the popup error message, simulator just closed and everything seemed OK (I could see all the simu threads properly exit).  Until I tried to close Companion... then it crashed, again with some barely traceable memory garbage. 

**The point of this story** is that some memory corruption issues were fixed, most likely due to the bad QString allocations which were happening in the firmware import/export routines -- valgrind really had a fit about those debug streams into what was essentially NULL (including "uninitiated value" errors).  The conversion tables Cache objects were leaking like a sieve (and not actually being a cache), and some other large temporary objects were being left in limbo.   And it was easy to crash the simulated firmware if it was told to start/stop while already running/stopped, which is possibly what was happening when Simulator was exiting unexpectedly.

bsongis:
To be tested in next RC as discussed on the chat

bsongis:
Thanks!

PR #4647, opened 2017-03-20
Proposal to fix #4525

### LINKED ISSUE #4525: In 2.1.9, why is MULTIPLY allowed in the first mix line?
It seem to me that MULTIPLY should automatically be changed to ADD when it winds up in the first mix line of a channel, since multiplying something by zero can never produce a useful result.

Am I missing something, or has it already been changed for 2.2?

bsongis:


kilrah:
Maybe display "MULT!" instead? Probably more explicit...

bsongis:
OK with me

PR #4649, opened 2017-03-21
Tools for 2.2 release handling

3djc:
- for nightlies and release : massively increase the default simu options as per Kilrah's research (maybe windows and mac builds need to be adjusted too)

- set release sdcard  folders as required by : https://github.com/opentx/opentx/commits/kilrah/companion_sdcard_download

- setup release build mechanics


bsongis:
THanks!

PR #4654, opened 2017-03-21
Changed the shutdown delay to 1,5 seconds for Horus.

RealTadango:
With 1.5 seconds delay the shutdown time is exactly the same as my Spektrum DX18. I have been using 1.5 seconds for a while now and it works much nicer.


3djc:
Clearly not

1Sven:
This will be perfect! Nice!!

RealTadango:
Yes but that needs an eeprom change, and will take a long time. Why not create a better delay right now?

kilrah:
Because not everybody likes it the way you do.

PR #4666, opened 2017-03-22
[Simulator] Virtual joystick tweaks.

mpaperno:
Highlights: 
 * Create a reusable RadioTrimWidget (this will be used again later to crate the extra Horus trims); 
 * Prettier trim tab toolbuttons with icons instead of pushbuttons with text.
 * New buttons & icons for axis hold/fix toggles; 
 * Joystick ball is now slightly scaled to available stick area size;
 * Virtual joystick now "follows" mouse clicks within the stick area (this is my favorite). Left-click anywhere inside a joystick widget and the stick node will jump to your mouse cursor and hold/follow until you release. 

The last 3 items apply to trainer simu as well.

vjoy-update


kilrah:
This is way cooll again! 👍

bsongis:
Thanks!

PR #4667, opened 2017-03-22
Allow neg value for telem comparison fonctions

### LINKED ISSUE #4660: Taranis QX 7 Logical Switch Issue
When setting up a logical switch for a variometer you cannot set negative FPS or M/S values. For example I want the vario to active if the sink rate is less than -.5 FPS. It works for positive numbers but not negative. I can set a>x at .1 FPS but I cannot set a<x where -.5 is the value I want, it only allows positive numbers.

3djc:
This fixes #4660

### DIFF
--- a/radio/src/gui/128x64/model_logical_switches.cpp
+++ b/radio/src/gui/128x64/model_logical_switches.cpp
@@ -161,12 +161,10 @@ void menuModelLogicalSwitchOne(event_t event)
           if (v1_val >= MIXSRC_FIRST_TELEM) {
             drawSourceCustomValue(CSWONE_2ND_COLUMN, y, v1_val, convertLswTelemValue(cs), attr|LEFT);
             v2_max = maxTelemValue(v1_val - MIXSRC_FIRST_TELEM + 1);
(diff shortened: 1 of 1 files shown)


radio/src/gui/128x64/model_logical_switches.cpp
```diff
@@ -161,12 +161,10 @@ void menuModelLogicalSwitchOne(event_t event)
           if (v1_val >= MIXSRC_FIRST_TELEM) {
             drawSourceCustomValue(CSWONE_2ND_COLUMN, y, v1_val, convertLswTelemValue(cs), attr|LEFT);
             v2_max = maxTelemValue(v1_val - MIXSRC_FIRST_TELEM + 1);
-            if (cs->func == LS_FUNC_DIFFEGREATER)
-              v2_min = -v2_max;
-            else if (cs->func == LS_FUNC_ADIFFEGREATER)
+            if (cs->func == LS_FUNC_ADIFFEGREATER)
```
bsongis: Not sure this is right on the AVR boards

bsongis:
👍

PR #4670, opened 2017-03-22
[build] Only build companion/simulator resources files once.

mpaperno:
The RCC takes "a while" (precious seconds! ;-) ) to re/build and it was repeating that for each of the two targets.  Funny thing with the variable name though, very strange (see comment @ 165).


mpaperno:
Solved mystery of the strange linking problems (where the Q_INIT_RESOURCE is placed), had nothing to do with the CMake variable name (naturally).  Also found out what Q_INIT_RESOURCE is really for -- only needed when a resource file is in a separate library, to ensure linkage.  :)

PR #4676, opened 2017-03-23
[Companion] Translation system improvements.

### LINKED ISSUE #4637: Missing Qt translation
Hi folks,

When closing Companion without saving the modifications, the popup windows got some fields in English.
I didn't see them with Qt Linguist. A guess there is a Qt "tr(xx)" missing somewhere.
tmp

mpaperno:
* Introduce new Translations class to centralize all related functionality;
  * Add ability to search multiple locations for .qm translation binaries (local folders, resources, system paths);
  * Add ability to reload translations dynamically w/out application restart;
  * Detect and use system language settings to load correct translations by default;
  * Improve support for loading Qt translations (if available);
  * Look for and include pre-built Qt translations with other translation resources (fixes #4637);
  * Generate translations.qrc dynamically with rest of build files;  
  * Add EN "translation" file (allows for simpler UI string edits w/out rebuild and also possible abstractions, eg. use a "filler" string in the source and then add full string in translation file).
  * Make language options menu dynamic based on available translations, and always show native language name; 
  * Allow changing language of some elements w/out restarting (esp. menus/toolbar buttons);

To clarify, with the Qt translations included, all the standard Qt UI elements like button texts and such are translated (elements handled by Qt itself, not the operating system), even if there is no Companion translation for that language.  

This does somewhat significantly increase the size to the generated executable... eg. MSVC Companion gains ~2.2MB and Simulator ~1.6MB, or ~25% increase in both cases.  (One way around this would be to have the translation files installed separately and make an installer with language options... so maybe in some future.)

**Translators** especially can make use of the 2nd feature to test their translations (after generating .qm file).  The new .qm can be placed in either:
  1. In any folder specified via `OPENTX_APP_TRANSLATIONS_PATH` environment variable.
  2. Inside a `translations` folder inside the main Companion folder (wherever Companion is running from);

Either of those takes precedence over the translation compiled into the companion/simulator executables.  The env. variable has highest priority.

I'll also add some changes to the language switching UI in MainWindow so it uses the new static language list and also translates (most of) the UI on-the-fly instead of requiring a restart.


mpaperno:
Oh, also with these changes the default system language is automatically detected at startup and the correct translations loaded (Companion-specific translations only if we have them, of course).  The Companion language option overrides the default system language, including Qt translations but not OS translations.

PR #4677, opened 2017-03-23
simuicons

mhotar:


mpaperno:
Martin, could you post a preview of the images please?  Unfortunately (or not) GH just shows the SVG text.

Also could we get rid of all the unnecessary Inkscape XML in the files (save as "plain SVG")?  This all gets compiled to binary and included in the executables.  Maybe maintain a separate version if any of that metadata is actually useful

Thanks!

mhotar:
snimek obrazovky porizeny 2017-03-23 21-51-42

mhotar:
OK, I am done for today :) Time for my German lessons. Gute Nacht!

mpaperno:
Danke schön!  Those look great!

My one aesthetic suggestion would be to maybe use the radio image inside the circle of "restart radio" icon to make it more obvious and different from the reload scripts icon.

Also if you create any other icons with the `-active` suffix they'll automatically be used for the mouse-over hover effect... eg. `telemetry-active.svg` or `trainer-active.svg`  (I just didn't have those before because of time/laziness/omission).  Same with `-on` except that is only for "checkable" buttons like word wrap, eg. `word_wrap-on.svg` would be shown when ww is enabled (the button is pushed down).

mhotar:
@mpaperno  Max, I still have an isuue with those -on and -active images. While it is all OK with buttons, It does not work well with toggle-buttons. See the fixed_x.svg and fixed_x-on.svg, there is no -active image at all, but it still changes the icon when pointer is hovered over it. While the button is in -on state, hovering the mouse pointer chnages it back to fixed_x.svg.

mpaperno:
Martin, what you describe is a bug in Qt < 5.3 on Linux (or at least Ubuntu).  I noticed it also on my Mint install when building with "native" (and ancient) Qt 5.2.  I think we could solve it by adding another button state -- `-active-on`.  But I'm not sure it's worth it...? since it otherwise works on every other OS and newer Qt versions on Linux.  But if you're up for it, that would be fine with me.

One change would be needed in the code:  
```diff
--- a/companion/src/simulation/simulator.h
+++ b/companion/src/simulation/simulator.h
@@ -79,6 +79,10 @@ namespace Simulator
         if (QFile(addfile).exists())
           addFile(addfile, QSize(), QIcon::Active, QIcon::Off);
 
+        addfile = QString("%1-active-on.svg").arg(baseFile);
+        if (QFile(addfile).exists())
+          addFile(addfile, QSize(), QIcon::Active, QIcon::On);
+
         addfile = QString("%1-disabled.svg").arg(baseFile);
         if (QFile(addfile).exists())
           addFile(addfile, QSize(), QIcon::Disabled, QIcon::Off);
``

mhotar:
Thanks, I am not going to go this -active-on way if It is fixed in newer Qt.
In this case I am finished here, feel free to merge it (In case Travis will not complain :)

mpaperno:
Plain SVG please.

PR #4687, opened 2017-03-25
Updated French translation

LapinFou:
Easily updated French translation.
Thanks a lot to @mpaperno !! It is now so easy for translator. 💯 
See #4637


3djc:
Merci !

PR #4688, opened 2017-03-25
Add firmware update tools location for FreeBSD

tjouan:
Add paths for avrdude and dfu-util programs under FreeBSD. SAM-BA does
not exist yet in the FreeBSD port collection.


schwabe:
Looks good!

Issue #3096, opened 2015-11-23
Speedup the creation of QTableWidget pages


projectkk2glider:
They are used in Oputputs, Logical switches and Special functions since #2861. 

An idea from @griebd https://github.com/opentx/opentx/pull/3091#issuecomment-158797813 is:

> in logsdialog.cpp I tried a lot to improve QTableWidget (logTable) when filling it up. Look at lines 485 forward... 

Not sure it will work here, because each cell has its own widget and layout.

483917:
Hello, i have a laptop Windows 10 pro 32bit ,2Gig ram ,Intel 4x1,46Ghz processor
and with Companion 2.1.6N530 "open/edit" a  model takes 1 - 2 Sec 
but with Companion 2.1.6N550 "open/edit" a model takes 30 - 40 Sec 
this QTableWidget seems to bee way to slow to use,
mabye go back to the previous lists and put some auto moving/resizing Labels above the columns ?

kilrah:
Hmm there's an issue indeed, also takes about 7 sec to open a model edit window for Taranis Plus on Windows.
Seems the issue is mostly on the LS page as each time I click on that I have a 2-sec freeze while SFs and outputs come up instantly. 
Missing lock?

projectkk2glider:
>Missing lock?

I always wondered what was `lock` used for :smile:  Now it makes more sense. This is very possible reason for slowness.

kilrah:
When you change the value of an object (whether manually or programmatically) it triggers its valueChanged method or equivalent. Said method calls update(), which changes values of other objects, which triggers their changed methods, which without lock would call update() again... i.e. changing one thing could cause an insane amount of cascaded calls.

projectkk2glider:
I've added some timing code and came up with this (on my machine opening empty model):

The overall time was  (first time is time taken from the previous printout or from start, second number in [] means cumulative time):
```
"global" "ModelEdit creation" "0 ms [0 ms]" 
  "ModelEdit" "Outputs" "232 ms [359 ms]" 
  "ModelEdit" "LS" "298 ms [657 ms]" 
  "ModelEdit" "CF" "917 ms [1574 ms]" 
  "ModelEdit" "end" "73 ms [1647 ms]" 
"global" "ModelEdit shown" "20 ms [1668 ms]" 
```

Outputs take 230 ms most of the time is spent in the loop adding elements to the QTableWidget, update() takes only 3ms:
```
[...]

PR #4689, opened 2017-03-25
[Companion] Speed up loading of model editor

mpaperno:
By only displaying Logical Switches and Special Functions which are actually used.  I get ~2-3x faster startup times, which is significant.  

Personally I think it looks a bit cleaner as well.  But please correct me if there's actually a reason to show the lines when no function/switch is assigned.  

I have some other changes which help but this was the biggest impact with the least code.   Most of the diff is due to indent changes.  It simply skips a lot of the code now if a function/switch isn't set for a line.

image

image

### DIFF
--- a/companion/src/modeledit/logicalswitches.cpp
+++ b/companion/src/modeledit/logicalswitches.cpp
@@ -214,7 +214,7 @@ void LogicalSwitchesPanel::edited()
 
     if (oldFuncFamily != newFuncFamily) {
       model->logicalSw[i].clear();
-      model->logicalSw[i].func = csw[i]->itemData(csw[i]->currentIndex()).toInt();
+      model->logicalSw[i].func = csw[i]->currentData().toInt();
       if (newFuncFamily == LS_FAMILY_TIMER) {
         model->logicalSw[i].val1 = -119;
         model->logicalSw[i].val2 = -119;
@@ -249,7 +249,7 @@ void LogicalSwitchesPanel::edited()
         updateTimerParam(cswitchOffset[i], model->logicalSw[i].val2, 0.1);
         break;
       case LS_FAMILY_EDGE:
-        cswitchOffset2[i]->setSpecialValueText(tr("(instant)"));
+        cswitchOffset2[i]->setSpecialValueText(" " + tr("(instant)"));
         if (sender() == cswitchOffset[i]) {
           model->logicalSw[i].val2 = TimToVal(cswitchOffset[i]->value());
           updateTimerParam(cswitchOffset[i], model->logicalSw[i].val2, 0.0);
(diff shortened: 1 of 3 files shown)


projectkk2glider:
Brilliant idea, I actually prefer this, because it is more clear which switches are used and which are not.

kilrah:
Nice! It also actually matches the radio!

2-3x seems a LOT... are we sure we don't have an `update()` loop somewhere?

mpaperno:
YMMV  :)

Pretty sure no loop, ya (but not ruling it out).  Drawing all the GUI stuff especially with the massive amounts of options in the switches/sources seems to be the biggest culprit.  Interestingly it's always much faster on my Linux VM than on the parent Windows OS (VM has half the CPUs and way less RAM).  Runs slower on the OSX VM.

I have another change which uses data Models to populate the source/switch combos, which allows the data to be cached (instead of building the list each time).  This also helps a bit but it's still nothing compared to actual GUI drawing, it seems.

companion/src/modeledit/logicalswitches.cpp
```diff
@@ -374,9 +383,12 @@ void LogicalSwitchesPanel::setSwitchWidgetVisibility(int i)
   cswitchOffset2[i]->setVisible(mask & VALUE3_VISIBLE);
   cswitchTOffset[i]->setVisible(mask & VALUE_TO_VISIBLE);
   if (firmware->getCapability(LogicalSwitchesExt)) {
-    cswitchDelay[i]->setEnabled(mask & DELAY_ENABLED);
+    cswitchDuration[i]->setVisible(mask & DURATION_ENABLED);
+    cswitchDelay[i]->setVisible(mask & DELAY_ENABLED);
+    if (mask & DURATION_ENABLED)
     cswitchDuration[i]->setValue(model->logicalSw[i].duration/10.0);
```
mpaperno: I found it first  :)

projectkk2glider:
I did some work on speed in #3096 but never finished.

mpaperno:
Ah, now I see where the timers came from and the table widget option... cool.

I liked your [suggestion](https://github.com/opentx/opentx/issues/3096#issuecomment-160450758) of lazy population of tabs only when actually used.  That was actually my first thought as well to speed up the initial load.  But I had/have no idea what can of worms that may open up....

I just pushed some other changes... [this one](https://github.com/opentx/opentx/commit/f5d0932021bb9ba991273be2bd2ee6fcf02dea8c) switches over to use data models instead of directly populating the combo boxes (which it sounds like maybe you already tried once).  I do see a small improvement but not anywhere as huge as not populating the combos at all in the first place.

@ All:
[This new branch](https://github.com/opentx/opentx/commits/mpaperno/cpn_src_names) incorporates the changes here plus all the changes needed to display custom switch & source names in Companion.

PR #4693, opened 2017-03-25
[Companion] Show custom control names throughout the UI (& internal refactoring)

mpaperno:
Pending comments from @bsongis on the [new `Boards` class](https://github.com/opentx/opentx/commit/663db79433ae1c7b6d6a0fbd3411051427458a63) (comments at bottom), but submitted for your review overall.  This also incorporates the changes in #4689 (speed up loading of model editor) and improves on that.

As far as UI, basically you can now see custom stick/switch/etc names everywhere (unless I missed some place).  Input/mixer lines, all selection lists, model printer, & warnings setup.

Internally, it is now simple to get custom control names because RawSwitch and RawSource are already aware of them.  See https://github.com/opentx/opentx/commit/668a4ec4263d4632a4eb47d81fea247af1f5b18c

Also internally we now use a different method to get data for populating switch and source combo boxes -- instead of writing directly to each CB, the Helpers functions now return a data model, which in turn can be used to populate one or more combo boxes (and optionally cached by whatever requested it).  So instead of building the list over and over for every CB, it is only built a few times and then re-used.  This seems to improve UI response a bit, but is also more flexible for future uses (the data model can be used in any compatible UI control or as a lookup value for a column in another data model).  Details in this commit: https://github.com/opentx/opentx/commit/f5d0932021bb9ba991273be2bd2ee6fcf02dea8c


mpaperno:
image

mpaperno:
Rebased onto `next`.

bsongis:
Thanks!

PR #4694, opened 2017-03-26
Tiny French cosmetic changes

LapinFou:
Added some tiny cosmetic changes in order to help user to make the connection between Companion and the radio menu.
This may (or not) be a good idea. I just got the idea because some recurrent questions/confusions on the French forum.
It is as you want to accept or not. 😄


bsongis:
Pas bête

LapinFou:
Merci. =)

PR #4712, opened 2017-03-29
Make default category translatable (with dev warning)

### LINKED ISSUE #4702: Missing FR translation on the Horus
Hi folks,

When using Companion (French version), the default created category with a Horus profile is always in English.
I guess this one is a special one (common with the radio firmware).
tmp

3djc:
Thix fixes #4702

### DIFF
--- a/companion/src/mdichild.cpp
+++ b/companion/src/mdichild.cpp
@@ -415,7 +415,8 @@ void MdiChild::generalEdit()
 
 void MdiChild::categoryAdd()
 {
-  CategoryData category("New category");
+  /*: Translators do NOT use accent for this, this is the default category name on Horus. */
+  CategoryData category(qPrintable(QObject::tr("Models")));
   radioData.categories.push_back(category);
   setModified();
   emit copyAvailable(false); // workaround : nothing is selected after model creation
--- a/companion/src/mdichild.cpp
+++ b/companion/src/mdichild.cpp
@@ -416,7 +416,7 @@ void MdiChild::generalEdit()
 void MdiChild::categoryAdd()
 {
   /*: Translators do NOT use accent for this, this is the default category name on Horus. */
-  CategoryData category(qPrintable(QObject::tr("Models")));
+  CategoryData category(qPrintable(tr("New category")));
   radioData.categories.push_back(category);
   setModified();
   emit copyAvailable(false); // workaround : nothing is selected after model creation
--- a/companion/src/modeledit/modeledit.cpp
+++ b/companion/src/modeledit/modeledit.cpp
@@ -115,7 +115,7 @@ void ModelEdit::launchSimulation()
   RadioData * simuData = new RadioData();
   simuData->generalSettings = generalSettings;
   /*: Translators do NOT use accent for this, this is the default category name on Horus. */
-  CategoryData category(qPrintable(QObject::tr("Models")));
+  CategoryData category(qPrintable(tr("Models")));
   simuData->categories.push_back(category);
   if (simuData->models.size() == 0) {
     simuData->models.push_back(model);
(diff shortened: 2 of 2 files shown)


LapinFou:
Nice for the translators!!
👍

companion/src/mdichild.cpp
```diff
@@ -415,7 +415,8 @@ void MdiChild::generalEdit()
 
 void MdiChild::categoryAdd()
 {
-  CategoryData category("New category");
+  /*: Translators do NOT use accent for this, this is the default category name on Horus. */
+  CategoryData category(qPrintable(QObject::tr("Models")));
```
mpaperno: Do not qualify with `QObject::` here.  You only need or want that when the code/class you're in doesn't already inherit from QObject.  When you qualify like this, Linguist will group the translations under a generic "QObject" heading (as you've seen before), which is not helpful, instead of with all the other translatable strings in ModelEdit.

`tr()` also accepts translator notes inline.  The better way to use `tr()` outside of QObjects is with [QCoreApplication::translate()](http://doc.qt.io/qt-5/qcoreapplication.html#translate)  which also accepts a "context" name (to group translations by subject/area) instead of putting everything in the default "QObject" translation group.

projectkk2glider: Perhaps you meant `"New category"`

mpaperno:
One thing to watch out for is that the translation system will auto-translate some texts based on existing translations.  So, eg., if the single word "Models" is already translated somewhere else in the code, that will get pulled in automatically during `lrelease` run, accents (if any) and all.

LapinFou:
@mpaperno I started a wiki page to help the translators: [Flow for the Translators](https://github.com/opentx/opentx/wiki/Flow-for-the-Translators)
😉

3djc:
So ideally we need to use something that's not used anywhere else then, is that what's implied ?

mpaperno:
@LapinFou 👍   Thanks, looking good!  This is OT here, but... My one suggestion would be to create a separate page for Git checkout instructions (BTW the code can also be downloaded from GH as a zip file).  Right now Git instructions are repeated multiple times throughout the docs.  It would be much simpler to maintain a single page for that and just link to it from different places.  I like your setup instructions for SourceTree, they seem very thorough.  Though I think it would also be good to include some generic git instructions as well since there are many different clients out there.

@3djc I don't think we need/want to change the name of "Models" I was just bringing it up as a caution.  I'm not sure, but it's possible that adding a "disambiguation" to the `tr()` call would make that translation unique (not subject to auto-heuristics).  I can't seem to find any info on that. There's a `-disable-heuristic` flag for `lrelease` but that would also disable the beneficial ones.

mpaperno:
I'd like to suggest (and have tested) a way to replace `ModelEdit::launchSimulation()` contents entirely with one line.  Should I just use this branch or start a new one?

```
void ModelEdit::launchSimulation()
{
  startSimulation(this, radioData, modelId);
}
```

(But needs to save `radioData` in a member variable in the `ModelEdit()` initializer, which means also can get rid of stored `generalSettings` and `model` references.)

This is exactly the same as using "Simulate Model" from the MdiChild UI, so it loads the whole radio but sets edited model to current.  I'm not sure simulating a model w/out the rest of the radio data makes sense... what if there are global settings which affect the model?

mpaperno:
Or maybe we merge this and I suggest change in new PR.

LapinFou:
@mpaperno OK. Good idea.
First, I will finish what I'm doing (+ correcting all English mistakes). The idea is to describe step by step the procedure to easily make a translation on Windows. Even for the none-computer guy.
 
When it will be done, I will try to clean-up and reorganize this part.

mpaperno:
Sounds good, thanks!  Perhaps I can help add or edit some parts when you're done, just let me know.

LapinFou:
OK. Thanks.

PR #4714, opened 2017-03-29
Instead of copying current profile, start from a blank one

### LINKED ISSUE #4586: Calculated sensor with no name is still saved
(On the Horus discovered, might apply to others)
Add a calculated sensor, choose totalize as function, exit without entering a name. Now there is a new sensor without a name, that is not shown in the sensor list. The sensor can be selected as a source on the Horus. Companion does list it also
image

### LINKED ISSUE #4686: Companion profile behavior incorrect
- When deleting a profile, another one is loaded. The window contents are however still those of the now deleted profile, so exiting with "OK" uverwrites the settigns of the newly selected profile with that of the old.

- When creating a new profile it starts as a copy of the currently selected one (which might have inappropriate settings configured), should start blank / with defaults.

3djc:
Thix fix #4686 

Thix changes current behaviour of copying current profile when creating a new one. Given the differences in radio capabilities and working principles, this protects from carrying not relevant, sometime incompatible data

### DIFF
--- a/companion/src/mainwindow.cpp
+++ b/companion/src/mainwindow.cpp
@@ -1403,15 +1403,12 @@ void MainWindow::updateProfilesActions()
 
 void MainWindow::copyProfile()
 {
-  int i;
-  for (i=0; i<MAX_PROFILES && g.profile[i].existsOnDisk(); i++);
-  if (i==MAX_PROFILES)  //Failed to find free slot
+  int previous = g.id();
+
+  createProfile();
+  if (g.id() == previous) //Failed to find free slot
     return;
-
-  g.profile[i] = g.profile[g.id()];
-  g.profile[i].name(tr("New Radio"));
-
-  g.id(i);
+  g.profile[g.id()] = g.profile[previous];
   updateMenus();
 }
 
--- a/companion/src/mainwindow.cpp
+++ b/companion/src/mainwindow.cpp
@@ -1405,8 +1405,7 @@ void MainWindow::createProfile()
   if (i==MAX_PROFILES)  //Failed to find free slot
     return;
 
-  // Copy current profile to new and give it a new name
-  g.profile[i] = g.profile[g.id()];
+  g.profile[i].init(i);
   g.profile[i].name(tr("New Radio"));
 
   g.id(i);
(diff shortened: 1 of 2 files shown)


mpaperno:
I'm still not sure I understand the underlying issue (or rather how this fixes it).  What happens now when someone switches radio type in their existing profile settings?  Wouldn't creating a new profile and then changing the radio type be exactly the same thing (cause the same problem(s))?  OTOH if I don't want to change radio type for the new profile, I don't have to and all my data is carried over...

3djc:
Because what creates an issue is data relevant to one type of radio been copied to one that does not handle them, then you change type, you can't see them, but those data are still there.

By starting fresh, you enter value you choose , it not done on your back

mpaperno:
The it's the same issue if you change radio type on an existing profile.  Any conversion issues should be fixed there, IMHO.

3djc:
Those should not be mutually exclusive :)

mpaperno:
But this is not a fix.  When/if the radio type of the new (or any) profile is changed, that is when any conversion issues should be addressed (or all the data wiped out if that is really the "best" way to do it).

mpaperno:
What if I just want to create an exact clone of my current profile (for testing or whatever)?  This change just made it a lot more work.  If the conversion/data wipe runs only when radio type is actually changed, no problem (probably with a warning though since there's no "undo" in this case).

3djc:
then we need to also do a proper copy profile function, but this is 'add' not copy

mpaperno:
Semantics aside, the actual problem (right now just with Horus splash images, as I understand it) still happens when changing the radio type, not when creating/copying profiles.  

To quickly fix the Horis issue it would be enough to simply set `profiles[i].splashFile("")` when radio type changes to Horus.

companion/src/mainwindow.cpp
```diff
@@ -1398,21 +1400,35 @@ void MainWindow::updateProfilesActions()
...
+  if (i==MAX_PROFILES)  //Failed to find free slot
+    return;
+
+  g.profile[i].init(i);
+  g.profile[i].name(tr("New Radio"));
+
+  g.id(i);
+  updateMenus();
+}
+
```
mpaperno: This is cool, I like the new option for adding vs. copying.  👍 

But shouldn't `copyProfile()` first call `createProfile()` and then just copy the settings from the old one to new?  To avoid duplicating all that code.

kilrah:
> To quickly fix the Horis issue it would be enough to simply set profiles[i].splashFile("") when radio type changes to Horus.

+1

PR #4720, opened 2017-03-29
[Companion] Improve handling of non-supported splash image (Horus) in radio profile settings page.

mpaperno:
Re: #4686

(Didn't test that the actual flashing works now, but it does wipe the splash image name correctly from profile settings when Horus radio type is selected).


mpaperno:
This also hides/shows the splash image fields whenever the radio type is changed, not only on the initial load.

PR #4721, opened 2017-03-29
[Companion] Change simulator startup from ModelEdit panel.

mpaperno:
Now acts the same as simulating a model from the models list widget using "Simulate Model" menu item (that is, including the other radio data as well, whereas before a default radio data object was created and the model inserted into that).  

Besides simplifying the code, making it more consistent with the other way to simulate a model, and removing the extra category name (re: #4712), I'm not sure it makes sense to simulate a model w/out the rest of the radio data -- what if the model relies on general settings like global functions?


3djc:
I also think that it makes little sense to simulate without corresponding radio settings

kilrah:
Indeed, always assumed it did actually...

PR #4740, opened 2017-04-02
Updated French translation.

LapinFou:
Updated French translation once again... 😉

### DIFF
--- a/companion/src/translations/companion_fr.ts
+++ b/companion/src/translations/companion_fr.ts
@@ -9373,42 +9373,42 @@ Si cette option est cochée, la voie des gaz est inversée: le ralenti  est &apo
         <location filename="../simulation/simulateduiwidget.cpp" line="150"/>
         <source>screenshot</source>
         <comment>Simulator LCD screenshot file name prefix</comment>
-        <translation type="unfinished"></translation>
+        <translation>capture</translation>
     </message>
     <message>
         <location filename="../simulation/simulator_strings.h" line="6"/>
         <source>ENTER</source>
-        <translation type="unfinished"></translation>
+        <translation>Entrer</translation>
     </message>
     <message>
         <location filename="../simulation/simulator_strings.h" line="7"/>
         <source>PG-UP</source>
-        <translation type="unfinished"></translation>
+        <translation>PgPréc</translation>
     </message>
     <message>
         <location filename="../simulation/simulator_strings.h" line="8"/>
         <source>PG-DN</source>
-        <translation type="unfinished"></translation>
+        <translation>PgSuiv</translation>
     </message>
     <message>
         <location filename="../simulation/simulator_strings.h" line="9"/>
         <source>DEL</source>
-        <translation type="unfinished"></translation>
+        <translation>Suppr</translation>
     </message>
     <message>
         <location filename="../simulation/simulator_strings.h" line="10"/>
         <source>BKSP</source>
-        <translation type="unfinished"></translation>
+        <translation>RetArr</translation>
     </message>
     <message>
         <location filename="../simulation/simulator_strings.h" line="11"/>
         <source>ESC</source>
-        <translation></translation>
+        <translation>Échap</translation>
     </message>
     <message>
         <location filename="../simulation/simulator_strings.h" line="12"/>
         <source>INS</source>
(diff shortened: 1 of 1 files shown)


companion/src/translations/companion_fr.ts
```diff
@@ -4340,265 +4320,315 @@ Nous recommandons de lire attentivement les notes de version à l&apos;aide du b
...
+    </message>
+    <message>
+        <location filename="../mainwindow.cpp" line="1469"/>
+        <source>Confirm Delete Profile</source>
+        <translation>Confirmer la suppression du profil</translation>
+    </message>
+    <message>
+        <location filename="../mainwindow.cpp" line="1470"/>
+        <source>Are you sure you wish to delete the &quot;%1&quot; radio profile? There is no way to undo this action!</source>
+        <translation>Etes-vous sûr de vouloir effacer le profil radio %1 ? C&apos;est action est irréversible !</translation>
```
bsongis: Cette

companion/src/translations/companion_fr.ts
```diff
@@ -4340,265 +4320,315 @@ Nous recommandons de lire attentivement les notes de version à l&apos;aide du b
...
+    </message>
+    <message>
+        <location filename="../mainwindow.cpp" line="1090"/>
+        <source>Delete Current Radio Profile</source>
+        <translation>Effacer le profil radio actuel</translation>
+    </message>
+    <message>
+        <location filename="../mainwindow.cpp" line="1090"/>
+        <source>Delete the current Radio Settings Profile</source>
+        <translation>Affacer les réglages radio du profil actuel</translation>
```
bsongis: Effacer

companion/src/translations/companion_fr.ts
```diff
@@ -9328,420 +9370,205 @@ Si cette option est cochée, la voie des gaz est inversée: le ralenti  est &apo
...
+        <location filename="../simulation/simulator_strings.h" line="27"/>
+        <source>&lt;font size=+3&gt;&amp;#x21d1;&lt;/font&gt;</source>
         <translation></translation>
     </message>
     <message>
-        <location filename="../simulation/simulateduiwidgetX9.cpp" line="48"/>
-        <source>-/DN</source>
-        <translation>-/BAS</translation>
+        <location filename="../simulation/simulator_strings.h" line="29"/>
+        <source>&lt;font size=+3&gt;&amp;#x21d5;&lt;/font&gt;</source>
```
bsongis: Those strings should not be translatable!

LapinFou: Do you mean all strings in the `SimulatedUIWidget`?

companion/src/translations/companion_fr.ts
LapinFou: Oups !! Done.

companion/src/translations/companion_fr.ts
LapinFou: Done.

LapinFou:
Updated done.

companion/src/translations/companion_fr.ts
mpaperno: They keyboard keys can/should certainly be translatable (to match local keyboards, if necessary).  They radio key labels should probably match the actual radio hardware (assuming that's not localized in most cases).

LapinFou:
Updated keyboard shortcuts according @mpaperno comment.

companion/src/translations/companion_fr.ts
bsongis: Indeed but I don't think html should be translatable! We can change this later

mpaperno: Perhaps, but the HTML in this case is part of the formatting which is going to depend on the size of the characters used (eg. some use font size +3, others +2, others none, etc).  This way the whole indicator string can be replaced if necessary (eg. with an icon) w/out being tied to a particular font size (or any other attributes included in the HTML).  Granted this possibly requires a bit more "comprehension" on the part of the translator (or at least testing), but OTOH it is a lot more flexible.

LapinFou: Maybe, a `Developer comments` should be added to avoid any confusion?

PR #4749, opened 2017-04-04
X7 : Add GVAR handling similar to X9

### LINKED ISSUE #4744: QX7 openTX N366 Spezial Funktons GVx   Inc/Dec  only +/-1
Hy,
at QX7 I can only set GVx Inc/ Dec +/-1
at X9D, X9E, X12  I can set GVx Inc/ dec with a free value, thats fine for free steps settings

3djc:
Please review thoroughly

This fixes #4744

### DIFF
--- a/radio/src/gui/128x64/model_special_functions.cpp
+++ b/radio/src/gui/128x64/model_special_functions.cpp
@@ -71,6 +71,36 @@ void onCustomFunctionsFileSelectionMenu(const char * result)
 #endif
 
 #if defined(PCBX7)
+
+void onAdjustGvarSourceLongEnterPress(const char * result)
+{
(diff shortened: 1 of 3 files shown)


radio/src/gui/128x64/model_special_functions.cpp
```diff
@@ -71,6 +71,36 @@ void onCustomFunctionsFileSelectionMenu(const char * result)
 #endif
 
 #if defined(PCBX7)
+
+void onAdjustGvarSourceLongEnterPress(const char * result)
+{
+  CustomFunctionData * cfn = &g_model.customFn[menuVerticalPosition];
```
bsongis: I know this code is only written for X7, but I think removing the HEADER_LINE is good

Issue #3334, opened 2016-03-01
add field unit to LUA getFieldInfo


lshems:
Currently there is no way to know in the LUA environment which unit is assigned to a telemetry field. This causes problems when writing scripts for for instance launch altitude announcement.

In 2.0 there was automatic conversion on playValue for instance, which sort of "solved" this problem.

My suggestion would be to add the telemetry field unit, as specified in the telemetry settings, to the getFieldInfo() command.

lshems:
To be more specific:

in 2.1, sensor discovery will assign the units according to the general.imperial setting.
getValue(sensorfield) will return the corresponding value, so meters if set up in meters, and feet if set up in feet.

There is however no way to detect if the user has changed the sensor units after discovery.

So when determining what to play using playValue, or what to show using lcd.drawNumber, one can only guess based on the imperial setting what the sensor units probably are, and act accordingly.

This is actually worse then in 2.0, where at least the conversion was not done at sensor value level, so one could use the imperial setting to know how and if to convert for lcd.drawnumber, and playvalue had automatic conversion. 

by adding the unit number to the getFieldInfo, this could be solved.

lshems:
this would allow to use playValue() as in the special function, where the value unit is added in the announcement.

nigelsheffield:
I think it would make sense that changing the tx to imperial or metric would then change the units too and not only if you rediscover sensors.
Maybe there is some reason to do it like it is now?

lshems:
@nigel

I would prefer having a warning pop-up if changing to or from imperial settings, asking if you want to rediscover sensors.

There may be good reasons to have settings differ per sensor from the global setting.

kilrah:
The global setting is now only used to choose what unit is selected by default when a new sensor is discovered. They can be changed later by the user or even have the same value displayed in more than one unit, so the global setting can't be used anymore.

lshems:
Please leave as default sensor units.

kilrah:
I was talking of removing it from the lua interface, not from the radio. It's now useless to be able to read that from lua. 

> It would allow to choose your settings and just rediscover sensors to be in sync again.

No it doesn't work like that, you'd have to delete all of them and anything that's set up around them before rediscovering.

lshems:
Please leave it in the LUA interface. It is rather usefull to knwo what the users preferences are.

And without a LUA access to the sensor units, this is the best guess you get at knowing what the units are you are getting when doing a getvalue.

lshems:
[...]

PR #4750, opened 2017-04-04
Add unit to lua getFieldInfo

### LINKED ISSUE #4747: LUA Widgets: It is not possible to get the units for sensors
I have a Widget that needs to do some calculations on the vario 'Alt' and 'VSpd'.  The problem is that I need to know if the user set the sensor to feet or meters to do the calculation properly.  I have worked around this by creating 2 Widget options (AltitudeIsInFeet, VSpeedIsInFeet) so that the user can sync up those to Widget option settings to match what they set the sensors to.  This of course requires the user to make those Widget option changes before the sensor works properly.

It would be great to be able to automatically handle this with LUA code so that the user does not have to tell the Widget how they set up units on the sensor (feet vs meters).  Maybe something like getFieldInfo('Alt').units or something like that.  I don't know if that is the right way to handle it... only an idea that I thought of.

I am aware that lcd.drawChannel() will automatically append m, ft, etc but that is only useful if you want to display the value and doesn't help if you want to get the value through getValue() and manipulate it as you don't know if they set the sensor to ft or m.

This request came out of the discussion of another feature request...
https://github.com/opentx/opentx/issues/4741#issuecomment-291209813

3djc:
Allow to query a telemetry sensor unit using lua getFieldInfo()

this closes #4747 #3334

### DIFF
--- a/radio/src/lua/api_general.cpp
+++ b/radio/src/lua/api_general.cpp
@@ -478,6 +478,7 @@ static int luaCrossfireTelemetryPush(lua_State * L)
 Return detailed information about field (source)
 
 The list of valid sources is available:
+
 | OpenTX Version | Radio |
 |----------------|-------|
 | 2.0 | [all](http://downloads-20.open-tx.org/firmware/lua_fields.txt) |
@@ -490,7 +491,7 @@ The list of valid sources is available:
  * `id`   (number) field identifier
  * `name` (string) field name
  * `desc` (string) field description
- * 'unit' (number) unit identifier (https://opentx.gitbooks.io/opentx-2-2-lua-reference-guide/content/appendix/units.html)
+ * 'unit' (number) unit identifier [Full list](../appendix/units.html)
 
 @retval nil the requested field was not found
 
@@ -673,7 +674,7 @@ Play a numerical value (text to speech)
 
 @param value (number) number to play. Value is interpreted as integer.
 
-@param unit (number) unit identifier ((https://opentx.gitbooks.io/opentx-2-2-lua-reference-guide/content/appendix/units.html))
+@param unit (number) unit identifier [Full list]((../appendix/units.html))
 
 @param attributes (unsigned number) possible values:
  * `0 or not present` plays integral part of the number (for a number 123 it plays 123)
@@ -983,7 +984,7 @@ static int luaDefaultStick(lua_State * L)
 
 @param value fed to the sensor
 
-@param unit unit of the sensor.(https://opentx.gitbooks.io/opentx-2-2-lua-reference-guide/content/appendix/units.html)
+@param unit unit of the sensor [Full list](../appendix/units.html)
 
 @param precision the precision of the sensor
  * `0 or not present` no decimal precision.
--- a/radio/src/lua/api_general.cpp
+++ b/radio/src/lua/api_general.cpp
@@ -478,13 +478,11 @@ static int luaCrossfireTelemetryPush(lua_State * L)
 Return detailed information about field (source)
 
 The list of valid sources is available:
-* for OpenTX 2.0.x at http://downloads-20.open-tx.org/firmware/lua_fields.txt
-* for OpenTX 2.1.x at http://downloads-21.open-tx.org/firmware/lua_fields.txt (depreciated)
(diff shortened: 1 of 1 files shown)


radio/src/lua/api_general.cpp
```diff
@@ -578,6 +578,72 @@ static int luaGetValue(lua_State * L)
...
+| 18  | RPM                |
+| 19  | Gee                |
+| 20  | Degrees            |
+| 21  | Radians            |
+| 22  | Milliliters        |
+| 23  | Fluid Ounces       |
+| 24  | Hours              |
+| 25  | Minutes            |
+| 26  | Seconds            |
+
```
schwabe: Maybe add reference to the luaSetTelemetry function whihc also documents the value types?

3djc: Wait, since playValue also documents this, that's now 3 places, we need to put this straight, I'll do a separate PR on that

radio/src/lua/api_general.cpp
```diff
@@ -504,6 +538,11 @@ static int luaGetFieldInfo(lua_State * L)
     lua_pushtableinteger(L, "id", field.id);
     lua_pushtablestring(L, "name", what);
     lua_pushtablestring(L, "desc", field.desc);
+    if (field.id >= MIXSRC_FIRST_TELEM && field.id <= MIXSRC_LAST_TELEM) {
+      div_t qr = div(field.id-MIXSRC_FIRST_TELEM, 3);
```
schwabe: I don't think you need qr

radio/src/lua/api_general.cpp
schwabe: Okay then :)

radio/src/lua/api_general.cpp
```diff
@@ -482,13 +482,17 @@ The list of valid sources is available:
 * for OpenTX 2.1.x at http://downloads-21.open-tx.org/firmware/lua_fields.txt (depreciated)
 * for OpenTX 2.1.x Taranis and Taranis Plus at http://downloads-21.open-tx.org/firmware/lua_fields_taranis.txt
 * for OpenTX 2.1.x Taranis X9E at http://downloads-21.open-tx.org/firmware/lua_fields_taranis_x9e.txt
+* for OpenTX 2.2.x Taranis and Taranis Plus at http://downloads-22.open-tx.org/firmware/lua_fields_x9d.txt
+* for OpenTX 2.2.x Taranis X9E at  http://downloads-22.open-tx.org/firmware/lua_fields_x9e.txt
+* for OpenTX 2.2.x Horus at  http://downloads-22.open-tx.org/firmware/lua_fields_x12s.txt
```
projectkk2glider: Please use the table formatted links for this from the docs of `luaGetValue()` function.

radio/src/lua/api_general.cpp
```diff
@@ -482,13 +482,17 @@ The list of valid sources is available:
...
 
 @param name (string) name of the field
 
 @retval table information about requested field, table elements:
  * `id`   (number) field identifier
  * `name` (string) field name
  * `desc` (string) field description
+ * 'unit' (number) unit identifier (https://opentx.gitbooks.io/opentx-2-2-lua-reference-guide/content/appendix/units.html)
```
projectkk2glider: Proper link is `[units](../appendix/units.md)` and add text `added in 2.2.0`.

Also you should add to the `@status` section notice:
```
@status current Introduced in 2.0.8, filed `unit` added in 2.2.0
```

radio/src/lua/api_general.cpp
bsongis: Yes just an integer division is enough

radio/src/lua/api_general.cpp
```diff
@@ -504,6 +507,13 @@ static int luaGetFieldInfo(lua_State * L)
     lua_pushtableinteger(L, "id", field.id);
     lua_pushtablestring(L, "name", what);
     lua_pushtablestring(L, "desc", field.desc);
+    if (field.id >= MIXSRC_FIRST_TELEM && field.id <= MIXSRC_LAST_TELEM) {
+      TelemetrySensor & telemetrySensor = g_model.telemetrySensors[(int)((field.id-MIXSRC_FIRST_TELEM)/3)];
```
bsongis: Why a cast here?

PR #4755, opened 2017-04-05
Horus : adjust INVERS background for all text sizes

### LINKED ISSUE #4751: [Horus] Inverted text is not displayed correctly at MID and DBL Size
The parameter INVERS doesn't have the correct dimensions of the background area for the font size MID and DBL.

text_invers

3djc:
This fixes #4751 

image


3djc:
What I have done currently is math the style that was already in place (ie top and bottom space depends on font size), but my personal preference would be one pixel top and bottom, much like left and right, but I would appreciate more comment before changing that

bsongis:
SMLSIZE seems OK, but the others have more padding on top / bottom than left / right
But I am not sure that you won't encounter problems if you change it, it needs to be tried

3djc:
image

This is newer version, much cleaner in all aspects I think

PR #4758, opened 2017-04-06
Check for confirming inversion of throttle channel is no more possibl…

### LINKED ISSUE #4757: Horus/N366 Can't reverse direction of output
A user of one of my sailplane setups found this issue which I can also reproduce on my production Horus (mode 2) running N366

1. Go to Outputs menu
2. Highlight CH4
3. Skip to Direction col
4. Press Enter to change direction.  

At this point, a throttle reverse confirmation dialog appears with two options Enter and Rtn.

5. Press Enter to confirm

Expected: arrow changes direction
Actual: arrow remains unchanged.

I tried altering the Throttle.Src in Model.Setup from Thr to some other control but no difference.

As an aside, I'm puzzled as to why OpenTx thinks that CH4 is associated with the throttle stick as there is no explicit assignment in my setup.

kilrah:
…e since Virtual Inputs got introduced (Fix #4757)

### DIFF
--- a/radio/src/gui/128x64/model_outputs_arm.cpp
+++ b/radio/src/gui/128x64/model_outputs_arm.cpp
@@ -20,16 +20,6 @@
 
 #include "opentx.h"
 
-bool isThrottleOutput(uint8_t ch)
-{
-  for (uint8_t i=0; i<MAX_MIXERS; i++) {
-    MixData *mix = mixAddress(i);
-    if (mix->destCh==ch && mix->srcRaw==MIXSRC_Thr)
-      return true;
-  }
-  return false;
-}
-
 enum MenuModelOutputsItems {
   ITEM_OUTPUTS_OFFSET,
   ITEM_OUTPUTS_MIN,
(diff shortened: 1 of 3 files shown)


radio/src/gui/128x64/model_outputs_arm.cpp
```diff
@@ -156,17 +146,10 @@ void menuModelLimitsOne(event_t event)
...
       case ITEM_OUTPUTONE_DIR:
       {
-        uint8_t revert = ld->revert;
         lcdDrawTextAlignedLeft(y, TR_LIMITS_HEADERS_DIRECTION);
-        lcdDrawTextAtIndex(LIMITS_ONE_2ND_COLUMN, y, STR_MMMINV, revert, attr);
+        lcdDrawTextAtIndex(LIMITS_ONE_2ND_COLUMN, y, STR_MMMINV, ld->revert, attr);
         if (active) {
-          uint8_t revert_new = checkIncDecModel(event, revert, 0, 1);
-          if (checkIncDec_Ret && isThrottleOutput(k)) {
-            POPUP_CONFIRMATION(STR_INVERT_THR);
```
bsongis: Is the string still needed elsewhere?

kilrah: Yup, AVR :)

PR #4769, opened 2017-04-10
[Companion] Restore model drag/drop/compare, many UI improvements to main and sub-windows.

mpaperno:
* Refactor MdiChild & TreeModel, restore drag/drop, and more:
  * Add full drag-and-drop functions to the models list, including model copy, move, and insert actions (also between windows);
  * Add general radio settings copy/paste across file windows;
  * Full MIME support with distinct types for various data (eg. models vs. general settings), also add new type for metadata (header);
  * Copied/cut models can now be inserted in any spot (similar to drop-insert);
  * Deleted/cut models can now be removed entirely (including the empty slot) for all radio types (any models after the removed one move up, like for Horus); This can be disabled via new option in app settings;
  * Default model can be deleted (new default is picked);
  * Completely revamp the actions system (copy/paste/edit/etc) to make it fully functional:
      * Actions exist in the window, not just context menu, therefore all keyboard shortcuts actually work, w/out any duplicated code;
      * Provides access from MainWindow in a comprehensive manner: all actions are available and properly updated based on selections in current editor window, w/out awkward event monitoring;
      * More actions available from context menu, and somewhat reorganized;
      * Also show all actions on mini toolbars right in the MDI window, which can also be hidden;
      * Updated English localization file to account for plurals usage in new action item names (other languages need to be updated);
  * Keyboard navigation in the model list now works/practical;
  * MDI window uses all available height by default, saves/restores last used size (or maximized state);
  * Newly added models/categories are selected and highlighted; Category names are immediately editable; New option to launch model editor instead of wizard (or nothing);
 * MainWindow updates:
      * Edit toolbar and menu now showing all available edit actions, with proper context;
      * File menu properly updates with all "general" radio actions: edit/copy/paste settings and simulate radio;
      * Added Window menu showing list of open windows (eg. to switch when full screen) and tabbed window option;
      * Added some more shortcuts and actions (eg. close current file window).
      * Immediate update of icon size preference;
* Model Compare dialog:
  * Restore model compare functionality by using new MIME data exchange system; 
  * Add ability to compare any number of models at a time; 
  * Multi-model printer now shows custom hardware settings for each dropped model (even from different radios).
  * Remember size/position of compare window between uses.
* Misc.
  * Model wizard starts with default model name populated.
  * Properly re-translate UI instead of recreating all the actions/etc.; 
  * Add missing monoblue & monowhite theme sdsync icons.
  * Tell Mac specifically where to put some of our actions (seems to make it happier) -- eg. there was no main app settings action in the menus;  (Though it still wants to put an "Emojis" option in our Edit menu... bleh.)


mpaperno:
image

image

image

mpaperno:
Yes, new translator "challenge" is singular vs. plural forms for some of the menu items.  :)   But most of the current translations (for menu items and such) are preserved already, even with all the code moving around, which is nice.

image

LapinFou:
Good to know. 😄

PR #4771, opened 2017-04-11
Updated FR translation for Companion 2.2.0N368

LapinFou:
Updated the French translation.

I got a doubt (see snapshot below), if the field `New model` can support or not foreign accent. There is no developer comment... For now, I didn't change it.
tmp

### DIFF
--- a/companion/src/translations/companion_fr.ts
+++ b/companion/src/translations/companion_fr.ts
@@ -4,27 +4,27 @@
 <context>
     <name>AileronsPage</name>
     <message>
-        <location filename="../wizarddialog.cpp" line="432"/>
+        <location filename="../wizarddialog.cpp" line="437"/>
         <source>No</source>
         <translation>Non</translation>
     </message>
     <message>
-        <location filename="../wizarddialog.cpp" line="433"/>
+        <location filename="../wizarddialog.cpp" line="438"/>
         <source>Yes, controlled by a single channel</source>
         <translation>Oui, contrôlés par une seule voie</translation>
     </message>
     <message>
-        <location filename="../wizarddialog.cpp" line="434"/>
+        <location filename="../wizarddialog.cpp" line="439"/>
         <source>Yes, controlled by two channels</source>
         <translation>Oui, contrôlés par 2 voies</translation>
     </message>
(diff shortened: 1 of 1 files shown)


LapinFou:
Ooops!! I think I've some bad.
The 1st branch I pushed was *LapinFou/N386_Fr_Translation*. When I realized I've done a typo, I deleted the 1st one and push a new branch *LapinFou/N368_Fr_Translation*.
However, the Travis scripts tried to run on the deleted branch *LapinFou/N386_Fr_Translation*.
I don't have the skills to clean-up this mess... Sorry guys.

LapinFou:
False alert, finally Travis CI did see the branch swap. 😄

mpaperno:
> I got a doubt (see snapshot below), if the field New model can support or not foreign accent.

Ah yes, I meant to add a comment there but forgot with all the other changes.  Actually there are (at least) 3 places where a default model name is created, and for some reason Horus and non-Horus use different default names  (eg. for non-Horus it's "MODELXX" and Horus is just "Model").  Some of this could/should probably be cleaned up, but I wasn't sure why the names are treated differently to begin with.

3djc:
On taranis, models have a number,  as there are numbered model slots, but this doesn't exist on Horus, due to the very different way model are stored, hence the difference.

mpaperno:
Well there are still "slots" in the background (array indexes) they are just not numbered on Horus so user can see.  But for example if you have an "unnamed" Horus model "Model1" and move it to 3rd place, the name gets updated to "Model3".  So it's still tracking the position for some reason.  If you create multiple new models on Horus they all have the same default name... but other radios you get the automatic number appended (which at least helps ID the different models if you don't change name).  I just think it can be consistent between radios and we should only need to provide default model name in one place in the code, where the other defaults are set  -- https://github.com/opentx/opentx/blob/next/companion/src/eeprominterface.cpp#L1623  (that one place could make exceptions based on type of radio, of course).

3djc:
I made a recent change to Horus to actually display that number, but that was driven by the introduction of Companion and how was handling import from X9. Before that, that wasn't there at all

PR #4790, opened 2017-04-13
Fix key variable not initialised

schwabe:
@mpaperno you touched that function, can you approve or make a proper fix?


radio/src/targets/simu/opentxsimulator.cpp
```diff
@@ -284,6 +284,9 @@ void OpenTxSimulator::rotaryEncoderEvent(int steps)
     key = KEY_MINUS;
   else if (steps < 0)
     key = KEY_PLUS;
+  else
+   // Should not happen but Clang complains that key is unset otherwise
```
bsongis: wrong indentation

PR #4793, opened 2017-04-14
X9E : better manage additional switches declared as toggle or 2pos

### LINKED ISSUE #4650: Better handling of additional toggle switches
Currently, toggle switches are well handled ONLY if they are in the predefined slot.

If a toggle switch is installed in an additional slot or replacing an existing 3 pos, there are issues, mostly in autodetect, but in other places too::

- if the switch is physically an 3 pins inverter it works (position is either up or down)

- if the switch is made of a 2 pins toggle between pin1 and pin2, and a resistor between 2 and 3 : position is seen between up and middle (and this creates issue in companion)

- if the switch is made of a 2 pins toggle between pin1 and pin2 and nothing else : position is seen between up and middle SF

A working fix to to declare the switches as 2POS in the source code as I have done for my radio (https://github.com/opentx/opentx/commit/ff8e74bb4c4c8022a0172c3052c5c8a22512cfd8) but that obviously is not a proper solution.

Creating this issue to foster discussion on the subject

3djc:
This fixes #4650 

Successfully tested with:

image

with the pull down resistor present or not.

WARNING : this change will require X9E owners using extra 2pos or toggle switches to review their settings. Previous version (2.1 or 2.2) where reading incorrectly those switches toggling between middle (which should not exist for a 2pos/toggle) and either up or down (depending on how they are wired). This was generally not seen as an issue because Companion was not able to display/manage those, and the functionality seemed ok

### DIFF
--- a/radio/src/targets/taranis/keys_driver.cpp
+++ b/radio/src/targets/taranis/keys_driver.cpp
@@ -2,7 +2,7 @@
  * Copyright (C) OpenTX
  *
  * Based on code named
- *   th9x - http://code.google.com/p/th9x 
+ *   th9x - http://code.google.com/p/th9x
  *   er9x - http://code.google.com/p/er9x
  *   gruvin9x - http://code.google.com/p/gruvin9x
  *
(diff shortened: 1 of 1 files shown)


bsongis:
Does it change something for people who didn't modify their radio (I mean SF = 2-POS and SH = 2-POS-TOGGLE)?

3djc:
No, those are hard-coded 2pos (https://github.com/opentx/opentx/blob/3djc/x9e-additional-switches/radio/src/targets/taranis/keys_driver.cpp#L171). Nothing changes for the default switches

bsongis:
👍

PR #4808, opened 2017-04-16
Multimodule has a pulldown resistor and does not work with open drain…

schwabe:
…, use push pull instead

### DIFF
--- a/radio/src/targets/sky9x/pulses_driver.cpp
+++ b/radio/src/targets/sky9x/pulses_driver.cpp
@@ -163,21 +163,25 @@ void disable_ppm(uint32_t port)
 // TD is on PA17, peripheral A
 void init_ssc(uint8_t baudrateDiv1000)
 {
-  Ssc * sscptr;
+  Ssc *sscptr;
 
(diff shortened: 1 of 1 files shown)


radio/src/targets/sky9x/pulses_driver.cpp
```diff
@@ -163,21 +163,25 @@ void disable_ppm(uint32_t port)
...
-  sscptr->SSC_CMR = Master_frequency / (1000*baudrateDiv1000*2);
+  sscptr->SSC_CMR = Master_frequency / (1000 * baudrateDiv1000 * 2);
   sscptr->SSC_TCMR = 0;
   sscptr->SSC_CR = SSC_CR_TXEN;
 
 #if defined(REVX)
-  PIOA->PIO_MDER = PIO_PA17;						// Open Drain O/p in A17
+  if (IS_MODULE_MULTIMODULE(EXTERNAL_MODULE)) {
+    PIOA->PIO_MDDR = PIO_PA17;                 // Push Pull O/p in A17
+  } else {
```
bsongis: `else {` should go to the next line

PR #4818, opened 2017-04-18
Fix DS3231 RTC read/write on AR9X

sreichholf:
There was some feedback recently on the original issue stating that the rtc is broken on ar9x.
This fixes the issue by properly initializing the address buffer to {0,0} (plus i've changed the define to CHIPID instead of ADDR to avoid confusion).


sreichholf:
Just discovered some issues with setting the time, will have to investigate further before merge...

schwabe:
The original version worked because 0x0 just happend to have 0 as content
then I broke it with using an unitialised array and now it is fixed again by having the array to contain 0.

PR #4819, opened 2017-04-18
Fix DS3231 RTC read/write on AR9X

sreichholf:
2nd try...
Fixes DS3231 RTC read/write on AR9X


schwabe:
Looks good!

schwabe:
The original version worked because 0x0 just happend to have 0 as content.
Then I broke it with using an unitialised array and the fix initiliases the array with 0 again

PR #4835, opened 2017-04-21
X7 : fix calibration interuption

### LINKED ISSUE #4833: X7 calibration interuption issue 2.2N369
2.2N369
X7 calibration - long menu long page - calibration menu. 
press enter to start - pressed
press enter when stick and pots in center - pressed
now not continue press exit it skip to home screen 
process is canceled
from this time is not posible to use page to browse the main setting radio menu or model menu. page button not do any action. 

Radio must be restarted to make it work again.

3djc:
It does seem to do the job, but not overly happy with the fix

This fixes #4833

### DIFF
--- a/radio/src/gui/128x64/navigation.cpp
+++ b/radio/src/gui/128x64/navigation.cpp
@@ -633,14 +633,10 @@ void check(event_t event, uint8_t curr, const MenuHandlerFunc * menuTab, uint8_t
 
     case EVT_KEY_LONG(KEY_EXIT):
       s_editMode = 0; // TODO needed? we call ENTRY_UP after which does the same
-      reusableBuffer.calib.state = 0;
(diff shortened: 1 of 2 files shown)


bsongis:
Do we have the same problem on X9D? on X12S?

3djc:
Nop

bsongis:
Do we have the same hack for those other boards?

3djc:
No, it's much cleaner on those other boards, but there is already something 'odd', since menuCalibrationState is mentionned only once in x9/x12 navigation routine, and already 3 in x7 before this fix.

So do I like this fix : no, but I was not able to come with a better one, but I'm sue I know someone who can ;)

bsongis:
Excellent, it looks like others boards code now!

PR #4850, opened 2017-04-24
X7 : Differentiate message for external flashing

### LINKED ISSUE #4847: New Feature: Use IS_PCBREV_40 in error message of flashing on X7D
The number of people trying to flash their sport rxs with the module is becoming more and more. As idea to slience some of these questions:

We have the IS_PCBREV_40 function which tells the revision of the X7. When flashing fails add a text.

if (IS_PCBREV_40), add to the error message
- Make sure module is connected to bottom sport connector, *NOT* to the module bay pins.

3djc:
Use hardware revision to help troubleshoot user error.

This fixes #4847


3djc:
Another alternative is : "Bottom SPort error" instead of "Bottom pin no resp"

PR #4855, opened 2017-04-24
Projectkk2glider/issue 4827

### LINKED ISSUE #4827: Companion2.20N369 Simulator crash with a multidimensional array
When using a multidimensional array in a lua-script, the simulator crashes when the script is called. 
When the script is called at TX Horus, the Horus goes into the emergency mode.

Q: Is the array to big? If I restrict the array to 2 value series, then the script doesn't crash.

Example Array:
local freqLookup = {}
    freqLookup[1] = { 5865, 5845, 5825, 5805, 5785, 5765, 5745, 5725 }
    freqLookup[2] = { 5733, 5752, 5771, 5790, 5809, 5828, 5847, 5866 }
    freqLookup[3] = { 5705, 5685, 5665, 5645, 5885, 5905, 5925, 5945 }
    freqLookup[4] = { 5740, 5760, 5780, 5800, 5820, 5840, 5860, 5880 }
    freqLookup[5] = { 5658, 5695, 5732, 5769, 5806, 5843, 5880, 5917 }

projectkk2glider:
Closes #4827 

This reverts part of commit f94b4c7 and restores back the original table hash size calculation. The code before this allowed only 265 hash nodes.

The code now potentially uses more RAM (because it allocates node array in bigger steps), but at the same time there are a lot less node allocations and copying of existing nodes in the new array. Previously each node number increase caused the whole malloc/copy/free process, now it only happens when size overflows a power of two (ie 1, 2, 4, 8, 16, 32, etc,...). So there are a lot less calls to the `rehash()` (which grows a table).

### DIFF
--- a/radio/src/debug.h
+++ b/radio/src/debug.h
@@ -70,8 +70,18 @@ extern "C" {
 
 #if defined(TRACE_LUA_INTERNALS_ENABLED)
   #define TRACE_LUA_INTERNALS(f_, ...)     debugPrintf(("[LUA INT] " f_ "\r\n"), ##__VA_ARGS__)
+
+  #define TRACE_LUA_INTERNALS_WITH_LINEINFO(L, f_, ...)   do { \
+                                                            lua_Debug ar; \
+                                                            if (lua_getstack(L, 1, &ar)) {  \
+                                                              lua_getinfo(L, ">Sl", &ar); \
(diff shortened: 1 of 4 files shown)


bsongis:
Thanks!
Just the RAM issue, I hope that both S6R and CROSSFIRE scripts will be still OK

PR #4856, opened 2017-04-24
Proposal to fix #4831 (CZ TTS problem with PREC1 and units)

### LINKED ISSUE #4831: Czech speech audio issue 2.2N369
There is issue in 2.2N 364-369. Selected czech sound pack not reproduce the values well. 
the value 10.7V should say "ten point seven volts" 
we got only  "ten seven"

English audio works well. Tested on X7, X9E 

On SD we have sthe sound files well and inclide point, volts ....

bsongis:


mhotar:
10 volts is OK:
playFile("/SOUNDS/cz/SYSTEM/0010.wav", flags=0, id=1)
playFile("/SOUNDS/cz/SYSTEM/volt2.wav", flags=0, id=1)

but 10.1volts is not:
playFile("/SOUNDS/cz/SYSTEM/0010.wav", flags=0, id=1)
pushUnit: out of bounds unit : 114
playFile("/SOUNDS/cz/SYSTEM/0001.wav", flags=0, id=1)
playFile("/SOUNDS/cz/SYSTEM/mamp0.wav", flags=0, id=1)

bsongis:
mamp0.wav, then we need another -1 to play volts instead of amps, right?

3djc:
Seems ok to me

bsongis:
Just this big switch ... we should replace the values by constants!

radio/src/translations/tts_cz.cpp
```diff
@@ -42,9 +42,9 @@ enum CzechPrompts {
 #if defined(VOICE)
 
 #if defined(CPUARM)
-  #define CZ_PUSH_UNIT_PROMPT(p, u) cz_pushUnitPrompt((u), (p), id)
+  #define CZ_PUSH_UNIT_PROMPT(p, u) cz_pushUnitPrompt((p), (u), id)
```
bsongis: it should be u first, no? (on both sides!)

bsongis:
:+1 Thanks!

PR #4866, opened 2017-04-27
3djc/fix wrong fix

3djc:
Fix for #4814 was (very very) wrong

Appologies


schwabe:
That fix looks much better. My clang compiler is now happy too.

3djc:
wrong wrong wrong

PR #4874, opened 2017-04-28
[Companion] Show custom channel, G-Var, and Curve names throughout UI.

mpaperno:
And only a little refactoring  ;-)  More changes on tap, based on these.


mpaperno:
Also propose to rename logical switches to "LSw##" in selections lists to avoid confusion with "L#" and "LS" hardware controls.

PR #4883, opened 2017-04-30
Fix server compile

3djc:
Fixes opentx-9x-autoswitch-autosource-dblkeys-battgraph-telemetrez-sp22-stickrev-pgbar-en.hex

TELEMETREZ handling is done line 113


zmoothrc:
stickrev don't seem to work anymore on 2.2.0 & 9X

PR #4884, opened 2017-04-30
Fix #4881

### LINKED ISSUE #4881: Possible incorrect external PXX pulses
In pxx_arm.cpp, the pulse duration is set to either 48 or 32. This value gets put into the ARR register of the timer (via DMA). Counters count from 0 to the ARR value, then go back to 0. This will result in either 49 or 33 counts for the pulse width, thus making the pulses 0.5uS too long.
The "rest" value is calculated using one less than the duration value, so the total period will also be incorrect.
The duration should be set to 47 or 31, with the "rest" value being decremented by (duration+1).

This may also apply to serial protocols (DSM and Multi). Worth checking all with an oscilloscope.

schwabe:
Thanks Mike for noticing: In pxx_arm.cpp, the pulse duration is set to either 48 or 32. This value gets put into the ARR register of the timer (via DMA). Counters count from 0 to the ARR value, then go back to 0. This will result in either 49 or 33 counts for the pulse width, thus making the pulses 0.5uS too long.

I tested only with external PXX with my logic analyzer but this code also affects internal pxx.

### DIFF
--- a/radio/src/pulses/pxx_arm.cpp
+++ b/radio/src/pulses/pxx_arm.cpp
@@ -134,7 +134,7 @@ void pxxPutPcmTail(uint8_t port)
 #else
 void pxxPutPcmPart(uint8_t port, uint8_t value)
 {
-  pulse_duration_t duration = value ? 48 : 32;
+  pulse_duration_t duration = value ? 47 : 31;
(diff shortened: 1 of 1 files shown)


bsongis:
Thanks both!

Issue #4879, opened 2017-04-30
[Cosmetic issue] Input are not displayed correctly in mixes tab


LapinFou:
With the N371, the inputs are not displayed properly.
It is `:Input` or `[Ix]:Input` instead `(input)` or `[Ix]Input`
![tmp](https://cloud.githubusercontent.com/assets/1685030/25563263/6f9495e4-2d98-11e7-9992-970eb53dd623.png)

mpaperno:
Why is this not "proper?"  The inputs were always displayed with the `[ ]` brackets (presumably to mimic the inverted appearance in the radio UI).  I added a colon to a) make it easier to read (especially in the selector lists) and b) make it more consistent with how the other sources are shown (eg. `CH01:Name`).  

The channels were never in brackets, but some separator was needed to show the name.  Parenthesis are a) extra characters (3, with the space) and b) sometimes harder to read because in proportional fonts they align very closely with the characters they surround (again, especially in selector lists).

I'm open to opinions and changing it to whatever most people prefer, but the above is my rationale for the changes.

Now, looking at the actual Inputs screen (Entrées) I do see that there in no colon in the left column.  So maybe one should be added there to make it more consistent.  Is that what you meant?

![image](https://cloud.githubusercontent.com/assets/1366615/25563541/e7dbdb34-2d6b-11e7-8dfd-2582a7540ebf.png)

LapinFou:
Sorry, my explanation wasn't accurate enough.
What I mean is with the N371 the display in **Mixes** (not **Inputs**) is not as before.

kilrah:
It's not as before becasue he added the `:`...

mpaperno:
"There can be no progress without change."  :)   The question is, is it better this way?  

A colon is also used in the mixer lines to denote replacement, but IMHO it still looks OK (replacement could just be `=` but kinda late to change it now I guess).  

The other advantage of the changes is that there is only one place in the code where the channel (etc.) display names are defined (less to translate/maintain  ;-) ). Except now I see where I missed the input name line printer (for Inputs screen), so I should fix that to be consistent (either way).

kilrah:
I actually think it's clearer as it was before without the `:`...

mpaperno:
Even in the dropdown selector lists?

kilrah:
Yes, I don't see them add anything, just distracting..

mpaperno:
[...]

PR #4888, opened 2017-05-03
[Companion] Input names formatting 

mpaperno:
Re: #4879 

With colons after input number:
image
image

Without colons (current version of this code):
image
image

(Those are from someone's actual radio file.)

I'd rather just pick a version and stick to it, vs. making exceptions in the code for the different view types.  If we're to have exceptions then they should at least be systematic, eg. a "verbose" view vs. "short" view, or whatever.  IMHO.

### DIFF
--- a/companion/src/eeprominterface.cpp
+++ b/companion/src/eeprominterface.cpp
@@ -508,7 +508,7 @@ QString RawSource::toString(const ModelData * model, const GeneralSettings * con
   int genAryIdx = 0;
   switch (type) {
     case SOURCE_TYPE_VIRTUAL_INPUT:
-      result = QObject::tr("[I%1]").arg(index+1);
+      result = QObject::tr("[I%1]").arg(index+1, 2, 10, QChar('0'));
       if (model && model->inputNames[index][0])
           result.append(":" + QString(model->inputNames[index]).trimmed());
       return result;
@@ -550,7 +550,7 @@ QString RawSource::toString(const ModelData * model, const GeneralSettings * con
       return QObject::tr("CYC%1").arg(index+1);
 
     case SOURCE_TYPE_PPM:
-      return QObject::tr("TR%1").arg(index+1);
+      return QObject::tr("TR%1").arg(index+1, 2, 10, QChar('0'));
 
     case SOURCE_TYPE_CH:
       result = QObject::tr("CH%1").arg(index+1, 2, 10, QChar('0'));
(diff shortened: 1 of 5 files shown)


bsongis:
I would like to keep **[Ixx] Name**
And change to **[CHxx] Name**

mpaperno:
> I would like to keep **[Ixx] Name**

You want to *add* a space there?

Adding length to the channel number will change some layouts (eg. Simulator outputs screen again, maybe others places) and add length to the input/mixer lines... could be a(nother) can of worms.

bsongis:
No, sorry, no space needed there:
**[Ixx]Name**
**[CHxx]Name**

kilrah:
How about having the space only in the dropdowns where everything is small and tight?

mpaperno:
I'm not sure what you guys have against a separator.  Even in your reply it's not as easy to read as with a separator.

**[I11]Name**
**[CH11]Name**

**[I11]:Name**
**[CH11]:Name**

Anyway, I personally don't want to touch the global channel name format right now for reasons mentioned above.

> How about having the space only in the dropdowns where everything is small and tight?

As mentioned, I'm trying to avoid exceptions in the code where possible (central place to change & translate things) and have a consistent display throughout.  But yes we could have conditional formatting... in next version.

Honestly I didn't think it would be so controversial... might be easier to revert the whole change for now and do a design committee later.  The GVars and curves also follow the `NNxx:Name` format.... seems nice to pick a convention and stick to it.  I'd also get rid of the `[ ]` around `Ixx` entirely, but I figured such a drastic change would bring protests (and `I` is a nasty character to have next to a digit).

**IN11:Name**
**CH11:Name**

bsongis:
I think a space is nice when using a proportional font, but not when we are short in horizontal space.
I don't have any good reason, but I don't like too much the ":" separator

mpaperno:
Not sure what to do here.  I think spaces between the ID## and custom name are vague in the input/mixer lines (the body, not the left column) because it's not always clear what the name belongs to (to the left source or the right?).  They work better in selector lists where each item is on one line anyway, but still I prefer a definitive (unambiguous) separator (I did try spaces at first, as mentioned).  The [brackets] don't do it for me at all, they look too much like I/l/1/L and crowd whatever they surround (esp. in tight propo. font).  The colon is practically the only choice AFAICT since it is small, clear, typically used in such a role already, and doesn't carry any mathematical significance.

Should I revert the input/mixer lines to the way they were (with `CH# (name)` format), or....?  I think that's the main/only difference from how they were formatted before the colon-gate.  :)   Or someone else can take a stab at it, I won't take it personally.

elecpower:
I haven't looked at every use case but conceptually agree with @mpaperno.
+1 for colons and removing brackets. Referencing INnn sources differentiates from raw hardware inputs.
-1 for LSwnn as the 'w' visually inconsistent. Understand why to differentiate from LS, RS. Maybe down the track change Logical Switches to Programmed/Virtual/? Switches PS/VS/?S and yes understand how significant that is to coding, users and documentation.

bsongis:
We will discuss and improve things later, thanks @mpaperno for this PR

PR #4889, opened 2017-05-03
Italian audio timers

### LINKED ISSUE #4186: Horus OpenTX 2.2 N360 IT voice error
Hello i see that on Horus with last N360 build and the correct SD card content there is a problem with IT language voice. For example in the timer if i have 5:51 the voice say 5 minutes and 5 second so it will read the decimal value as second and the not the real value in this case it would be fiftyone seconds. Also there are some error in the traduction like minutes instead of minuti and seconds instead of secondi.

I picked the csv file in the SD and i make my own traduction and created the language pack with TTSAutomate but the errors are still present even if now with the right spelling so seconds are now secondi.
Also on spalsh screen if italian is selected the horus hasn't sound instead in english it say welcome to opentx. 

If i use english voice there is no problem and the values are read in the right way.
I hope you understand the problem

3djc:
- generate proper translation for hours minutes and seconds (this WILL require a sdcard version increment)

- Switch the time mechanic to english one, with italian words. It is likely not perfect, but probably usable, while current italian audio timers are not working at all

This fixes #4186 until further details are given

### DIFF
--- a/radio/src/translations/tts_it.cpp
+++ b/radio/src/translations/tts_it.cpp
@@ -179,63 +179,32 @@ I18N_PLAY_FUNCTION(it, playNumber, getvalue_t number, uint8_t unit, uint8_t att)
 
 I18N_PLAY_FUNCTION(it, playDuration, int seconds PLAY_DURATION_ATT)
 {
+  if (seconds == 0) {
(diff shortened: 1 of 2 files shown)


bsongis:
@romoloman may we ask you to review @3djc PR? Thanks in advance!

PR #4898, opened 2017-05-06
Italian speller fix

klamath1977:
some italian translator fix for plural unit and general speller fix


radio/util/tts_it.py
```diff
@@ -14,40 +14,40 @@
...
-                            (u"chilometro ora", "kph0"), ("kilometers per hour", "kph1"),
+                            (u"chilometro ora", "kph0"), ("chilometri orari", "kph1"),
                             (u"miglio per ora", "mph0"), ("miglia per ora", "mph1"),
                             (u"metro", "meter0"), ("metri", "meter1"),
                             (u"piede", "foot0"), ("piedi", "foot1"),
                             (u"grado celsius", "celsius0"), ("gradi celsius", "celsius1"),
                             (u"grado fahrenheit", "fahr0"), ("gradi fahrenheit", "fahr1"),
                             (u"percento", "percent0"), (u"percento", "percent1"),
-                            (u"milliampere ora", "mamph0"), (u"milliampere ora", "mamph1"),
+                            (u"milliampèr? ora", "mamph0"), (u"milliampèr? ora", "mamph1"),
```
3djc: This can't be right

radio/util/tts_it.py
```diff
@@ -14,40 +14,40 @@
 systemSounds.append(("mille", filename(PROMPT_SYSTEM_BASE + 102)))
 for i, s in enumerate(["virgola", "un", "e", "meno", "ora", "ore", "minuto", "minuti", "secondo", "secondi"]):
     systemSounds.append((s, filename(PROMPT_SYSTEM_BASE + 103 + i)))
-for i, (s, f) in enumerate([(u"volt", "volt0"),
-                            (u"ampere", "amp0"),
-                            (u"milliampers", "mamp0"),
+for i, (s, f) in enumerate([(u"volts", "volt0"),
+                            (u"ampèr?", "amp0"),
```
3djc: This can't be right

radio/util/tts_it.py
```diff
@@ -14,40 +14,40 @@
...
+                            (u"milliampèr?", "mamp0"),
```
3djc: Can't be right

radio/util/tts_it.py
klamath1977: Why can't be right? 
please listen these sample and tell me what is the best Italian spell for "milliampere",

[sample.zip](https://github.com/opentx/opentx/files/981115/sample.zip)

3djc: I completely trust you on the sound since I do not speak Italian at all. The issue is that those string are used for generation of nightlies sd card pack (which might be using different tools) and also are going into the cvs files, to be then used in even more different tools on various platform, so you fine tuned spelling for one specific engine is likely to not sound good on another one, or break another one (some don't like ? without space for example)

Next RC will have new sound generations for Italian

3djc:
Most of those changes have been applied through: 
 https://github.com/opentx/opentx/commit/301e33a27147aa808c4857ed0f53dfd209a4679c

I'm closing this, but please let us know the result when the next RC is released.

Thanks a lot for your help

PR #4899, opened 2017-05-06
gTTS

3djc:
- move to gTTS (need to be field tested on server on a RC)
- add CZ and PT generation
- review french and english


radio/util/tts_en.py
```diff
@@ -91,29 +91,29 @@
...
                             (u"acro. mode!, on", "acro"),
-                            (u"flight. mode!, one", "fm-1"),
-                            (u"flight. mode!, two", "fm-2"),
-                            (u"flight. mode!, three", "fm-3"),
-                            (u"flight. mode!, four", "fm-4"),
-                            (u"flight. mode!, five", "fm-5"),
-                            (u"flight. mode!, six", "fm-6"),
-                            (u"flight. mode!, seven", "fm-7"),
-                            (u"flight. mode!, eight", "fm-8"),
+                            (u"flight mode one", "fm-1"),
```
bsongis: Are we using Python3 or Python2 here? Python3 doesn't need the u prefix, and it would be nice to avoid them on Python2 with `from __future__ import unicode_literals`

tools/nightly22/tts.py
```diff
@@ -27,15 +27,18 @@
...
+        command = "sox %s -r 32000 %s reverse silence 1 0.1 0.1%% reverse" % (output, filename)
+        os.system(command.encode('utf-8'))
     else:
         output = "output.mp3"
         command = 'gtts-cli -l %s -o %s "%s"' % (voice[:2], output, str)
-    os.system(command.encode('utf-8'))
-    command = "sox %s -r 32000 %s reverse silence 1 0.1 0.1%% reverse" % (output, filename)
-    os.system(command.encode('utf-8'))
+        os.system(command.encode('utf-8'))
+        command = "sox %s -r 32000 %s" % (output, filename)
```
bsongis: No need to cut some silence?

tools/nightly22/build-sdcard.sh
```diff
@@ -33,6 +33,8 @@ else
   python3 -B ${workdir}/code/tools/nightly22/tts.py es csv files
   python3 -B ${workdir}/code/tools/nightly22/tts.py it csv files
   python3 -B ${workdir}/code/tools/nightly22/tts.py de csv files
+  python3 -B ${workdir}/code/tools/nightly22/tts.py cz csv files
```
bsongis: Excellent!

tools/nightly22/tts.py
```diff
@@ -70,6 +73,11 @@ def generate(str, filename):
         directory = "es"
         voice = "es-ES"
 
+    elif "cz" in sys.argv:
+        from tts_es import systemSounds, sounds
+
+        directory = "cz"
+        voice = "cs-CZ"
```
bsongis: is cs instead of cz done on purpose?

tools/nightly22/tts.py
3djc: Not with gtts

tools/nightly22/tts.py
3djc: Yes required like this

radio/util/tts_en.py
3djc: Will have a look later today

Issue #4848, opened 2017-04-24
Companion & Horus - model files are getting corrupt by SD sync


dk7xe:
Companion - any version (current 2.2.0N369)
TX - FrSky Horus

**Issue**
model files on TX are getting corrupt in case of SD card sync initiated before "Read models and Settings from Radio" is executed. Only the current selected model is ok still.
This happens especially when upgrading Companion to a newer version and doing the SD card sync before upgrading the TX for backup reasons.
The issue occured several times in the past with different versions of Companion & OpenTX. 

**Proposal**
Sync *.bin files only in one direction from TX to PC or exclude them from SD card sync at all.
And do update of *.bin files on TX via "Write Models and Settings to Radio" (and "Read Models and Settings from Radio") only.

kilrah:
What do you mean by "corrupt" exactly?
The only thing that could in theory happen is that if you opened your SD card copy in the simulator models from these are now newer than what's on the card and they get copied across.

dk7xe:
Corrupt means that oTX is not able to load them any longer. In the model selector the models are displayed as being invalid.

3djc:
This means that data has gotten damaged somehow during transport. It can be many things, an opentx issue (radio or companion side), a usb cable issue,, a pc issue,...

At this point in time, we are considering disabling the sd sync feature altogether for 2.2.0, because thats not something we want to rush before 2.2.0 release

projectkk2glider:
The issue #4848 might be related.

dk7xe:
No, Companion is not crashing at all in na case.

powercycle:
If I delete a file that isn't working on the SD, then choose sync, that file works as expected.
Not sure why the "write" portion of the code is giving us different results than the "copy" portion of the code.
If it does choose to write instead of copy, it will break the file.

Makes me wonder if QTextStream doesn't deal well with non-text files.

mpaperno:
Indeed, looking at [the code](https://github.com/opentx/opentx/blob/next/companion/src/process_sync.cpp#L115), if a destination file doesn't already exist then it simply copies it from source folder directly.  However if it does exist, _for some reason_ it tries to copy the file byte-by-byte, using QTextStream.  Which indeed is for... text, but none of the text-specific features are used anyway (there's a QDataStream for more generic content).

Does anyone know/remember why it does this?

Otherwise it's really basic code, just copies files/folders recursively... it should be easy to fix if the byte-level copy can be eliminated.

[...]

Issue #4837, opened 2017-04-22
Crash when syncronizing SD card


Freddan-67:
First time use on a new computer.
Go to settings->settings and set up a writable directory for sd card backup (in my case %userprofile\opentx\sd-backup\ )
Go to file->syncronize SD
left-click outside of the checkbox called "show details" 
The application crash.
Update: This seems to happen during backup of directory Of (sd card drive)\Info\Video, and the file explorer also crashes in this directory. The explorer screenshot was added after having disconnected the taranis and re-connected it to the computer.
![skarmklipp 2017-04-22 08 56 22](https://cloud.githubusercontent.com/assets/7888909/25302227/b91bea74-2739-11e7-9918-534b63fb4af4.png)

![screenshot2017-04-22 08 40 04](https://cloud.githubusercontent.com/assets/7888909/25302133/7e1825f2-2737-11e7-95f9-045fbf6ddaf8.png)
![skarmklipp 2017-04-22 08 43 43](https://cloud.githubusercontent.com/assets/7888909/25302146/dc9e5c2c-2737-11e7-8bd7-98ca36ae2f6d.png)

projectkk2glider:
@Freddan-67  Which version of Companion?

Freddan-67:
As can be seen on top of the 3:rd image I am using 2.1.9 (windows 10), Lenovo X1 Carbon (2016) with Intel USB 3 host.

PR #4906, opened 2017-05-08
[Companion] SD Sync reworked.

mpaperno:
* Complete re-write of SyncProcess, now more robust, UI-independent, runs in own thread, large speed boost;
 * Use file hash comparisons instead of old method with QStrings and text stream (probable fixes for #4837 and #4848);
 * New pre-sync options dialog with selectable folders and other settings (settings are remembered until app exit);
 * Option to choose sync direction(s) (up/down/bi) and the order in which bi-directional sync runs;
 * Option to choose file comparison mode (timestamp/hash/both) or force overwrite (for one-way sync);
 * Option to set a maximum file size to copy;
 * Option to do a test run w/out actually copying any files;
 * Better reporting of progress, total counts (copied/updated/skipped), and errors, with color coding;
 * Much more reliable to cancel long-running/hung copy process;
 * ProgressWidget can now display rich-text in the text browser, added some new functions for usage;
 * Misc. small layout tweaks to ProgressDialog and ProgressWidget to make them behave/look better;

One thing it doesn't do is delete "extra" files at destination (which aren't present in source).  I'm not entirely sure that option is a good idea (at least w/out another option to back up deleted files)... but it would make the tool more complete in terms of true sync.  Just don't ask for a "backup to cloud" option  :)

One issue I have on my virtu-Mac is that the dialog takes quite a while to open while it searches for the removable drive in `findMassstoragePath()`.  Don't know if  it's just my setup though... could easily be a VM thing.  I never tried the sync on OS X until now.  

No such issues on the other platforms for me, and detection of course works with any removable drive, not just a radio.  As long as a `SOUNDS` folder is present.

image

image

image

image

image

### DIFF
--- a/companion/src/mainwindow.cpp
+++ b/companion/src/mainwindow.cpp
@@ -44,6 +44,7 @@
 #include "process_sync.h"
 #include "radiointerface.h"
 #include "progressdialog.h"
+#include "progresswidget.h"
 #include "storage.h"
 #include "translations.h"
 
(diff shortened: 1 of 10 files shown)


bsongis:
👍

dk7xe:
Great, thank you!

PR #4910, opened 2017-05-09
Updated Wizard for X9 radios.

LapinFou:
Centered startup icons.
Renamed some mixers.
Removed some useless combo box.

### DIFF
--- a/radio/sdcard/taranis-x9/SCRIPTS/WIZARD/delta.lua
+++ b/radio/sdcard/taranis-x9/SCRIPTS/WIZARD/delta.lua
@@ -122,7 +122,7 @@ local function init()
 end
 
 -- Engine Menu
-local engineModeItems = {"No", "Yes..."}
+local engineModeItems = {"No", "Yes"}
 local function drawEngineMenu()
   lcd.clear()
   lcd.drawText(1, 0, "Has your model got an engine?", 0)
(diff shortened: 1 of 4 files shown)


3djc:
I have mixed feelings about this. Some of the wizard users have only a basic knowledge of OpenTx operations, and changing the names of things will likely make them unhappy or lost, and when comparing model generated before and after, they would appear different, That would be ok if there was a good benefit for them to compensate that trouble, but I'm not sure there is.

LapinFou:
I changed the names in order to help the basic user. To me, it make more sense to name the mixes xxx_Left/Right (matching the wizard's pictures) rather xxx_1/2. 
I understand the point that it could be confusing for the cons part, however on the pros side the names are more meaningful.

schwabe:
I agree, users which are confused by _left/_right instead of _1/_2, should probably learn about how mixers/channel etc. work anyway.

bsongis:
@3djc are you ok with this PR? you said you had mixed feelings

3djc:
My concerns was renaming things. Some folks follow guides and video, and when one letter is different, they get lost.  It probably should have been left/right from the begining, the question is : is it worth the pain now.

Also, that seem in places to depart from the rest of the firmware : it wants "Long [ENT] to confirm" when we have '[Enter Long] to confirm" in the radio firmware.

But I'm fine either way (but at least we should make [Enter Long]  consistent with the firmware, I think)

yanivasy:
Maybe the modified names can be a combination of old and new names:  left1/_right2

bsongis:
everybody seems ok to merge it now, I press the button

PR #4916, opened 2017-05-11
Introduce S6R calibration tool to x7

### LINKED ISSUE #4911: No S6R_Calibrate.lua available for X7
sdcard content for X7 has the S6R.lua configuration script, but is missing the S6R_Calibrate.lua script. The X9 version is incompatible due to use of lcd.drawPixmap

3djc:
Simplify positioning instructions

Introduce real time position feedback (it will disable [enter] until your S6R is in the right position and tell you what is wrong)

Autosense between S6R and S8R (not same sensor orientation)

thanks to @cprinos for the bulk of the work

sdcard increment already dealt with in another PR

This fixes #4911

### DIFF
--- a/radio/sdcard/taranis-x7/S6R/S6R_Calibrate.lua
+++ b/radio/sdcard/taranis-x7/S6R/S6R_Calibrate.lua
@@ -1,0 +1,283 @@
+local VALUE = 0
+local COMBO = 1
+
+local COLUMN_2 = 150
+
+local edit = false
+local page = 1
+local current = 1
+local refreshState = 0
+local refreshIndex = 0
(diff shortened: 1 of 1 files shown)


cprinos:
@3djc maybe a dumb question... can the sensor be relied upon for position feedback during it's own calibration? In other words if it was really uncalibrated to start with, how can it determine if you have positioned it correctly during calibration?

3djc:
It's part for that reason that this was only introduced to x7 (and also because you can't have pictures on x7). We believe the 30% error factor should cover that aspect (it's never "fully" uncalibrated) but only time will tell for sure.

PR #4933, opened 2017-05-17
Don’t spam simulator with dummy values on telemetry reset anymore

kilrah:
It wasn't possible to simulate a radio with the equivalent of no receiver powered when the radio is turned on, and on ARM we have the telemetry simu now anyway.
Kept for AVR since there isn't one for that.

### DIFF
--- a/radio/src/telemetry/telemetry.cpp
+++ b/radio/src/telemetry/telemetry.cpp
@@ -374,31 +374,6 @@ void telemetryReset()
   telemetryData.hub.current = 55;
   telemetryData.hub.maxCurrent = 65;
 #endif
-
-/*Add some default sensor values to the simulator*/
-#if defined(CPUARM) && defined(SIMU)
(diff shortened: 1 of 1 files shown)


radio/src/telemetry/telemetry.cpp
```diff
@@ -381,7 +374,6 @@ void telemetryReset()
   telemetryData.hub.current = 55;
   telemetryData.hub.maxCurrent = 65;
 #endif
-#endif
 
 /*Add some default sensor values to the simulator*/
 #if defined(CPUARM) && defined(SIMU)
```
mpaperno: I think this block can also be removed entirely.  I've been running it like this for some time now and haven't noticed any side effects.

kilrah: Indeed, that looked more recent so thought that might upset someone - but yeah no point.

mpaperno:
👍

PR #4934, opened 2017-05-17
Minor fixes

kilrah:


radio/src/targets/horus/board.h
```diff
@@ -506,12 +506,14 @@ void hapticOff(void);
 void hapticOn(uint32_t pwmPercent);
 
 // GPS driver
+#if defined(PCBX12S) && defined(INTERNAL_GPS)
```
bsongis, line 509: it shouldn't be needed to test PCBX12S here

bsongis: do you need some help?

projectkk2glider:
@kilrah what is up with this PR?

kilrah:
I don't even remember... but I think this was intended to fix a combination that didn't build. Oh well, will fix if it ever pops up again.

PR #4940, opened 2017-05-22
[Companion] Improve first use of Companion & upgrade notice

mpaperno:
Please see commit notes.  While working on this I discovered that by checking for old version settings, the top-level keys were being created even if they didn't exist.  Eg. calling `QSettings settings("OpenTX", "Companion 2.0");` will actually create *OpenTx/Companion 2.0* file/keys if they don't already exist.  This commit makes sure those are cleaned up.

Here's what the new first-use looks like:
image

And the updated upgrade message:
image

### DIFF
--- a/companion/src/warnings.h
+++ b/companion/src/warnings.h
@@ -20,9 +20,33 @@
 
 #ifndef _WARNINGS_H_
 #define _WARNINGS_H_
-#define WARNING_LEVEL 0
-#define WARNING_ID 7
-#define WARNING QObject::tr("Thank you for upgrading to OpenTX 2.0. This is a major upgrade that adds and modifies a lot of things, so please make sure that you read release notes carefully to learn about the changes, and thoroughly check each of your models for proper function.\n\nThe OpenTX Team.")
+
+#define CPN_STR_MSG_WELCOME   QCoreApplication::translate("Companion", \
+  "<p><b>Welcome to OpenTX v%1.</b></p>" \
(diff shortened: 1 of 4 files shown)


companion/src/warnings.h
```diff
@@ -20,9 +20,33 @@
...
-#define WARNING_LEVEL 0
-#define WARNING_ID 7
-#define WARNING QObject::tr("Thank you for upgrading to OpenTX 2.0. This is a major upgrade that adds and modifies a lot of things, so please make sure that you read release notes carefully to learn about the changes, and thoroughly check each of your models for proper function.\n\nThe OpenTX Team.")
+
+#define CPN_STR_MSG_WELCOME   QCoreApplication::translate("Companion", \
+  "<p><b>Welcome to OpenTX v%1.</b></p>" \
+  "<p>As the first step, please configure the initial Radio Profile by selecting your Radio Type, Menu Language, and Build Options.</p>" \
+  "<p>You may also want to take this time to review the other available options in the displayed Settings dialog.</p>" \
+  "<p>After saving your settings, we recommend you download the latest firmware for your radio by using the <i>File -&gt; Download</i> menu option.</p>" \
+  "<p>Please visit <a href='http://www.open-tx.org'>open-tx.org</a> for latest news, updates and documentation. Thank you for choosing OpenTx!</p>" \
```
bsongis: OpenTX instead of OpenTx

PR #4945, opened 2017-05-23
Fixed syntax error in UDEV rules

zumbrunnen:
Extraneous semicolon results in an error saying: `invalid key/value pair in
file /lib/udev/rules.d/45-companion22-taranis.rules on line 1, starting
at character 83 (';')`


schwabe:
Looks good!

schwabe (approved):

PR #4955, opened 2017-05-27
Schwabe/lua set curve

schwabe:


radio/src/datastructs.h
```diff
@@ -367,7 +367,7 @@ PACK(struct FlightModeData {
 PACK(struct CurveData {
   uint8_t type:1;
   uint8_t smooth:1;
-  int8_t  points:6; // TODO conversion
+  int8_t  points:6; // TODO conversion (?!)  (describes number of points - 5)
```
bsongis, line 370: you may remove the TODO

Issue #4961, opened 2017-05-28
companion 2.2 and 9XR-PRO


benoit35690:
I edit a model using companion and write it to the radio.
The copy to the radio is corrupted.

[9XR-symaX5.zip](https://github.com/opentx/opentx/files/1034082/9XR-symaX5.zip)
![dsc_0544](https://cloud.githubusercontent.com/assets/9080382/26527768/0b0670e8-439b-11e7-8a99-b9a0584a2cec.JPG)

schwabe:
Is your attached file before or after the corruption?

benoit35690:
before.
You can reproduce the problem starting from an empty model and writting it.

schwabe:
with current 2.2 and the fix for your other issue I cannot reproduce this. Probably fixed in current code.

schwabe:
Should be closed by #4963

PR #4963, opened 2017-05-28
Fix issue #4961

schwabe:
I am not sure if that the right solution but at least it fixes the crash. Can anyone who has more experience with the trim logic take a good look at it?


bsongis:
👍

PR #4965, opened 2017-05-29
Adjust logical switch abbreviations.

mpaperno:
`L#` and `L##` were changed to `LSw##` in switch/source selection lists to avoid confusion with L1/L2/LS hardware controls.  The logical switches editor list still shows `L#`/`L##` for the line labels.  It has been suggested that `LSw` is therefore confusing, doesn't follow other naming conventions (though it's not the only one), nor matches the radio.  

Both `LS##` and `L##` (ensuring zero-padding) have been suggested, both have positives and negatives.  Radio has the advantage of using icons to distinguish logic switch vs. hardware, whereas Companion doesn't, so difference between "L1" and "L01" becomes pretty subtle.  But "LS##" doesn't currently match the radio.

This PR uses the `LS##` convention for both selector lists and the editor line screen labels, but it also centralizes the string to one location, so any further changes can be made in just one place.

### DIFF
--- a/radio/src/strhelpers.cpp
+++ b/radio/src/strhelpers.cpp
@@ -294,7 +294,8 @@ char * getSwitchString(char * dest, swsrc_t idx)
 #endif
 
   else if (idx <= SWSRC_LAST_LOGICAL_SWITCH) {
-    strAppendUnsigned(strAppend(s, "L"), abs(idx-SWSRC_FIRST_LOGICAL_SWITCH+1), 2);
+    *s++ = 'L';
+    strAppendUnsigned(s, idx-SWSRC_FIRST_LOGICAL_SWITCH+1, 2);
   }
   else if (idx <= SWSRC_ONE) {
     getStringAtIndex(s, STR_VSWITCHES, IDX_ON_IN_STR_VSWITCHES + idx - SWSRC_ON);
--- a/radio/src/strhelpers.cpp
+++ b/radio/src/strhelpers.cpp
@@ -294,7 +294,7 @@ char * getSwitchString(char * dest, swsrc_t idx)
 #endif
 
   else if (idx <= SWSRC_LAST_LOGICAL_SWITCH) {
-    strAppendStringWithIndex(s, "L", idx-SWSRC_FIRST_LOGICAL_SWITCH+1);
+    strAppendUnsigned(strAppend(s, "L"), abs(idx-SWSRC_FIRST_LOGICAL_SWITCH+1), 2);
   }
   else if (idx <= SWSRC_ONE) {
     getStringAtIndex(s, STR_VSWITCHES, IDX_ON_IN_STR_VSWITCHES + idx - SWSRC_ON);
--- a/companion/src/eeprominterface.cpp
+++ b/companion/src/eeprominterface.cpp
@@ -682,7 +682,7 @@ QString RawSwitch::toString(Board::Type board, const GeneralSettings * const gen
         }
 
       case SWITCH_TYPE_VIRTUAL:
-        return QObject::tr("LS%1").arg(index, 2, 10, QChar('0'));
+        return QObject::tr("L%1").arg(index, 2, 10, QChar('0'));
 
       case SWITCH_TYPE_MULTIPOS_POT:
         if (!getCurrentFirmware()->getCapability(MultiposPotsPositions))
--- a/companion/src/eeprominterface.cpp
+++ b/companion/src/eeprominterface.cpp
@@ -544,7 +544,7 @@ QString RawSource::toString(const ModelData * model, const GeneralSettings * con
       return result;
 
     case SOURCE_TYPE_CUSTOM_SWITCH:
-      return QObject::tr("LSw%1").arg(index+1, 2, 10, QChar('0'));
+      return RawSwitch(SWITCH_TYPE_VIRTUAL, index+1).toString();
 
     case SOURCE_TYPE_CYC:
       return QObject::tr("CYC%1").arg(index+1);
@@ -682,7 +682,7 @@ QString RawSwitch::toString(Board::Type board, const GeneralSettings * const gen
         }
 
       case SWITCH_TYPE_VIRTUAL:
-        return QObject::tr("LSw%1").arg(index, 2, 10, QChar('0'));
+        return QObject::tr("LS%1").arg(index, 2, 10, QChar('0'));
 
       case SWITCH_TYPE_MULTIPOS_POT:
         if (!getCurrentFirmware()->getCapability(MultiposPotsPositions))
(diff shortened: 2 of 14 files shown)


mhotar:
@mpamerno Hallo. Are you willing to add icons to companion? I already have them in svg format. I however do not know how easy/time consuming task it is :)

radio/src/strhelpers.cpp
```diff
@@ -294,7 +294,7 @@ char * getSwitchString(char * dest, swsrc_t idx)
 #endif
 
   else if (idx <= SWSRC_LAST_LOGICAL_SWITCH) {
-    strAppendStringWithIndex(s, "L", idx-SWSRC_FIRST_LOGICAL_SWITCH+1);
+    strAppendUnsigned(strAppend(s, "L"), abs(idx-SWSRC_FIRST_LOGICAL_SWITCH+1), 2);
```
bsongis: abs not needed (done some lines above)

bsongis: *s++ = 'L';
strAppendUnsigned(s, idx-SWSRC_FIRST_LOGICAL_SWITCH+1, 2);

mpaperno:
@mhotar Hi Martin,  yes, I think it would be great to have those resources available as SVGs for use in Companion.   I've thought about using the bitmaps directly but SVG would be better.  I'm talking about the ones currently in `sdcard/horus/themes/default` and `radio/src/bitmaps/212x64/std` (though I think there's some overlap), and maybe there are some icons in Horus font_stdsize.png which aren't anywhere else...?

The icons can be placed in various places, like the editor tabs and so on.  They don't work very well in the actual selector lists when next to each item -- it gets quite busy looking and ultimately doesn't help.  But they do look good when an items is already selected (the combo box dropdown/popup is not being shown).  They would also look good in the category titles (bold text in this example).

image

mpaperno:
>We had decided on "Lxx"

We had?  Seems like this just brings back the original complaint of confusion between hardware and logic switch.  All other radio sources use (at least) two letters... etc.  The biggest argument against using "LS##" was that it wouldn't match the radio, but you went ahead and changed radio side anyway...   Well, either way, just be aware that the original change was made due to a user request to begin with, which others seemed to think was a good idea at the time (not sure I can find it now, was part of some other issue IIRC).

I've heard several times recently that "Companion should show/work exactly the same as the radio" and while in general I agree with that in order to provide continuity for the user, I don't agree with it entirely because if that was the case why bother with Companion at all?  Just use the simulator.  Using obscure symbols or acronyms can be OK for a tiny radio UI, but seems often counterproductive in a desktop application.

kilrah:
The point is that we want radio and companion to match since that seems to be what people want, and as mentioned yesterday on the chat the radio can't have more than 3 characters. Lxx is thus the only solution to the problem if we want to avoid the slider/LS confusion (L1 for the slider, L01 for the switch).

From reports LSw brought more confusion than anything, people seem to expect to see the same thing on both sides for references (I would too).
Companion's additional space is nice for more complex, one-off, less-descriptive features... and lots of people (not me) just prefer to use a keyboard and mouse on a PC with the entire page in sight than the radio interface, that's what companion is (or "became") for nowadays.

radio/src/strhelpers.cpp
kilrah: I used the call from `strAppendStringWithIndex()` as is, IMO brings less potential confusion in the future (pattern can be easily found). OK to change if you feel differently.

elecpower:
This is not a vote for or against this PR but one perspective from a relatively new entrant to RC and thus OpenTX. 
1st radio 2nd hand JR XG7 and only the radio UI to program which I used for about a year then decided to buy a Taranis X9D+ 18 months ago which out of box only came with FrSky radio UI which lasted 5 minutes after I installed OpenTX and Companion (cudos to the devs past and present). Have built every model using Companion as the UI is far superior, especially for a newcomer to RC, due to the available UI size, GUI controls and changed/improved representation of the same data and the model saving Simulator. Only use the radio UI for field tweaks. I fly for the enjoyment, friendships and would like programming tools that make my life easier and Companion fits the bill. I test, report issues and have fixed a few minor ones to support the devs and all users.
Agree that certain elements should be the same between radio and Companion which unfortunately the radio limits in certain cases. However clearer representation of the data is where Companions UI capabilities should not be restricted by the radio eg FMs and GVARs are already represented differently and improves over radio so why stop there. As an example I have been kicking around an idea for a graphical representation of a model config but based on some views would be a waste of my time as it would be rejected as not matching the radio.
I know We (royal) are all not going to agree all the time but past decisions should not be etched in stone, if with the passing of time, technologies, rational discussion, etc a change of policy is warranted.
Keep up the good work.

radio/src/strhelpers.cpp
bsongis: Yes I used strAppend there because LS is 2 letters long, but now with "L" it's not needed anymore

kilrah:
@mpaperno's upcoming "groups in the dropdown" is a perfect example of making use of companion's ability to make things clearer. It precisely allows for some handholding without requiring naming something differently on fw/companion.

The "LSwXX" has only been there for a couple of RCs but we got maybe 10 people asking/complaining about it. Some people are way more rigid than we'd think, e.g. when we innocently renamed "custom functions" into "special functions" years ago you wouldn't believe how many people would get stopped in their tracks and unable to do anything when the video they were watching referred to the former and they were unable to realise the entire screen and everything was identical apart from that one word. 
That certainly won't stop us, but we need to keep it in mind and at least try to minimize the impact...

> I have been kicking around an idea for a graphical representation of a model config but based on some views would be a waste of my time as it would be rejected as not matching the radio.

Why not talk about it?

mpaperno:
>> I have been kicking around an idea for a graphical representation of a model config but based on some views would be a waste of my time as it would be rejected as not matching the radio.

> Why not talk about it?

I created a new label for Issues -- `Request for Comments`.  Perhaps an RFC with a rough sketch of a new idea (graphically/verbally) would be a good way to judge response before getting too far into design/code.  I think I should start doing this also...  ;-)   I'm not entirely sure Issues is the best place for that in terms of exposure, but I can't think of a better place offhand (and one could always post a link to issue in forums/chat/etc).

rdeanchurch:
On the one hand what Kilrah said about SF vs CF is certainly true, on the other, it is never going to be less painful to change than now.  A very gentle vote for going to LS, (But I really want BS...which is for Boolean Switch of course;-)  
Just B would work, as they switch nothing, they just return a true or false.  What is done with that could be thought of as a switch or enable or custom/Special...or nothing.

3djc:
I think 90%+ of opentx will find boolean way too cryptic and not making sense at all

kilrah:
99% of people will have no idea of what a Boolean Switch is though. Not to mention that "BS" will make people think about something we already see too much of ;)

PR #4981, opened 2017-05-31
Allow to disable telemetry alarams

schwabe:
This mainly useful for cheap transceivers that have telemetry but no power amplifier and very limited range of often 10-20 meters. This includes fishpepper's tinyFISH FC (and its legal/illegal clones) and also Spektrum BNF models.

The commit reuses the used level fields on ARM. I know this commit might be controversal as it disables a safety feature. To remedy this, I added a Warning when selecting the model. (This model has RSSI alarms disabled. This is for the affected receivers still much better than a constant telemetry lost/recovered/critical stream of sounds.


radio/src/gui/212x64/model_telemetry.cpp
```diff
@@ -25,6 +25,7 @@ enum MenuModelTelemetryFrskyItems {
   ITEM_TELEMETRY_RSSI_LABEL,
   ITEM_TELEMETRY_RSSI_ALARM1,
   ITEM_TELEMETRY_RSSI_ALARM2,
+  ITEM_TELEMETRY_DISALBE_ARLAMS,
```
bsongis: double typo

radio/src/datastructs.h
```diff
@@ -468,11 +468,18 @@ PACK(struct ScriptData {
 /*
  * Frsky Telemetry structure
  */
-
+#if defined(CPUARM)
+PACK(struct FrSkyRSSIAlarm {
```
bsongis: perhaps there is another place for this! because it looks like a hack ...
unless we completely remove this struct from CPUARM and replace it by the `disabled` bit + a bigger spare

radio/src/gui/212x64/model_telemetry.cpp
bsongis: many typos in the whole PR in fact ;)

radio/src/telemetry/telemetry.cpp
```diff
@@ -188,27 +188,29 @@ void telemetryWakeup()
...
 
-    if (TELEMETRY_STREAMING()) {
-      if (getRssiAlarmValue(1) && TELEMETRY_RSSI() < getRssiAlarmValue(1)) {
-        AUDIO_RSSI_RED();
-        SCHEDULE_NEXT_ALARMS_CHECK(10/*seconds*/);
-      }
-      else if (getRssiAlarmValue(0) && TELEMETRY_RSSI() < getRssiAlarmValue(0)) {
-        AUDIO_RSSI_ORANGE();
-        SCHEDULE_NEXT_ALARMS_CHECK(10/*seconds*/);
+    if (!g_model.frsky.rssiAlarms[0].disabled) {
```
bsongis: don't you need "sensors lost" in addition?

radio/src/translations/en.h.txt
```diff
@@ -1027,6 +1027,9 @@
 #define TR_ANTENNACONFIRM2             "Make sure antenna is installed!"
 #define TR_LOWALARM                    INDENT "Low alarm"
 #define TR_CRITICALALARM               INDENT "Critical alarm"
+#define TR_RSSIALARM_WARN              TR("RSSI","TELEMETRY RSSI")
+#define TR_NO_RSSIALARM                "Audio alarms disabled"
```
bsongis: why not "Telemetry alarms"?

schwabe: It actually only disables audio and Telemetry alarms disabled does not fit on X7. We would need to use Telemetry alarms dis. or something like that.

radio/src/gui/212x64/model_telemetry.cpp
schwabe: copy and paste :)

radio/src/telemetry/telemetry.cpp
schwabe: Oh  yeah, forgot about that one.

schwabe:
I refactored the FrksyRSSIAlarm[2] into a RSSIAlarm structure that is still eeprom compatible. This should make doing this the proper way later better.

bsongis:
Thanks for the refactoring, it makes sense to remove this struct which was not anymore used in 2.2

radio/src/gui/128x64/model_telemetry.cpp
```diff
@@ -701,14 +702,18 @@ void menuModelTelemetryFrsky(event_t event)
...
         if (attr && s_editMode>0) {
-          CHECK_INCDEC_MODELVAR(event, g_model.frsky.rssiAlarms[alarm].value, -30, 30);
-        }
+          if (warning)
+            CHECK_INCDEC_MODELVAR(event, g_model.rssiAlarms.warning, -30, 30);
+          else
+            CHECK_INCDEC_MODELVAR(event, g_model.rssiAlarms.critical, -30, 30);
+     }
 #else // CPUARM
+        int alarm = (k-ITEM_TELEMETRY_RSSI_ALARM1);
```
bsongis: uint8_t (or it uses more flash, int is 16 bits)

schwabe: Oops. Changed back to uin8_t

bsongis:
Thanks!

PR #4985, opened 2017-06-01
Align Taranis 128*64 stats to its larger sister

### LINKED ISSUE #4950: Global Persistent Timer
There's no simple way I can find to create a persistent global timer that runs whenever the transmitter is on.  Would like to be able to keep track of how long the radio runs on a charge.  Any persistent timers created using a model revert to the last time recorded with that model active so it can't be done there.  Running Companion on Windows OS, OpenTX version 2.2.0 N371.

3djc:
image

This fixes #4950

### DIFF
--- a/radio/src/gui/128x64/view_statistics.cpp
+++ b/radio/src/gui/128x64/view_statistics.cpp
@@ -20,6 +20,11 @@
 
 #include "opentx.h"
 
+#define STATS_1ST_COLUMN               1
+#define STATS_2ND_COLUMN               7*FW+FW/2
+#define STATS_3RD_COLUMN               14*FW+FW/2
+#define STATS_LABEL_WIDTH              3*FW
+
 void menuStatisticsView(event_t event)
 {
   TITLE(STR_MENUSTAT);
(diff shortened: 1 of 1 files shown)


radio/src/gui/128x64/view_statistics.cpp
```diff
@@ -41,15 +46,38 @@ void menuStatisticsView(event_t event)
       break;
   }
 
+#if defined(PCBTARANIS)
```
schwabe: Shouldn't this be CPUARM? Or do the other ARM radio not have a globalTimer?

3djc: That and number of timers I'm also unsure of

schwabe: Just checked. TIMERS defaults to 3 on all ARM platforms. You can set timers to 2 by manually compiling but that probably also breaks in Taranis/Horus world. (We might just remove that TIMER=2 option).

The only platform which is a bit strange is sky9x with reva board. There the timer might just always be zero. But I think showing a zero timer isn't problematic there. 

I would rather have all arm platforms the same here.

bsongis:
👍

PR #4988, opened 2017-06-01
Fix clang warnings

schwabe:
- signed char on AVR vs unsigned char on ARM
- use references instead of points in calculation (same code size)
- Ignore packed address warning for mavlink includes
- Turn error on packed addresses to warning for Horus (Cannot fix that without major refactoring, might even need eeprom change)

### DIFF
--- a/radio/src/maths.cpp
+++ b/radio/src/maths.cpp
@@ -129,18 +129,18 @@ uint16_t isqrt32(uint32_t n)
 
 #if defined(FRSKY_HUB) && !defined(CPUARM)
 // convert latitude and longitude to 1/10^6 degrees
-void extractLatitudeLongitude(uint32_t & latitude, uint32_t & longitude)
+void extractLatitudeLongitude(uint32_t * latitude, uint32_t * longitude)
 {
   div_t qr = div(telemetryData.hub.gpsLatitude_bp, 100);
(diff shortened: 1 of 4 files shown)


mpaperno:
> signed char on AVR vs unsigned char on ARM

Is this a `clang` thing? I noticed a similar change you did recently.  There may be a compiler setting to treat all chars as unsigned?  That is typically the default for the compilers we use, but perhaps not for clang? 

Also wondering which version of clang you use since I never see these warnings...  just tried again, 9X, MEGA, and 9X-PRO builds of libsimulator (and firmware just for kicks).  Release and "blank" build types.  Builds clean.   Hmm... looks like this is using `clang-800.0.42.1`

schwabe:
```clang -v
Apple LLVM version 8.1.0 (clang-802.0.38)```

Maybe the ```-Waddress-of-packed-member``` argument isn't default on normal Clang versions?

The signess is in the API. And it is actually the avr-gcc 6.2 here that complains.

bsongis:
Will you fix AVR compiler issue?

schwabe:
I will look into it. My avr-gcc 6.3 has no problem with that syntax, seem older avr-gcc version don't like that syntax. Probably better to also ignore that warning for now.

schwabe:
I remove the AVR fixes until we have a working version on the build server/travis. I couldn't find a solution that has no warning and also does not increase avr code size.

PR #5004, opened 2017-06-05
Fix SDcard size and format issue

### LINKED ISSUE #4924: Qx7  report wrong sd card size
If a micro SD card bigger than 4Gig is use, 8Gig etc. The Q X7 information screen report a max of 3800M and refuse to format it, returning a cd card error. 

The card can be formatted on a pc and use as a normal 4Gig card. 

I tested the cards with h2testw and they are fine.

3djc:
- fix overflow in SDcard size computation
- fix format function
- implement more informative format error handling

This fixes #4924

### DIFF
--- a/radio/src/gui/common/stdlcd/radio_sdmanager.cpp
+++ b/radio/src/gui/common/stdlcd/radio_sdmanager.cpp
@@ -175,12 +175,32 @@ void menuRadioSdManager(event_t _event)
     audioQueue.stopSD();
 #endif
     BYTE work[_MAX_SS];
-    if (f_mkfs(0, FM_FAT32, 0, work, sizeof(work)) == FR_OK) {
-      f_chdir("/");
-      REFRESH_FILES();
-    }
-    else {
-      POPUP_WARNING(STR_SDCARD_ERROR);
+    FRESULT res = f_mkfs("", FM_FAT32, 0, work, sizeof(work));
+    switch(res) {
+      case FR_OK :
+        f_chdir("/");
+        REFRESH_FILES();
+        break;
+      case FR_DISK_ERR:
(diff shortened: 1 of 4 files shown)


3djc:
I have left error message untranslatable, as users should not see them, and when/if it comes to  us I would prefer it in english, but I'm opened to different opinions

radio/src/sdcard.cpp
```diff
@@ -472,7 +472,7 @@ uint32_t sdGetNoSectors()
 
 uint32_t sdGetSize()
 {
-  return (sdGetNoSectors() * BLOCK_SIZE) / 1000000;
```
bsongis, line 475: 👍

radio/src/gui/common/stdlcd/radio_sdmanager.cpp
```diff
@@ -175,12 +175,32 @@ void menuRadioSdManager(event_t _event)
     audioQueue.stopSD();
 #endif
     BYTE work[_MAX_SS];
-    if (f_mkfs(0, FM_FAT32, 0, work, sizeof(work)) == FR_OK) {
-      f_chdir("/");
-      REFRESH_FILES();
-    }
-    else {
-      POPUP_WARNING(STR_SDCARD_ERROR);
+    FRESULT res = f_mkfs("", FM_FAT32, 0, work, sizeof(work));
```
bsongis: This code is common to all ARM platforms, perhaps it's possible to add a function and avoid redundant code?

bsongis: ... at least the code which deals with error codes

bsongis:
Thanks!

PR #5008, opened 2017-06-06
Fully adjustable splash timer on x7, matching x9 implementation

### LINKED ISSUE #5007: Very long startup time of QX7 after update to official 2.2.0
After update of my QX7 to official release 2.2.0 it takes about 3 seconds for startup. During that startup time the display shows "OPEN TX". In the Tx setup page I have not activated the option to show a startup picture. Can change to any model, also without SD card no change in loading time. When pressing a key e.g. ENTER it immediately skips/terminates the splash screen. My last OTx version before this 2.2.0 was the nightly N368, startup time there was much shorter.

3djc:
image

This fixes #5007

### DIFF
--- a/radio/src/gui/128x64/radio_setup.cpp
+++ b/radio/src/gui/128x64/radio_setup.cpp
@@ -467,6 +467,7 @@ void menuRadioSetup(event_t event)
 #if defined(SPLASH) && !defined(FSPLASH)
       case ITEM_SETUP_DISABLE_SPLASH:
       {
+#if defined(CPUARM)
         lcdDrawTextAlignedLeft(y, STR_SPLASHSCREEN);
         if (SPLASH_NEEDED()) {
           lcdDrawNumber(RADIO_SETUP_2ND_COLUMN, y, SPLASH_TIMEOUT/100, attr|LEFT);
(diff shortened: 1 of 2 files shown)


radio/src/gui/128x64/radio_setup.cpp
```diff
@@ -467,8 +467,15 @@ void menuRadioSetup(event_t event)
 #if defined(SPLASH) && !defined(FSPLASH)
       case ITEM_SETUP_DISABLE_SPLASH:
       {
-        uint8_t b = 1-g_eeGeneral.splashMode;
-        g_eeGeneral.splashMode = 1 - editCheckBox(b, RADIO_SETUP_2ND_COLUMN, y, STR_SPLASHSCREEN, attr, event);
+        lcdDrawTextAlignedLeft(y, STR_SPLASHSCREEN);
```
bsongis: this uses some flash on 9X for nothing, am I mistaken?

radio/src/gui/128x64/radio_setup.cpp
```diff
@@ -467,8 +467,15 @@ void menuRadioSetup(event_t event)
...
+        if (SPLASH_NEEDED()) {
+          lcdDrawNumber(RADIO_SETUP_2ND_COLUMN, y, SPLASH_TIMEOUT/100, attr|LEFT);
+          lcdDrawChar(lcdLastRightPos, y, 's');
+        }
+        else {
+          lcdDrawMMM(RADIO_SETUP_2ND_COLUMN, y, attr);
```
bsongis, line 477: Isn't it better to hide the whole line then?
This would remove the 9X issue in the same time...

bsongis:
:+1

Issue #5006, opened 2017-06-05
companion 2.2 and 9XR-PRO


benoit35690:
issue  #4961 was not fixed

try to write the "phoenixRC" model
[issue4961.zip](https://github.com/opentx/opentx/files/1052429/issue4961.zip)

in radio
model page "INPUTS" 4/11
- channels are now in order "Rud, Ele Thr Ail"
- look at inputs : there are now random offsets

schwabe:
I created a pull to fix the problem. Note that the otx file already is "broken", i.e. the wrong offset are already in the otx file. The commit fixes Companion to correctly read and write those
![image](https://cloud.githubusercontent.com/assets/1038539/26824714/be4b7dc4-4ab2-11e7-9128-250fea11ce79.png)

PR #5009, opened 2017-06-06
Fix issue #4961

schwabe:
Fix yet another reference to HORUS_OR_TARANIS that should be IS_ARM
Also rename C9X function to export/import to be inline with the rest


schwabe:
Oops, should be fix issue #5006

companion/src/firmwares/opentx/opentxeeprom.cpp
```diff
@@ -1369,13 +1369,13 @@ class InputField: public TransformedField {
 
     virtual void beforeExport()
     {
-      _weight = smallGvarToEEPROM(expo.weight);
+      _weight = smallGvarImport(expo.weight);
 
-      if (IS_HORUS_OR_TARANIS(board) && version >= 216) {
-        _offset = smallGvarToEEPROM(expo.offset);
+      if ((IS_HORUS_OR_TARANIS(board) && version >= 216) || (IS_ARM(board) && version >= 218)) {
```
bsongis, line 1384: 👍

companion/src/firmwares/opentx/opentxeeprom.cpp
```diff
@@ -1369,13 +1369,13 @@ class InputField: public TransformedField {
...
+        _offset = smallGvarImport(expo.offset);
       }
 
-      if (!IS_TARANIS(board) || version < 216) {
+      if ((!IS_ARM(board) && version < 218) || (!IS_TARANIS(board) && version < 217)|| version < 216) {
```
bsongis: It changes SKY9X in VERSION 217
Shouldn't it be <= instead of < here?

companion/src/firmwares/opentx/opentxeeprom.cpp
```diff
@@ -1386,33 +1386,31 @@ class InputField: public TransformedField {
...
       else if (expo.mode) {
         expo.srcRaw = RawSource(SOURCE_TYPE_STICK, expo.chn);
       }
 
-      expo.weight = smallGvarToC9x(_weight);
+      expo.weight = smallGvarExport(_weight);
 
-      if (IS_STM32(board) && version >= 216) {
-        expo.offset = smallGvarToC9x(_offset);
+      if ((IS_STM32(board) && version >= 216) || (IS_ARM(board) && version >=218)) {
```
bsongis: Typo: one space missing between >= and 218

companion/src/firmwares/opentx/opentxeeprom.cpp
schwabe: I think the right thing here is 
```
      if ((!IS_ARM(board) && version >= 218) || (!IS_TARANIS(board) && version == 217)|| version < 216) {
```

That also changes Horus in 218. But my quick test shows that the import is right and I would be very suprised if horus is different from other arm platforms in 2.2

bsongis:
I think it's still not good, but I am too tired to be sure of what I am saying

kilrah:
Will fix #5098 too

companion/src/firmwares/opentx/opentxeeprom.cpp
```diff
@@ -1369,13 +1369,13 @@ class InputField: public TransformedField {
...
+      _weight = smallGvarImport(expo.weight);
 
-      if (IS_HORUS_OR_TARANIS(board) && version >= 216) {
-        _offset = smallGvarToEEPROM(expo.offset);
+      if ((IS_HORUS_OR_TARANIS(board) && version >= 216) || (IS_ARM(board) && version >= 218)) {
+        _offset = smallGvarImport(expo.offset);
       }
 
-      if (!IS_TARANIS(board) || version < 216) {
+      if ((!IS_ARM(board) && version >= 218) || (!IS_TARANIS(board) && version == 217) || version < 216) {
```
bsongis: I propose this fix:
`if (!IS_ARM(board) || (!IS_TARANIS(board) && version < 218) || version < 216) {`

companion/src/firmwares/opentx/opentxeeprom.cpp
```diff
@@ -1386,33 +1386,31 @@ class InputField: public TransformedField {
...
       else if (expo.mode) {
         expo.srcRaw = RawSource(SOURCE_TYPE_STICK, expo.chn);
       }
 
-      expo.weight = smallGvarToC9x(_weight);
+      expo.weight = smallGvarExport(_weight);
 
-      if (IS_STM32(board) && version >= 216) {
-        expo.offset = smallGvarToC9x(_offset);
+      if ((IS_STM32(board) && version >= 216) || (IS_ARM(board) && version >= 218)) {
```
bsongis, line 1417: 👍

companion/src/firmwares/opentx/opentxeeprom.cpp
```diff
@@ -1386,33 +1386,31 @@ class InputField: public TransformedField {
...
+        expo.offset = smallGvarExport(_offset);
       }
 
-      if (!IS_ARM(board) || (!IS_STM32(board) && version < 218) || version < 216) {
```
bsongis, line 1423: What is the use case which needed a fix here?

companion/src/firmwares/opentx/opentxinterface.cpp
```diff
@@ -559,7 +559,7 @@ int OpenTxFirmware::getCapability(::Capability capability)
     case ChannelsName:
       return (IS_ARM(board) ? (HAS_LARGE_LCD(board) ? 6 : 4) : 0);
     case HasCvNames:
-      return (IS_HORUS_OR_TARANIS(board) ? 1 : 0);
+      return (IS_ARM(board) ? 1 : 0);
```
bsongis, line 564: 👍

companion/src/firmwares/opentx/opentxeeprom.cpp
schwabe: I think !IS_STM32 and !IS_TARANIS should be identical for 217, changing this makes sense since it s then the same if on the import and export.

companion/src/firmwares/opentx/opentxeeprom.cpp
schwabe: Your statement would also be true for sky9x on 217. I am not sure what sky9x on 217. I am not sure if that change is right for them on 217.

bsongis: If I am not mistaken we did the port of this feature from Taranis to Sky9x in EEPROM 218, so that sounds right to me

companion/src/firmwares/opentx/opentxeeprom.cpp
bsongis: You are right for IS_STM32
But I think the !IS_ARM(board) should remain as it was (no test for 218), as I did above
And yes this would be the same test in import and export

companion/src/firmwares/opentx/opentxeeprom.cpp
```diff
@@ -1379,13 +1379,13 @@ class InputField: public TransformedField {
...
+      _weight = smallGvarImport(expo.weight);
 
-      if (IS_HORUS_OR_TARANIS(board) && version >= 216) {
-        _offset = smallGvarToEEPROM(expo.offset);
+      if ((IS_HORUS_OR_TARANIS(board) && version >= 216) || (IS_ARM(board) && version >= 218)) {
+        _offset = smallGvarImport(expo.offset);
       }
 
-      if (!IS_TARANIS(board) || version < 216) {
+      if (!IS_ARM(board) || (!IS_TARANIS(board) && version < 217) || version < 216) {
```
bsongis: It should be `(!IS_TARANIS(board) && version < 218)` as in the other test, right?

schwabe: yes, my fault changed == to < and forgot to change 217 to 218 :(

PR #5011, opened 2017-06-06
Model ID between internal and external module do not conflict

schwabe:
An internal xjt with model 2 and a ext xjt with model 2 are not binding the same receiver, other external modules do not even use the same protocol.


3djc:
Closing this one as we are working on #5018

PR #5033, opened 2017-06-16
x7/x9e startup review

3djc:
This changes both x7 and x9 startup sequence and bring them on par.

- both x7 and x9 now start after a fairly short 1 second push (long enough to avoid unwanted 'on' during transportation and the likes, but also short enough to not be bothering).

- turn off is unchanged timing wise, but graphical sequences is reversed (power on fills in, turning off fills out).

We understand some x7 users will be frustrated by this change, but we have had reports of people arriving on field with empty x7 because it did turn itself off during manutention.


bsongis:
👍

Issue #1954, opened 2014-12-26
[Wish]Companion: tool to generate notes files


MicheleVilla:
Notes files are very useful options but sometimes is difficult to insert and visualize some font (i.e. arrow  up and down)
A simple tool for generate notes files may be helpful in Companion,in this way could be used custom font

WavyDavy1960:
I would also like a Companion tool for writing/managing Checklist notes. It would be a tab in Companion that shows "Checklist Notes" for selected model and allow writing/editing of notes. This would be very useful when creating/modifying settings in Companion, and would be saved with model settings.

PR #5047, opened 2017-06-20
Add model checklist edit to Companion

elecpower:
First attempt for feedback.
There are a few requests for editors. I decided a basic editor for model checklist would be useful for the majority.
Further improvements could include:
- limiting line length to the display area of each radio model
- handling model name renames. However, renaming the checklist file would need to be delayed until the otx is saved as the user can discard the changes. Maybe just too much effort for little gain. The Import feature can be used to retrieve the file contents with the old name and load the file for the new name.
Import can also be used to load a template.

screenshot from 2017-06-20 17-56-30

screenshot from 2017-06-20 17-56-41

### DIFF
--- a/companion/src/modeledit/checklistdialog.h
+++ b/companion/src/modeledit/checklistdialog.h
@@ -37,12 +37,14 @@ public:
 private slots:
   void import();
   void update();
+  void changed();
 
 private:
   Ui::ChecklistDialog *ui;
   QString mModelChecklist;
   QString mChecklistFolder;
-  QString readFile(const QString filepath, const bool exists);
+  QString readFile(const QString & filepath, const bool exists);
+  bool mDirty;
 };
 
 #endif // _CHECKLISTDIALOG_H_
--- a/companion/src/modeledit/checklistdialog.h
+++ b/companion/src/modeledit/checklistdialog.h
@@ -1,0 +1,48 @@
+/*
+ * Copyright (C) OpenTX
+ *
+ * Based on code named
+ *   th9x - http://code.google.com/p/th9x
+ *   er9x - http://code.google.com/p/er9x
+ *   gruvin9x - http://code.google.com/p/gruvin9x
+ *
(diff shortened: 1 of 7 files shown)


powercycle:
I was just thinking about how i might be able to use a feature like this.  Thanks for giving this the effort.

companion/src/modeledit/setup.cpp
```diff
@@ -891,6 +892,8 @@ SetupPanel::SetupPanel(QWidget * parent, ModelData & model, GeneralSettings & ge
 
   if (!firmware->getCapability(HasDisplayText)) {
     ui->displayText->hide();
+    ui->editText->hide();
+    connect(ui->editText, SIGNAL(clicked()), this, SLOT(on_editText_clicked()));
```
elecpower: else missing between hide and connect

companion/src/modeledit/checklistdialog.h
```diff
@@ -0,0 +1,48 @@
...
+
+private slots:
+  void import();
+  void update();
+
+private:
+  Ui::ChecklistDialog *ui;
+  QString mModelChecklist;
+  QString mChecklistFolder;
+  QString readFile(const QString filepath, const bool exists);
```
bsongis: const QString &

bsongis (commented):
Cool function, thanks!

companion/src/modeledit/checklistdialog.h
```diff
@@ -37,12 +37,14 @@ class ChecklistDialog : public QDialog {
...
   void import();
   void update();
+  void changed();
 
 private:
   Ui::ChecklistDialog *ui;
   QString mModelChecklist;
   QString mChecklistFolder;
-  QString readFile(const QString filepath, const bool exists);
+  QString readFile(const QString & filepath, const bool exists);
```
bsongis: not sure this const is really useful, but it won't hurt

bsongis:
👍

elecpower:
Fixes #1954 part

PR #5068, opened 2017-06-30
Bsongis/x7 bluetooth

bsongis:


radio/src/targets/taranis/bluetooth_driver.cpp
```diff
@@ -179,18 +196,27 @@ void bluetoothWakeup(void)
...
-        for (int i=0; i<len; i++) {
-          btTxFifo.push(idx2char(g_eeGeneral.bluetoothName[i]));
+        if (len > 0) {
+          for (int i = 0; i < len; i++) {
+            btTxFifo.push(idx2char(g_eeGeneral.bluetoothName[i]));
+          }
+        }
+        else {
+#if defined(PCBHORUS)
+          bluetoothWriteString("Horus");
```
bsongis: Don't we have those strings somewhere else?

PR #5071, opened 2017-07-01
Hide checked but unused flight modes in inputs and mixes lists

### LINKED ISSUE #5057: Companion suggestion to supresss unused FMs from inputs and mixes lists
Just a cosmetic suggestion. Currently unused FMs i.e. those without a switch, display in the lists when one or more FMs are unchecked for a input or mix line. FM0 would always display.

capture

elecpower:
Fixes #5057 

screenshot from 2017-07-02 09-39-26

In this example only FMs 1 & 2 have switches assigned.

Before
screenshot from 2017-07-02 09-19-05

After
screenshot from 2017-07-02 09-20-18

### DIFF
--- a/companion/src/apppreferencesdialog.cpp
+++ b/companion/src/apppreferencesdialog.cpp
@@ -71,6 +71,7 @@ void AppPreferencesDialog::writeValues()
   g.simuSW(ui->simuSW->isChecked());
   g.removeModelSlots(ui->opt_removeBlankSlots->isChecked());
   g.newModelAction(ui->opt_newMdl_useWizard->isChecked() ? 1 : ui->opt_newMdl_useEditor->isChecked() ? 2 : 0);
+  g.displayAllFMs(ui->opt_displayAllFMs->isChecked());
   g.historySize(ui->historySize->value());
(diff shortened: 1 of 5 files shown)


lshems:
What happens if you assign switches to only flight modes 1 and 2? if the associated switches are both off, flight mode 0 gets activated, but I would not see its effect in the mixer line.

What happens if only 0 and 2 are assigned? 1 gets defaulted.

Pay attention to these scenario's. They are often used.

mpjf01:
See issue #5057

elecpower:
Since there was so much angst to suppress I am proposing that the suppression is a user choice controlled by an application setting. By default there will be no change.
screenshot from 2017-07-10 21-04-09

mpjf01:
By default there will be a change, there will be an additional check box in the settings that has to be explained and documented. Inevitably people will select it without realising what they are doing and then wonder why their FM displays have changed generating more user interaction, complaint and negativity. 

Any change to any system needs to be justified because it always affects someone and the benefits need to outweigh any negative impacts. There are no benefits from this proposed change as far as I can tell.

elecpower:
No support so closing

PR #5078, opened 2017-07-04
Adds CrossTrim Special Function

mtwomey:
Adds a function that swaps left and right trims. This allows the following:

1. Easy / quick way to enable / disable cross-trims
1. Visually, the trims still show up under the sticks they're affecting
1. Can be assigned globally or as a special function per model
1. Doesn't impact any other setting in the radio (just switches the buttons effectively)

As seen here: https://youtu.be/JKQgAdqzfmY

### DIFF
--- a/radio/src/dataconstants.h
+++ b/radio/src/dataconstants.h
@@ -997,6 +997,7 @@ enum Functions {
   FUNC_BACKLIGHT,
 #if defined(PCBTARANIS)
   FUNC_SCREENSHOT,
+  FUNC_CROSSTRIMS,
 #endif
 #if defined(DEBUG)
   FUNC_TEST, // should remain the last before MAX as not added in Companion
--- a/radio/src/functions.cpp
+++ b/radio/src/functions.cpp
@@ -578,6 +578,9 @@ void evalFunctions()
               mainRequestFlags |= (1 << REQUEST_SCREENSHOT);
             }
             break;
+          case FUNC_CROSSTRIMS:
+            newActiveFunctions |= (1 << FUNC_CROSSTRIMS);
+            break;
 #endif
 
 #if defined(DEBUG)
--- a/radio/src/targets/taranis/keys_driver.cpp
+++ b/radio/src/targets/taranis/keys_driver.cpp
@@ -53,22 +53,41 @@ uint32_t readTrims()
 {
   uint32_t result = 0;
 
-  if (~TRIMS_GPIO_REG_LHL & TRIMS_GPIO_PIN_LHL)
-    result |= 0x01;
-  if (~TRIMS_GPIO_REG_LHR & TRIMS_GPIO_PIN_LHR)
-    result |= 0x02;
-  if (~TRIMS_GPIO_REG_LVD & TRIMS_GPIO_PIN_LVD)
-    result |= 0x04;
-  if (~TRIMS_GPIO_REG_LVU & TRIMS_GPIO_PIN_LVU)
-    result |= 0x08;
-  if (~TRIMS_GPIO_REG_RVD & TRIMS_GPIO_PIN_RVD)
-    result |= 0x10;
-  if (~TRIMS_GPIO_REG_RVU & TRIMS_GPIO_PIN_RVU)
-    result |= 0x20;
-  if (~TRIMS_GPIO_REG_RHL & TRIMS_GPIO_PIN_RHL)
-    result |= 0x40;
-  if (~TRIMS_GPIO_REG_RHR & TRIMS_GPIO_PIN_RHR)
-    result |= 0x80;
+  if (isFunctionActive(FUNC_CROSSTRIMS)){
+    if (~TRIMS_GPIO_REG_LHL & TRIMS_GPIO_PIN_LHL)
+      result |= 0x40;
+    if (~TRIMS_GPIO_REG_LHR & TRIMS_GPIO_PIN_LHR)
+      result |= 0x80;
+    if (~TRIMS_GPIO_REG_LVD & TRIMS_GPIO_PIN_LVD)
+      result |= 0x10;
+    if (~TRIMS_GPIO_REG_LVU & TRIMS_GPIO_PIN_LVU)
+      result |= 0x20;
+    if (~TRIMS_GPIO_REG_RVD & TRIMS_GPIO_PIN_RVD)
+      result |= 0x04;
+    if (~TRIMS_GPIO_REG_RVU & TRIMS_GPIO_PIN_RVU)
+      result |= 0x08;
+    if (~TRIMS_GPIO_REG_RHL & TRIMS_GPIO_PIN_RHL)
+      result |= 0x01;
+    if (~TRIMS_GPIO_REG_RHR & TRIMS_GPIO_PIN_RHR)
+      result |= 0x02;
+  } else {
+    if (~TRIMS_GPIO_REG_LHL & TRIMS_GPIO_PIN_LHL)
+      result |= 0x01;
+    if (~TRIMS_GPIO_REG_LHR & TRIMS_GPIO_PIN_LHR)
+      result |= 0x02;
+    if (~TRIMS_GPIO_REG_LVD & TRIMS_GPIO_PIN_LVD)
+      result |= 0x04;
+    if (~TRIMS_GPIO_REG_LVU & TRIMS_GPIO_PIN_LVU)
+      result |= 0x08;
+    if (~TRIMS_GPIO_REG_RVD & TRIMS_GPIO_PIN_RVD)
+      result |= 0x10;
+    if (~TRIMS_GPIO_REG_RVU & TRIMS_GPIO_PIN_RVU)
+      result |= 0x20;
+    if (~TRIMS_GPIO_REG_RHL & TRIMS_GPIO_PIN_RHL)
+      result |= 0x40;
+    if (~TRIMS_GPIO_REG_RHR & TRIMS_GPIO_PIN_RHR)
+      result |= 0x80;
+  }
 
   // TRACE("readTrims(): result=0x%02x", result);
 
--- a/radio/src/translations/en.h.txt
+++ b/radio/src/translations/en.h.txt
@@ -306,6 +306,7 @@
 
 #if defined(PCBTARANIS)
   #define TR_SF_SCREENSHOT             "Screenshot"
+  #define TR_SF_CROSSTRIMS             "CrossTrims"
 #else
   #define TR_SF_SCREENSHOT
 #endif
@@ -313,7 +314,7 @@
 #define TR_SF_RESERVE                  "[reserve]\0"
 
 #if defined(CPUARM)
-  #define TR_VFSWFUNC                  TR_SF_SAFETY "Trainer\0  ""Inst. Trim""Reset\0    ""Set \0     " TR_ADJUST_GVAR "Volume\0   " "SetFailsfe" "RangeCheck" "ModuleBind" TR_SOUND TR_PLAY_TRACK TR_PLAY_VALUE TR_SF_RESERVE TR_SF_PLAY_SCRIPT TR_SF_RESERVE TR_SF_BG_MUSIC TR_VVARIO TR_HAPTIC TR_SDCLOGS "Backlight\0" TR_SF_SCREENSHOT TR_SF_TEST
+  #define TR_VFSWFUNC                  TR_SF_SAFETY "Trainer\0  ""Inst. Trim""Reset\0    ""Set \0     " TR_ADJUST_GVAR "Volume\0   " "SetFailsfe" "RangeCheck" "ModuleBind" TR_SOUND TR_PLAY_TRACK TR_PLAY_VALUE TR_SF_RESERVE TR_SF_PLAY_SCRIPT TR_SF_RESERVE TR_SF_BG_MUSIC TR_VVARIO TR_HAPTIC TR_SDCLOGS "Backlight\0" TR_SF_SCREENSHOT TR_SF_TEST TR_SF_CROSSTRIMS
 #elif defined(PCBGRUVIN9X)
   #define TR_VFSWFUNC                  TR_SF_SAFETY "Trainer\0  ""Inst. Trim""Reset\0    " TR_ADJUST_GVAR TR_SOUND TR_PLAY_TRACK TR_PLAY_BOTH TR_PLAY_VALUE TR_VVARIO TR_HAPTIC TR_SDCLOGS "Backlight\0" TR_SF_TEST
 #else
--- a/radio/src/translations/en.h.txt
+++ b/radio/src/translations/en.h.txt
@@ -313,8 +313,10 @@
 
 #define TR_SF_RESERVE                  "[reserve]\0"
 
-#if defined(CPUARM)
+#if defined(PCBTARANIS)
   #define TR_VFSWFUNC                  TR_SF_SAFETY "Trainer\0  ""Inst. Trim""Reset\0    ""Set \0     " TR_ADJUST_GVAR "Volume\0   " "SetFailsfe" "RangeCheck" "ModuleBind" TR_SOUND TR_PLAY_TRACK TR_PLAY_VALUE TR_SF_RESERVE TR_SF_PLAY_SCRIPT TR_SF_RESERVE TR_SF_BG_MUSIC TR_VVARIO TR_HAPTIC TR_SDCLOGS "Backlight\0" TR_SF_SCREENSHOT TR_SF_TEST TR_SF_CROSSTRIMS
+#elif defined(CPUARM)
+  #define TR_VFSWFUNC                  TR_SF_SAFETY "Trainer\0  ""Inst. Trim""Reset\0    ""Set \0     " TR_ADJUST_GVAR "Volume\0   " "SetFailsfe" "RangeCheck" "ModuleBind" TR_SOUND TR_PLAY_TRACK TR_PLAY_VALUE TR_SF_RESERVE TR_SF_PLAY_SCRIPT TR_SF_RESERVE TR_SF_BG_MUSIC TR_VVARIO TR_HAPTIC TR_SDCLOGS "Backlight\0" TR_SF_SCREENSHOT TR_SF_TEST
 #elif defined(PCBGRUVIN9X)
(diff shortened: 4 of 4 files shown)


3djc:
Why on earth would you want that???

mtwomey:
Mainly the reasons I listed, but the initial reason was I didn't like the fact that when I assign different trims to the sticks, they show up, visually (on the screen), next to the sticks they aren't affecting.

kilrah:
> Doesn't impact any other setting in the radio (just switches the buttons effectively)

That you think... that's actually absolutely terrible, the trim switches can be referred to by a ton of other things and you're breaking all of that.

Trims are sources, the keys are switches, they control GVAR adjusters... they can do way too many things (and multiple at the same time).

mtwomey:
> and you're breaking all of that

There is no evidence of that... at all.

The swap is right above the hardware abstraction layer, the rest of openTX acts as it always does (GVARS, Mixes, ...etc - are all unaffected). This is one benefit you get from using a hardware abstraction layer.

I certainly appreciate these are valid concerns, and I'd be happy to answer any questions or perform any specific tests if there are specific inquires.

kilrah:
The point is that it is currently a given that the name a trim bears always relates to the matching stick, it's consistency that gets broken and I fear many could get confused over it.
But since it's only when the function is active... why not.

mtwomey:
Ah - ok, I can understand that concern (good question):

> the name a trim bears always relates to the matching stick

This does not change with this implementation (e.g. this is still true, still a given). All that changes is that the 4 physical trim switches on the left are swapped with the four on the right (right above the GPIO layer). For example - selecting Rud, Ele, Thr, Ail for an input Trim will match the same trims they always have. This function simply swaps which physical buttons are used to control them.

schwabe:
By switching the switches right aboves the gpio layer you are changing the switches names. 

E.g. whenever I reference the SA switch it is *always* the SA switch. When I reference the throttle trim, it should be *always* the trim of the throttle (in my current mode). Your patch breaks that.

I am with Kilrah on this. The idea what you are doing is good but the way this is implemented breaks too many assumptions

mtwomey:
@schwabe Yeah, true. If you're referencing the trim switches as switches (as opposed to the trim values, then they'd be switched).

Mainly I want to be able to easily see what's going on w/cross-trims without digging through menus. I'd also strongly prefer to have the visual sliders show up under the sticks they're trimming (even when in cross-trim mode). This achieves both those goals - however, I'm open to other thoughts on a better implementation or an improvement to this one?

bsongis:
I will have a closer look to this one. The idea is good, but I also think that it should not be implemented in the HAL, but a layer above

bsongis:
Why a function for that? Don't you want it a model setting? Or even a radio setting?

RCdiy:
As a user and helping with documentation I'd say model setting

* Setup
  * Check box - Cross Trims

Now how do we explain this?

When Cross Tims is selected the trims with the left stick are now trims for the tight sticks. The trims for the right sticks are now for the left sticks. This way you do not have to take your fingers off the stick that is being trimmed or reach across with the other hand. Your hands stay in place and you retain a form grip on the radio.

mtwomey:
@bsongis I also have it coded as a model setting and even a radio setting. I'd be happy to submit those as well. Ultimately, I switched over to implementing it as a function because I felt it was more flexible - you could assign it to a switch if you wanted to easily turn it on and off, or you could assign it to an "always on" virtual switch if you didn't want to risk changing the state accidentally.

If not the HAL, can you clarify a little more specifically where you think it would be best and I'll update the request with an implementation there?

@RCdiy Yes - this is exactly where I had it in my other implementation (as a radio setup). I think your explanation is good. I'd maybe also add that the visual indicators of the trims will remain next to the sticks they are impacting (this was the key thing for me, the ability to quickly know which trims I was affecting without having to go back into the menu - to make absolutely sure).

jobinseattle:
Hello, I am new to OpenTX and recently purchased a HorusX12S and I am trying to figure out how to setup crosstrim. I have the latest OpenTX. I can do this right?

mcddd:
I'm a big fan of crossed trims and would like to see a simple selection box to enable crossed trims as a general radio setting.  Since I want crossed trims activated on all of my models, there is no reason to make the selection on a model by model basis.

The only request I disagree with is leaving the trim displays uncrossed.  I prefer the display also gets crossed and follows the trim tab and not remain with the stick function.  Maybe this could be a selection option?

Krist370:
I would rather have trims for right stick on third joystick

kelkarakshay:
How to add cross trim function in global function as shown by mtwomey

kelkarakshay:
Let me know the process or new firmwares

kilrah:
@kelkarakshay please stop spamming. There is no way to have it unless you build his branch yourself, and do the programming work needed to port it to the current codebase.

kelkarakshay:
Okey thanks for reply @kilrah

mtwomey:
> There is no way to have it unless you build his branch yourself, and do the programming work needed to port it to the current codebase.

I am still interested in supporting this PR.  If I could get some guidance on what updates I could do to make it more acceptable, I'd be happy to update it to the current version. There has been some degree of support for this concept. Several folks have asked me about this in response to my Youtube video.

kelkarakshay:
Yes we need cross trim function in next version of open tx

MarkusPfaff:
Hi, I changed to OpenTx (2.2.3) on Horus X12S from MPX some 6 weeks ago and feel perfectly at home already. Very nice work!

I'd very much like to have cross trim as a radio setting with the trim graphics still showing up beneath the sticks they belong to (i.e. the behavior mtwomey proposed). My view of cross trimming is the same as my view of stick mode. You decide for one and don't change modes from model to model. Whether trim graphics should show crossed or stay put at the sticks they belong to could also be chosen as a radio setting.

I tried to cross trims in 2.2.3, but failed with the "Throttle Trim Idle Only" option needed for IC powered models (already a [2.3 milestone](https://github.com/opentx/opentx/issues/5635)). When this will be fixed cross trim can be implemented on a per model basis. So no additional settings would be needed for a per model setting. The option for having the trim graphics beneath the sticks they belong to would still be adequate as a radio setting I think. Having this as a per model option would lead to confusion.

Horus X12S specific: As a workaround I crossed ail and rud trims via input settings, but let throttle trim stay beneath the throttle stick (left side in my case, mode2). Elevator trim source was set to Trim5 on my X12S to get the trim tab in reach of free thumb  at least. This worked in principle, but was of no practical use, because Trim5 does have flight mode specific values (which would be a nice enhancement for future OpenTx versions for Trim5 and Trim6).

kilrah:
T5/T6 do have flight mode specific values, you just can't see/adjust them in companion but on the radio they're like the others.

MarkusPfaff:
Andre, sorry for that, I overlooked it. Because flightmodes are much easier edited in the companion I use the transmitter seldom for that. Then this is a companion, not an OpenTx issue.

MarkusPfaff:
I'd like to share some further thoughts on the why for (and where of) cross trim as a radio setting: I consider cross trim as a submode to the stick mode. Whether the source "Ail" is the horizontal movement of the left or right stick depends on the stick mode (1, 2, 3, 4). There are lots of other theoretically possible stick modes one could think of, but only these have practical meaning (in OpenTX and other radio systems). The trim knobs are arranged in a similar fashion to the sticks (vertical vs. horizontal, left vs. right) and should also be assigned according to the chosen stick mode, i.e. with uncrossed trim the trim knob should be the one beneath the stick it sets the trim for. With crossed trim the according (horizontal, vertical) knob beneath the other stick sets this trim. Where crossed trims make it much easier to keep an untrimmed aircraft flying safely while at the same time trimming it which is why one would want the feature.

Somewhere in the software the stick mode is used to assign the left stick horizontal value sent from the hardware to either source Ail or Rud. Trim knobs should be handled the same way, not only taking stick mode into account, but also the non-crossed or crossed trim setting. I don't know enough about the structure of the OpenTx software to point directly to the file and the lines this is done in, but it should be evident that the trim knob mapping should take place in a similar place in this structure if one agrees with the idea of cross/noncross trim as a submode to stick mode.

When learning to use OpenTX some weeks ago I accepted that it does not let the user assign a horizontal stick movement to the source "Ele", because that does not make sense (at least for "non-square" persons). On the other hand I was puzzled why it is possible to use a horizontal trim knob to trim for "Ele", thus offering a far more general use model for the trim knobs than for the sticks. I don't talk about Trim5 and 6 which are not affiliated to any stick source by default, but may be used as general input devices of course.

Whether trim graphics should be where the trim knobs are or beneath the sticks they belong to could also be chosen as a radio setting.  One is trimming the function controlled by an input source of course, not the trim knobs. Suppose some future version of OpenTx where you might be able to trim by voice commands. Where would you put the trim graphics? Around your mouth or around the sticks for the functions you trim?

kilrah:
> When learning to use OpenTX some weeks ago I accepted that it does not let the user assign a horizontal stick movement to the source "Ele", because that does not make sense (at least for "non-square" persons). On the other hand I was puzzled why it is possible to use a horizontal trim knob to trim for "Ele", thus offering a far more general use model for the trim knobs than for the sticks. 

That's just a default name for the stick and is mostly irrelevant. If you want to use a horizontal stick for moving your elevator you do it in the same place you assign the trim, in inputs. In your Ele input you assign the Rud stick and there we go.
Flexibility is the same.

> Somewhere in the software the stick mode is used to assign the left stick horizontal value sent from the hardware to either source Ail or Rud. Trim knobs should be handled the same way

It already does obviously.

Really the Crosstrim option would bring nothing other than change the displayed location.

MarkusPfaff:
> > When learning to use OpenTX some weeks ago I accepted that it does not let the user assign a horizontal stick movement to the source "Ele", because that does not make sense (at least for "non-square" persons). On the other hand I was puzzled why it is possible to use a horizontal trim knob to trim for "Ele", thus offering a far more general use model for the trim knobs than for the sticks.
> 
> That's just a default name for the stick and is mostly irrelevant. If you want to use a horizontal stick for moving your elevator you do it in the same place you assign the trim, in inputs. In your Ele input you assign the Rud stick and there we go.
> Flexibility is the same.

From the user perspective the name a stick has is quite relevant. Also not only the result is relevant, but the way leading to a result is relevant too. The simpler the better. Consistency is also important. Even better if you can have both.

Using Ele source for an input that controls the ailerons would be possible, but rather confusing. If that scenario would be encountered more often (than almost never) in practical use sources would never have been named Ail, Ele,... I guess, but StickRH, StickLV or something like that and a radio setting of the mode would simply not be needed then, because you can associate sticks to inputs freely anyway. Of course you set the mode of a radio even physically by deactivating neutralization of the throttle stick, which makes mode a rather static and global setting. OpenTX reflects that and it should do so I think.

On the contrary argumenting that source names are irrelevant calls for removing the global mode setting and associating sticks to input functions without using the global mode setting, doesn't it?

The radio setting "mode" changes the association between a physical stick and its name as a source. From the user perspective I'd expect that trim knobs will also change accordingly (which they do already for uncrossed trim). In a cross trim setting trim knobs should simply change in a different way. Isn't this simply a submode to the stick mode? Wouldn't it be inconsistent to handle trims different than sticks?

Of course you still should be able to freely associate trim knobs to whatever input you want if a special use case has a need for that. Using the input setting on a per model basis makes that still possible.

> > Somewhere in the software the stick mode is used to assign the left stick horizontal value sent from the hardware to either source Ail or Rud. Trim knobs should be handled the same way
> 
> It already does obviously.

You may get the same result (although not in 2.2.3, but hopefully after the "Throttle Trim Idle Only" issue has gotten fixed), but you do it on a per model basis and by using a special setting for every input line of every model. This adds up to quite a lot of "special" instead of a single radio setting.
Using the input setting to implement cross trim the trim knob named for Ail use is actually used for Rud. 
Seen from the user perspective when using cross trim the trim knob below the aileron stick simply _is_ the Rud trim knob but it _is not_ the Ail trim knob. Thus the naming is or at least strongly feels wrong.
This is not the case for the global mode setting where the stick naming is already changed before it is put into an input line, i.e. on the source level. I see inconsistency for the cross trim situation between the way the sticks and the way the trim knobs are associated to their names.

> Really the Crosstrim option would bring nothing other than change the displayed location.
 
I see two possibilities for where the trim settings are displayed: 1. beneath the stick the trim controls, 2. in the location where the trim tabs are situated.
If you don't use crosstrim 1 and 2 simply merge.
If you use crosstrim you either might want option 1 _or_ option 2.
Display location is a subsetting of crosstrim. Crosstrim does not imply option 2 for the display location.

kilrah:
> If that scenario would be encountered more often (than almost never) in practical use sources would never have been named Ail, Ele,... I guess, but StickRH, StickLV or something like that

You can assign custom names to sticks if you prefer them different. 

> argumenting that source names are irrelevant calls for removing the global mode setting and associating sticks to input functions without using the global mode setting, doesn't it?

Sure, but for legacy reasons that's a bit hard to go muck around with. It works well for people who expect that to be there. Even the radios I have used that have "generic" stick names like J1-J4 still have a mode setting.

For me crosstrim as a radio setting makes no sense. People don't all use just fixed wing models on a radio.

If anything it has to be per model, and its only effect would be cosmetic, swapping trim names and display location.

MarkusPfaff:
> 
> 
> > If that scenario would be encountered more often (than almost never) in practical use sources would never have been named Ail, Ele,... I guess, but StickRH, StickLV or something like that
> 
> You can assign custom names to sticks if you prefer them different.
> 
For the sticks the mode setting does the job.

> > argumenting that source names are irrelevant calls for removing the global mode setting and associating sticks to input functions without using the global mode setting, doesn't it?
> 
> Sure, but for legacy reasons that's a bit hard to go muck around with. It works well for people who expect that to be there. Even the radios I have used that have "generic" stick names like J1-J4 still have a mode setting.
> 
> For me crosstrim as a radio setting makes no sense. People don't all use just fixed wing models on a radio.

That is definitely an argument for a model specific tick box of course. Even better would it be to use the model category for this setting (as far as I know categories do not exist in all OpenTx radios unfortunately). Category settings would be a nice enhancement for other uses also.

> 
> If anything it has to be per model, and its only effect would be cosmetic, swapping trim names and display location.

As I said I'd prefer not only to swap trim names but the associated function also. Perhaps I just misinterpret the exact meaning of your sentence. The behavior that mtwomey proposes seems attractive to me, although the way it achieves the effect making changes to the HAL layer may be inappropriate.

mcddd:
> For me crosstrim as a radio setting makes no sense. People don't all use just fixed wing models on a radio.

For me crosstrim as a radio setting makes perfect sense.  I can't think of any reason to revert back to standard trims no matter whether I'm flying fixed wing, rotor wing, or any aircraft.  Crosstrim even works great on cars and boats.  Once you become a fan of crosstrim, it feels very awkward to go back to standard trims.

Also thank you MarcusPfaff for reviving the crosstrim discussion!

MarkusPfaff:
> 
> 
> > For me crosstrim as a radio setting makes no sense. People don't all use just fixed wing models on a radio.
> 
> For me crosstrim as a radio setting makes perfect sense. I can't think of any reason to revert back to standard trims no matter whether I'm flying fixed wing, rotor wing, or any aircraft. Crosstrim even works great on cars and boats. Once you become a fan of crosstrim, it feels very awkward to go back to standard trims.

A cross trim checkbox in the radio settings would users who want it still only offer the _option_ to tick it. And users who don't want it not to tick it.

If you rather need to have cross trim on a per model basis, simply don't tick it, but instead go down the smaller and more rough road and use the input settings flexibility of OpenTx to implement your cross trim setting on every input line (again: only possible for IC powered models no earlier than from version 2.3 on, not possible in 2.2.3) .

If you don't like cross trim at all, then simply stay with the default, don't tick the option and you are done.

There are other radio options in OpenTx for which one might argue they should be model specific. Play Delay (really a good idea for every situation?), Stick Mode (if you use the radio as a trainer radio for a person using another stick mode), FAI mode, Voice Language (depending on who in the family is using the radio) are examples for such settings. And in most cases you are not even able to circumvent those settings on a per model basis like you would be with a radio global cross trim check box.

MarkusPfaff:
What about cross trims (the way you can do it in 2.2.3) and instant trim?
The last posting in [this thread](https://openrcforums.com/forum/viewtopic.php?t=9104) says it affects the wrong functions when trims are crossed via input line setting. The same for [this thread](https://openrcforums.com/forum/viewtopic.php?t=8574). This of course would be a severe flaw in OpenTx which will make a crash rather probable. I'm glad I've not "tested" this so far (and thus cannot confirm nor doubt the described behavior). It could be easily overlooked when setting up the model, but will for sure lead to a disastrous maiden flight.

Both threads also discuss whether cross trim should be a radio option by the way. The radio option is widely asked for by nearly all participants. All participants which mention cross trim in both threads except Kilrah wanted to setup _all_ their models cross trimmed. Most of them even argued explicitly for that.

RCdiy:
Once (if) cross trims becomes an easier to configure feature I see more people using it. The reason I would prefer it as per model (maybe with a default option to be crossed or not when a new model is created) is that if two people use the same radio then they can each have their own model. For example parent child, spouses. Per model allows flexibility. If as a radio option then it is all or nothing. 

Once one understands what cross trims are and used it a few times it makes total sense. You don’t have to have someone else click your trims or move the other hand to reach over. 

Maybe do it in companion first and gauge response.

mcddd:
If the crossed trim option can be set as the default per model, with the option to change it back to standard trims on a per model basis, then I'd be happy with that solution.  But that solution isn't really any different that selecting crossed trims as a main radio setting, with the option to change it on a per model basis.

If the ultimate solution to the crossed trim problem can only be solved by making it a main radio setting without the option to change it per model, then that solution works for me.  I can't think of any scenario where I would want to go back to standard trims on any RC vehicle.

Sharing one transmitter between two or more people to operate their own vehicle would be a rare occurrence.  Having different trim location preferences between those who share a single transmitter would be even rarer.  I don't see this argument as valid to avoid setting up the main radio itself for crossed trims.

mcddd:
The crossed trims issue when the TIO box is checked is not stale and still needs to be corrected.

MarkusPfaff:
Same for me. Cross trim is still high on my wish list, it is still much too hard to set up and it is still impossible to set up for IC powered models due to the Trim Idle Only error.

dgatf:
> The crossed trims issue when the TIO box is checked is not stale and still needs to be corrected.

This is a feature enhancement request... can not be corrected

mcddd:
> > The crossed trims issue when the TIO box is checked is not stale and still needs to be corrected.
> 
> This is a feature enhancement request... can not be corrected
>

I don't see how this is a feature enhancement since the software interface screens already allow for crossing the trims in the INPUTS, and the TIO function already uses a checkbox to enable or disable it.

But let's assume you are correct that it's not a bug, but a feature enhancement instead.  So what needs to happen differently here on Github in order for crossed trims combined with the Trim Idle Only (TIO) function to work correctly?

3djc:
Superseeded

PR #5099, opened 2017-07-20
Review SxR script to handle lastest receiver firmware revision

### LINKED ISSUE #5067: "convenient mode" missing from S6R Lua script
On June 5th, FRSKY introduced a new firmware with a "simple" or "convenient" mode that only allows for 2 flight modes and "off".

https://www.rcgroups.com/forums/showpost.php?p=37669372&postcount=3106

The Lua script appears to be missing the checkbox to enable this functionality. The PC app has a checkbox called "convenient mode" while FRSKY referred to the feature as "simple mode".

In other words, if a user updates the firmware on an s6r with this firmware file, they are still unable to select convenient mode by using the lua script.

3djc:
- Add 'quick' mode compatibility
- Add compatibility with previous receiver firmware (not blocking anymore, unsupported field/functions just keep indicating  '---')
- rename S6R to SxR to acknowledge S8R is supported too

This also fixes #5067


kilrah:
👍

Jmelz:
Tried this on a q x7 - the enable/disable is not displayed on the screen for simple mode. 

When scrolling the cursor, it acts as if there is a value there, but does not display on the screen (scrolling from top to bottom).

Using this

 firmware: https://www.frsky-rc.com/wp-content/uploads/2017/07/Firmware/Beta-S6R/S6R%20FW%20Convenient170615.zip

which is the beta firmware from this page: https://www.frsky-rc.com/s6r/ 

s6r

EDIT: Looks like I had the script copied incorrectly. I grabbed raw text and recreated - all is good (even on my other receivers that I had, so I didn't have to reflash anything). Thanks again for your patience.
--

PR #5107, opened 2017-07-24
Improve settings migration and management

kilrah:
Users now have a decent way to remove settings during uninstall, and import from older version is no more mandatory.


mpaperno:
You should be able to translate by wrapping the string in `QCoreApplication::translate()` function.

Though personally I'd avoid using GUI/interactive functions in a non-GUI class like AppData altogether.

kilrah:
The translate comment was about the NSIS uninstaller, not Qt

Agreed about the location, but I haven't been able to figure out how to do it cleanly...

bsongis:
I would split `importSettings` into 2 functions: the first one for checking if we can import some settings, and the second to do the job, so that the confirmation message box thing is not in this file

kilrah:
Thanks for the hint

bsongis:
Perhaps you could remove the whole part of code which crashes, it's so old ...

companion/src/companion.cpp
```diff
@@ -71,6 +71,20 @@ int main(int argc, char *argv[])
 
   CustomDebug::setFilterRules();
 
+  if (!g.hasCurrentSettings()) {
+    QString previousVersion;
+    if (g.findPreviousVersionSettings(&previousVersion)) {
+       QMessageBox msgBox;
+       msgBox.setText("We have found existing settings for Companion version: " + previousVersion + ".\nDo you want to import them?");
```
bsongis: No translations?

bsongis (approved):

companion/src/companion.cpp
kilrah: Ah yes forgot that...

kilrah:
> Perhaps you could remove the whole part of code which crashes, it's so old ...

Remove the whole import from c9x then? Cause it'll likely be pretty broken without that...

bsongis:
👍

PR #5110, opened 2017-07-24
Fixes #5090: periodically update sensor ratios from the radio when th…

### LINKED ISSUE #5090: Some sensors created with telemetry simulator do not update
Discovered while looking into problem reported by rcyoda on  http://openrcforums.com/forum/viewtopic.php?f=94&t=10355&p=130160#p130129

Tested with Companion2.2 x9d simulator
If a rxbat, A1, or A2 sensor is created by using the telemetry simulator it will not update.  Steps to recreate:

1. In Companion create or simulate a model with no sensors detected
2. Open telemetry simulator and add values for rxbt, A1, or A2
3. On radio simulator go to telemetry screen and "discover new sensors"
4. Sensors are discovered, but do not update.

  The reason for this is that the telemetry simulators ratio field is read and set from the existing sensors.  If there are no sensors it is set to 0.0.  This field is not editable so the sensor is created but will not update (shows 0).  Even if the ratio is changed on the radio simulator it will not update because the ratio on the telemetry simulator is still 0 and not editable.

I will create a pull request to allow the telemetry simulator rxbt, A1, and A2 ratio fields be editable.  This will allow the user to change those fields and add a ratio that will then have the sensor update.

Question:  Is this enough or should I add default ratios for the rxbat, A1, and A2 sensors and if so what should the defaults be.

projectkk2glider:
…e telemetry simulator is running.

Closes #5090

### DIFF
--- a/companion/src/simulation/telemetrysimu.cpp
+++ b/companion/src/simulation/telemetrysimu.cpp
@@ -211,9 +211,7 @@ void TelemetrySimulator::setupDataFields()
   SET_INSTANCE(accy_inst,     ACCY_FIRST_ID,          0);
   SET_INSTANCE(accz_inst,     ACCZ_FIRST_ID,          0);
 
-  ui->rxbt_ratio->setValue(simulator->getSensorRatio(BATT_ID) / 10.0);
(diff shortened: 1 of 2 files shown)


bsongis:
👍

PR #5111, opened 2017-07-24
Fixes #5093: usbasp udev were rules missing from the installation pac…

### LINKED ISSUE #5093: no device /dev/ttyACM0
Hi

I put my arduino in isp mode and the device /dev/ttyACM0 is not seen under Linux.
Anywhere I can put it in the .config file and it work fine but can you add it to the GUI.

Thanks

projectkk2glider:
…kage

I haven't tested this - no radio. But the file appears in the generated deb.

Closes #5093


bsongis:
Seems good

PR #5122, opened 2017-07-30
[Companion] Sensor switches support added

bsongis:
+ some refactoring to hide irrelevant switches in the combobox

More refactoring to come:
 - have only one SwitchModel per model instead of one per tab
 - only do the update work if something has changed


companion/src/eeprominterface.cpp
```diff
@@ -718,6 +718,15 @@ QString RawSwitch::toString(Board::Type board, const GeneralSettings * const gen
       case SWITCH_TYPE_TIMER_MODE:
         return CHECK_IN_ARRAY(timerModes, index);
 
+      case SWITCH_TYPE_SENSOR:
+        if (modelData && index <= CPN_MAX_SENSORS && strlen(modelData->sensorData[index-1].label) > 0)
+          return QObject::tr("Sensor%1 (%2)").arg(index).arg(modelData->sensorData[index-1].label);
```
3djc: I think this SensorX display is very confusing, since it doesn't exist one the radio, nor is it existing in telemetry screen. I think displaying only the name is more in line with the rest of the implementation, specialy on the radio

PR #5148, opened 2017-08-12
Neutralise BT on X9E, unless DEBUG is defined

3djc:
This is WIP but brings things is a somewhat 'releasable' state :)

### DIFF
--- a/radio/src/bluetooth.cpp
+++ b/radio/src/bluetooth.cpp
@@ -64,6 +64,7 @@ char * bluetoothReadline()
 
   while (1) {
     if (!btRxFifo.pop(byte)) {
+      TRACE("NO RESPONSE FROM BT MODULE");
       return NULL;
     }
     TRACE_NOCRLF("%02X ", byte);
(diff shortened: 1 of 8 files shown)


radio/src/bluetooth.cpp
```diff
@@ -64,6 +64,7 @@ char * bluetoothReadline()
 
   while (1) {
     if (!btRxFifo.pop(byte)) {
+      TRACE("NO RESPONSE FROM BT MODULE");
```
bsongis: It will fire a lot of traces, no?

radio/src/opentx.h
```diff
@@ -194,7 +194,7 @@
   #define CASE_PCBX9E(x)
 #endif
 
-#if defined(BLUETOOTH)
+#if defined(BLUETOOTH) && !(defined(PCBX9E) && !defined(DEBUG))
```
bsongis: difficult to read!

```
#if defined(BLUETOOTH) && (!defined(PCBX9E) || defined(DEBUG))
```

radio/src/bluetooth.cpp
3djc: It did not at all on X9E, only when sending a CLI command to module and module would not respond. Will do some tests with Horus

3djc: Given I cannot test with two radio connected, I made a change that should ease your concern

bsongis:
👍

PR #5153, opened 2017-08-13
Option to boot directly to telemetry screen

papaxiclas:
Seems there is no way to boot directly to a telemetry screen?
If we are able to customize a telemetry screen for our needs I think it should be possible to power the radio directly to that screen instead of booting to a main screen that doesn't have any of the info we need.  There are times that I find myself pressing the page button before even powering the radio.
Looking on rcgroups for this option a few users showed interest in this.
Thanks


schwabe:
Creating a pull request is not the right way to report an issue.

papaxiclas:
Is not an issue, is a feature request?!

2017-08-13 12:35 GMT+01:00 Arne Schwabe <notifications@github.com>:

kilrah:
Pull requests are a way to submit your own code for inclusion to the project. Say if you coded that feature you want and wanted to submit it to us.

Issues are for both problems and enhancement suggestions.

papaxiclas:
aaaa ok, sorry for my mistake and to waste your time:)
O will re-submite in the Issues section.

thanks


Em 13/08/2017 14:35, "Andre Bernet" <notifications@github.com> escreveu:

PR #5179, opened 2017-08-22
Add SQT5 that was missing in 2.2

### LINKED ISSUE #5142: sqt5font
No font in 2.2.0 release. Flashed FW with font, but used default

3djc:
This closes #5142

image

image


radio/src/CMakeLists.txt
```diff
@@ -336,6 +337,10 @@ if(FRSKY_STICKS)
   add_definitions(-DFRSKY_STICKS)
 endif()
 
+if(FONT)
```
bsongis: I don't like too much the option, I would prefer something -DFONT=SQT5 (with automatic conversion from SQT5 to sqt5)

bsongis (changes requested):

bsongis (approved):

Issue #2050, opened 2015-01-30
Less memory available for Lua in 2.0.15


projectkk2glider:
People on forums report that some Lua scripts won't run anymore on 2.0.15 and that the reason is less memory. We need to check this. 

On the other hand it might be something else in script that is wrong.

bsongis:
NANO is available

on4mj:
I'm one of those people :-)
My <a href="http://rcsettings.com/index.php/viewdownload/13-lua-scripts/179-f3k-training">F3K training script</a> stopped working on the TX when 2.0.15 was released (but it would still work fine in Companion).

I didn't keep the notes I made when I solved the problem, so I don't have exact figures anymore.
What I can tell is that I used the new debugging feature in Companion to get memory usage data from the GC.  Sorry about that.

I first reduced the memory usage in my previous version by about 400 bytes.  The script would start again, but crashed along the way as more memory was allocated.

In the end, I gained about 1.000 bytes with diverse space optimizations (mostly removing 'closures') and that definitely solved the problem.  I could even reintroduce a couple of new features.

In the process, it seems that I also hit an arbitrary limit put on the variables amount (number of vars and functions global to the script, I guessed).

That those errors are reported as "Lua syntax error" doesn't help to identify the problem :-/

bsongis:
OK. I am afraid it's the NANO=YES option that we introduced in 2.0.15 which makes the difference. Is it the same in the nightly builds (stable ones, 2.0)? If yes, I will try to go back to NANO=NO to see if it fixes your problem. I wait for your test on nightly build then.

on4mj:
Did it.  Unfortunately, I can't draw any conclusion from that test.  Let me explain :

1/ I got the nightly (21 march 15) build and installed it in the radio.
2/ I uploaded the previous version of my script (the one which didn't run anymore in 2.0.15) in my test model.
3/ Result : "syntax error".
4/ Then I tried the latest version of my script (the one which runs in 2.0.15).  It doesn't work either in the nightly build.  So I checked and saw that that build includes the heli support.

Someone reported to me once that the script didn't work on his T+ in 2.0.15.  After checking the noheli option, and reflashing, he got it working.

Did the script already break in 2.0.13 with the heli support ?  I have no idea :-(

Conclusion : I can't really compare if there's a difference between the official build and the nightly one, unless there's one with the noheli option.

mpaperno:
[...]

PR #5180, opened 2017-08-22
Switch to nano version of newlib

schwabe:
save about 50k of flash and 2k of RAM on Taranis


projectkk2glider:
Looks good. 

I remember we had some issues with newlib-nano before, but look like it was used in 2.1 branch in the end. I will just reference these issues here so we can found them easier if any problems arise in the future:
 * #2050 (comparison of actual usable memory for LUA scripts with different libs)
 * #1597 (memory leak, probably related to specific nano lib version)

schwabe:
I changed the pull request so that you have to use -DNANO to use newlib nano and binalloc. That should allow to test the feature until we deem it safe it enough to enable it always

PR #5185, opened 2017-08-23
Show telemetry UART errors in statistics debug page

schwabe:


projectkk2glider:
Have you considered putting this into Debug page (menuStatisticsDebug), it seems more logical place for it and you could then also implement the reset of the value on long key press ENT. Unless there is not enough place on LCD on all radios.

radio/src/gui/128x64/radio_diaganas.cpp
```diff
@@ -70,6 +70,9 @@ void menuRadioDiagAnalogs(event_t event)
     lcdDrawTextAlignedLeft(MENU_HEADER_HEIGHT + 1 + (NUM_STICKS+NUM_POTS+NUM_SLIDERS+1)/2 * FH + 1 * FH + 2, "RAS");
     lcdDrawNumber(LEN_CALIB_FIELDS*FW+FW, MENU_HEADER_HEIGHT + 1 + (NUM_STICKS+NUM_POTS+NUM_SLIDERS+1)/2 * FH + 1 * FH + 2, telemetryData.swr.value, LEFT);
   }
+    // UART statistics
```
bsongis: Not here, but rather in DEBUG page

radio/src/gui/212x64/radio_diaganas.cpp
```diff
@@ -40,13 +40,15 @@ void menuRadioDiagAnalogs(event_t event)
     lcdDrawNumber(x+10*FW-1, y, (int16_t)calibratedAnalogs[CONVERT_MODE(i)]*25/256, RIGHT);
 #endif
   }
+  // UART statistics
```
bsongis: would also be better in DEBUG page

radio/src/gui/480x272/view_statistics.cpp
```diff
@@ -186,6 +186,10 @@ bool menuStatsAnalogs(event_t event)
 
   }
 
+  // UART statistics
```
bsongis: same comment here

schwabe:
Debug statistics page has no more space on X9D+ at least. Should I add another debug page?

schwabe:
New version with a second debug page.

kilrah:
Hmm it seems navigation is a bit broken on X7, can't get back to the STATS screen once on DEBUG...

EDIT: Going back and forth between the 2 debug screens on X7 causes a crash due to the repeated pushMenus that don't return (menus stack overflow is my guess)

3djc:
Navigation for x7 and x9 changed to:
- UP/DOWN or PAGE/PAGE LONG to move from one screen to another
- on statistics screen : long MENU does reset total timer (need to be kept for historical reasons)
- on statistics screen : long ENTER does also reset total timer (added for consistency)
- on debug screens : long ENTER resets data

PR #5187, opened 2017-08-29
Improve pilotPosition when internal GPS is available

3djc:
For radio with internal GPS, pilotPosition is updated using intenal GPS data.
When internal GPS doesn't lock works as before.


bsongis:
👍

PR #5189, opened 2017-08-31
Horus : improve add widgets

### LINKED ISSUE #4257: UI Behaviour proposition: Open Widget settings after adding widget with options
When i add a widget with options (Value for example) i always open the settings. It would save a lot of time (when adding multiple widgets) if the settings screen is opened automatically after adding the widget.

3djc:
Automaticaly enter widget settings when a widget is added

This fixes #4257


bsongis:
👍

PR #5192, opened 2017-09-02
Trying to make the code more readable

ghost:
Hi all,

I'm running some C code analysis tool that pointed me to this small code fragment. I'm suggesting this patch to try to make the source code more understandable.

Would you guys agree with this type of changes? The tools pointed me to almost hundreds cases like this one in the source code of opentx.

### DIFF
--- a/radio/src/debug.h
+++ b/radio/src/debug.h
@@ -348,8 +348,12 @@ private:
   uint32_t _start_loprec;
 
   void evalStats() {
-    if (min > last) min = last;
-    if (max < last) max = last;
+    if (min > last) { 
+        min = last;
+    }
+    if (max < last) {
+        max = last;
+    }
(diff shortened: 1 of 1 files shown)


powercycle:
The original code doesn't seem unreadable to me.  Why does adding braces make it more readable?  Seems like the tool you are using is overly strict.  BTW, which tool is it?

radio/src/debug.h
```diff
@@ -348,8 +348,12 @@ class DebugTimer
   uint32_t _start_loprec;
 
   void evalStats() {
-    if (min > last) min = last;
-    if (max < last) max = last;
+    if (min > last) {
```
bsongis, line 351: I don't know if it's more readable, but it doesn't respect our indentation (2 spaces)

PR #3606, opened 2016-06-17
3djc/Mixer monitor

3djc:
Add two mixer monitor screens to the monitor section

PR #5194, opened 2017-09-04
Crossfire baudrate selection for X7 support ( fixes #4922 )

### LINKED ISSUE #4922: X7 - Crossfire constantly telemetry warning
Had an interesting behaviour on a X7 from a friend. Seems that Crossfire telemetry generate constant report telemetry lost/regained.

X7 running RC18 (no-heli, lua, diy module)

The behavior is similar to described in. 
http://fpvlab.com/forums/showthread.php?42097-TBS-CROSSFIRE-Next-Generation-Long-Range-R-C-System&p=832994&viewfull=1#post832994

Crossfire module run on latest stable but even on the beta behavior is the same. This seems once the receiver connect (crossfire telemetry is enabled). Module works fine in X9D plus. PPM also run error free on X7

sirPerna:
Requires TBS Crossfire V2.07 or higher.


sirPerna:
fixes #4922

ghost:
@sirPerna What do you mean? have TBS? resolved this issue with upgrade to crossfire?

3djc:
I think what he means is the solution is to run at reduced speed, which requires both a change at TBS end (included in 2.07 and higher), AND ALSO one in OpenTx (this patch)

sirPerna:
@gustavlindgren0 Yes. This is more a workaround until FrSky fix properly it in hardware.  I will upload a crossfire update once this is merged together with some guidelines.

ghost:
@3djc @sirPerna ok thanks a lot guys :)

radio/src/gui/128x64/model_setup.cpp
```diff
@@ -921,6 +921,23 @@ void menuModelSetup(event_t event)
             }
           }
         }
+        else if (IS_MODULE_CROSSFIRE(moduleIdx)) {
+          lcdDrawTextAlignedLeft(y, INDENT "Baudrate");
+          lcdDrawNumber(MODEL_SETUP_2ND_COLUMN, y, CROSSFIRE_BAUDRATES[g_eeGeneral.mavbaud], attr|LEFT);
+          if (attr) {
+            g_eeGeneral.mavbaud = DIM(CROSSFIRE_BAUDRATES) - 1 - checkIncDecModel(event, DIM(CROSSFIRE_BAUDRATES) - 1 - g_eeGeneral.mavbaud, 0, DIM(CROSSFIRE_BAUDRATES) - 1);
```
schwabe: I don't think you should reuse mavbaudrate here. Module specific data should be in ModuleData. PXX and multi are already saving their extra settings there and I think crossfire should do the same.

radio/src/targets/taranis/telemetry_driver.cpp
```diff
@@ -114,6 +114,7 @@ void sportSendBuffer(uint8_t * buffer, uint32_t count)
   DMA_Cmd(TELEMETRY_DMA_Stream_TX, ENABLE);
   USART_DMACmd(TELEMETRY_USART, USART_DMAReq_Tx, ENABLE);
   DMA_ITConfig(TELEMETRY_DMA_Stream_TX, DMA_IT_TC, ENABLE);
+  USART_ClearITPendingBit(TELEMETRY_USART, USART_IT_TC);
```
schwabe, line 118: I am a bit unsure why this now needed and wasn't needed before.

radio/src/gui/128x64/model_setup.cpp
```diff
@@ -921,6 +921,23 @@ void menuModelSetup(event_t event)
...
+            if (checkIncDec_Ret) {
+              pauseMixerCalculations();
+              pausePulses();
+              EXTERNAL_MODULE_OFF();
+              CoTickDelay(10); // 20ms so that the pulses interrupt will reinit the frame rate
+              telemetryProtocol = 255; // force telemetry port + module reinitialization
+              EXTERNAL_MODULE_ON();
+              resumePulses();
+              resumeMixerCalculations();
+            }
```
schwabe: Messing with timers and so here seems very ugly. I would have have in telemetryWakeup check if the baudrate is still right as part of the checks if telemetry needs to reinitalised.

radio/src/telemetry/crossfire.h
```diff
@@ -41,4 +41,21 @@ void processCrossfireTelemetryData(uint8_t data);
 void crossfireSetDefault(int index, uint8_t id, uint8_t subId);
 bool isCrossfireOutputBufferAvailable();
 
+#if defined(PCBX7)
```
schwabe, line 44: Is there a good reason to restrict this to X7? I vaguely that some user with a Horus also had problem with Crossfire.

radio/src/gui/128x64/model_setup.cpp
kilrah: Very much not, this would (if the setting is 1) cause a control loss on Horus in case of emergency mode (that var is not backed up).
(I know it's only on X7 at the moment, but think it should be implemented on all platforms)

schwabe:
@gustavlindgren0 I think this need a major revision to be included, if you need help or pointers just drop by in our chat.

sirPerna:
@schwabe I made some changes. Please review them. Thank you.

radio/src/gui/128x64/radio_hardware.cpp
```diff
@@ -205,6 +206,24 @@ void menuRadioHardware(event_t event)
         break;
       }
 
+      case ITEM_RADIO_HARDWARE_SERIAL_BAUDRATE:
+        lcdDrawTextAlignedLeft(y, "Max bauds");
+        lcdDrawNumber(HW_SETTINGS_COLUMN2, y, CROSSFIRE_BAUDRATES[g_eeGeneral.mavbaud], attr|LEFT);
+        if (attr) {
+          g_eeGeneral.mavbaud = DIM(CROSSFIRE_BAUDRATES) - 1 - checkIncDecModel(event, DIM(CROSSFIRE_BAUDRATES) - 1 - g_eeGeneral.mavbaud, 0, DIM(CROSSFIRE_BAUDRATES) - 1);
```
projectkk2glider: It is true that `g_eeGeneral.mavbaud` is only used on AVR radios when MAVLINK compilation option is enabled. But using it here for Crossfire I think is not appropriate. At least it needs a generous comment in datastructs. 

I would much more prefer a more clean solution, perhaps an ifdef and a different name for this variable for ARM platforms.

radio/src/gui/128x64/radio_hardware.cpp
```diff
@@ -205,6 +206,24 @@ void menuRadioHardware(event_t event)
         break;
       }
 
+      case ITEM_RADIO_HARDWARE_SERIAL_BAUDRATE:
+        lcdDrawTextAlignedLeft(y, "Max bauds");
```
projectkk2glider: Translation missing

projectkk2glider:
@sirPerna I added a couple of comments. Tell us if you need help with translations.

I also think this should be available on all radios that have Crossfire option. Lets see what @bsongis has to say.

radio/src/gui/128x64/radio_hardware.cpp
sirPerna: ok

sirPerna:
@projectkk2glider Thank you for reviewing it. Any help with translations is appreciated.

radio/src/gui/128x64/radio_hardware.cpp
sirPerna: I propose to rename mavbaud to telemetryBaudrate. It might be more accurate.

projectkk2glider:
Try to see how translated strings are used in other similar functions. All translated strings begin with `STR_`. We have a helper script `addtr.py` to add new translation string to all languages, see https://github.com/opentx/opentx/pull/3606#r67580771

radio/src/gui/128x64/radio_hardware.cpp
projectkk2glider: Use
```c
#if defined(CPUARM)
#else
#endif
```

See how `SPLASH_MODE` is defined in `datastructs.h`.

bsongis:
Here is the command to add a translation to all translations files:
```
cd radio/src
../util/addtr.py TR_MAXBAUDRATE="Max bauds" TR_TIME
```

bsongis:
About the Horus (X12S) we use DMA on the serial port (for both Tx and Rx), I am still not sure there is an issue on this serial port, or an issue in the trigger (to switch from Rx to Tx and vice-versa)

radio/src/gui/128x64/radio_hardware.cpp
sirPerna: I used IF_AVR / IF_ARM macros instead. Hope this is fine.

t413:
@sirPerna you rock. Any details on how one gets the Crossfire 2.07 firmware? It's not mentioned in any of the TBS forums or available via unstable releases in TBS Agent.

Issue #5069, opened 2017-07-01
QX7 drawPixmap enable ?


aurelx03:
Hi,
I would like know if its possible to enable drawPixmap for the QX7, i know resolution of qx7 screen is small but need to upload small image for custom screen script or other. It is a pity to disable a function between 2 radios.

Kiss friends <3

3djc:
Not only is the resolution lower, but it does not do greyscale either, it is pure 1 bit bitmaps, not helpful at all

aurelx03:
yeah but for example you would draw an icon or a splash screen for your script u must draw the picture pixel by pixel its annoying

rdeanchurch:
One bit on the low res screen is not perfect, 
but it beats nothing by a country mile.
Please consider putting drawbitmap for Qx7.
![deansplashscreen](https://user-images.githubusercontent.com/4843199/28174154-8be8e9a6-67a5-11e7-89b3-c6d7d17bffd2.PNG)

RCdiy:
Better than nothing. At least it gives us a choice to live with the image or not.

3djc:
Done in https://github.com/opentx/opentx/commit/541aa459cf5e368dbff916553e4124abc43e2fb9

PR #5198, opened 2017-09-06
Taranis Q X7 optimized wizard

NiklasVoigt:
I redesigned the wizard for the X7!

Check this out:
https://www.youtube.com/watch?v=86H1YKcVyB4

After the implementation of issue #5069 drawbitmap will be implemented too! :-)

### DIFF
--- a/radio/sdcard/taranis-x7/SCRIPTS/WIZARD/delta.lua
+++ b/radio/sdcard/taranis-x7/SCRIPTS/WIZARD/delta.lua
@@ -1,0 +1,366 @@
+-- Delta Wizard pages
+local ENGINE_PAGE = 0
+local ELEVONS_PAGE = 1
+local RUDDER_PAGE = 2
+local CONFIRMATION_PAGE = 3
+
+-- Navigation variables
+local page = ENGINE_PAGE
+local dirty = true
+local edit = false
+local field = 0
+local fieldsMax = 0
+
+-- Model settings
+local engineMode = 1
+local thrCH1 = 0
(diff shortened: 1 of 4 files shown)


3djc:
Quite nice. There seems to be an issue when the drop down and other text :
image

NiklasVoigt:
Fixed :-)

3djc:
Yes, much nicer :). I think the botttom wiz botttom line should go away (the ModWiz...), and since we do not have heli, maybe that should not be looking like an option too (remove since there is not way to 'grey out' ?)

NiklasVoigt:
That sound good! I've removed the bottom bar and uncommented heli (since there is not way to 'grey out' ) for easier implementing in the future. :-)

mythreeleggedcat:
Just tried this on 2.2.  Still testing it but good job!  One issue I noticed is the final Long Press Enter to save the model always brings up the Backup|Copy|Move menu which you then have to dismiss!

RCdiy:
@NiklasVoigt could you help me get this set up prior to 2.2.1 release.... I'd like to document it...

schwabe:
@RCdiy it is part of the nightly builds (+nightly sdcard contents)

RCdiy:
@NiklasVoigt 
What is GV Use:
It shows up when Menu is long pressed
How does one remove it once it shows up
screenshot_x7_17-10-12_12-11-39

Mikey1968:
You should add retracts as well , Not remember if the flaps were there as well

PR #5218, opened 2017-09-16
Fix FrSky D cable telementry for 9XR PRO and DJT module.

rodericktaylor:
Telemetry  wasn't working on the 9XR Pro with the DJT module and the cable from connecting it to the  futaba trainer port.

I've assumed the NOBACKUP macro in datastructs.h prevents the data from being written out so I didn't have to bump the EEPROM_VER.

### DIFF
--- a/radio/src/datastructs.h
+++ b/radio/src/datastructs.h
@@ -954,6 +954,7 @@ PACK(struct TrainerData {
 #elif defined(PCBSKY9X)
   #define EXTRA_GENERAL_FIELDS \
     EXTRA_GENERAL_FIELDS_ARM \
+    NOBACKUP(uint8_t  serial2Mode:4); \
     int8_t   txCurrentCalibration; \
     int8_t   temperatureWarn; \
     uint8_t  mAhWarn; \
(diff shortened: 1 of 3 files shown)


radio/src/datastructs.h
```diff
@@ -954,6 +954,7 @@ PACK(struct TrainerData {
 #elif defined(PCBSKY9X)
   #define EXTRA_GENERAL_FIELDS \
     EXTRA_GENERAL_FIELDS_ARM \
+    NOBACKUP(uint8_t  serial2Mode:4); \
```
bsongis, line 957: ouch, really? you break all existing SKY9X / 9XR-PRO EEPROM with this line!

bsongis (changes requested):

radio/src/datastructs.h
schwabe: Also if eeprom is changed, we need changes to companion as well

kilrah:
NOBACKUP is a Horus thing and related to emergency mode, so that doesn't apply at all.

Besides I don't see where you'd be using that new setting...

schwabe:
I thinke the #define SERIAL2 triggers using that setting

rodericktaylor:
Thanks for the feedback. I'll have another go.

Issue #2894, opened 2015-09-18
Feature request: Selectable serial outputs instead of PPM


DTFUHF:
Expanding on the DSM2/DSMX serial protocol that is already available but limited to 6 channels, I'd like to see some serial protocols available on the PPM output pin in the module bay. Specifically SUMD, but also SBUS, expanded Spektrum, etc. would ensure a broad range of compatibility with radio modules that accept a serial input. PXX is difficult to decode on lower end hardware (AVR) and has not seen widespread acceptance. PPM latency is high vs. these serial protocols, especially on 16 channels.

fujin:
++++:100:

mlyle:
Or just turn up the DSM to use more channels. :P

DTFUHF:
I'm not able to do this, it's limited to 6 channels.

mlyle:
@dtfuhf -- not sure I understand.  http://wiki.paparazziuav.org/wiki/DSM  The actual protocol appears capable of carrying 18 channel streams.

DTFUHF:
Yes, but I'm unable to do this in OpenTX.

bsongis:
Why not SBUS on the PPM Pin.
Or perhaps we could use the S.PORT Pin for sending pulses in whatever serial format (there is an hardware serial on this pin), and Heartbeat Pin for receiving Telemetry. 
Because bitbanging on the PPM Pin uses CPU ...

DTFUHF:
I would definitely be open to any form of serial TX on the module bay pins.

kilrah:
The only 2 pins we can drive in there are pin 1 (usually PPM), which in terms of hardware driving possibilities is a timer output, and pin 5 (smart port), multiplexed, inverted serial port.

On pin 2 we can receive only, inverted serial port.

MikeBland:
Is this related to this project?: http://www.rcgroups.com/forums/showthread.php?t=2165676.
Er9x and ersky9x (including on the Taranis) are supporting this serial protocol, on the PPM pin. It is a variant of the DSM hack protocol.
Also, for serial receive, on ersky9x I have "bit-bang" code working to receive serial data, both inverted and non-inverted, on a processor pin. This works by making sure all other interrupts are a lower priority, then using an "interrupt on pin change" interrupt at the highest priority to detect the edges on the serial input and decode them into a serial data stream.

phobos-:
[...]

PR #5225, opened 2017-09-19
Implement SBUS output for external module bay.

schwabe:


schwabe:
image
Current implementation how warn users of the voltage level when SBUS is selected.

yanivasy:
By using the battery UART port as an another alternative for the SBUS output, its possible to get standard SBUS voltage levels.

If a beta version was already implemented,  I'll be happy to test it give a feedback.

kilrah:
Since the pull request hasn't been merged yet you need to build yourself.

companion/src/firmwares/opentx/opentxeeprom.h
```diff
@@ -98,15 +98,11 @@ class ProtocolsConversionTable: public ConversionTable
...
         addConversion(PULSES_DSMX, val++);
       }
-      if (IS_HORUS_OR_TARANIS(board)) {
+      if (IS_ARM(board)) {
         addConversion(PULSES_CROSSFIRE, val++);
         addConversion(PULSES_MULTIMODULE, val++);
         addConversion(PULSES_PXX_R9M, val++);
-      }
-      else if (IS_ARM(board)) {
-        val++;
```
bsongis, line 107: it was for removing CROSSFIRE support from SKY9X but perhaps we could add it soon

radio/src/opentx.h
```diff
@@ -447,14 +444,20 @@ void memswap(void * a, void * b, uint8_t size);
...
 #endif
 #define IS_MODULE_R9M(idx)                (g_model.moduleData[idx].type==MODULE_TYPE_R9M)
 #define IS_MODULE_PXX(idx)                (IS_MODULE_XJT(idx) || IS_MODULE_R9M(idx))
 
+#if defined(DSM2)
+  #define IS_MODULE_DSM2(idx)             (idx==EXTERNAL_MODULE && g_model.moduleData[EXTERNAL_MODULE].type==MODULE_TYPE_DSM2)
+  #define IS_MODULE_SBUS(idx)             (idx==EXTERNAL_MODULE && g_model.moduleData[EXTERNAL_MODULE].type==MODULE_TYPE_SBUS)
+#else
+  #define IS_MODULE_DSM2(idx)             (false)
+#endif
```
bsongis, line 459: no `IS_MODULE_SBUS()` here?

bsongis (approved):

companion/src/firmwares/opentx/opentxeeprom.h
schwabe: Yeah, but it only does the mapping. If a protocol is actually availabe is on another place. And on the Sky9x and Taranis the eeprom ids are identical anyway, so I remove that distinction.

yanivasy:
@bsongis , from what I see, your version in #2894 was using the S. Port pin  which later was used for CrossFire modules.
Correct me if I'm wrong but @schwabe new version uses the PPM pin.

bsongis:
Crossfire and SBUS are not the same protocol, Crossfire is bidirectional

yanivasy:
> Crossfire and SBUS are not the same protocol, Crossfire is bidirectional

@bsongis , I never claimed that (see my first comment on #5236). Just wanted to point out to difference in implementation of SBUS output with your version a couple of years ago and @schwabe's version now a days.

yanivasy:
I'll be very grateful for a compiled build for Taranis+ X9D plus along with Multimodule (including SBUS) & sqt5font as build options for testing

schwabe:
Just download the  nightly build it has sbus already in it

yanivasy:
> Just download the nightly build it has sbus already in it

@schwabe , Starting from Nightly build companion-windows-2.2.1N391.exe ( thru 2.2.1N392 and including latest 2.2.1N393) there is  a problem with Companion updates.

A popup error with "**Unable to check for updates**" appears.

(Although I set "Use firmware nightly builds" & "Use Companion nightly builds" check box within the "Application Settings" tab).

Currently I was able to build only using 2.2.1N390 or an older version of Companion (for windows).
I double checked it with two different OS (win7 & win10).

schwabe:
@yanivasy There is a problem with SSL in newer update url. Just install a firmware from your N390, the build server should still give you a current nightly

yanivasy:
I've tested the sbus functionality with Taranis+and it working well. Thank's a lot.
Please include the SBUS output feature within 2.2.1.

3djc:
Since it is merged, it will be in 2.2.1

Issue #5217, opened 2017-09-16
The ordinal 3873 could not be located in the dynamic link library ... when starting companion 2.2.0


dlondon2017:
Hi,

Just installed Companion 2,2,0 on Windows 10

I get popup errors when I start companion:

The ordinal 3873 could not be located in the dynamic link library C:\Program Files (x86)\iis express\PHP\v5.4\ssleay32.dll
then
The ordinal 3873 could not be located in the dynamic link library C:\Program Files (x86)\PHP\v5.4\ssleay32.dll

Any ideas on how to fix?

Regards,

David

projectkk2glider:
If we are using OpenSSL library, that dll should be included into installation dir. This way the system will not use other ssleay32.dll which in this case is apparently not-compatible - ie wrong verson.

@kilrah and @3djc  you are more familiar with windows build, any clue.

3djc:
We have add a couple of those; I think that's conflicting ssl stacks on the PC. I thought that @mpaperno made a change to have it incorporated

kilrah:
I think there were talks of it but nothing actually done. No problem to do it, question is what version should we bundle and where to source it from for "build environment friendliness"...

dlondon2017:
Thanks for looking into this.  I did try downloading openssl-1.0.2l-x64_86-win64.zip and placing ssleay32.dll in C:\Program Files (x86)\OpenTX\Companion 2.2 and also C:\Windows\System32 but I guess those crude attempts were not enough because the same errors came up

Any idea if Companion will work  despite this error message?  It does open.

Regards,

David

projectkk2glider:
I can't find why we even need SSL lib. Anyone know? 

Looking at the cmake we only try to include that ddl in mingw build:
```
  elseif(MINGW)
    get_filename_component(MINGW_DIR ${CMAKE_CXX_COMPILER} PATH)
    if(IS_DIRECTORY ${MINGW_DIR})
      install(FILES "${MINGW_DIR}/libgcc_s_dw2-1.dll" "${MINGW_DIR}/libstdc++-6.dll" "${MINGW_DIR}/libwinpthread-1.dll" DESTINATION ${INSTALL_DESTINATION})
      set(MINGW_DIR "${MINGW_DIR}/../opt/bin")
      # SSL support
      install(FILES "${MINGW_DIR}/libeay32.dll" "${MINGW_DIR}/ssleay32.dll" DESTINATION ${INSTALL_DESTINATION} OPTIONAL)
[...]

PR #5247, opened 2017-09-28
Move to https

3djc:
Server configured for ssl handling


projectkk2glider:
Now that we use https #5217 needs to be fixed!

PR #5251, opened 2017-09-29
[X10S] PWM sticks support - with autodetection

bsongis:
Review and comments appreciated!


3djc:
Hardware tested x9e,  no issue

bsongis:
it remains that sticks are inverted

schwabe:
@bsongis I can spend some time polishing @3djc and my approach to learn stick direction while doing hw calibration to make it ready to include in normal OpenTX.

We could then always do the STick direction detecting on X10 and optionally on the other remotes

bsongis:
It's not needed here, but why not!

PR #5051, opened 2017-06-22
x7/x9 GPS datetime fix

3djc:
This fixes #5026

image

image

PR #5254, opened 2017-09-30
GPS datetime display fixes

3djc:
Fixes glitches I introduced in #5051 

For x7, in telem setup, date and time are alternating, and in telem view, time is displayed in full, but only day-month are displayed

image

### DIFF
--- a/radio/src/gui/128x64/lcd.cpp
+++ b/radio/src/gui/128x64/lcd.cpp
@@ -1147,11 +1147,11 @@ void drawDate(coord_t x, coord_t y, TelemetryItem & telemetryItem, LcdFlags att)
      lcdDrawNumber(lcdNextPos, y, telemetryItem.datetime.sec, att|LEADING0, 2);
   }
   else {
-    lcdDrawNumber(x, y, telemetryItem.datetime.day, att|LEADING0|LEFT, 2);
+    lcdDrawNumber(x, y, telemetryItem.datetime.year, att|LEADING0|LEFT, 4);
     lcdDrawChar(lcdLastRightPos, y, '-', att);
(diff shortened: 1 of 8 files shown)


bsongis:
perhaps british will prefer 09-30, no?

bsongis:
@robthomson ?

robthomson:
I will check if this occurs on my setup.  Sofar my telemetry on the X7 with crossfire has been rock solid.

bsongis:
@robthomson just for the date format, the screenshot is above, if it's good for everybody I press the button

robthomson:
Just checked and I dont see these values at all.

I suspect they are being sent by whatever device is plugged into the receiver.

Crossfire rx can receive telemetry from a variety of sources - so my feeling is that this value is likely being sent from an alternative source than the base rx.   (I only use the micro rx)

It would help to know what equipment is plugged into @3djc receiver?

Its entirely possible the issue is not opentx - but the device that is the source for these vars?

kilrah:
This has nothing to do with crossfire, but is about which format to display date in OpenTX... which would for example be coming from a FrSky (or compatible) GPS connected to a receiver.

3djc:
Ok, back to some consistency :)

x7
image

image


x9
image

image

robthomson:
@kilrah well spotted.. I was muddling up my threads!

But for reference:  DD-MM-YYYY is normal for the UK.

Pretty sure its only USA that use MM-DD-YYYY.

Rob

3djc:
Yes, but right now, the date setting places use YYYY-MM-DD on the radio, so to keep consistency within opentx, I propsoe for 2.2.1 we fix display as above, and for later on, have a more thourough discusion of date (and well time also) format (since we havereceived one request for AM:PM display)

PR #5277, opened 2017-10-15
Companion Build Option Lua Toot Tip change wording from "model" to "custom"

RCdiy:
Nowhere does the user see any documentation on model scripts. This
change is similar to how no heli tool tip is.


RCdiy:
@bsongis @kilrah could you please review.

kilrah:
You should not touch the .ts files, remove them from the PR and it will be fine.

RCdiy:
Updated... how about now...
So I went to every file .ts  and reverses edited them... is there an easier way?

kilrah:
When you use a git client yes, just select the files and revert them

PR #5293, opened 2017-10-18
Add access to vBatWarn from luaGetGeneralSettings()

floaledm:
Min and max are accessible, but not the warning level... This would allow using the warning level in lua scripts.


schwabe:
Looks good 👍

PR #5309, opened 2017-10-29
Check coin battery on startup and display it in debug view

schwabe:
This pull request is tested on the X9D+. It should work on all other Taranis radios. Todo:

- For Horus X12S, the ADC3 should be replaced by ADC1 as these inputs are only available on ADC1 but the other inputs should be also on ADC1, cannot tested myself
- For X10 and X12s the menu implementation is missing as I did not understand how to add another debug screen.


bsongis:
I missed this PR and I implemented the RTC battery check. Your implementation is better in some points, I will open a new PR soon for the latest 2.3

radio/src/targets/common/arm/stm32/adc_driver.cpp
```diff
@@ -88,32 +89,35 @@ void adcInit()
 
   ADC_MAIN->CR1 = ADC_CR1_SCAN;
   ADC_MAIN->CR2 = ADC_CR2_ADON | ADC_CR2_DMA | ADC_CR2_DDS;
-  ADC_MAIN->SQR1 = (NUM_ANALOGS_ADC-1) << 20; // bits 23:20 = number of conversions
+  ADC_MAIN->SQR1 = (NUM_ANALOGS_ADC-1) << 20; // bits 23:20 = number of conversions, NUM_ANALOGS_ADC
 
 #if defined(PCBX10)
   if (ANALOGS_PWM_ENABLED()) {
-    ADC_MAIN->SQR2 = (ADC_CHANNEL_EXT1<<0) + (ADC_CHANNEL_EXT2<<5); // conversions 7 and more
+    ADC_MAIN->SQR2 = (ADC_CHANNEL_EXT1<<0) + (ADC_CHANNEL_EXT2<<5) + (ADC_Channel_TempSensor<<10) + (ADC_Channel_Vrefint<<15) + (ADC_Channel_Vbat<<20);      ; // conversions 7 and more
```
bsongis, line 96: Only ADC1, not ADC3

schwabe: Yes when I did that X12 and X10 was a bit fuzzy and I did not have HW to test on, so that might be wrong.

bsongis: No problem. Also on this board you cannot measure the RTC and the temperature in the same time as they share the same input. I am a little bit lazy, I will do the minimum

PR #5313, opened 2017-10-31
Add Lua getUsage() function that returns percent of already used up i…

projectkk2glider:
…nstructions in current script execution cycle. It can be used to avoid script being killed when doing long lasting tasks (like telemetry logs loading and parsing)

### DIFF
--- a/radio/src/lua/api_general.cpp
+++ b/radio/src/lua/api_general.cpp
@@ -1178,7 +1178,7 @@ Get percent of already used Lua instructions in current script execution cycle.
 
 @retval usage (number) a value from 0 to 100 (percent)
 
-@status current Introduced in 2.2.0
+@status current Introduced in 2.2.1
 */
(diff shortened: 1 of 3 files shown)


radio/src/lua/api_general.cpp
```diff
@@ -1171,6 +1171,21 @@ static int luaLoadScript(lua_State * L)
...
 }
 
+/*luadoc
+@function getUsage()
+
+Get the percent of already used Lua instructions in current script execution cycle.
+
+@retval usage (number) a value from 0 to 100 (percent)
+
+@status current Introduced in 2.2.0
```
schwabe: Needs to be 2.2.1

projectkk2glider: Thanks

schwabe:
I guess  the way we currently handling going over limits and killing scripts instead suspending them, this is a good idea.

Issue #5346, opened 2017-11-12
Wrong GPS position is parsed when replaying log file captured with crossfire telemetry


Harpalyke:
When trying to replay a log file in the simulator where the models GPS position was captured while using a crossfire micro module the GPS coordinates that are displayed in the telemetry simulator window are different than those contained within the log file.

The crossfire telemetry sensor for GPS data logs the data in a different format to the way that FrSky telemetry, it uses the format "53.279969 -7.701416" which isn't being parsed correctly when reading the log file.

I suspect this may have something to do with issue #5224

PR #5347, opened 2017-11-12
Adding regex check for crossfire gps format when reading logs in tele…

Harpalyke:
…metry simulator

Adding a regex check when the telemetry simulator parses gps cords from a log file to see if the if the cords are in the crossfire gps log format. 
The GPS telemetry sensor from crossfire logs the coordinates as latitude with a precision of six decimal places followed by a space and then longitude also to six decimal places. e.g. "53.279969 -7.701416"
When the cords are already in this format their is no need to call convertDegMin on the values

Addresses issue #5346

### DIFF
--- a/companion/src/simulation/telemetrysimu.cpp
+++ b/companion/src/simulation/telemetrysimu.cpp
@@ -23,6 +23,7 @@
 #include "ui_telemetrysimu.h"
 #include "simulatorinterface.h"
 #include "radio/src/telemetry/frsky.h"
+#include <QRegularExpression>
 
 TelemetrySimulator::TelemetrySimulator(QWidget * parent, SimulatorInterface * simulator):
   QWidget(parent),
@@ -627,6 +628,7 @@ TelemetrySimulator::LogPlaybackController::LogPlaybackController(Ui::TelemetrySi
 {
   TelemetrySimulator::LogPlaybackController::ui = ui;
   stepping = false;
+  logFileGpsCordsInDecimalFormat = false;
   // initialize the map - TODO: how should this be localized?
   colToFuncMap.clear();
   colToFuncMap.insert("RxBt(V)", RXBT_V);
@@ -690,6 +692,31 @@ QDateTime TelemetrySimulator::LogPlaybackController::parseTransmittterTimestamp(
     format = "M/d/yyyy hh:mm:ss.z";
   }
   return QDateTime::fromString(datePart + " " + timePart, format);
+}
+
+void TelemetrySimulator::LogPlaybackController::checkGpsFormat()
+{
+  // sample the first record to check if cords are in decimal format
+  logFileGpsCordsInDecimalFormat = false;
+  if(csvRecords.count() > 1) {
+    QStringList keys = csvRecords[0].split(',');
+    if(keys.contains("GPS")) {
+      int gpsColIndex = keys.indexOf("GPS");
+      QStringList firstRowVlues = csvRecords[1].split(',');
+      QString gpsSample = firstRowVlues[gpsColIndex];
+      QStringList cords = gpsSample.simplified().split(' ');
+      if (cords.count() == 2) {
+        // frsky and TBS crossfire GPS sensor logs cords in decimal format with a precision of 6 places
+        // if this format is met there is no need to call convertDegMin later on when processing the file
+        QRegularExpression decimalCoordinateFormatRegex("^[-+]?\\d{1,2}[.]\\d{6}$");
+        QRegularExpressionMatch latFormatMatch = decimalCoordinateFormatRegex.match(cords[0]);
+        QRegularExpressionMatch lonFormatMatch = decimalCoordinateFormatRegex.match(cords[1]);
+        if (lonFormatMatch.hasMatch() && latFormatMatch.hasMatch()) {
+          logFileGpsCordsInDecimalFormat = true;
+        }
+      }
+    }
+  }
 }
 
 void TelemetrySimulator::LogPlaybackController::calcLogFrequency()
@@ -760,6 +787,7 @@ void TelemetrySimulator::LogPlaybackController::loadLogFile()
     supportedCols.clear();
     recordIndex = 1;
     calcLogFrequency();
+    checkGpsFormat();
   }
   ui->logFileLabel->setText(QFileInfo(logFileNameAndPath).fileName());
   // iterate through all known mappings and add those that are used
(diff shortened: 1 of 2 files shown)


companion/src/simulation/telemetrysimu.cpp
```diff
@@ -868,8 +868,19 @@ QString TelemetrySimulator::LogPlaybackController::convertGPS(QString input)
...
   }
-  double lon = convertDegMin(lonLat[0]);
-  double lat = convertDegMin(lonLat[1]);
+  double lon, lat;
+  // crossfire GPS sensor logs cords in decimal format with a precision of 6 places
+  // if this format is met there is no need to call convertDegMin
+  QRegularExpression crossfireCoordinateFormatRegex("^([-+]?\\d{1,2}[.]\\d{6})$");
+  QRegularExpressionMatch latFormatMatch = crossfireCoordinateFormatRegex.match(lonLat[0]);
+  QRegularExpressionMatch lonFormatMatch = crossfireCoordinateFormatRegex.match(lonLat[1]);
+  if(lonFormatMatch.hasMatch() && latFormatMatch.hasMatch()){
```
bsongis: space between `if` and `(`

companion/src/simulation/telemetrysimu.cpp
```diff
@@ -868,8 +868,19 @@ QString TelemetrySimulator::LogPlaybackController::convertGPS(QString input)
...
+	  lat = lonLat[0].toDouble();
+	  lon = lonLat[1].toDouble();
+  }else{
```
bsongis: newline before `else` and space between `else` and `{`

bsongis (commented):
just cosmetics before I press the merge button so that it is in next release

Harpalyke:
Updated. Thanks!

companion/src/simulation/telemetrysimu.cpp
```diff
@@ -874,10 +874,12 @@ QString TelemetrySimulator::LogPlaybackController::convertGPS(QString input)
   QRegularExpression crossfireCoordinateFormatRegex("^([-+]?\\d{1,2}[.]\\d{6})$");
   QRegularExpressionMatch latFormatMatch = crossfireCoordinateFormatRegex.match(lonLat[0]);
   QRegularExpressionMatch lonFormatMatch = crossfireCoordinateFormatRegex.match(lonLat[1]);
-  if(lonFormatMatch.hasMatch() && latFormatMatch.hasMatch()){
+  if (lonFormatMatch.hasMatch() && latFormatMatch.hasMatch()){
 	  lat = lonLat[0].toDouble();
 	  lon = lonLat[1].toDouble();
-  }else{
+  }
+  else
```
bsongis: no newline after the else! sorry to be so picky!

Harpalyke:
No problem. I just misread it, I should have checked one of the other else stmts in the code anyway.

companion/src/simulation/telemetrysimu.cpp
```diff
@@ -868,8 +868,20 @@ QString TelemetrySimulator::LogPlaybackController::convertGPS(QString input)
...
-  double lon = convertDegMin(lonLat[0]);
-  double lat = convertDegMin(lonLat[1]);
+  double lon, lat;
+  // crossfire GPS sensor logs cords in decimal format with a precision of 6 places
+  // if this format is met there is no need to call convertDegMin
+  QRegularExpression crossfireCoordinateFormatRegex("^([-+]?\\d{1,2}[.]\\d{6})$");
+  QRegularExpressionMatch latFormatMatch = crossfireCoordinateFormatRegex.match(lonLat[0]);
+  QRegularExpressionMatch lonFormatMatch = crossfireCoordinateFormatRegex.match(lonLat[1]);
+  if (lonFormatMatch.hasMatch() && latFormatMatch.hasMatch()){
+	  lat = lonLat[0].toDouble();
```
bsongis: and now indentation is bad, we use 2 spaces in this project

companion/src/simulation/telemetrysimu.cpp
```diff
@@ -868,8 +868,20 @@ QString TelemetrySimulator::LogPlaybackController::convertGPS(QString input)
...
   }
-  double lon = convertDegMin(lonLat[0]);
-  double lat = convertDegMin(lonLat[1]);
+  double lon, lat;
+  // crossfire GPS sensor logs cords in decimal format with a precision of 6 places
+  // if this format is met there is no need to call convertDegMin
+  QRegularExpression crossfireCoordinateFormatRegex("^([-+]?\\d{1,2}[.]\\d{6})$");
+  QRegularExpressionMatch latFormatMatch = crossfireCoordinateFormatRegex.match(lonLat[0]);
+  QRegularExpressionMatch lonFormatMatch = crossfireCoordinateFormatRegex.match(lonLat[1]);
+  if (lonFormatMatch.hasMatch() && latFormatMatch.hasMatch()){
```
bsongis: and a space before the `{`

bsongis (commented):
cosmetics again

mpaperno:
The regex check for every line of the file will be a fairly expensive operation.  At the least I'd recommend only defining that regex once (as a class member or as static in just the parsing function) -- this will prevent the need for Qt to "compile" the expression on every execution.  

Another option, assuming all GPS data is formatted the same way in every line of a given log file, is to determine the format on the first line of the log and then use that determination for the rest of the lines -- so the regex needs to be executed only once per file.

Lastly, if the log file GPS format itself could be normalized (always log in same format), that may be a better option, but I haven't looked at the logging routines at all so maybe that is less practical in this case.

companion/src/simulation/telemetrysimu.h
```diff
@@ -27,6 +27,7 @@
 #include <QDateTime>
 #include <QtCore/qmath.h>
 #include <QFileDialog>
+#include <QRegularExpression>
```
mpaperno: Any particular reason to put this in the header file?  Since there's no QRegularExpression declared in the header, this would probably make more sense in the cpp file.  Not a big deal either way.

companion/src/simulation/telemetrysimu.cpp
```diff
@@ -868,8 +868,20 @@ QString TelemetrySimulator::LogPlaybackController::convertGPS(QString input)
   if (lonLat.count() < 2) {
     return ""; // invalid format
   }
-  double lon = convertDegMin(lonLat[0]);
-  double lat = convertDegMin(lonLat[1]);
+  double lon, lat;
+  // crossfire GPS sensor logs cords in decimal format with a precision of 6 places
+  // if this format is met there is no need to call convertDegMin
+  QRegularExpression crossfireCoordinateFormatRegex("^([-+]?\\d{1,2}[.]\\d{6})$");
```
mpaperno: Since the captured result is never used, the parenthesis (after `^` and before `$`) could be excluded for a slight optimization of the regex.  Also, if you haven't yet, I'd make sure that a comma is never being used here as a decimal separator (e.g. in EU locales).

Harpalyke:
@mpaperno, good point on executing the regex for every line of the file. I've updated the code so it samples the first line as suggested and tidied up the other bits you suggested. 

Also, the data I had been using for testing was generated with a crossfire module in an X9D. I had assumed that it was because of the fact I was using the crossfire that the GPS data had been logged in an incompatible format. However, I've run a quick test logging the GPS data when connecting via an FrSky XSR and it appears that the gps format being logged is the same as with the crossfire instead of the expected "DDmm.mmmmH DDDmm.mmmmH" format noted in the code. 
Maybe the logging format was changed on the radio side but the expected format on the telemetry simulator wasn't updated to match?

bsongis:
thanks! still not 100% compliant with our coding rules, but not so far.

Harpalyke:
@bsongis, OK, sorry about that, I'll make sure to follow the style guide more closely next time. Thanks!

PR #5351, opened 2017-11-12
Lua themes with no options wouldn't load anymore (got broken in b4939…

kilrah:
Please check if that's not dangerous, but don't believe so...

### DIFF
--- a/radio/src/lua/widgets.cpp
+++ b/radio/src/lua/widgets.cpp
@@ -235,13 +235,11 @@ void luaLoadThemeCallback()
 
   if (name) {
     ZoneOption * options = createOptionsArray(themeOptions, MAX_THEME_OPTIONS);
-    if (options) {
-      LuaTheme * theme = new LuaTheme(name, options);
-      theme->loadFunction = loadFunction;
-      theme->drawBackgroundFunction = drawBackgroundFunction;
-      theme->drawTopbarBackgroundFunction = drawTopbarBackgroundFunction;
-      TRACE("Loaded Lua theme %s", name);
-    }
+    LuaTheme * theme = new LuaTheme(name, options);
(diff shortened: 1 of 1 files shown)


bsongis:
This was done this way in case there is no RAM left for the malloc (options is 0)

kilrah:
Better now? Thing is options is also 0 when there are no options and we thus ask malloc to alloc 0 bytes...

radio/src/lua/widgets.cpp
```diff
@@ -235,11 +235,13 @@ void luaLoadThemeCallback()
 
   if (name) {
     ZoneOption * options = createOptionsArray(themeOptions, MAX_THEME_OPTIONS);
-    LuaTheme * theme = new LuaTheme(name, options);
-    theme->loadFunction = loadFunction;
-    theme->drawBackgroundFunction = drawBackgroundFunction;
-    theme->drawTopbarBackgroundFunction = drawTopbarBackgroundFunction;
-    TRACE("Loaded Lua theme %s", name);
+    if (options || (themeOptions == 0)) {
```
bsongis: perhaps we could avoid to call `createOptionsArray` when themeOptions is NULL, no?

bsongis:
I think this version would work:

```
  if (name) {
    ZoneOption * options = NULL;
    if (themeOptions) {
      options = createOptionsArray(themeOptions, MAX_THEME_OPTIONS);
      if (!options)
        return;
    }
    LuaTheme * theme = new LuaTheme(name, options);
    theme->loadFunction = loadFunction;
    theme->drawBackgroundFunction = drawBackgroundFunction;
    theme->drawTopbarBackgroundFunction = drawTopbarBackgroundFunction;
    TRACE("Loaded Lua theme %s", name);
  }
```

PR #5364, opened 2017-11-14
Shift 3rd column a bit left

### LINKED ISSUE #5363: Bars upper limit maybe aligned right (or shifted one caracter left)
Using 2.2.1N434 for Taranis X7

On the DISPLAY tab of a model, when the right value exceed 4 caracters (Voltage > 10V or timer) the last caracter is out of the screen.

For the voltage losing the unit is no big deal of course, for the timer you don't know what is the last seconds digit.

image

3djc:
Before
image

After
image

I do PR it because I'm unsure if there was a reason to have it that far right

This fixes  #5363

### DIFF
--- a/radio/src/gui/128x64/model_display.cpp
+++ b/radio/src/gui/128x64/model_display.cpp
@@ -50,7 +50,7 @@ enum MenuModelDisplayItems {
 #else
   #define DISPLAY_COL2                (8*FW)
 #endif
-#define DISPLAY_COL3                  (18*FW+2)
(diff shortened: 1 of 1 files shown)


capitaineflam25:
Thanks a lot. And sorry for the duplicate.

kilrah:
> I'm unsure if there was a reason to have it that far right

Most likely the code was copy/pasted from 212 and not adjusted, seems we only come across a "bars" user about once a year :P

PR #5367, opened 2017-11-14
small cosmetic colour changes

raphaelcoeffic:
This allows the RSSI/Tx-Battery/Volume display to be displayed with the proper background colour when changing the main theme colour. Also puts the background colour at some use, as I could not see it used anywhere.


raphaelcoeffic:
hmm.... this one test seems to fail here:
```
/home/travis/build/opentx/opentx/radio/src/tests/lcd_480x272.cpp:152: Failure
Value of: checkScreenshot_480x272("transparency")
  Actual: false
Expected: true
./tools/commit-tests.sh: line 181: 10863 Segmentation fault      (core dumped) ./gtests ${TEST_OPTIONS}
```

What clue what that means....

3djc:
It means during this test, it is comparing your screen with a reference screen and rejecting it because they do not match

raphaelcoeffic:
@3djc ok, so should I make a change to the reference screen as well? Or can I just let it go of it... (no real clue how to do that...) I changed the default colour a little bit, as you can see in the diff, but I could not tell the difference between both on the screen.

3djc:
It needs to be fixed before merge. Either by fixing 'the fix' to match reference screen (always prefered), or by changing reference screen (but then it needs probably to be documented a bit more)

kilrah:
Can you post the before and after screenshots to detail what you actually want to change?

raphaelcoeffic:
@kilrah no problem!

raphaelcoeffic:
The main change is to correct this little glitch:
screen shot 2017-11-15 at 10 47 09

And make it look like that:
screen shot 2017-11-15 at 10 45 54

EDIT: not quite happy about the background colour being used for the text/icons the title bar, let me change that one.

raphaelcoeffic:
After the second commit, it looks like that:
screen shot 2017-11-15 at 11 00 13
screen shot 2017-11-15 at 11 01 12

kilrah:
Indeed nicer, had never noticed... will need to adapt the tests then

raphaelcoeffic:
>Indeed nicer, had never noticed... will need to adapt the tests then

Done! See last commit. The first impression is always important, when you show someone your brand new shinny transmitter ;-)

kilrah:
What's the reason for the color change though?

raphaelcoeffic:
I have the amber X10, the red thing is looking like crap on that TX (for my taste). And well "because I can".

raphaelcoeffic:
Looks like there is an issue with the build/test env and `AVR9X`. Here is what Travis says:
```
$ sudo apt-get install --yes --force-yes -qq qt${QT_BASE}base qt${QT_BASE}multimedia qt${QT_BASE}svg qt${QT_BASE}tools; source /opt/qt${QT_BASE}/bin/qt${QT_BASE}-env.sh
E: Unable to locate package qt53base
E: Unable to locate package qt53multimedia
E: Unable to locate package qt53svg
E: Unable to locate package qt53tools
/home/travis/.travis/job_stages: line 57: /opt/qt53/bin/qt53-env.sh: No such file or directory
```

It's running through on my notebook...

kilrah:
I meant the slight change in red shade.

raphaelcoeffic:
Previously, it was set like that:
`RGB(130, 1, 5)`

The main `RED` is `RGB(229, 32, 30)` and the other shade is calculated with 70% of that. Basically, `RGB(130, 1, 5)` is hard to calculate starting from `RED`, so that it is applicable to other colours as well. Now I just use 50%, which is `RGB(114, 16, 15)`. You wouldn't be able to tell the diference when you see it on the horus screen.

Otherwise, I would land with something like:
```
if (color == RED)
   lcdColorTable[MENU_TITLE_DISABLE_COLOR_INDEX] = RGB(130, 1, 5);
else
  lcdColorTable[MENU_TITLE_DISABLE_COLOR_INDEX] = RGB(GET_RED(color)>>1, GET_GREEN(color)>>1, GET_BLUE(color)>>1);
```

kilrah:
Maybe @projectkk2glider can comment, I think there could have been a reason for hardcoded values...

kilrah:
Maybe can make the popup contour and monitor screens follow the color as well while you're at it?

bsongis:
👍

PR #5379, opened 2017-11-17
Export flight session to its own CSV log file

Harpalyke:
Currently the transmitter logs all flight sessions for the same day into a single log file and the log file viewer allows for switching between session data. 
It can be a bit of a tedious job to separate the records up if feeding the CSV data into anything but the the log file viewer though.
This change adds a button beside the session selector dropdown in the log viewer that allows the user to export the CSV data for the selected session to its own log file.


kilrah:
Cool!

PR #5385, opened 2017-11-19
[Horus] one-click model select

raphaelcoeffic:
When on the model select screen, pressing [enter] loads the model currently select, instead of having to press enter long + click “select model” in the pop up menu.

It might not be the case for everyone, but it seems to me the most intuitive action to execute, and should thus not require the user to trigger the pop-up menu.

I remember always doing to the wrong thing on my Taranis when trying to select a new model. I would always click on it, thus triggering the "copy" function instead of selecting a model. I guess this is something that could be changed on the Taranis series as well for a more intuitive UI experience.

EDIT:
PR has been changed to just trigger the pop-up menu with a single short click, just like on X7 (and possibly others, if #5386 is accepted).

### DIFF
--- a/radio/src/gui/480x272/model_select.cpp
+++ b/radio/src/gui/480x272/model_select.cpp
@@ -219,9 +219,8 @@ bool menuModelWizard(event_t event)
 }
 #endif
 
-void onModelSelectMenu(const char * result)
+void selectModel()
 {
-  if (result == STR_SELECT_MODEL) {
     // we store the latest changes if any
     storageFlushCurrentModel();
     storageCheck(true);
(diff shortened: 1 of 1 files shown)


1Sven:
I think it would be good: 1X short click = model choice (immediately), long press perhaps the old menu. There is nothing dangerous. (The discussion sounds a bit like switching off with the short? ..)

1Sven:
I think it would be good: 1X short click = model choice (immediately), long press perhaps the old menu. There is nothing dangerous. (The discussion sounds a bit like switching off with the short? ..)

1Sven:
Sorry this was one klich to much...

raphaelcoeffic:
Rebased on top of 2.2.

PR #5386, opened 2017-11-19
[Taranis] one-click model select

raphaelcoeffic:
When on the model select screen, pressing [enter] loads the model currently select, instead of having to press enter long + click “select model” in the pop up menu.

It might not be the case for everyone, but it seems to me the most intuitive action to execute, and should thus not require the user to trigger the pop-up menu.

This PR is the Taranis version of #5385.

EDIT:
So far, the following alternatives have come out of the discussion:
- one-click option (implemented by the PR) (removes the copy/move system from easy access).
- click triggers pop-up menu (like on the X7) (removes the copy/move system from easy access).
- anything else?

### DIFF
--- a/radio/src/gui/128x64/model_select.cpp
+++ b/radio/src/gui/128x64/model_select.cpp
@@ -219,7 +219,7 @@ void menuModelSelect(event_t event)
         s_copyMode = 0;
         event = EVT_ENTRY_UP;
       }
-      else if (event == EVT_KEY_LONG(KEY_ENTER) || IS_ROTARY_BREAK(event)) {
+      else if (event == EVT_KEY_BREAK(KEY_ENTER)) {
         s_copyMode = 0;
         killEvents(event);
 #if defined(NAVIGATION_MENUS)
@@ -253,6 +253,7 @@ void menuModelSelect(event_t event)
 #endif
       }
       else if (eeModelExists(sub)) {
+        killEvents(event);
         s_copyMode = (s_copyMode == COPY_MODE ? MOVE_MODE : COPY_MODE);
         s_copyTgtOfs = 0;
         s_copySrcRow = -1;
--- a/radio/src/gui/128x64/model_select.cpp
+++ b/radio/src/gui/128x64/model_select.cpp
@@ -252,12 +252,6 @@ void menuModelSelect(event_t event)
         }
 #endif
       }
-      else if (eeModelExists(sub)) {
-        killEvents(event);
-        s_copyMode = (s_copyMode == COPY_MODE ? MOVE_MODE : COPY_MODE);
-        s_copyTgtOfs = 0;
-        s_copySrcRow = -1;
-      }
       break;
 
 #if defined(PCBX7)
--- a/radio/src/gui/212x64/model_select.cpp
+++ b/radio/src/gui/212x64/model_select.cpp
@@ -162,7 +162,7 @@ void menuModelSelect(event_t event)
           s_copyMode = 0;
           event = EVT_ENTRY_UP;
         }
-        else if (event == EVT_KEY_LONG(KEY_ENTER)) {
+        else if (event == EVT_KEY_BREAK(KEY_ENTER)) {
           s_copyMode = 0;
           killEvents(event);
           if (g_eeGeneral.currModel != sub) {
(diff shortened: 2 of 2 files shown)


mpaperno:
IMHO a 2-step process to select a new active model is an important safety feature.  Changing at this point (when so many users are used to it) seems especially unexpected/dangerous.

A single click could bring up the menu though, perhaps with Copy being the first choice (for extra safety).  I also often just click before remembering to click+hold for the menu.  But I'd rather do that than switch models by mistake.

3djc:
I second Max comments

raphaelcoeffic:
@mpaperno @3djc Frankly, this did not come to my mind. But just so that I understand it properly: what is actually the scenario you are thinking of? What could happen? I don't think I would press long menu while flying. Please note that it will only switch the model when first going up/down before. If the current model is still selected (cursor on the same line), nothing will happen.

For my part, the current behaviour is just very counter-intuitive. I could however live with the pop-up menu being triggered by a simple click, if this is a general consensus.

raphaelcoeffic:
Another option would be to ask for confirmation when a model is currently ON (`isTelemetryStreaming()`).

kilrah:
Simple click for menu (and maybe long for model?) would be more in line with operation in other places, BUT that's used by the select/copy/move system. Which is the same in inputs/mixes for coherence.

3djc:
You cannot rely on telem streaming for safety, as many fly without (older gen receivers, logn range wihout telem, ...)

mpaperno:
> what is actually the scenario you are thinking of? ...  I don't think I would press long menu while flying.

Well, it's short-menu to get to model select, but either way: possible to confuse with PAGE, settings screens might be used in flight with eg. autonomous drone, and (especially) during bench testing/model setup when one is in and out of the menus a lot anyway.  On Horus it's a different story (though I wouldn't vote for one-click there either).

raphaelcoeffic:
>Simple click for menu (and maybe long for model?) would be more in line with operation in other places, BUT that's used by the select/copy/move system. Which is the same in inputs/mixes for coherence.

I understand the behaviour must be coherent with other pages, but selecting a model is completely different task compared to the mixer page. I find the behaviour on the mixer well done.

But most of the time (90%), when I want to switch model, here is what happens:
- press [MENU] to get to the "model select" page
- navigate to the model
- click on the model
- then I notice: Oh no! This is making a copy!
- so: [EXIT]
- think....
- long press on [MENU]
- finally click on [ENTER] and my model loads.

Gentlemen, this is called "security by obscurity", and really not otx-beginner friendly (because counter intuitive).

I'm open to any other workflow that would make it simpler for stupid people like me (and my 3 other buddies using Taranis/Horus at the field who pressed me to submit a PR).

raphaelcoeffic:
> Well, it's short-menu to get to model select

@mpaperno yep, sorry, I screwed that up because it's different on the Horus...

kilrah:
Basically things are how they are becasue of the presence of the copy/move system, so it boils down to "are we OK to remove it and require using the menu for it". 
That is how it is on the X7 so I quite agree to remove it.

raphaelcoeffic:
@Kilrah that's really a good summary. Do you think we should have the behaviour on the X9D be the same as on the X7?

The question is: easy model select vs. easy copy/move. I tend to think that easy model select is more important, as it's used way more often than the later.

EDIT: just looked on a X7, and it really kicks the pop-up menu when clicking somewhere. I could really live with that option if the one-click option is not desired.

rottweilertje:
Since changing models is done way more often than copying, I’d say please make it easier to switch models. 
Flying with many different models on a session I have had times I had to clean up my model memory after a session because I copied models around.

RipperDrone:
+1 for single-press to change models :-)

raphaelcoeffic:
@kilrah @3djc I just changed that PR to what we discussed some weeks/month ago, which is that it now works just like on X7:
- one click (enter) triggers the pop-up menu
- long press switches to copy-mode.
- another long press switches to move-mode.

3djc:
Hum,

I have personnaly grown used to long ent on x7, and wait for the popup to show, and now that triggers copy instead of the popup, and it is not obvious at all it does. You do not make it like x7, you change x7 too

raphaelcoeffic:
@3djc Well, I have no issue removing the short cut for the copy/move-mode! That was just an idea how it could be kept, as some seemed to like it.

EDIT: done! Copy/move-mode now only using the pop-up menu.

mpaperno:
Looks good, my only last "paranoid safety" suggestion would be to move the Select Model option away from the first/default choice on the popup menu.  It is safer since it would require 2 different buttons/controls to be used before switching models.  Too paranoid?  Perhaps.  :)

raphaelcoeffic:
Rebased on top of current 2.2.

PR #5387, opened 2017-11-19
[Simulator] Fix coordinate of encoder button in X10 skin.

mpaperno:


kilrah:
Ah yep, missed that one

PR #5389, opened 2017-11-19
x10 simulator skin fixes

mhotar:
I have used transparent images that works together with background gradient. Hopefully it works on all OSs.


kilrah:
Fine on Windows

Issue #5368, opened 2017-11-14
X10 + 2.2 (c1b0e9fea3511) / Blocks on debug statistics page


raphaelcoeffic:
While on the memory debug page (statistics), the Tx just blocked on the that screen, whereby the trim seemed to worked, as the haptic warning was present when trimming a bit left and then right. I could however not reproduce it. I had to unplug the battery.

I'll try to reproduce in the next days. I compiled with `-DPCB=X10 -DHELI=NO -DLUA=YES -DGVARS=YES`.

3djc:
Likely duplicate #5312, please test with N434 or later

raphaelcoeffic:
@3djc not going to change: I was already using HEAD of 2.2 branch, compiled yesterday, and I just checked that f6309138f6ae was already included. My HEAD is currently at c1b0e9fea3511.

kilrah:
More info needed, what setup (otx file?), any lua scripts or widgets used?

raphaelcoeffic:
No lua script involved, one widget (model picture with no picture configured) on main page. I'll upload my otx file ASAP.

raphaelcoeffic:
Tried to reproduce it, and it's not really easy, but I manage it ever couple of minutes while playing with that screen. What I noticed is: even if I don't get it to block, the "RTN" will not work every time. After pressing the wheel on the right a couple of time, it seems I need more presses on "RTN" before I can leave that screen. And sometime, it does not work anymore and I'm stuck on the screen.

[X10-files.zip](https://github.com/opentx/opentx/files/1475087/X10-files.zip)

raphaelcoeffic:
I might note that this might be related to the isues I had when using gcc 6.2.1. I'll have to retest to be sure it is somehow still happening.

raphaelcoeffic:
Rested with a firmware compiled on 2.2 HEAD with gnu-arm-gcc 4.7.4: **cannot reproduce anymore**.

BUT, I noticed that as soon as I used the rotary knob on the mem stats page, it takes at least TWO presses on [RTN] to exit the menu instead of ONE normally.

3djc:
@raphaelcoeffic could you try https://github.com/opentx/opentx/tree/3djc/Colorlcd-stats-menu-fix, I think it fixes the behaviour you mention

raphaelcoeffic:
@3djc thx, works for me in the simulator (could reproduce it there as well).

PR #5390, opened 2017-11-20
Move to stats scrrens to simple menu 

3djc:
This prevents the need for dual RTN when scroll wheel is used while on those screen


raphaelcoeffic:
As discussed previously, this fixes #5368 (tested on simulator only, but I guess that's ok).

3djc:
Well one of the behaviour you mention at least

raphaelcoeffic:
@3djc the other issue was most probably caused by the compiler (gcc 6.2.1), as I cannot reproduce it anymore with firmwares compiled with gcc 4.7.4.

PR #5395, opened 2017-11-21
Add getUID() lua function

floaledm:
This allows access in lua to the radio UID as a string (e.g., "12345678 55AA55AA 87654321").
Tested in lua on an actual X9D+

### DIFF
--- a/radio/src/lua/api_general.cpp
+++ b/radio/src/lua/api_general.cpp
@@ -48,6 +48,23 @@
 #define FIND_FIELD_DESC  0x01
 
 /*luadoc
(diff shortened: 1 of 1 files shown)


kilrah:
Interesting, what's the purpose?

projectkk2glider:
I would prefer to add this as another return value to the existing getValue function.

schwabe:
I am closing this as 

a) another approach is preffered as projectkk2gilder pointed out
b) the purpose of this is unclear. (I suspect something but you probably don't want say that openly)

PR #5435, opened 2017-11-28
Use X10 picture during X10 calibration

3djc:
image

I have left horus.bmp named as it is instead of renaming X12.bmp for compatibility with existing sd card content


raphaelcoeffic:
Just thinking we should have picked the X10S carbon instead of the silver: fits better to any color profile :-D

schwabe:
@raphaelcoeffic Nah, show the X10 in all its ugly glory!

raphaelcoeffic:
https://opentx.rocket.chat/file-upload/WCF68RgeRdpzcQJMS/X10S.bmp

PR #5437, opened 2017-11-28
[Companion] Adjust source value ranges and editors in LS/CF/Telemetry

mpaperno:
Went down a rabbit hole with regards to timer and RTC ranges and how they're used/limited/edited in regards to LS (comparison functions), CF (set timerN action), and telemetry display bar limits.  Seems there's a bit of confusion here both in Companion and on radio (probably due to legacy issues).

Fixes #2984
Fixes #5362  (also negative values were not possible)
Fixes #5374
Fixes #5375
Fixes #5527
Fixes timer and RTC time ranges to comply with firmware limits.
Fixes not being able to use RTC time in LSs.
Fixes newly-added LS not being available in source lists (except when pasting).
Fixes bug with timer/rtc comparison values getting reset to value of spin box instead of time editor.
New TimerEdit GUI component for negative and other typically "invalid" times.
Removed the short limits on timer values.
GVars, trims, script outputs, and sticks/switches/cyclic/MAX get proper limits.
Fixes #3581 (Companion part).
Fixes sensor value offset and range editing (mentioned in 3581).

There are some issues/questions on the firmware side... I found a couple small bugs and will open a separate PR and bring them up there.  

But the main question is why the limit of +/- 23m59s on timer values for LS comparison functions and telem. bar display min/max since the variable types (in firmware) can handle a lot more than that.

PR #5445, opened 2017-11-30
[Radio] Adjust value limits for mix sources and update UI for LSs, CFs & telem.

### LINKED ISSUE #3581: Logical Switches: a>x etc.: x limited to le. 150 if a is the output of a lua script
For me, this is an interesting behaviour. I tried to set up a logical switch, which is triggered, if the calculated distance from a LUA-Script output is larger then 250(m). But 150 is the biggest number I can fill in for x. With other parameters for a (for ex. SWR and others) higher values for x are possible.
Would it be possible to increase the x-values for lua-scripts to higher values as well?

Best Regards Paul

mpaperno:
Adjust some value limits for mix sources and update editing UI for LSs, CFs, & telem. bars:

  * Increase timer limits to +/-8:59:59 when used with LSs, CFs, & bars;
  * Increase Lua script output limits to +/-30K (see #3581);
  * Trim limits based on extended limit model setting;
  * Limit TX voltage to 25.5;
  * Limit TX time comparison value to 29:59m;
  * Use proper GVar limits (user-configured), display precision, & unit in LSs & telem. bars;
  * Display timer values properly with hours (instead of mmm:ss);
  * getTimerString() would fail with times >= 100min if hours weren't shown.

Fixes #3581 (fw part)

This PR relates to #5437 and should be merged together.

Notes/discuss:

Negative times always blink, even in the editor screens.  Obviously distracting especially when it's selected for editing.  I'd vote for changing this, but somehow keep the old default of blinking when timer is not used on editing screen.

I think the X7 (small ARM screen) time display function could be revisited.  When viewing values >=60m it goes into a "HHhMM" display mode where seconds can't be seen (but are still changed when editing).  I understand the problem with space, but at least in the editor screens I think the full h:mm:ss format would fit just about everywhere (there's no place 2-digit hours are shown).  One possible exception is the CF editor where we don't yet have a dedicated "pop up" dialog to edit the values. @3djc ?

### DIFF
--- a/radio/src/gui/128x64/model_display.cpp
+++ b/radio/src/gui/128x64/model_display.cpp
@@ -193,18 +193,18 @@ void menuModelDisplay(event_t event)
 
         if (IS_BARS_SCREEN(screenIndex)) {
           FrSkyBarData & bar = g_model.frsky.screens[screenIndex].bars[lineIndex];
-          source_t barSource = bar.source;
-          drawSource(DISPLAY_COL1, y, barSource, menuHorizontalPosition==0 ? attr : 0);
-          int barMax = getMaximumValue(barSource);
-          int barMin = -barMax;
-          if (barSource) {
-            if (barSource <= MIXSRC_LAST_CH) {
-              drawSourceCustomValue(DISPLAY_COL2, y, barSource, calc100toRESX(bar.barMin), (menuHorizontalPosition==1 ? attr : 0) | LEFT);
-              drawSourceCustomValue(DISPLAY_COL3, y, barSource, calc100toRESX(bar.barMax), (menuHorizontalPosition==2 ? attr : 0) | LEFT);
+          drawSource(DISPLAY_COL1, y, bar.source, menuHorizontalPosition==0 ? attr : 0);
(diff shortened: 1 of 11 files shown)


bsongis:
Wondering if you tested this PR on the good old 9X? Even on the simu. Does it increase flash usage on 9X?

mpaperno:
Yessir I did indeed. No functional change on any AVR platforms.  No memory usage changes.

9X build with many options enabled (gvars, frsky, etc), current 2.2:
```
AVR Memory Usage
----------------
Device: atmega64

Program:   64730 bytes (98.8% Full)
(.text + .data + .bootloader)

Data:       3227 bytes (78.8% Full)
(.data + .bss + .noinit)
```

With this PR:
```

AVR Memory Usage
----------------
Device: atmega64

Program:   64730 bytes (98.8% Full)
(.text + .data + .bootloader)

Data:       3227 bytes (78.8% Full)
(.data + .bss + .noinit)
```

bsongis:
Cool! I need to have a closer look to it then. I propose we don't include it in 2.2.1

PR #5455, opened 2017-12-02
Horus bootloader

raphaelcoeffic:
As the title says, this introduces a bootloader for Horus X10 / X12S, allowing for flashing the firmware from SD card or over USB.

Please that this pull request is based on @schwabe 's branch to provide firmware version over USB (see #5442).

I tested it locally on:
- X12S
- X10
- X9D+
- X7 

More tests are needed, so please, feel free to report here how that works for you so that the last bugs can be ironed out.

Here a couple of screenshots:
main_screen
fw_select
flash_start
flashing
flash_done
usb_connected


raphaelcoeffic:
@bsongis I'll wait for #5442 to be merged first and rebase on top of that (makes the history somehow easier to understand). It should be good to merge afterwards.

bsongis:
some merge work needed as you said

Issue #5434, opened 2017-11-27
Feature Request: Custom Splash Screen for Horus X10


trojanGoat:
I don't see options for using custom splash screens for Horus X10 when powering up the radio.

https://imgur.com/a/vJYeH

raphaelcoeffic:
Yep, I'd really prefer to have custom splash images being just stored on the SD card, so the process is way easier:
- no need to patch the firmware on the fly
- automatically there when flashing a new version.

commonspace:
Bump. 
Is this feature going to be added?

3djc:
Will be in 2.2.2

R3M1X972:
Hi guys, anyone got news on this feature ? I still can see the "edit splash screen" button when connecting my Horus X12S to Companion 2.2.2 freshly updated, but the button is grey and can't be used, any ideas ?

3djc:
maybe you could read 2.2.2 release notes

raphaelcoeffic:
@R3M1X972 you can now select a splash screen directly in the SD explorer. Just click on a PNG image, and it will appear in the context menu.

R3M1X972:
@raphaelcoeffic & @3djc : thanks guys. However I did created a splash.png and placed it in the SD as suggested in release notes (/IMAGES/splash.png). I can see the file displaying when navigating through the menu but for some reason it doesn't shows up on Horus start (OpenTx screen is still showing up instead)... did I miss something again ?

raphaelcoeffic:
@R3M1X972 you are using 2.2.2, right?

R3M1X972:
@raphaelcoeffic yes, indeed, I'm on 2.2.2. could png image size be the solution?

kilrah:
It must be the the right size for the screen or smaller.

raphaelcoeffic:
@kilrah in fact no, it will work with bigger images, will just get cropped.

R3M1X972:
Dammit, I'm sure i'm missing something but can't figure what it is. Created the splash.png in 3 different sizes and placed it subsequently in the above mentionned folder, but still I'm getting the "opentx" logo when powering up the Horus :-/ 
Should you have any idea it'll be really appreciated ! 👍

raphaelcoeffic:
@R3M1X972 could you please come to our chat (chat.open-tx.org) and post me your image?

R3M1X972:
[...]

PR #5463, opened 2017-12-04
allow splash image to be loaded from SD on Horus

raphaelcoeffic:
This basically implements issue #5434 as described. The image `/IMAGES/splash.png` is loaded from the SD card while booting and used as a splash image. If the image is not present, nothing changes and the splash image embedded in the firmware is used.

Things tested:
- no file: nothing happens (default is loaded: it's optional after all).
- wrong file (some text inside): nothing happens (default is loaded).
- big file (640x480): picture is probably cropped, because I see the background.

While testing, I did not notice any usual delay compared to the normal splash image.

The splash screen can be modified by using the new menu item "Splash screen" in the SD card manager. That will copy the file to `/IMAGES/splash.png`, thus overwriting any existing file with the same name.


Barbaryann:
It is only working on the next 2.2.2 version ?
On the 2.2.1 we can't change the slpash screen on our X10 ?

raphaelcoeffic:
@Barbaryann exactly. If you want it earlier, you'll have to merge/compile yourself.

Barbaryann:
Thanks but i don't really know how to do that, I can wait some week to have a nice Splash screen.
Thanks

radio/src/gui/480x272/splash.cpp
```diff
@@ -19,13 +19,36 @@
...
+
 void drawSplash()
 {
-  lcd->clear();
-  lcd->drawBitmap((LCD_W-BMP_SPLASH.getWidth())/2, (LCD_H-BMP_SPLASH.getHeight())/2, &BMP_SPLASH);
+    static bool     loadImgFromSD  = true;
+    static BitmapBuffer* splashImg = NULL;
+
+    if (loadImgFromSD && splashImg == NULL) {
+        bool sd_mounted = sdMounted();
```
bsongis, line 33: is it possible to go there? if yes would you add a comment?

raphaelcoeffic: It definitely is, that's what happens the very first time the splash function is called. The goal is to execute this code only once, whether it failed or not, we should not re-try later.

raphaelcoeffic: But normally, the SD card should be unmounted when splash() is called the first time. So basically the case where the SD is already mounted is only handled here as a safeguard.

trojanGoat:
Hey Guys, 

This is only cosmetic improvements, but it would be very cool to be able to use a .gif file for the splash screen.  

Actually if you guys are interested, I can volunteer my time and make a 3d animated/rendered splash screen for open tx.  Here's a little example of what I made for our group.  https://www.youtube.com/watch?v=s06MtPKOwKc

sly1231:
_The splash screen can be modified by using the new menu item "Splash screen" in the SD card manager. That will copy the file to /IMAGES/splash.png, thus overwriting any existing file with the same name._


Where does one find ""Splash screen" in the SD card manager"

kilrah:
In yet unreleased version 2.2.2

PR #5465, opened 2017-12-04
Add sane defaults for X10/X12. Also set battery warnings when creatin…

schwabe:
…g general settings in simulator/companion


raphaelcoeffic:
Looks good to me, but I didn’t test though.

schwabe:
I tested X12 and X10 in simulator and that looks fine.

PR #5471, opened 2017-12-06
Show multi failsafe to x7 UI

### LINKED ISSUE #5470: No option for setting failsafe for Multi AFDS2A on 2.2.1 rc1
On Companion and QX7 there is no option to set failsafe values on Multi AFDS2A. The settings only show on the internal module.

multi

2.1.1 rc1 Feature

Multi: allow syncing multi output with Multimodule radio frames (reduces latency) and allow setting failsafe for AFDS2A, requires experimental Multi code

3djc:
This fixes #5470

### DIFF
--- a/radio/src/gui/128x64/model_setup.cpp
+++ b/radio/src/gui/128x64/model_setup.cpp
@@ -1059,7 +1059,7 @@ void menuModelSetup(event_t event)
       case ITEM_MODEL_EXTERNAL_MODULE_FAILSAFE: {
         uint8_t moduleIdx = CURRENT_MODULE_EDITED(k);
         ModuleData &moduleData = g_model.moduleData[moduleIdx];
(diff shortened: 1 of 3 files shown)


radio/src/gui/128x64/model_setup.cpp
```diff
@@ -1059,7 +1059,7 @@ void menuModelSetup(event_t event)
       case ITEM_MODEL_EXTERNAL_MODULE_FAILSAFE: {
         uint8_t moduleIdx = CURRENT_MODULE_EDITED(k);
         ModuleData &moduleData = g_model.moduleData[moduleIdx];
-        if (IS_MODULE_PXX(moduleIdx)) {
+        if (IS_MODULE_PXX(moduleIdx) || IS_MODULE_MULTIMODULE(moduleIdx)) {
```
schwabe: I think we should get rid of this if completely. Other ui (horus, x9) also don't have an if here

PR #5473, opened 2017-12-07
added support for min/max contrast to eeprom interface and use it

### LINKED ISSUE #5468: Limit contrast value in Companion
Companion 2.2.1N453 / Taranis X7

After converting my eeprom from my Sky9X radio to my new X7radio, i transfered it and get a blank screen after the splash, making the radio unusable.

I realised that it was due to the contrast value imported from the profile of my Sky9x radio (Which is set to "38", and the limit on the X7 is "30")

I don't know if a limit on the Companion field or a security on the firmware (when reading the eeprom value) could avoid that.
Now i know that situation i can correct it quickly, but it take time to understand what's happening the first time.

image

raphaelcoeffic:
fixes #5468

### DIFF
--- a/companion/src/firmwares/opentx/opentxinterface.cpp
+++ b/companion/src/firmwares/opentx/opentxinterface.cpp
@@ -543,6 +543,13 @@ int OpenTxFirmware::getCapability(::Capability capability)
         return 0;
     case MaxVolume:
       return (IS_ARM(board) ? 23 : 7);
+    case MaxContrast:
+      if (IS_TARANIS_X7(board)) return 30;
+      return 45;
+    case MinContrast:
+      if (IS_TARANIS_X7(board)) return 10;
+      if (IS_TARANIS(board)) return 0;
+      return 10;
     case HasSoundMixer:
       return (IS_ARM(board) ? 1 : 0);
     case ExtraInputs:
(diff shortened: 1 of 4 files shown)


companion/src/firmwares/opentx/opentxinterface.cpp
```diff
@@ -543,6 +543,13 @@ int OpenTxFirmware::getCapability(::Capability capability)
         return 0;
     case MaxVolume:
       return (IS_ARM(board) ? 23 : 7);
+    case MaxContrast:
+      if (IS_TARANIS_X7(board)) return 30;
```
mpaperno: Line break after `if` please (also below).

companion/src/radiodata.cpp
```diff
@@ -212,6 +212,16 @@ void GeneralSettings::convert(Board::Type before, Board::Type after)
       memcpy(switchName[5], switchName[7], sizeof(switchName[5]));
     }
   }
+
+  if (IS_TARANIS(after)) {
+      unsigned int min_contrast = getCurrentFirmware()->getCapability(MinContrast);
```
mpaperno: Wrong indents in this block, 4 spaces instead of 2.  Also doubled in the `if/else` just below.

companion/src/firmwares/opentx/opentxinterface.cpp
bsongis: why? the `return 45;` does the job

bsongis (commented):
Cosmetics issue as @mpaperno says

bsongis:
Thanks!

PR #5477, opened 2017-12-07
[Companion] Fix GVar min corrupted when upgrading from firmware <= v217

### LINKED ISSUE #5466: 2.2.1 RC1: GV 'min' property corrupted during conversion from 2.1.9 to 2.2.1
Just noticed a couple of GV related issues, after converting an eeprom image from 2.1 to 2.2.1 using 2.2.1 RC1. The eeprom targets the X9D.

Steps to reproduce:
1. Download .eepe file [from here](https://www.dropbox.com/s/tfs867jop4r36m5/GV_Bug_21.bin?dl=0)
2. Open .eepe file into Companion 2.1.9 with X9D profile. Note the GV values.
3. Open .eepe file, this time in Companion 2.2.1 RC1 also with X9D profile. Eeprom appears to convert okay, but GV definitions are corrupted, with min set to +1024.

Also SF1, SF2 and SF3 fail to update the GVs correctly as the source channels change. This is easiest to see with GV9: it should track the TrmR (rudder trim) value, instead the value jumps to 1024 immediately the flight mode is changed (GV9 is configured as 'own value' in each flight mode).

mpaperno:
Fixes #5466

### DIFF
--- a/companion/src/firmwares/opentx/opentxeeprom.cpp
+++ b/companion/src/firmwares/opentx/opentxeeprom.cpp
@@ -3227,15 +3227,16 @@ OpenTxModelData::OpenTxModelData(ModelData & modelData, Board::Type board, unsig
     for (int i=0; i<MAX_GVARS(board, version); i++) {
       if (version >= 218) {
         internalField.Append(new ZCharField<3>(this, modelData.gvarData[i].name, "GVar name"));
-        internalField.Append(new SignedField<12>(this, modelData.gvarData[i].min));
-        internalField.Append(new SignedField<12>(this, modelData.gvarData[i].max));
+        internalField.Append(new UnsignedField<12>(this, (unsigned &)modelData.gvarData[i].min));
+        internalField.Append(new UnsignedField<12>(this, (unsigned &)modelData.gvarData[i].max));
         internalField.Append(new BoolField<1>(this, modelData.gvarData[i].popup));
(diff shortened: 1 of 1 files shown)


companion/src/firmwares/opentx/opentxeeprom.cpp
```diff
@@ -3227,15 +3227,16 @@ OpenTxModelData::OpenTxModelData(ModelData & modelData, Board::Type board, unsig
...
+        internalField.Append(new UnsignedField<12>(this, (unsigned &)modelData.gvarData[i].max));
         internalField.Append(new BoolField<1>(this, modelData.gvarData[i].popup));
         internalField.Append(new UnsignedField<1>(this, modelData.gvarData[i].prec));
         internalField.Append(new UnsignedField<2>(this, modelData.gvarData[i].unit));
         internalField.Append(new SpareBitsField<4>(this));
       }
       else {
-        internalField.Append(new ZCharField<6>(this, modelData.gvarData[i].name, "GVar name"));
+        internalField.Append(new ZCharField<3>(this, modelData.gvarData[i].name, "GVar name"));
+        internalField.Append(new SpareBitsField<3*8>(this));
```
bsongis, line 3239: Perhaps a SpareBytesField<N> later ...

bsongis:
Thanks!

PR #5406, opened 2017-11-22
Bsongis/r9 m fixes

bsongis:

PR #5479, opened 2017-12-08
[Companion][R9M] Adjust for R9M changes

mpaperno:
Allow module type selection; 
Adjust UI to better match radio; 
Improve model printer details. 

Related to #5406, based on (and PR target): https://github.com/opentx/opentx/tree/bsongis/R9M_fixes

A new PR could be done against `2.2` instead, there are no changes on radio side or in the import/export, it is GUI-only.


bsongis:
Thanks!

PR #5485, opened 2017-12-08
Updated Wizard

LapinFou:
- Reduced name of mixers to 6 characters.
- Added license GPLv2 header.
- Renamed some mixes in order to keep the same naming convention between all radios.


3djc:
I will merge this just before next RC, as this requires sdcard version increment (and it is better if we group it, shall something else need one too (nothing on the radar requiring it so far, but who knows whats to come !))

PR #5486, opened 2017-12-08
[X9] Allow LSs 33-64 in main view & fix X9E view of channels 17-32

### LINKED ISSUE #5476: Add the ability to see Logical Switches (LS) 33-64 from the main view screen. XD9+
Taranis X9D+

On the 3rd page of the main view screen (the Logical Switches screen), only switches 1-32 are visible (LS 1-32).

It seems intuitive that by pressing the + key, you would then see 'page 2' of the logical switches: LS 33-64, just like how the Channel Monitor screen operates.

Can this be made to happen?

Thank you.

mpaperno:
+/- keys or left/right encoder will switch between LS groups 1-32 and 33-64.
Also with X9E (encoder only) there was no way to navigate in channels/mixes monitor to the higher channels.

Closes #5476

screenshot_x9d _17-12-08_19-33-14
screenshot_x9d _17-12-08_11-33-30

### DIFF
--- a/radio/src/gui/212x64/view_channels.cpp
+++ b/radio/src/gui/212x64/view_channels.cpp
@@ -40,6 +40,10 @@ void menuChannelsView(event_t event)
       break;
     case EVT_KEY_FIRST(KEY_RIGHT):
     case EVT_KEY_FIRST(KEY_LEFT):
+#if defined(ROTARY_ENCODER_NAVIGATION)
+    case EVT_ROTARY_LEFT:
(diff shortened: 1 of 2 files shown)


Kennamer:
Hello mpaerno,

Can you read my reply in my original post #5476?

Thanks!

Kennamer:
Nevermind... I see you already handled it! LOL. You are too fast! ;-)

mpaperno:
We aim to please...  ;-)   It was a good point, looks much better this way.

PR #5487, opened 2017-12-08
Better support for R9M EU

3djc:
IMPORTANT : this make former BINDING_OPTIONS part of basic OpenTx

- R9M Eu option are choosen aftezr pushing bind

image

When internal module is not off

image


geofrancis:
is there a eu version or does the r9m do both 868 and 900?

kilrah:
It does one or the other based on what firmware is loaded onto the module/receiver.

PR #5493, opened 2017-12-10
[Companion] Adjust X10/S hardware options to match radio (no L1/L2/JSx/JSy)

### LINKED ISSUE #5488: [X10] LS & RS controls mismatch between Companion and firmware
On my X10 I assigned the RS for volume control. After reading the TX settings to Companion the parameter for volume control was not transferred.
image

I tested it also the other way. If the parameter is set to RS in companion volume control works on the TX (after writing settings to the TX), but in reverse (down make it louder).
Nevertheless, the Parameter-field for volume on the TX is empty. After re-assign RS the volume control works as expected (RS up for volume up).

OTX-file for reference:
[FrSky X10.zip](https://github.com/opentx/opentx/files/1544627/FrSky.X10.zip)

mpaperno:
Removes L1/L2/JSx/JSy as possible inputs on X10, including in radio HW and calibration settings.

Fixes #5488  (tested only in sim)

Does not deal with any conversion issues.  Especially the "silent" conversion between X10 and X12, which is not currently detected at all.

When going from X12 to X10, L1/L2 become LS/RS, and vice-versa when going the other way (same issue with any X9E/X12 <-> X9D/X10 conversion).

### DIFF
--- a/companion/src/boards.cpp
+++ b/companion/src/boards.cpp
@@ -172,37 +172,29 @@ const int Boards::getCapability(Board::Type board, Board::Capability capability)
   switch (capability) {
     case Sticks:
       return 4;
+
     case Pots:
-      if (IS_HORUS(board))
-        return 3;
-      else if (IS_TARANIS_X7(board))
+      if (IS_TARANIS_X7(board))
         return 2;
       else if (IS_TARANIS_X9E(board))
         return 4;
-      else if (IS_TARANIS(board))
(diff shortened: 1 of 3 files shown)


raphaelcoeffic:
@mpaperno Thx! Works for me on the X10.

bsongis:
We are missing a FourCC, I forgot it!

PR #5496, opened 2017-12-10
Fix problem with multi  frsky fallback

schwabe:


radio/src/telemetry/multi.cpp
```diff
@@ -416,14 +422,35 @@ void processMultiTelemetryData(const uint8_t data)
       processMultiTelemetryByte(data);
       break;
 
+    case MultiStatusOrFrskyData:
+      // Check len byte if it makes sense for multi
+      if (data >= 5 && data <= 10) {
+        multiTelemetryBufferState = ReceivingMultiStatus;
+        telemetryRxBufferCount = 0;
+      } else
```
bsongis: code style!

bsongis:
I am rather reticent for 2.2.1, what is the issue?

schwabe:
When multi code is compiled with multi status (officially more for erksy9x and er9x than opentx), the current code switches to frsky when it sees a 0x7e and never switches back to multi telemetry. 

It is not important for 2.2.1 as users can always also compile their modules with MULTI_TELEMETRY which works fine.

bsongis:
ok I will merge it just after 2.2.1 is released then

bsongis:
Time to merge this one as 2.2.1 is out

PR #5497, opened 2017-12-10
[X10] Accept old OTX_FOURCC for X12 as valid when reading data.

mpaperno:
Due to new OTX_FOURCC for X10 in RC2, existing X10 data is being detected as incompatible.


mpaperno:
Alternate, more verbose version which doesn't abuse O9X_FOURCC:  https://github.com/opentx/opentx/commit/8ca34850954ae347c43d04f7121190d290ad1d38

3djc (commented):
Maybe instead of only allowing it, we should also change the fourcc to genuine x10 ?

bsongis:
oops, my mistake when copy / pasting the hex value!

PR #5503, opened 2017-12-11
Fixes #5500

### LINKED ISSUE #5500: RSSI Shutdown timeout on X9D+/SHUTDOWN_CONFIRMATION on other platforms
SHUTDOWN_CONFIRMATION disables RSSI check on shutdown via #idef, this seem wrong and counter intuative.

On the X9D+ if you don't switch power back on and just let the screen sit there, the TX turn off after almost exactly 10s. If you compile the X9D+ with PWR_BUTTON_PRESS the timeout disappear.

bsongis:
Quick fix, not tested

### DIFF
--- a/radio/src/opentx.cpp
+++ b/radio/src/opentx.cpp
@@ -2897,6 +2897,7 @@ uint32_t pwrCheck()
           return e_power_on;
         }
         else if (readKeys() == (1 << KEY_ENTER)) {
+          timeForcePowerOffPressed = 0;
           return e_power_off;
         }
       }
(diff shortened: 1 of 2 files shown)


radio/src/opentx.cpp
```diff
@@ -2897,6 +2897,7 @@ uint32_t pwrCheck()
           return e_power_on;
         }
         else if (readKeys() == (1 << KEY_ENTER)) {
+          timeForcePowerOffPressed = 0;
```
schwabe: Nah, it is wrong here. If you press Enter it switches off anyway. The 10s timeout happens you if press nothing.

bsongis: stupid me, you are right here, I did it too quickly

PR #5507, opened 2017-12-12
Rework rssi code a bit

### LINKED ISSUE #5500: RSSI Shutdown timeout on X9D+/SHUTDOWN_CONFIRMATION on other platforms

schwabe:
Fixes #5500

### DIFF
--- a/radio/src/targets/taranis/CMakeLists.txt
+++ b/radio/src/targets/taranis/CMakeLists.txt
@@ -2,8 +2,8 @@ option(SHUTDOWN_CONFIRMATION "Shutdown confirmation" OFF)
 option(LCD_DUAL_BUFFER "Dual LCD Buffer" OFF)
 option(TARANIS_INTERNAL_PPM "Taranis internal module hack to output PPM" OFF)
 
+set(PWR_BUTTON_DEFAULT "PRESS")
 if(PCB STREQUAL X9E)
-  set(PWR_BUTTON "PRESS" CACHE STRING "Pwr button type (PRESS/SWITCH)")
   set(STICKS "STANDARD" CACHE STRING "Sticks type (STANDARD/HORUS)")
   set(CPU_TYPE STM32F4)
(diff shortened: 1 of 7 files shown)


radio/src/targets/taranis/CMakeLists.txt
```diff
@@ -34,6 +33,7 @@ elseif(PCB STREQUAL X9D+)
   set(HAPTIC YES)
   set(LUA_EXPORT lua_export_x9d)
   set(FLAVOUR x9d+)
+  set (PWR_BUTTON_DEFAULT "SWITCH")
```
bsongis: It's not possible to have PRESS here

radio/src/targets/taranis/CMakeLists.txt
```diff
@@ -56,7 +56,6 @@ elseif(PCB STREQUAL X9D)
   set(SERIAL2_DRIVER ../common/arm/stm32/serial2_driver.cpp)
   set(GVAR_SCREEN model_gvars.cpp)
 elseif(PCB STREQUAL X7)
-  set(PWR_BUTTON "PRESS" CACHE STRING "Pwr button type (PRESS/SWITCH)")
   set(CPU_TYPE STM32F2)
```
bsongis: the definition below should be removed as well

bsongis:
I will revert the CMakeLists.txt for merging it quickly

bsongis:
Tested on X9D+

radio/src/targets/taranis/CMakeLists.txt
schwabe: That only sets the default the real setting is further down

PR #5510, opened 2017-12-13
improved Horus rotary encoder driver and generic behavior

### LINKED ISSUE #5492: Scroll wheel issue
I have a problem with the scroll wheel on my Horus X12S, it's working erratically in Open-TX, if I scroll down, it will scroll down than jump up, than down again, the same if I scroll UP, it will go UP than down, than UP again.

With FrOS this is not happening, I have reinstalled Open-TX 2.2.0 3 times now, this is still happening, very annoying!

raphaelcoeffic:
Heavily tested on X10 and X7.

fixes #5492

### DIFF
--- a/radio/src/opentx.cpp
+++ b/radio/src/opentx.cpp
@@ -135,27 +135,44 @@ void per10ms()
 #if defined(ROTARY_ENCODER_NAVIGATION)
   if (IS_ROTARY_ENCODER_NAVIGATION_ENABLE()) {
     static rotenc_t rePreviousValue;
+    static bool cw = false;
     rotenc_t reNewValue = (ROTARY_ENCODER_NAVIGATION_VALUE / ROTARY_ENCODER_GRANULARITY);
-    int8_t scrollRE = reNewValue - rePreviousValue;
+    rotenc_t scrollRE = reNewValue - rePreviousValue;
     if (scrollRE) {
+      static uint32_t lastEvent;
       rePreviousValue = reNewValue;
-      putEvent(scrollRE < 0 ? EVT_ROTARY_LEFT : EVT_ROTARY_RIGHT);
-    }
+
+      bool new_cw = (scrollRE < 0) ? false : true;
+      if ((g_tmr10ms - lastEvent >= 10) || (cw == new_cw)) { // 100ms
+
+        putEvent(new_cw ? EVT_ROTARY_RIGHT : EVT_ROTARY_LEFT);
+
+        if (new_cw) TRACE("+1");
+        else TRACE("-1");
+
 #if defined(CPUARM)
-    // rotary encoder navigation speed (acceleration) detection/calculation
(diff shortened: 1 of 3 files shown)


raphaelcoeffic:
To be able test properly the *before* and *after* state, I found useful to use the following traces:
```
diff --git a/radio/src/opentx.cpp b/radio/src/opentx.cpp
index 54a38340a..d85ef72d1 100644
--- a/radio/src/opentx.cpp
+++ b/radio/src/opentx.cpp
@@ -140,6 +140,8 @@ void per10ms()
     if (scrollRE) {
       rePreviousValue = reNewValue;
       putEvent(scrollRE < 0 ? EVT_ROTARY_LEFT : EVT_ROTARY_RIGHT);
+      if (scrollRE < 0) TRACE("-1");
+      else TRACE("+1");
     }
 #if defined(CPUARM)
     // rotary encoder navigation speed (acceleration) detection/calculation
```

radio/src/opentx.cpp
```diff
@@ -136,7 +136,7 @@ void per10ms()
   if (IS_ROTARY_ENCODER_NAVIGATION_ENABLE()) {
     static rotenc_t rePreviousValue;
     rotenc_t reNewValue = (ROTARY_ENCODER_NAVIGATION_VALUE / ROTARY_ENCODER_GRANULARITY);
-    int8_t scrollRE = reNewValue - rePreviousValue;
+    rotenc_t scrollRE = reNewValue - rePreviousValue;
```
raphaelcoeffic: Please note that this one is probably not necessary (works either way), but I found it cleaner to use the same type here.

bsongis: TODO: check if we use this code on a 8bits platform (we are in an interrupt) and in this case what is rotenc_t type

raphaelcoeffic:
@bsongis I'm testing a way better code right now (using state table), as I could still see some glitches while further testing.

bsongis:
no rush, we will keep it for 2.2.2

radio/src/opentx.cpp
raphaelcoeffic: There is a define for that, it's basically 8bit an anything but ARM.

raphaelcoeffic:
exactly, no rush, this time I want it to be as perfect as possible ;-)

raphaelcoeffic:
@bsongis ok, now it's really rock solid. Works 100% for me on X10 and X7.

raphaelcoeffic:
[X12S_rot_enc_fix.zip](https://github.com/opentx/opentx/files/1564645/X12S_rot_enc_fix.zip)

raphaelcoeffic:
[X10_rot_enc_fix.zip](https://github.com/opentx/opentx/files/1564654/X10_rot_enc_fix.zip)

bsongis:
We will need more tests from nightly builds, but this seems good

PR #5518, opened 2017-12-15
[X10] CPPM trainer input in module bay

raphaelcoeffic:
This enables the use of CPPM receiver plugged into the heartbeat, batt and gnd pins of the module bay. However, unlike the taranis, it requires a small hardware mod, as the trace to the heartbeat pin is not connected by default. I used 500 Ohm (2x 1k Ohm in parallel), to have some kind of protection against the MCU pins.

img_5282

The small issue with this PR is however that it shifts the values used for other trainer input/output settings, especially the bluetooth ones.

Please note that this PR is a proof-of-concept for now, as it is missing some items:
- [ ] Modify companion to enable CPPM trainer on external module
- [ ] Conversion routines in firmware and companion once a new EEPROM version has been spawned.


hrenz:
What is that?
The Heartbeat signal is an open source/open drain signal, 
a very small impuls, every 9ms, from the internal Tx Modul, after each transmitting a frame

It can be used for synchronisation internal/external Tx moduls
but not for CPPM trainer inputs

kilrah:
Yup don't understand the HW mod either.

raphaelcoeffic:
It’s used exactly the same on the taranis for Trainer input. There is even a module from aloft hobbies using that possibility.

schwabe:
I think the mod is that in his x10 the R1 isn't there at all and @raphaelcoeffic put in a 500 Ohm resistor

raphaelcoeffic:
HEre is the link for the ready made module: https://alofthobbies.com/linker.html

raphaelcoeffic:
@schwabe I looked at the internal module motherboard sold as a spare part for the X12S, and it seems to be missing R1 as well. Did not test there, but I believe the X12S is built the same in that matter.

hrenz:
ok, I see 
the LINKER from alofthobbies don´t use a CPPM-Signal 
it uses a SBUS-Signal  for Input at Pin 2 (Heartbeat)

Does anyone have a detail of this part of the shematic round the Pin 2
Heartbeat-Impulssignal from TX  as Output  and  SBUS as Input

raphaelcoeffic:
@hrenz you do not have  use SBUS. The receivers support CPPM as well. However, to support SBUS on that input, we need to implement softserial on that pin first....

radio/src/dataconstants.h
```diff
@@ -284,8 +284,8 @@ enum BeeperMode {
     TRAINER_MODE_SLAVE,
 #if defined(PCBTARANIS)
     TRAINER_MODE_MASTER_SBUS_EXTERNAL_MODULE,
-    TRAINER_MODE_MASTER_CPPM_EXTERNAL_MODULE,
 #endif
+    TRAINER_MODE_MASTER_CPPM_EXTERNAL_MODULE,
```
bsongis, line 288: this needs an EEPROM conversion!

bsongis (changes requested):

radio/src/dataconstants.h
raphaelcoeffic: @bsongis Is it still possible to add to 2.3? What exactly is necessary?

robustini:
@raphaelcoeffic have you run aground somewhere or will you continue in this direction?
I've both the X10S Express and the X12S waiting for this implementation, I've many smd resistors to solder to activate that pin, I would put a 1K.
I don't understand why FrSky didn't solder the R1 for HB.

kilrah:
I expect this to be ditched in favor of multiprotocol module RX function.

robustini:
> 
> 
> I expect this to be ditched in favor of multiprotocol module RX function.

I'm disappointed... :(
"in favor of multiprotocol", if it worked! :-)

robustini:
The "multiprotocol module RX function" it has a big limit, being unable to share the telemetry pin, which I use in all my drones with Ardupilot (S-Port Telemetry with LUA script).
This integration does not have that limit, keep it in mind, thanks!

raphaelcoeffic:
I still use this on my X10 with 2.2.4, until I find the time to test what happens on 2.3.x with the heartbeat sync. There is a better chance of using the extra uart for this purpose. The X10 express has actually a uart RX right on the heartbeat pin. It might be useful to source one free uart rx and connect it there to make it useable on the X10 and X12S, this enabling the new access external modules at the same time.

robustini:
Awesome @raphaelcoeffic, continue with the development of this feature, don't stop! :-)

brainbubblersbest:
Yes, we need this to mount a isrm spare Part. From the express origin horus😉

kilrah:
This would do nothing for an ISRM.

brainbubblersbest:
Whats about here with making uart accessable😆
Im thinkIng about the same for x10 and x12 @raphaelcoeffic.
A physical twoway switch mounted accesible, from the bay inside, would reduce incompabilitys.
(switch between heartbeat and full Serial)

brainbubblersbest:
Can this be enabled with applied external access mod?

brainbubblersbest:
> I still use this on my X10 with 2.2.4, until I find the time to test what happens on 2.3.x with the heartbeat sync. There is a better chance of using the extra uart for this purpose. The X10 express has actually a uart RX right on the heartbeat pin. It might be useful to source one free uart rx and connect it there to make it useable on the X10 and X12S, this enabling the new access external modules at the same time.

As the externalaccessMOD is there, which exactly does what you described a year ago. ( and does it also on other Horus based Radios) is that not a Chance to solve this second oldest pull request in combination with the eamod.? 😉

raphaelcoeffic:
>As the externalaccessMOD is there, which exactly does what you described a year ago. ( and does it also on other Horus based Radios) is that not a Chance to solve this second oldest pull request in combination with the eamod.? 😉

@brainbubblersbest that sounds indeed interesting. However, I'm using a MPM now for the trainer input, which is much more convenient. However, it still requires a hardware mod to be able to work the internal module, as they needlessly block the SPORT pin for telemetry even if the upgrade module (ISRM) is installed. Nothing is perfect.... This config works a threat though on T16 & TX16S with internal and external module (both using telemetry at the same time).

PR #5521, opened 2017-12-16
Implement s.port polling for multi

schwabe:
Sport polling is implemented quite low level for regular s.port.
Go another approach for multi and add another multi type for s.port polling


radio/src/telemetry/multi.cpp
```diff
@@ -146,6 +147,14 @@ static void processMultiTelemetryPaket(const uint8_t *packet)
     case ConfigCommand:
       // Just an ack to our command, ignore for now
       break;
+    case FrskySportPolling:
+      #if defined(LUA)
```
bsongis, line 151: not sure about the code style in `#ifdef` but well...

PR #5530, opened 2017-12-18
Merge companion and firmware notes, and get them from the server

3djc:
Since we are, hum, quite bad at it !

### DIFF
--- a/companion/src/mainwindow.cpp
+++ b/companion/src/mainwindow.cpp
@@ -1020,17 +1020,10 @@ void MainWindow::sdsync()
 
 void MainWindow::changelog()
 {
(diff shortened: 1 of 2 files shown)


companion/src/mainwindow.cpp
```diff
@@ -1019,18 +1019,11 @@ void MainWindow::sdsync()
 }
 
 void MainWindow::changelog()
-{
-  ReleaseNotesDialog * dialog = new ReleaseNotesDialog(this);
```
mpaperno, line 1023: This class is no longer used so maybe remove the files also?  

Though of course this is a better name than "ReleaseNotesFirmwareDialog" since that one now shows both.  :)

bsongis:
why not!

Issue #4264, opened 2017-01-21
Improve cross-radio conversions


kilrah:
Some examples:

X9D->Horus: 
- Pot remapping can be done with some sense

Horus->Taranis:
- Throttle source shift
- Trim selections become mangled due to the 2 extra trims

Various issues with switch and source lists

In general when conversions are not straightforward or result in an invalid selection we could choose a default value and notify the user of what was changed in a messagebox or text file "conversion report"

kilrah:
Examples in #4570

raphaelcoeffic:
@kilrah #5534 has already provided some improvements (at least it is linked to this issue). Should we still leave this here open?

kilrah:
#5534 added a mechanism to do conversions from radio A to radio B, this issue was about doing something useful with it

raphaelcoeffic:
> this issue was about doing something useful with it

Thx. Do you think this is something that could be done for 2.3? (I mean, someone you know will have time do it in the very next future?)

kilrah:
It is something we should care about.

raphaelcoeffic:
> It is something we should care about.

Does that mean you think 2.3 should be rather delayed than being released without this feature? Just trying to figure out if that can be moved to another release.

PR #5534, opened 2017-12-18
[Companion] Radio conversion improvements

mpaperno:
The goal here is twofold: 1) Provide better conversion and validation of new settings, and 2) inform the user what exactly was done and (especially) what couldn't be converted properly.  Related: #4264 

I added a rudimentary conversion state tracker/logger class which accumulates conversion details throughout the process and afterwards can return a full log of events.  This state is passed around between the conversion functions of the various modules.

After conversion, if any messages are generated then the log data is presented in a table view dialog which I enhanced to provide text and HTML export functions (clipboard copy or save to file).  The dialog is modeless so you can keep it open while looking at the converted models.  Errors (e.g. missing hardware controls) are highlighted in red, conversions (e.g. SE->SG) in orange, and "adjustments" (e.g. X9 SF -> X7 SF) in black.  Suggestions to improve the wording or presentation here are welcome.

Since the state tracker keeps a copy of the old radio data, we could theoretically launch model compare after conversion, or other ideas.  That will need some UI work since obviously we don't want to launch a bunch of compare windows at the same time or if user isn't interested.  Thoughts for later.

I also spent some time making sure conversions can't happen by accident on an unsaved file (if switching radio type user is prompted to save or cancel), and letting the user opt-out of conversion altogether.  The system will detect if you're changing radio type via either a profile switch or when editing profile settings.  It also will not let you delete the active profile if you have unsaved file changes.

And finally the conversion process itself got some additions and fixes:
  * Make sure all switches/pots/knobs have default configs for destination board type;
  * Try to move more custom control names, intelligently;
  * Convert model timers and throttle source;
  * Adjust for Horus 6P switch in place of Taranis S2 knob;
  * Adjust for extra sliders on X12 and X9E to move LS and RS to proper slots as needed;
  * Make sure ALL controls are validated, eg. to always account for extra trims/analogs/switches/etc which don't exist on destination;
  * Properly indicate invalid RawSource::toString() items with "???".

Diffs should be easier to absorb per-commit vs. all at once... I tried to break them up logically.

This will conflict a little with my other open PR #5437, so I will need to rebase after that is merged (already tested), or vice versa.

"Extreme" example converting a couple X12 radio files to X7:

image

image

image

### DIFF
--- a/companion/src/storage/storage.h
+++ b/companion/src/storage/storage.h
@@ -67,47 +67,6 @@ class StorageFormat
       return board;
     }
 
-    static uint32_t getFourCC(Board::Type board)
-    {
-      switch (board) {
-        case Board::BOARD_X12S:
-          return 0x3478746F;
-        case Board::BOARD_X10:
-          return 0x3778746F;
-        case Board::BOARD_TARANIS_X7:
-          return 0x3678746F;
-        case Board::BOARD_TARANIS_X9E:
-          return 0x3578746F;
-        case Board::BOARD_TARANIS_X9D:
-        case Board::BOARD_TARANIS_X9DP:
-          return 0x3378746F;
-        case Board::BOARD_SKY9X:
-        case Board::BOARD_AR9X:
-        case Board::BOARD_9XRPRO:
-          return 0x3278746F;
-        case Board::BOARD_MEGA2560:
-        case Board::BOARD_GRUVIN9X:
-          return 0x3178746F;
-        default:
-          return 0;
-      }
-    }
-
-    uint32_t getFourCC()
-    {
-      return getFourCC(board);
-    }
-
-    static bool isBoardCompatible(Board::Type board1, Board::Type board2)
-    {
-      return (getFourCC(board1) == getFourCC(board2));
-    }
-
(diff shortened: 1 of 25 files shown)


mpaperno:
This is the prompt that happens if you try to change radio type in profile with open modified file(s):

image


And if you try to switch to an incompatible radio via the profiles menu:

image

kilrah:
Wow that's awesome again!

companion/src/storage/storage.h
```diff
@@ -98,9 +98,14 @@ class StorageFormat
       return getFourCC(board);
     }
 
-    virtual bool isBoardCompatible(Board::Type board)
+    static bool isBoardCompatible(Board::Type board1, Board::Type board2)
```
mpaperno: @bsongis Should we move this to `Boards` class?  With the getFourCC()?  `Boards` already manages getEEPromSize() and getFlashSize().  If anything, seems like all those go together.

bsongis: yes it makes sense

bsongis:
Thanks for this one!

PR #5542, opened 2017-12-19
1 too short

3djc:
Fix compilation issue since #5442

### DIFF
--- a/radio/src/stamp.cpp
+++ b/radio/src/stamp.cpp
@@ -56,8 +56,8 @@ const char* getOtherVersion()
 }
 #elif defined(STM32)
 
-__attribute__ ((section(".fwversiondata"), used)) const char firmware_version[33] = "opentx-" FLAVOUR "-" VERSION " (" GIT_STR ")";
-__attribute__ ((section(".bootversiondata"), used)) const char boot_version[33] = "opentx-" FLAVOUR "-" VERSION " (" GIT_STR ")";
+__attribute__ ((section(".fwversiondata"), used)) const char firmware_version[] = "opentx-" FLAVOUR "-" VERSION " (" GIT_STR ")";
+__attribute__ ((section(".bootversiondata"), used)) const char boot_version[] = "opentx-" FLAVOUR "-" VERSION " (" GIT_STR ")";
 
 const char* getOtherVersion()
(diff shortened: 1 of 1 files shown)


radio/src/stamp.cpp
```diff
@@ -56,8 +56,8 @@ const char* getOtherVersion()
 }
 #elif defined(STM32)
 
-__attribute__ ((section(".fwversiondata"), used)) const char firmware_version[32] = "opentx-" FLAVOUR "-" VERSION " (" GIT_STR ")";
-__attribute__ ((section(".bootversiondata"), used)) const char boot_version[32] = "opentx-" FLAVOUR "-" VERSION " (" GIT_STR ")";
+__attribute__ ((section(".fwversiondata"), used)) const char firmware_version[33] = "opentx-" FLAVOUR "-" VERSION " (" GIT_STR ")";
```
bsongis: why do you give a length here? the compiler should do it for you!

radio/src/stamp.cpp
```diff
@@ -56,8 +56,8 @@ const char* getOtherVersion()
 }
 #elif defined(STM32)
 
-__attribute__ ((section(".fwversiondata"), used)) const char firmware_version[33] = "opentx-" FLAVOUR "-" VERSION " (" GIT_STR ")";
-__attribute__ ((section(".bootversiondata"), used)) const char boot_version[33] = "opentx-" FLAVOUR "-" VERSION " (" GIT_STR ")";
+__attribute__ ((section(".fwversiondata"), used)) const char firmware_version[] = "opentx-" FLAVOUR "-" VERSION " (" GIT_STR ")";
+__attribute__ ((section(".bootversiondata"), used)) const char boot_version[] = "opentx-" FLAVOUR "-" VERSION " (" GIT_STR ")";
```
raphaelcoeffic: @schwabe didn't you assume some fixed size somewhere while reading that tag? I'm not sure anymore.

3djc: Yeah, I was worried about that too, that's why I took the PR route

schwabe: @raphaelcoeffic I would assume so too. But everything seems fine.

PR #5544, opened 2017-12-19
Fix #5543

### LINKED ISSUE #5543: OTX 2.2.1 R9M How to enable telemetry on TX
2.2.1, X9D, R9M LBT (EU)
With checkbox telemetry enabled in companion telemtry works. The radio itself has no chance to enable telemetry. It is offered in bind-options, but does not work, if one selects it.
You have to use companion first, to enable telemetry there, then it works on the radio. Why not enable  as default? You can always disable with bind option (i hope ;)).

3djc:
This fixes Fix #5543

I am unsure why/if we should not automate module telemetry setting. It is automaticaly disabled if internal module is on, it could easily be automaticaly activated when internal is off, removing the need to worry about that (it's status could still be displayed as a read only status)


Allerhopp:
User view: He sees bind options with and without telemetry. Is it possible, to switch module telemetry accordingly and to show a warning, if internal module is active?

mpaperno:
I can take care of CPN once final decision is made (though this makes sense to me).

3djc:
Issue on my side, redoing

PR #5549, opened 2017-12-20
[FW] Cosmetics in definitions.h

mpaperno:


bsongis:
nice boy scout work!

mpaperno:
I felt guilty for adding that ugly single-use macro for the quick fix earlier...   :-/

PR #5551, opened 2017-12-20
[Companion] Radio data refactoring

mpaperno:
@bsongis As promised.... (a while ago).  What do you think of this approach?

This builds fine in all OSs as far as I can tell.


elecpower:
Hi @mpaperno @bsongis since I am working (albeit slowly) on adding more fields to model print & compare, is there an intent to refactor the model print field value functions into the data classes as  fieldnameToString? I've found a few instances where additional field values have been added to radio data but not replicated in modelprint. I've been adding the additional values to the function returns and adding new print field value to string functions for fields previously not printed.

companion/src/firmwares/curvereference.cpp
```diff
@@ -0,0 +1,63 @@
...
+ */
+
+#include "curvereference.h"
+
+#include "helpers.h"     // for Helpers::getAdjustmentString()
+#include "radiodata.h"  // for ModelData
+
+QString CurveReference::toString(const ModelData * model, bool verbose) const
+{
+	if (value == 0) {
```
bsongis: 2 chars indent, please :)

bsongis (commented):
I need a coffee before I review this, but it seems really nice :)

mpaperno:
Hi Neil, that's a good question.  This is going to be a matter of opinion.  My own personal preference has been to have the individual items be responsible for naming themselves (with "toString()" etc).  With some exceptions, because we also don't want to clutter up the individual classes with special cases.  I think one exception is when the only reason to show a string is when "printing" (or comparing) the model.  

So what's been happening is that whenever we need to show a name for an item somewhere besides the printer, that "toString" gets moved/created to/in the item class.  I think ModelPrinter should be reserved for printable/overview output, not populating elements all over the UI.

Another approach would be to create a dedicated class for providing names for everything... this has some valid arguments in favor and against... I'm not sure I see anything compelling enough to switch now, but I could be convinced.

My pet-peeve in relation to this is seeing the same string repeated multiple times all over the code... not only is it more work for translation, but also changing the name in the future becomes a chore.

companion/src/firmwares/curvereference.cpp
mpaperno: Gack!

I was wondering if I should just move `CurveReference` into `io_data` along with all the classes that actually use it. Though `io_data` is getting a bit crowded already.  Need to balance between too many files and too many classes in one file...

elecpower:
Hi Max, whilst reasonably new to c++ and having to effectively reverse engineer the data fields and their text meanings for printing from UI and radio, I favour your personal preference. I feel there should be consideration for enumerating valid values in the data classes as I've found raw numeric value = string to populate the UI. There are other cases of the UI calling print functions rather than fieldToString contrary to your paragraph 2. So probably needs to be an agreement by the devs and, I hate to say it but, update the coding style guide/rules. Then refactor accordingly.
Having a tighter standard would make maintaining the radio and Companion data code sets less painful and error prone. How many times have issues been raised where Companion doesn't mimic the radio.

mpaperno:
> I feel there should be consideration for enumerating valid values in the data classes as I've found raw numeric value = string to populate the UI.

Raw, or "magic," numbers are aberrations to be eradicated with extreme prejudice.  But sometimes you just have to wince and save a fix for a later date or you never get anything actually finished (BTDT!).

> There are other cases of the UI calling print functions rather than fieldToString contrary to your paragraph 2.

Very true.  Feels like an endless chore sometimes.  I move/centralize things as I come across them for other reasons, but have never went through all the code looking for these instances specifically.

PR #5552, opened 2017-12-20
Files refactoring + cosmetics

bsongis:


radio/src/targets/taranis/bootloader/CMakeLists.txt
```diff
@@ -1,84 +0,0 @@
-# TODO if the files order is different (these 2 SRC sections exchanged), the bootloader hangs for 20s in USB init. Why?
```
bsongis, line 1: This file seems not used anymore

PR #5553, opened 2017-12-20
R9M : activate automatically MODULE telem when possible

### LINKED ISSUE #5543: OTX 2.2.1 R9M How to enable telemetry on TX

3djc:
With this change, user doesn't have to deal with module telem anymore : if possible, module telem is enabled. Status of said telemetry is displayed as read only

@mpaperno, if accepted, could you do the Companion part ? I'm unsure if we just want to hide module telem altogether, or if we want to replate radio behaviour (read only status). The previously used setting bit is back to spare

This also fixes #5543


radio/src/datastructs.h
```diff
@@ -679,8 +679,7 @@ PACK(struct ModuleData {
       uint8_t receiver_telem_off:1;     // false = receiver telem enabled
       uint8_t receiver_channel_9_16:1;  // false = pwm out 1-8, true 9-16
       uint8_t external_antenna:1;       // false = internal antenna, true = external antenna
-      uint8_t sport_out:1;
-      uint8_t spare2;
+      uint8_t spare3;
```
schwabe: This does not look right. We should have spare1:1 instead sport_out. We always document spare bits

radio/src/gui/128x64/model_setup.cpp
```diff
@@ -1154,23 +1154,13 @@ void menuModelSetup(event_t event)
...
+        if (IS_MODULE_R9M(moduleIdx)) {
+          lcdDrawTextAlignedLeft(y, STR_MODULE_TELEMETRY);
           if (IS_TELEMETRY_INTERNAL_MODULE()) {
-            lcdDrawTextAlignedLeft(y, STR_MODULE_TELEMETRY);
             lcdDrawText(MODEL_SETUP_2ND_COLUMN, y, STR_DISABLE_INTERNAL);
           }
           else {
-            lcdDrawTextAlignedLeft(y, STR_MODULE_TELEMETRY);
-            lcdDrawText(MODEL_SETUP_2ND_COLUMN, y, STR_BINDING_OPTION);
+            lcdDrawText(MODEL_SETUP_2ND_COLUMN, y, STR_MODULE_TELEM_ON);
```
schwabe: 👍

radio/src/gui/gui_common.h
```diff
@@ -157,7 +157,7 @@ const mm_protocol_definition *getMultiProtocolDefinition (uint8_t protocol);
 #define IS_R9M_OR_XJTD16(x)            ((IS_MODULE_XJT(x) && g_model.moduleData[x].rfProtocol== RF_PROTO_X16) || IS_MODULE_R9M(x))
 
 #define FAILSAFE_ROWS(x)               ((IS_MODULE_XJT(x) && HAS_RF_PROTOCOL_FAILSAFE(g_model.moduleData[x].rfProtocol)) || MULTIMODULE_HASFAILSAFE(x) || IS_MODULE_R9M(x))  ? (g_model.moduleData[x].failsafeMode==FAILSAFE_CUSTOM ? (uint8_t)1 : (uint8_t)0) : HIDDEN_ROW
-#define R9M_FCC_OPTION_ROW             (IS_TELEMETRY_INTERNAL_MODULE() ? TITLE_ROW : (uint8_t)0)
+#define R9M_FCC_OPTION_ROW             (IS_TELEMETRY_INTERNAL_MODULE() ? TITLE_ROW : (uint8_t)-1)
```
schwabe: TITLE_ROW is -1 so R9M_FCC_OPTION_ROW is always TITLE_ROW now.

radio/src/gui/gui_common.h
```diff
@@ -157,7 +157,7 @@ const mm_protocol_definition *getMultiProtocolDefinition (uint8_t protocol);
...
 #define R9M_OPTION_ROW                 (IS_MODULE_R9M_FCC(EXTERNAL_MODULE) ? R9M_FCC_OPTION_ROW : TITLE_ROW)
 #define EXTERNAL_MODULE_OPTION_ROW     (IS_MODULE_R9M(EXTERNAL_MODULE) ? R9M_OPTION_ROW : (IS_MODULE_SBUS(EXTERNAL_MODULE) ? TITLE_ROW : MULTIMODULE_OPTIONS_ROW))
```
schwabe: So just replace this with 
```
((IS_MODULE_R9M(EXTERNAL_MODULE) || IS_MODULE_SBUS(EXTERNAL_MODULE)  ? TITLE_ROW : MULTIMODULE_OPTIONS_ROW)
```

And remove the other R9M_*_OPTION_ROW

radio/src/datastructs.h
```diff
@@ -679,8 +679,7 @@ PACK(struct ModuleData {
       uint8_t receiver_telem_off:1;     // false = receiver telem enabled
       uint8_t receiver_channel_9_16:1;  // false = pwm out 1-8, true 9-16
       uint8_t external_antenna:1;       // false = internal antenna, true = external antenna
-      uint8_t sport_out:1;
-      uint8_t spare2;
+      uint8_t spare2:2;
```
bsongis: no!

schwabe: :2 means two bits. Nothing will break here because the other structs in the union are larger but the right we is to have
```
[...]
uint8_t external_antenna:1
uint8_t spare2:1
uint8_t spare3;

radio/src/datastructs.h
schwabe: This looks better thanks!. NO idea why github thinks my old comment is still valid

bsongis:
Double checked with FrSky, this PR looks right. Just the Companion part is missing

mpaperno:
Waiting on the final-final[-final] confirmation before CPN changes...  :)

3djc:
This last version includes some important changes. S.Port link (aka module telemetry) is managed automaticaly and displayed as information only.

image

R9M Eu has been reworked. Because for EU version, mode change is done on bind ONLY, the UI has been enhanced to tell you how you last did binfd as readonly info

image

mpaperno:
If I got this right, we need to remove the "Module Telemetry" checkbox in Companion, but still preserve the pxx bits for round-trip back to radio.  Yes?

As far as radio UI, so it will always show "S.Port [link]" as on/off based on internal module status, regardless of bind option, right?  With the EU version it is then more obvious that telemetry is actually off because the bind options are shown.  But for FCC, they're not shown, so actual telemetry status is more of a mystery.  Would it make sense to show the bind options for FCC as well?  Also nice to see the channel range right there.

mpaperno:
Do we care about showing the bind mode string in CPN? Or S.Port enabled/disabled based on internal module selection?  Info-only of course, not changeable.

kilrah:
> With the EU version it is then more obvious that telemetry is actually off because the bind options are shown. But for FCC, they're not shown, so actual telemetry status is more of a mystery. 

It is prone to confusion anyway since this is only taken at bind time. So what's selected there may or may not match what the receiver is actually doing at a given moment.
I'm not sure it's good to display it for that reason.

mpaperno:
You could rebind it using jumpers with different options (I presume), or the Rx might not be bound anymore at all, I get that.  But then why display it for the EU version?  Seems inconsistent.  (Or did you mean display it in CPN?)

Unfortunately "Last Known Bind Options" is a bit long for a field label....  LKBO?  :)

3djc:
This display only applies to R9M EU, so there is not jumper option at work. The issue with this one is that power is set on bind, so previous way of working gives you no info at all on power set. This display is not totaly foolprof (while writing this, I'm thinking of waywe could explore to try to ensure the bind did actually complete), but if you use the radio 'logicaly', it gives you the info of last bind. It seems to me way better that relying on you memory to try to recall what was your last bind option

PR #5555, opened 2017-12-20
Bsongis/sonar

bsongis:


.travis.yml
```diff
@@ -82,6 +82,7 @@ addons:
     #   secure: "Qkmp4VrQJtKgZ/pz65SNnCDiF4slgFfQS/3whyLXLcDbFb1Hv1oj7Hqz9EVDUSO2eEhgQGaOKkRc0CfAAOBwW6OvhE0xh7P1CJEjzOp6+3gI/LyZUNsqa+awVRZTsfm1YZaOPaPWJ7cI0uz3/OoSUeIecs4PDWPrQMGRlsrq7Q8="
     branches:
       - "2.2"
+      - bsongis/sonar
```
bsongis: should be removed at the end

companion/src/modeledit/setup.cpp
```diff
@@ -522,7 +522,12 @@ void ModulePanel::update()
     ui->label_option->setText(qApp->translate("Multiprotocols", qPrintable(pdef.optionsstr)));
   }
 
-  // Failsafes
+  ui->autoBind->setVisible(mask & MASK_MULTIMODULE);
```
bsongis: I will remove this

bsongis (commented):
can be merged now!

companion/src/modeledit/setup.cpp
```diff
@@ -522,7 +522,12 @@ void ModulePanel::update()
...
+  ui->autoBind->setCheckState(module.multi.autoBindMode ? Qt::Checked : Qt::Unchecked);
+  ui->lowPower->setVisible(mask & MASK_MULTIMODULE);
+  ui->lowPower->setCheckState(module.multi.lowPowerMode ? Qt::Checked : Qt::Unchecked);
```
mpaperno: This block is duplicating what is just above it.

radio/src/CMakeLists.txt
```diff
@@ -443,7 +443,7 @@ set(CMAKE_EXE_LINKER_FLAGS "")
 set(CMAKE_C_LINK_EXECUTABLE "<CMAKE_C_COMPILER> <FLAGS> <LINK_FLAGS> <OBJECTS>  -o <TARGET>")
 set(CMAKE_CXX_LINK_EXECUTABLE "<CMAKE_CXX_COMPILER> <FLAGS> <LINK_FLAGS> <OBJECTS>  -o <TARGET>")
 
-use_cxx11()  # ensure gnu++11 in CXX_FLAGS with CMake < 3.1
+use_cxx14()  # ensure c++14 in CXX_FLAGS with CMake < 3.1
```
mpaperno, line 446: So which compiler versions does this knock out?

radio/src/gui/480x272/menu_model.cpp
```diff
@@ -118,7 +118,7 @@ void editName(coord_t x, coord_t y, char * name, uint8_t size, event_t event, ui
             killEvents(event);
             break;
           }
-          // no break
+          [[fallthrough]];
```
mpaperno, line 121: This is legal cpp?

.travis.yml
mpaperno: Probably don't want this in 2.2 branch?

radio/src/CMakeLists.txt
```diff
@@ -393,7 +393,7 @@ endforeach()
 add_definitions(-DCORRECT_NEGATIVE_SHIFTS)
 
 if(NOT MSVC)
-  set(WARNING_FLAGS "${WARNING_FLAGS} -Wall -Wno-strict-aliasing -Wformat -Wreturn-type -Wunused -Wuninitialized -Wunknown-pragmas -Wno-switch -Wtype-limits")
+  set(WARNING_FLAGS "${WARNING_FLAGS} -Wall -Wno-strict-aliasing -Wformat -Wreturn-type -Wunused -Wuninitialized -Wunknown-pragmas -Wno-switch -Wtype-limits -Wno-attributes")
```
mpaperno, line 396: What does disabling this warning do?

radio/src/CMakeLists.txt
mpaperno: Here's an answer for gcc: https://stackoverflow.com/questions/31965413/compile-c14-code-with-g
I wonder if CMAKE knows which flag to use for older gcc with `CMAKE_CXX_STANDARD 14` (but the `use_cxx14()` macro is a bit simplistic).

radio/src/gui/480x272/menu_model.cpp
bsongis: Yes this is what I learnt last week
Sonar doesn't understand `// no break` but it understands `[[fallthrough]];`

radio/src/CMakeLists.txt
bsongis: some compilers will raise a warning when seeing `[[fallthrough]];` (which is called an attribute)

.travis.yml
bsongis: perhaps an `*/sonar` would work here

radio/src/targets/horus/audio_spi_driver.cpp
```diff
@@ -136,34 +136,34 @@ void audioSpiInit(void)
...
-      AUDIO_SPI->CR1 |= 1<<3; // Fsck=Fpclk/4=18Mhz
+      AUDIO_SPI->CR1 |= 0x01 << 3; // Fsck=Fpclk/4=18Mhz
       break;
     case SPI_SPEED_8:
-      AUDIO_SPI->CR1 |= 2<<3; // Fsck=Fpclk/8=9Mhz
+      AUDIO_SPI->CR1 |= 0x02 << 3; // Fsck=Fpclk/8=9Mhz
       break;
     case SPI_SPEED_16:
-      AUDIO_SPI->CR1 |= 3<<3; // Fsck=Fpclk/16=4.5Mhz
+      AUDIO_SPI->CR1 |= 0x03 << 3; // Fsck=Fpclk/16=4.5Mhz
```
bsongis, line 151: here we had a warning from sonar

radio/src/gui/480x272/menu_model.cpp
schwabe: yes. That is C++17. But similar things exist in C+11 already: http://en.cppreference.com/w/cpp/language/attributes

bsongis: I think this attribute is C++14

mpaperno:
So, I know this is merged now and we'll have to deal with any fallout from bumping C++ version requirement, but what do the Sonar changes actually accomplish?  Is Sonar usable now?  A little description would be great.

I'm skeptical because before this it had a log of bogus warnings from other code it doesn't understand, like Qt signal/slots (and ISRs?).  Or very particular about code format, and has the "complexity threshold" of a pigeon (no offense to pigeons).  I have Sonar e-mails blocked at my server (dunno if it still sends them now), and if it starts leaving a ton of bogus warnings on PRs that's going to get annoying.  Is the scan then just for reviewing the issues @ Sonar site?

radio/src/CMakeLists.txt
mpaperno: So we're silencing a potentially useful compiler warning so that the code scanner doesn't issue a bogus warning?

cmake/Macros.cmake
```diff
@@ -42,11 +42,11 @@ macro(git_id RESULT)
   endif()
 endmacro(git_id)
 
-macro(use_cxx11)
+macro(use_cxx14)
   if (CMAKE_VERSION VERSION_LESS "3.1" AND CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
-    set (CMAKE_CXX_FLAGS "--std=gnu++11 ${CMAKE_CXX_FLAGS}")
+    set (CMAKE_CXX_FLAGS "--std=c++14 ${CMAKE_CXX_FLAGS}")
```
mpaperno, line 47: As indicated at the [SO post](https://stackoverflow.com/questions/31965413/compile-c14-code-with-g) I quoted earlier, sounds like this won't work for gcc < v5.  All the build automated build system (and all of my own) are still on 4.x as far as I can tell.  Luckily I think we're all moved past cmake 3.1?  But then maybe we should just require 3.1 up front in the companion CMake file?

radio/src/CMakeLists.txt
bsongis: no, this is gcc which issues a bogus warning, not sonar!

bsongis:
Sonar is still not useable, we just remove some warnins which is not bad after all. Sonar already found some potential issues, I would like we succeed in having 0 blocker bugs so that we have a real check on each PR. I disabled all automatic checks (with ton of emails).

cmake/Macros.cmake
bsongis: Hm. yes not wrong, this needs to be checked (including Windows builds), if I did something wrong I will revert. But we are at c++17 now, c++14 should not be too wrong ...

mpaperno: So far so good on MSVC2015 and MinGW 5.3 (gcc 5.3.0, I guess this one is >5 after all :) ).  CMake 3.10.

radio/src/CMakeLists.txt
mpaperno: I understand, I just meant that `[[fallthrough]]` was to silence a Sonar warning and then this switch is needed to silence warnings about `[[fallthrough]]`.  :)

PR #5557, opened 2017-12-22
[Companion] Qualify all translation strings

mpaperno:
Gets rid of all remaining generic `QObject::tr()` calls and replaces with class/object-qualified ones.  In most cases the `QObject::` prefix was unnecessary to begin with since the classes already had their own `tr()` methods.

I also centralized some strings which get repeated a lot in dialog titles ("Warning", "Error", "Companion").

This should be the last of the big translation rebases.  No functionality changes here.

@LapinFou The good news is that all existing translations are preserved (here and the previous translation PR). But the ones that have moved to their own class names (from QObject) are now marked "unverified" (with the yellow `?`).  A search/replace in the .ts file (with text editor) would probably work to mark them all verified, though it could cause some inadvertent verification.  Let me know if I can help.

### DIFF
--- a/companion/src/burnconfigdialog.cpp
+++ b/companion/src/burnconfigdialog.cpp
@@ -57,7 +57,7 @@ burnConfigDialog::burnConfigDialog(QWidget *parent) :
       ui->label_sb1->hide();
       ui->label_sb3->hide();
       ui->samba_location->hide();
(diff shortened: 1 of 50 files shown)


LapinFou:
Thanks for the notification. I will check all _unverified_ fields.

PR #5563, opened 2017-12-24
128x64 Display RAS when needed

### LINKED ISSUE #5520: TGY 9XR PRO-OpenTx 2.2. – the value of the RAS is not displayed.
The value of the RAS is not displayed (indicated). To be on the side of the 7/9 "Analogs"??
The value of the SWR was already sorted. For the correct setting of the antenna TX is the value of the RAS (SWR) important.

3djc:
Add display to non x7 ARM radio when XJTcd build-9x

image

This fixes #5520


radio/src/gui/128x64/radio_diaganas.cpp
```diff
@@ -56,6 +56,24 @@ void menuRadioDiagAnalogs(event_t event)
...
   }
 
+  // SWR
+#if defined(PCBX7)
+  if(IS_MODULE_XJT(EXTERNAL_MODULE) && !IS_INTERNAL_MODULE_ON()) {
+    coord_t y = MENU_HEADER_HEIGHT + 1 + (NUM_STICKS+NUM_POTS+NUM_SLIDERS+1)/2 * FH + 1 * FH + 2;
+      
+    lcdDrawText(1, y, "RAS:");
+    lcdDrawNumber(1 + 4*FW, y, telemetryData.swr.value, LEFT);
+  }
```
schwabe: But isn't RAS always invalid on X7 XJT as it does not have the measurement circuit?

PR #5568, opened 2017-12-26
Model print enhancements - General settings

### LINKED ISSUE #5462: [Companion] Print model enhancements
There are many fields that are currently not included in the model report. Most notably for general settings and telemetry.

elecpower:
Next batch of enhancements.
Tested using X9D+ settings.
Some formatting compromises to support model comparison.
Telemetry next
X12 will need extra fields printed and additional values specific to radio
@mpaperno will come back and refactor fieldToString and 'magic'/raw throughout Companion once all the fields are printed
Partly fixes #5462

### DIFF
--- a/companion/src/modelprinter.cpp
+++ b/companion/src/modelprinter.cpp
@@ -21,6 +21,7 @@
 #include "helpers.h"
 #include "modelprinter.h"
 #include "multiprotocols.h"
+#include "boards.h"
 
 #include <QApplication>
 #include <QPainter>
@@ -45,12 +46,9 @@ ModelPrinter::~ModelPrinter()
 {
 }
 
-QString ModelPrinter::printBoolean(bool val)
+QString formatTitle(const QString & name)
 {
-  if (val)
-    return tr("Y");
-  else
-    return tr("N");
+  return QString("<b>" + name + "</b>&nbsp;");
 }
 
 void debugHtml(const QString & html)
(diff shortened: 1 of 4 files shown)


mpaperno:
Looks great!  Tested with several models, didn't see any issues. 👍 

My only comment/request so far is to make the default text color black instead of gray...  nothing really to do with this PR, I know.  But it's hard to read.  I think the headings/labels being bold will be enough of a distinction.  I believe the default text color is set in `multimodelprinter.h:54` as the default argument value to `endCompare()` (was a bit hard to find).

I guess the other minor thing is how in a lot of cases the item details are appended in parenthesis but no space.  (Not a criticism of this PR as I understand it is just following current convention.)  E.g.:

    1: Internal Radio System, Protocol(FrSky XJT (D16)), Channels(1-16), Receiver(0)

I understand this is/was desirable in the Inputs/Mixes editor to save space (which uses that part of ModelPrinter), but in the other cases it seems unnecessary.  For example I'd propose:

    1: Internal Radio System; Protocol: FrSky XJT (D16); Channels: 1-16; Receiver: 0;

I don't want to get into a formatting debate, but I do wonder if having a shared helper function to format those parts would be useful.  e.g.:

    QString formatItem(const QString & name, const QString & value) {
      return QString("%1(%2); ").arg(name, value);
    }

Anyway, great work, thanks!

elecpower:
Agree with you @mpaperno on formatting inconsistencies and these need attention. Agree with some Helper functions /  macros as lots of repeat formatting. Also took into account your recent translation refactoring. There is conflicted formatting when values used for model comparison (red, green, grey, black, regular, bold).
I toyed with various layouts and whilst they worked for a single model they often resulted in corrupt formatting when used for model comparisons so took the 'safer' approach when this occurred.
With the devs working so fast on lots of changes and me plodding along, my strategy was to print out the values with some structure as something is better than nothing. Then have a second pass review the various layouts used and tune it up based on feedback like yours.
Also I'm probably going to create separate print for Inputs and Mixes to avoid breaking the lists. IMHO the lists should not be using print functions.

mpaperno:
That all sounds good to me and makes sense.  Do you want to work more on this PR, or merge it now and refine later?

elecpower:
I would prefer to merge as is and revisit once I have finished the first pass. This will also provide others the opportunity to find any bugs in my reverse engineering of the radio variants which I can address in the next pass.
I have already finished the failsafes and basic telemetry. Working on telemetry sensors (these will be 'fun' to format)

mpaperno:
Thanks!

PR #5574, opened 2017-12-28
Bsongis/frsky xlite

bsongis:


bsongis:
I have done a lot of refactoring with `#ifdef`

radio/src/mixer.cpp
```diff
@@ -534,13 +534,6 @@ void evalInputs(uint8_t mode)
     if (v < -RESX) v = -RESX;
     if (v >  RESX) v =  RESX;
 
-#if defined(PCBTARANIS) && !defined(PCBX7) && !defined(SIMU)
-    // TODO why not in the driver?
```
bsongis, line 538: @3djc I have removed this `TODO`, I have only checked X9D+, would you check X9E?

radio/src/targets/taranis/haptic_driver.cpp
```diff
@@ -20,18 +20,19 @@
 
 #include "opentx.h"
 
-#if defined(PCBX9E) || defined(PCBX9DP)
+#if defined(HAPTIC_PWM)
+// TODO test Haptic on X7 (I added PWM support)
```
bsongis, line 24: `HAPTIC_PWM` needs to be tested on X7

radio/src/translations/en.h.txt
```diff
@@ -429,14 +429,17 @@
 #define TR_VSWASHTYPE                  "---\0""120\0""120X""140\0""90\0"
 
 #if defined(PCBHORUS)
-#define LEN_VKEYS                      "\006"
-#define TR_VKEYS                       "Menu\0 ""Exit\0 ""Enter\0""Up\0   ""Down\0 ""Right\0""Left"
+  #define LEN_VKEYS                    "\005"
+  #define TR_VKEYS                     "Menu\0""Exit\0""Enter""Up\0  ""Down\0""Right""Left\0"
```
bsongis, line 433: I wonder why length was 6 here

bsongis (commented):
Still needs Companion support (almost the same than X7 except 2 switches instead of 6)

radio/src/mixer.cpp
mpaperno: Is this in the driver now?  There's an exception somewhere in Simulator/simu for this also... have to find it.

mpaperno:
For Companion/Sim, differences from X7: 

 - [x] only SA and SB switches
 - [x] only R & E trims (named TrH and TrV)
 - [ ] navigation like 9X - ?
 - [ ] new simulator skin (@mhotar ?)
 - [x] new FourCC
 - [x] new Variant - ?

Anything else?

bsongis:
Exact, I think trims should be named TrH (Horizontal) and TrV (Vertical)

radio/src/targets/taranis/hal.h
```diff
@@ -61,8 +78,8 @@
   #define ROTARY_ENCODER_POSITION()     (ENC_GPIO->IDR >> 12) & 0x03
 #elif defined(PCBX7)
   #define ENC_GPIO                      GPIOE
-  #define ENC_GPIO_PIN_A                GPIO_Pin_9  // PE.09
-  #define ENC_GPIO_PIN_B                GPIO_Pin_11 // PE.11
+  #define ENC_GPIO_PIN_A                GPIO_Pin_10 // PE.10
+  #define ENC_GPIO_PIN_B                GPIO_Pin_12 // PE.12
```
mpaperno: Intentional change of pins for X7?

mpaperno:
Something in this branch breaks the X7 Enter key (rotary press), at least in Simulator.  It acts instead as a rotary-left or a PLUS key, meaning e.g. it scrolls the view up instead of selecting an item.  Was not able to chase it down after short look, but doesn't seem like a Simulator-specific problem.

mpaperno:
Updated the Companion [checklist](https://github.com/opentx/opentx/pull/5574#issuecomment-354369371), above.

radio/src/targets/taranis/hal.h
bsongis: no it's a mistake

mpaperno: That fixed the enter key issue  👍

PR #5579, opened 2017-12-30
Save memory (move usb mass storage buffer to reusableBuffer)

schwabe:
I still feel guilty about giving X9D+ builds about 4k less memory per default as USB mass storage is always included. This commit does

- Lower that buffer to 512 bytes per default on F2 targets, makes usb speed slightly slower ~600 kB/s instead ~700 KB/s in my tests, saving 3584 bytes
- Save another ~300 bytes by putting that buffer into reusableBuffer
- Put a 128byte variable on stack that was global but does not need to be


raphaelcoeffic:
Thx! That’s going to help a lot of people fighting with lua & memory!

radio/src/mixer.cpp
```diff
@@ -1109,6 +1108,7 @@ uint8_t   flightModeTransitionLast = 255;
 
 void evalMixes(uint8_t tick10ms)
 {
+  int32_t sum_chans512[MAX_OUTPUT_CHANNELS] = {0};
```
bsongis, line 1111: The initialization seems not needed as it is done below (when fadde is in use)

bsongis:
I will only remove the unneeded initialization in case it increases flash usage on 9X

bsongis:
If I am not mistaken it uses 26bytes for nothing

PR #5585, opened 2018-01-02
x10 antenna is either int or ext

### LINKED ISSUE #5564: X10S openTx V2.2.1  antenna selection intern extern  both  FrOs only intern or extern
at FrOS I can select intern OR extern antenna, but not both
at openTx I can select intern, extern, AND both

Mybe only an error at selection gui at openTx?, comming from the X12 gui selection

3djc:
Behavior confirmed by FrSky.

This fixes #5564


kilrah:
Just the alignment...

PR #5595, opened 2018-01-04
Align python crossfire parser to c equivalent

3djc:
Since I do not have access to crossfire specs, I have updated the python scripts based on OpenTx 'c' crossfire parser.

The changes include support for negative values and GPS packets parsing

### DIFF
--- a/radio/util/crossfire-parse.py
+++ b/radio/util/crossfire-parse.py
@@ -79,13 +79,8 @@ def crc8(buffer):
 # note : in python arrays slicing payload[x:y] means from 'x' included to 'y' not included,
 #        y - x gives the number of bytes sliced.
 def ParseGPS(payload):
-    lat = struct.unpack('>i',bytes(bytearray(payload[0:4])))[0] / 1e7          #(degree / 10`000`000)
-    long = struct.unpack('>i',bytes(bytearray(payload[4:8])))[0] / 1e7         #(degree / 10`000`000)
-    speed = struct.unpack('>H',bytes(bytearray(payload[8:10])))[0] / 100       #(km/h / 100)
-    head = struct.unpack('>H',bytes(bytearray(payload[10:12])))[0] / 100       #(degree / 100)
-    alt = struct.unpack('>H',bytes(bytearray(payload[12:14])))[0] - 1000       #(meter - 1000m offset)
-    numsat = payload[14]
-    return "[GPS] lat:%f long:%f speed:%d heading:%d alt:%d numsat:%d" % (lat, long, speed, head, alt, numsat)
+    lat, long, speed, head, alt, numsat = struct.unpack('>iiHHHB',bytes(bytearray(payload[0:15])))          # bytes(bytearray) casting is required for python 2.7.3 compatibility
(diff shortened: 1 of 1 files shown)


3djc:
This was tested based on logs submitted by tgreer, so it would be nice to test again different logs too, just to be sure

radio/util/crossfire-parse.py
```diff
@@ -70,28 +71,47 @@ def dump(data, maxLen=None):
...
     crc = 0
     for c in buffer:
         crc = crc8tab[crc ^ c]
     return crc
 
+# note : python arrays slicing is a bit strange payload[x:y] means from 'x' included to 'y' NOT included
+#        therefore y - x gives the number of bytes sliced.
 def ParseGPS(payload):
-    pass
+    lat = getCrossfireValue(True, payload[0:4]) / 1e7
```
bsongis: do you mean 0x1e7?

radio/util/crossfire-parse.py
```diff
@@ -70,28 +71,47 @@ def dump(data, maxLen=None):
...
+    else:
+        return value.uint
+
 def crc8(buffer):
     crc = 0
     for c in buffer:
         crc = crc8tab[crc ^ c]
     return crc
 
+# note : python arrays slicing is a bit strange payload[x:y] means from 'x' included to 'y' NOT included
```
bsongis: why is it strange? I prefer you remove this comment ;)

radio/util/crossfire-parse.py
```diff
@@ -70,28 +71,47 @@ def dump(data, maxLen=None):
     0xAD, 0x78, 0xD2, 0x07, 0x53, 0x86, 0x2C, 0xF9
 ]
 
+# logs data are BigEndian
+def getCrossfireValue(signed, data=[], *args):
```
bsongis: I have to find a nicer way to do this tomorrow :)

radio/util/crossfire-parse.py
3djc: Not quite, I mean 10 000 000

Python 2.7.13 (default, Jan 19 2017, 14:48:08) 
[GCC 6.3.0 20170118] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> print 1e7
10000000.0
>>>

radio/util/crossfire-parse.py
3djc: Well, it seems a lot of peop;le (including myself) bump into this all the time with python slicing, but I'll make it politicaly correct :)

radio/util/crossfire-parse.py
3djc: Well, you could easily remove the signed argument and assume that 4 bytes data are signed and less then 4 is unsigned, which seems to be true in current version, but as this is more like a tool for debuging, I wanted it to be crystal clear (unless thats not the 'signed' bothering you ?)

radio/util/crossfire-parse.py
mpaperno: Much easier to read than counting zeros, IMHO.  And another reason hex should be written with uppercase chars...  :)

bsongis:
Thanks!

Issue #2354, opened 2015-06-03
Request, Nativ Mavlink Telemetry in OpenTX 


ilihack:
Thank you allot for you creat work on the OpenTX firmware fur us an for my Taranis!

but i miss one think, its the nativ mavlink Telemetry compatibility in the comming OpenTX 2.1 like this great work: http://droni.eu/mag/2014/10/12/mavlink-data-displayed-on-taranis-without-converter/

with the next big openTX 2.1 Telemetry update, makes it maybe easy to integrate mavlink due to its flexibility: 
http://openrcforums.com/forum/viewtopic.php?f=45&t=6887&start=90

dk7xe:
Would be really a big step forward for all us APM users. And we can get rid of additional HW that is currently needed to have nice Telemetry as it is done with [MAVLink to FrSky S.Port Protocol Converter](https://github.com/Clooney82/MavLink_FrSkySPort)

Gerald

bsongis:
Why would it be an OpenTX issue? Is there a Mavlink dedicated module that could be plugged in the module bay of the Taranis?

kilrah:
Mavlink telemetry is typically received through a simple transparent serial link, I believe that issue is about integrating native mavlink protocol like there is as an option on the 9x (i.e. in 2.1 a new telemetry handler to decode the stream and populate the generic sensors).

bsongis:
But it would be a D-Telemetry, not SPORT, right?

kilrah:
No it would be raw mavlink! Like on the 9x. No FrSky gear involved, just receive raw mavlink protocol either on UART3 or sport

Clooney82:
Would be a great enhancement for many people.
Best way would if it could be possible to send mavlink data via s.port to RC, but don´t know if this is possible.
Or connect a 433/915 Telemetry Modul direct to RC via serial port and read data from this.

If possible I would prefer a solution without additional hardware.

kilrah:
A link on the referenced issue shows that Ardupilot now supports putting out both the hub and the smart port protocol, so I'm not sure adding mavlink support makes much more sense anymore.

Clooney82:
Yes ardupilot supports both protocols more or less good, but there are more FCs out there.
For me it would be a nice to have feature which could save weight on smaller copters/helis/planes.

dk7xe:
Ardupilot supports currently a few parameters via FrSky protocol as i can see on http://copter.ardupilot.com/wiki/common-frsky-telemetry/
[...]

PR #5600, opened 2018-01-05
MAVLink telemetry support - WIP

hsteinhaus:
Most important thing first: This PR is not ready for merge yet - furthermore it should be a base for discussion and review.

## What's this PR about
The code of this PR receives a native MAVLink stream on one of the serial ports of the transmitter. Lots of discussion about the desire to add this functionality and about use cases are found in #2354 .

## Serial port handling
Currently, the implementation is tied to and only tested on UART3 of the X9D+, but the code is kept as extensible as possible here. More than one receiver on more than one serial port is conceivable for the future. At the moment,  I prefer to run the MAVLink telemetry code in parallel to conventional  (S.PORT/CRSF) telemetry to have both sources of data available at a time. But this might have to be discussed, i.e. from a perspective of resources.

## Data access
All MAVLink data is provided to telemetry scripts in tables. There is currently exactly one table per message type, named latest_MESSAGETYPE. The table contains all fields of the generic header as well all payload fields. E.g., a  telemetry script could access the latest known vehicle status via "latest_HEARTBEAT.system_status" .

## Code generation
The low level code for parsing the MAVLink stream is provided by the MAVLink C bindings, version 2.0, which have been added as a submodule in radio/src/thirdparty/GCS_MAVLink/include_v2.0 . The code works on a fixed message buffer. After successful parsing of a complete message, a generated message handler will be invoked. This handler (found in radio/src/lua/mavlink2lua.[cpp|h] ) is generated by a custom Python script (found in tools/mavgen_lua - derived from the original MAVLink bindings generator). Additionally, a Lua unit containing all symbolic enums is generated for convenience. I have added the generated code to git, as I assume it will change not too often in the long run (after it is stabilized) and I do not see a real benefit in regenerating at during every build.

## What's missing (and in the queue)
1. message/stream subscription is missing completely yet, so an autopilot will probably not deliver any useful data
2. a blacklisting mechanism to block out obscure and outdated MAVLink messages to conserve some valuable resources (mainly program memory). 
3. correct MAVLink2 handshake. The code is MAVLink2 ready, but other systems might refuse to switch to v2 if we do not shake hands properly.

## Test status
Only one message type has been intensively dealt with so far: HEARTBEAT. The code of this PR is receiving HBs and sending them. Beneath that, I have replayed some recorded ArduCopter tlogs, which at least did not crash anything on the TX. All replayed message typed have been counted as successfully decoded in mavling_status_t. 

An example telemetry script could be found in tools/mavgen_lua/mavtel.lua The script should produce output like this:

image

## Known problems
1. I have seen crashes, if a lot of incoming traffic is present on the serial port during bootup. I doubt that this is directly related to the MAVLink part - possibly some race condition related to the DMA fifo? Any help appreciated here.
2. there is no callback mechanism between telemetry receiver code and Lua. It would be very nice if a telemetry script could be woken up when new data has arrived instead of polling the latest_* tables continuously. 
3. The code providing the Lua tables seems to produce still lots of temporaries, which have been to GCed away later. I am not deep enough into Lua to judge if this is a problem or not and how to fix it best.

Latest memory footprint for Tatanis X9D+
```
---- Size report for firmware.elf (MCU: STM32F205xE) ----

Section Details:
section              size(B)         addr   loc
.text                 449032   0x08000000   Flash
.init_array               40   0x0806da08   Flash
.fini_array                4   0x0806da30   Flash
.data (added)            356   0x0806da34   Flash
.data                    356   0x20000000   SRAM1
.bss                   65308   0x20000180   SRAM1
._user_heap_stack       1024   0x2001009c   SRAM1

Totals         KB    usage  of ttl	KB free
  Flash:   438.90   85.72%     512	  73.10
  SRAM1:    65.12   58.15%     112	  46.88
  SRAM2:     0.00    0.00%      16	  16.00


hsteinhaus:
Rebased to latest 2.2 and trying to make Travis happy.

PR #5609, opened 2018-01-08
Fixes 5602 Special Functions Inc/Dec Gvar

### LINKED ISSUE #5602: Misbehaviour of GV's increment setting - X9D+, OpenTx 2.2.1
I set two Adjust GV special functions:
vg1

The GV limits are initially those (unmodified):
vg1bis

It’s working fine: GV1 increases by 10 or -10 when pressing trim tab right or left.

Then I change the minimum value of GV1:
vg2

And going back to the SF tab, I see that:
vg3

The increments values have been modified but SF1 and SF2 are still working as intended: GV1 increase by 10 or -10 when pressing trim tab right or left.

elecpower:
Fixes #5602 
I made the decision to limit inc/dec to +/-1024 (+/-102.4) to avoid possible unexpected consequences.
In theory if a gvar has range of -1024 to +1024 the max inc/dec is +/-2048.
Though relative to current value even a +/-1024 inc/dec could push the calculated value outside the gvars min/max range. This should be handled by the radio mixer.

### DIFF
--- a/radio/src/opentx.h
+++ b/radio/src/opentx.h
@@ -1078,6 +1078,16 @@ inline void getMixSrcRange(const int source, int16_t & valMin, int16_t & valMax,
 }
 #endif
 
+#if defined(GVARS)
+inline void getGVarIncDecRange(int16_t & valMin, int16_t & valMax)
+{
+  int16_t rng = abs(valMax - valMin);
+  rng = rng > GVAR_MAX ? GVAR_MAX : rng;
+  valMin = -rng;
+  valMax = rng;
+}
+#endif
+
 // Curves
 enum BaseCurves {
   CURVE_NONE,
--- a/radio/src/opentx.h
+++ b/radio/src/opentx.h
@@ -1078,7 +1078,7 @@ inline void getMixSrcRange(const int source, int16_t & valMin, int16_t & valMax,
 }
 #endif
 
-#if defined(GVARS)
+#if defined(CPUARM) && defined(GVARS)
 inline void getGVarIncDecRange(int16_t & valMin, int16_t & valMax)
 {
   int16_t rng = abs(valMax - valMin);
--- a/radio/src/opentx.h
+++ b/radio/src/opentx.h
@@ -1080,7 +1080,6 @@ inline void getMixSrcRange(const int source, int16_t & valMin, int16_t & valMax,
 inline void getGVarIncDecRange(int16_t & valMin, int16_t & valMax)
 {
   int16_t rng = abs(valMax - valMin);
-  rng = rng > GVAR_MAX ? GVAR_MAX : rng;
   valMin = -rng;
   valMax = rng;
 }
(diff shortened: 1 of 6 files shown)


elecpower:
~~Do not merge yet as the issue also exists on the radio. Sorry I should have checked as part of investigation into Companion side.
Before I perform the radio PR part I would like your views on the~~ min and max for the increment options.
1. Always +/- 1024 or +/- 102.4 (depending on precision)
2. Range of gvar limited to +/- 1024  or +/- 102.4 **(as PR)**
3. Range of gvar no limit with max limits +/- 2048 or +/- 204.8
4. Range +/- suitable value for the function eg +/- 100
5. +/- 1 as 2.1.x

elecpower:
Sorry will rebase to clean up the mess

elecpower:
Radios and Companion apply same rule for setting the editable range in special functions.
Tested on Companion Simulator for Taranis X9D+.
The radio mixer applies limits to each inc/dec to keep within the user defined gvar range.

elecpower:
Update model print special functions for inc/dec
Fix compilation error.

radio/src/opentx.h
```diff
@@ -1068,6 +1068,16 @@ inline void getMixSrcRange(const int source, int16_t & valMin, int16_t & valMax,
 }
 #endif
 
+#if defined(CPUARM) && defined(GVARS)
```
mpaperno: Minor quibble, but could we combine this with the `#if defined(CPUARM)` just above?  Also that block needs a closing comment. `#endif   // defined(CPUARM)`

Also technically you only care if `GVAR_MAX` is defined, not `GVARS` per-se.  If the function is unused it will get optimized out anyway.  I realize one probably won't exist w/out the other anyway, so it's a minor point again.

Otherwise looks good to me!

elecpower:
Code change as per @mpaperno suggestion

radio/src/opentx.h
```diff
@@ -1066,6 +1066,15 @@ inline void getMixSrcRange(const int source, int16_t & valMin, int16_t & valMax,
     valMin = -valMax;
   }
 }
+#if defined(GVAR_MAX)
+inline void getGVarIncDecRange(int16_t & valMin, int16_t & valMax)
```
bsongis, line 1080: there is something strange in this function: rng (the range between valMin and valMax) is tested against GVAR_MAX, while valMax can be set to +GVAR_MAX and valMin can be set to -GVAR_MAX
it means the range should be 2 * GVAR_MAX, no?

elecpower: For the maximum increment step I coded option 2 and what you are suggesting is option 3. If you would like I can modify.

elecpower: @bsongis removed GVAR_MAX limit test with new commit

elecpower:
@bsongis assuming this is going to be merged into 2.2.4 would you like me to submit a separate PR for 2.3?

bsongis:
Yes, I will take care of it once you have submitted a patch for 2.3

PR #5610, opened 2018-01-08
X12S : give internal GPS access to LUA

### LINKED ISSUE #5599: [Feature request] Make  Horus X12 GPS position available via getValue () function
As discusssed in the rocket forum 2 weeks ago, the Horus X12S GPS radio position is only available  when a telelemety link to the modelgps exists. 
An idea from an developer was to handle the radios gps position like a timer...
I am writing this to keep this issue tracked...

Would it be possible to make the inclination information of the radio accessible as well?
Best solution would be a new value like Joystick Jx Jy

3djc:
This introduce a function to get intenal GPS data

This can be tested using this (very crappy) widget

[main.zip](https://github.com/opentx/opentx/files/1611526/main.zip)

This fixes #5599

### DIFF
--- a/radio/src/lua/api_general.cpp
+++ b/radio/src/lua/api_general.cpp
@@ -655,20 +655,25 @@ static int luaGetRAS(lua_State * L)
 Return the internal GPS position or nil if no valid hardware found
 
 @retval table representing the current radio position
- * tx-lat : internal GPS lattiude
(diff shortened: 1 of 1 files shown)


Gewe99:
Wow.....
That was fast! GREAT.
Thanks

PR #5645, opened 2018-01-23
Model print telemetry

elecpower:
Adds telemetry to model print and compare.
Increased use of helpers and macros to simplify future changes thus the size and complexity of the diffs.
Next step is to move modelPrint internal to display conversions to data class valueToString functions and also reference these elsewhere.


bsongis:
Thanks!

PR #5658, opened 2018-01-28
Improve BMP handling

### LINKED ISSUE #5643: LUA lcd.drawPixmap() on X7 shows incorrect reulsts for images < 25px large
Have any of you tried drawPixmap() on the X7 with images narrower than 25 pixels? I can't seem to get any image of 24 pixels wide or less to draw correctly... No problem on the height (as long as it is 64 pixels or less).
Here's the code, the files I used, and a screenshot of the result. test1.bmp is 25 pixels wide. test2.bmp is the same image cropped by one pixel on the right, so it is 24 pixels wide.
```
lcd.drawPixmap(1, 1, "/test1.bmp")
lcd.drawPixmap(30, 1, "/test2.bmp")
```

[test.zip](https://github.com/opentx/opentx/files/1577519/test.zip)

result

3djc:
This needs to be thourougly tested, 

This fixes #5643

### DIFF
--- a/radio/src/gui/128x64/bmp.cpp
+++ b/radio/src/gui/128x64/bmp.cpp
@@ -20,7 +20,7 @@
 
 #include "opentx.h"
 
-uint8_t * lcdLoadDraw1BitBitmap(const char * filename, uint8_t x, uint8_t y )
+uint16_t load1BitBMPHeader(const char * filename, uint16_t &w, uint16_t &h)
 {
   FIL bmpFile;
   UINT read;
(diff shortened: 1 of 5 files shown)


kilrah:
pad**d**ing has 2 d's :P

radio/src/gui/128x64/bmp.cpp
```diff
@@ -20,54 +20,50 @@
 
 #include "opentx.h"
 
-uint8_t * lcdLoadBitmap(uint8_t * bmp, const char * filename, uint8_t width, uint8_t height)
+uint16_t load1BitBMPHeader(const char * filename, uint16_t &w, uint16_t &h)
```
bsongis: why 16 bits here?

radio/src/gui/128x64/bmp.cpp
```diff
@@ -78,43 +74,106 @@ uint8_t * lcdLoadBitmap(uint8_t * bmp, const char * filename, uint8_t width, uin
...
+
+uint8_t lcdLoadDrawBitmap(const char * filename, uint8_t x, uint8_t y )
+{
+  uint16_t w, h, hsize;
+  FIL bmpFile;
+  UINT read;
+  uint8_t bmpBuf[LCD_W]; /* maximum with LCD_W */
+  uint8_t * buf = &bmpBuf[0];
+
+  if(!(hsize =load1BitBMPHeader(filename, w, h)))
```
bsongis: it means the file is opened 2 times?

radio/src/gui/212x64/bmp.cpp
```diff
@@ -20,117 +20,202 @@
...
+uint8_t lcdLoadDrawBitmap(const char * filename, uint8_t x, uint8_t y )
+{
+  uint16_t w, h, hsize, depth;
+  FIL bmpFile;
+  UINT read;
+  uint8_t palette[16];
+  uint8_t bmpBuf[LCD_W]; /* maximum with LCD_W */
+  uint8_t * buf = &bmpBuf[0];
+
+  if(!(hsize =load1BitBMPHeader(filename, w, h, depth, &palette[0])))
```
bsongis: I would give an handle to this function instead of the filename (it means the FAT is read again to find where the files is etc.)

kilrah:
Was that bad?

3djc:
yeah, some parts where ok, but we need a much better version. Will make another attenpt later :)

PR #5661, opened 2018-01-30
Add mouseover tool tips to label fields with context menus

elecpower:


mpaperno:
Perhaps "Context menu available"? 

I actually have some commits which add toolbars, actual shortcuts, and selected-row indicators to some of those screens, but they're stale and need to be refreshed for the current state of affairs :-/   This would be helpful in the meantime though.

elecpower:
I chose "Popup menu available" for the non-programmers but not fussed.
In your toolkit could also do with a generic context menu set of functions, if practical, as the code is copied, pasted and tweaked for each data object type.

mpaperno:
Is "context menu" too geeky?  You may be right... I can't tell.

Well the actual action handlers need to be local to the individual modules (and also which actions are really available), but the whole action creation and management is definitely streamlined and centralized.

PR #5666, opened 2018-02-04
Fixed bad translation.

LapinFou:
The translation wasn't correct.
BTW, it seems this field has not been traduced in the radio firmware...
tmp


LapinFou:
I opened a ticket → #5667

PR #5684, opened 2018-02-11
Horus timer widget : display only timer name if defined

### LINKED ISSUE #5683: Problem with timer widget
This issue is still present in 2.2.1N and occurs with the Companion and the tx. If the user has set a timer name, when the timer widget is used, both the preset timer name and the user timer name appear together, and as in this instance, overlap.
image

3djc:
image

This fixes #5683


kilrah:
👍

PR #5688, opened 2018-02-12
Lua theme support

keereecat:
This is an attempt to bring near-parity of in-firmware themes to the Lua themes API.

Added were several new API functions, all documented. I've done the best I could to ensure there's nothing any _less_ safe than the prior API, however, it was quite broken.

This has been tested with a theme based on Darkblue, in the simulator. I don't yet have hardware to test on, so would be appreciated.

Theme file is available at [keereecat/opentx-horus-themes/KTheme](https://github.com/keereecat/opentx-horus-themes/tree/master/KTheme)


keereecat:
Having considered the theme-writing side of this further, would it make more sense to write a theme similar to a Lua module? For example:

```Lua
myTheme = {}
myTheme.name = "My Theme",
myTheme.options = {
  ...
}
function myTheme.drawBackground()
  ...
end

function myTheme.drawTopbar()
  ...
end
  
...

return myTheme
```

Since themes are currently broken, a more Lua-like API might be nice, though it won't match the rest of the API.

3djc:
The LUA themes do not have to match the C ones, and I personnaly would ratherhave  a more straight forward way to write LUA theme than matching the C one for the sake of it, So yes, it would make sense to define a simpler syntax for LUA theme and build the required API for that rather than starting with duplicating our current C theme

keereecat:
I'll work on prototyping a more Lua-ish theme API then.

3djc:
That would be cool, then we can include a sample LUA theme to help people making their own (I think that needs to be part of the PR)

keereecat:
Per discussion in Rocket.Chat, closing this PR in favor of an upcoming pull.

PR #5689, opened 2018-02-12
Fixed Wizard Lua scripts for Horus radios.

LapinFou:
Fixed few things in the wizards for the Horus radios.

**Plane wizard:**
- Fixed summary page when V-Tail is selected.

**Glider wizard:**
- Added missing copyright.
- Fixed summary page when V-Tail is selected.
- Align naming convention between plane & glider wizards


3djc:
👍

lshems:
am i missing something? I get only a plane wizard for the x10? Used the 2.2.1 sdcard versions.

I made a copy to make a start for a quad wizard. two choices are shown on model create, as expected, but i cannot move between the two??

Thanks for clarifying

3djc:
Seems to work ok here with rot enc

PR #5693, opened 2018-02-13
Replicate center trim beep behavior also when crossing trims

3djc:
Currently, center trim beep is purposefully disabled for throttle when t.trim is active.
This change make this behavior consistent when cross trims is used

This solve the issue reported on rocket:
> rcdiy 19:46
> Cross Trims - works now. Noticed that the trim centre does not work as expected. Chen cross trimmed the trilT does not have a trim centre (remember this has been crossed so actually is ELE trim)
> Also the TrimE, which is now THR trim has a centre when it probably should not... Tried this on the latest companion nightly..

### DIFF
--- a/radio/src/mixer.cpp
+++ b/radio/src/mixer.cpp
@@ -644,7 +644,7 @@ int getStickTrimValue(int stick, int stickValue)
     return 0;
 
   int trim = trims[stick];
-  if (stick == virtualInputsTrims[THR_STICK]) {
+  if (IS_THROTTLE_TRIM(stick)) {
     if (g_model.thrTrim) {
       int trimMin = g_model.extendedTrims ? 2*TRIM_EXTENDED_MIN : 2*TRIM_MIN;
       trim = ((g_model.throttleReversed ? (trim+trimMin) : (trim-trimMin)) * (RESX-stickValue)) >> (RESX_SHIFT+1);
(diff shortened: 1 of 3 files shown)


mcddd:
"Allows me to setup my radio so that it does what I want it to do exactly the way I want it to do it: No compromise, no workarounds, no having to think like someone else."

"If I can dream it, I can make it happen!"

So do those first two paragraphs sound familiar? They should. These are the top two reasons for using OpenTx as listed on the OpenTx University site on the Introduction page.
http://open-txu.org/home/undergraduate-courses/introduction/why-upgrade/

The center trim beep provides a quick audible indicator of the center trim position, regardless of which trim is being adjusted. This is especially true while the aircraft is inflight and it's inconvenient to look at the trim displays since glow/gas engines often require further throttle trim adjustments to achieve a slow enough idle speed for landing.

Since the center throttle trim beep can be turned on or off from the SETUP screen without the Trim Idle Only activated, why wouldn't the same selection box be able to turn the center beep on or off with the Trim Idle Only activated? Those that don't want the center beep can simply uncheck the box, and those (like me) that do want the center beep can check the box. Eliminating the choice goes against the top two reasons for using OpenTx that were listed at the beginning of this reply.

3djc:
Your comment as nothing to do with this PR, which ensures consistent behaviour when cross triming

RCdiy:
Thank you :)

PR #5698, opened 2018-02-17
Issue #5696

### LINKED ISSUE #5696: GV limits wrong when trim is reused
When I use a special function to adjust a GV with a trim button, then limits for the trim are applied, where limits for the GV should  be applied instead.
I have attached an OTX to test the issue. Use A up or down to adjust GVs with TrmR and TrmA, I have removed the other files, as I created a pull request.

Jesper

[Test.otx.zip](https://github.com/opentx/opentx/files/1733618/Test.otx.zip)

jfrickmann:
When an "Adjust GV" special function is reusing trim buttons to increase/decrease GVs, then the trim limits are applied instead of GV limits. This patch applies the GV limits.

This fixes #5696

### DIFF
--- a/radio/src/opentx.cpp
+++ b/radio/src/opentx.cpp
@@ -1369,35 +1369,53 @@ uint8_t checkTrim(event_t event)
       AUDIO_TRIM_MIDDLE();
       pauseEvents(event);
     }
-    else if (before>TRIM_MIN && after<=TRIM_MIN) {
-      beepTrim = true;
-      AUDIO_TRIM_MIN();
-      killEvents(event);
-    }
-    else if (before<TRIM_MAX && after>=TRIM_MAX) {
-      beepTrim = true;
-      AUDIO_TRIM_MAX();
-      killEvents(event);
-    }
-
-    if ((before<after && after>TRIM_MAX) || (before>after && after<TRIM_MIN)) {
-      if (!g_model.extendedTrims || TRIM_REUSED(idx)) after = before;
-    }
-
-    if (after < TRIM_EXTENDED_MIN) {
-      after = TRIM_EXTENDED_MIN;
-    }
-    if (after > TRIM_EXTENDED_MAX) {
-      after = TRIM_EXTENDED_MAX;
-    }
 
 #if defined(GVARS)
     if (TRIM_REUSED(idx)) {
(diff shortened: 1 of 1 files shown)


bsongis:
Seems good but I need to take a coffee and take time on it before pressing the button

jfrickmann:
Good to hear! I was wondering if we should also make the increment/decrement equal to 1 for GV, independently of trimInc? E.g. right after line 1362:
    if (thro) v = 4; // if throttle trim and trim trottle then step=4
insert:
#if defined(GVARS)
    if (TRIM_REUSED(idx)) v = 1;
#endif

jfrickmann:
So, how was the coffee? :-)

radio/src/opentx.cpp
```diff
@@ -1369,35 +1369,58 @@ uint8_t checkTrim(event_t event)
...
-    if (after > TRIM_EXTENDED_MAX) {
-      after = TRIM_EXTENDED_MAX;
-    }
 
 #if defined(GVARS)
     if (TRIM_REUSED(idx)) {
-      SET_GVAR_VALUE(trimGvar[idx], phase, after);
+      int8_t gvar = trimGvar[idx];
+#if defined(CPUARM)
+// TODO: This was necessary to make it compile for AVR9X target - maybe there is a better solution?
```
bsongis: There is no MIN and MAX on GVARS on AVR9X
The comment may be removed

bsongis (approved):
Good job, I think everything is right

jfrickmann (commented):
I have removed the comment as suggested, and added the lines that I suggested above in 1363-65.

radio/src/opentx.cpp
```diff
@@ -1360,6 +1360,9 @@ uint8_t checkTrim(event_t event)
     int8_t trimInc = g_model.trimInc + 1;
     int8_t v = (trimInc==-1) ? min(32, abs(before)/4+1) : (1 << trimInc); // TODO flash saving if (trimInc < 0)
     if (thro) v = 4; // if throttle trim and trim trottle then step=4
+#if defined(GVARS)
+    if (TRIM_REUSED(idx)) v = 1;
```
bsongis, line 1364: yes, exact

bsongis:
Merged, thanks for your help!

PR #5723, opened 2018-02-24
Fixed: compilation if SIMU_DISKIO = ON

mr-pokryshkin:
SIMU_DISKIO is really helpful option to debug fatFS and diskio related tasks. I use it to add support of cyrillic chars. 
I've tested SIMU_DISKIO with QX7 simulator, works fine.


bsongis:
Perfect!

PR #5724, opened 2018-02-24
added case for UNIT_TEXT used by crossfire FM Sensor

### LINKED ISSUE #5582: FM Sensor returns 0
I've got some Lua scripts running on 2.2.1 (cca71808) on a X9D.
I have inav running on an omnibus F4 Pro, and CRSF setup both ends.

My FM sensor returns 0 even though it has a value which can be seen on the Telemetry page. 

I've tried calling it by name and by ID
```
modes = getValue("FM")
lcd.drawText(textualRow, y, modes, SMLSIZE)
```
Even calling it with the following gets me 0
```
modes = getValue("258")
lcd.drawText(textualRow, y, modes, SMLSIZE)
```
I got the ID 258 using the following code, which also confirms the right name:
```
       y = y + 8
	local fieldinfo = getFieldInfo('FM')
	lcd.drawText(textualRow, y, fieldinfo['id'], SMLSIZE)
	y = y + 8
	lcd.drawText(textualRow, y, fieldinfo['name'], SMLSIZE)
```

Am I missing something?

tgreer:
Fixes #5582 
This fixes an issue with Lua scripts being unable to read the FM sensor generated by crossfire which is set to text. Thanks to @schwabe for the fixes, I'm just doing some admin. 

Tested with Taranis X9D, Crossfire Micro TX, Nano RX and inav supplying the telemetry.

### DIFF
--- a/radio/src/lua/api_general.cpp
+++ b/radio/src/lua/api_general.cpp
@@ -232,6 +232,9 @@ void luaGetValueAndPush(lua_State* L, int src)
         case UNIT_DATETIME:
           luaPushTelemetryDateTime(L, telemetrySensor, telemetryItems[qr.quot]);
           break;
+        case UNIT_TEXT:
+          lua_pushstring(L, telemetryItems[qr.quot].text);
+          break;
         case UNIT_CELLS:
           if (qr.rem == 0) {
             luaPushCells(L, telemetrySensor, telemetryItems[qr.quot]);


schwabe:
Since this a small and easy bug fix and already tested I am taking the liberty of merging this.

3djc:
@tgreer we do not have any equipement to test your config, could you check that this text sensor is displayed correctly elsewhere and the LS do work fine ?

PR #5737, opened 2018-03-01
Allow HorusBT chip install on  X9E and X7 (non S)

3djc:
This allows users to had a Horus chip to X9E and X7. Added chip should be activated even with their default 9600 bps

Somewhat tested, but need further tests.

THIS COMPLETELY WIPES previous unused X9E BT handling

### DIFF
--- a/radio/src/bluetooth.h
+++ b/radio/src/bluetooth.h
@@ -19,11 +19,6 @@
  */
 
 enum BluetoothStates {
-#if defined(PCBX9E) && !defined(USEHORUSBT)
-    BLUETOOTH_INIT,
-    BLUETOOTH_WAIT_TTM,
-    BLUETOOTH_WAIT_BAUDRATE_CHANGE,
-#endif
     BLUETOOTH_STATE_OFF,
     BLUETOOTH_STATE_FACTORY_BAUDRATE_INIT,
     BLUETOOTH_STATE_BAUDRATE_SENT,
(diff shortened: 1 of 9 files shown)


hrenz:
Hy,
BT für X9E 
very interesting, where can I find the hardware and shematic and firmware for the BT-Modul

kilrah:
Buy a spare X12S module and replace the existing one.

radio/src/bluetooth.h
```diff
@@ -19,11 +19,6 @@
  */
 
 enum BluetoothStates {
-#if defined(PCBX9E) && !defined(USEHORUSBT)
```
bsongis, line 22: I would prefer we change macros to BLUETOOTH and BLUETOOTH_XXXX here, either BLUETOOTH_HORUS or BLUETOOTH_X9E (or any else if needed)

bsongis:
We completely remove native BT support on X9E, is it a good idea?

radio/src/bluetooth.h
bsongis: obviously my remark is valid only if we have many BLUETOOTH protocols supported, if we keep only one, you may forget it

3djc:
We debated it with Kilrah, and we think there is no pratical use for it (and I think we have broken x9e BT already in 2.2.1, but nobody did notice).

If that worries you too much, I can do a compile option, but we are trying to move away from that model

kilrah:
It never did anything!

bsongis:
No compatible Android / iPhone app that we don't know?

kilrah:
With the new one yes, but not the old!

raphaelcoeffic:
@3djc do I understand it correctly that I can now solder a HorusBT module to a plain normal X7? I have one out of the first series.

3djc:
@raphaelcoeffic , on X7, you should be  able to with plain 2.2.1 already (this PR is more for X9E now)

3djc:
Given the  feedbacks on user experience of BT trainer, this is not worth it

PR #5754, opened 2018-03-09
Added: fonts with Cyrillic (Russian) chars

mr-pokryshkin:
Changed: FatFs codepage (if RU lang)
Added: dummy translation to Russian (copy of en.h.txt), so users can use Cyrillic chars (model name), can play wav files from cyrillic folders, let's say name.wav from modelname folder. Menus and texts are in English!


mr-pokryshkin:
In /radio/utils/ I see script font2png.py
I have added my own fonts into "sqt5" folder. I think I have to copy fonts to "std" folder, right? or should font2png.py generate PNGs automatically?

kilrah:
font2png.py is used for color screen radios to generate png's from a system font instead of manually doing the files like on std screen.

3djc:
I do not think it would be wise to allow the use of Cyrillic chars in 'data' elements, like model name, wav filenames, ..., as it would prevent sharing with others, which is part of the whole concept.

Actrually, for existing languages, 'language specific' characters are not allowed for those either, you cannot put french  'é' or german 'ß'

I'm all for Russian sound and UI translation, but not for data
ps : I have not looked at the code yet, just responding to the stated intent

kilrah:
+1. 

Actually the pull request in this state does not build for standard screen radios with RU selected, only X10/X12.

kilrah:
Just checked and it appears we do accept language specific chars but only on Horus.

But you can't add that many, it causes an overflow.

screenshot_x10_18-03-10_12-49-55

mr-pokryshkin:
Re: Actrually, for existing languages, 'language specific' characters are not allowed for those either, you cannot put french 'é' or german 'ß'

I think 1-2 chars is not a big deal. Unfortunatelly, there are 32 Cyrillic chars, and only 11 of 32 are looks like Latin chars, so we can not use 21 chars from our alphabet. :(

mr-pokryshkin:
2kilrah: Ok, I'll check X10/X12 settings and number of chars, 

To reduce number of chars I'll use uppercase only.

mr-pokryshkin:
Re: I do not think it would be wise to allow the use of Cyrillic chars in 'data' elements, like model name, wav filenames, ..., as it would prevent sharing with others, which is part of the whole concept.

It should not prevent, just may look weird. 3rd party library (FAT FS) allows to switch codepage, so there is no problem with filenames.

kilrah:
Actually did you even try to build and use this? There is literally nothing working...

This can only be used on Horus radios, but you did not touch or update the files Horus radios use. You only created standard screen fonts, but standard screen radios don't build since you missed adding the new files to the generation process. 
Number of chars should work on Horus.

> It should not prevent, just may look weird. 

No it is dangerous. On standard screen radios only one language's special chars are included. If someone selects the 30th char form Russian, then that gets loaded into French firmware which does not have that many then you're going to access something that doesn't exist. That's why standard screen radios do not allow language-specific chars in model names.

mr-pokryshkin:
> Actually did you even try to build and use this? There is literally nothing working...
As I said, I added files into "sqt5" folder only. It works fine with "sqt5" on X7. 
I did not test using X10/12. This is one of my next tasks.

> If someone selects the 30th char form Russian, then that gets loaded into French firmware which does not have that many then you're going to access something that doesn't exist. 

I'll change code to display "square" instead of random values from memory, if char does not exist. Like Chrome does if font unavailable.
Let's discuss in chat.

radio/src/thirdparty/FatFs/ffconf.h
```diff
@@ -89,8 +89,11 @@ extern "C" {
 /*---------------------------------------------------------------------------/
 / Locale and Namespace Configurations
 /---------------------------------------------------------------------------*/
-
+#if defined(TRANSLATIONS_RU)
+#define _CODE_PAGE  866
```
bsongis, line 93: visual alignment issue on this line

PR #5758, opened 2018-03-12
Disable pulse while model notes are displayed

### LINKED ISSUE #5747: Confirm antenna is connected before powering external module.
Pretty much any uhf system will be destroyed if the handset powers on without an antenna connected. So if there was a warning that makes you confirm the antenna is connected before it powers on the external module. Would stop a lot of modules being damaged.

Ideally it should be a configurable option since 2.4 modules generally have short antennas and can be left in place.

Graupner radios do something similar where they won't transmit untill you hit ok on the screen.

3djc:
This keep pulses off while model notes are displayed. This allow users to writes things like "Don't forget to connect external antenna', or 'this model needs module xxx' before those module are turned on.

This works at  boot and also when changing model.

Unlike other warning, getting out of this one requires exit/rtn

This closes #5747

### DIFF
--- a/radio/src/gui/480x272/menus.cpp
+++ b/radio/src/gui/480x272/menus.cpp
@@ -66,8 +66,8 @@ void pushMenu(MenuHandlerFunc newMenu)
   TRACE("pushMenu(%d, %p)", menuLevel, newMenu);
 }
 
-void readModelNotes() {
-
+void readModelNotes()
+{
   LED_ERROR_BEGIN();
 
   strcpy(s_text_file, MODELS_PATH "/");
--- a/radio/src/gui/480x272/menus.cpp
+++ b/radio/src/gui/480x272/menus.cpp
@@ -80,7 +80,7 @@ void readModelNotes() {
 
   clearKeyEvents();
   event_t event = EVT_ENTRY;
-  while (event != EVT_KEY_BREAK(KEY_EXIT)) {
+  while (event != EVT_KEY_FIRST(KEY_EXIT)) {
     lcdRefreshWait();
     lcdClear();
     menuTextView(event);
--- a/radio/src/gui/480x272/menus.cpp
+++ b/radio/src/gui/480x272/menus.cpp
@@ -66,6 +66,32 @@ void pushMenu(MenuHandlerFunc newMenu)
   TRACE("pushMenu(%d, %p)", menuLevel, newMenu);
 }
 
+void readModelNotes() {
+
+  LED_ERROR_BEGIN();
+
+  strcpy(s_text_file, MODELS_PATH "/");
+  char *buf = strcat_currentmodelname(&s_text_file[sizeof(MODELS_PATH)]);
+  strcpy(buf, TEXT_EXT);
+  if (!isFileAvailable(s_text_file)) {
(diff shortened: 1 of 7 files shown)


LapinFou:
Just to be sure I understood correctly, this modification will impact only **external** module.
This modification will not impact the **internal** module?
The reason I'm asking the question is for the case you unintentionally switch off you radio. In this case you need to switch it back on very quickly in order to take bake your model's controls.
😄

3djc:
No, it stops both module, but let’s review your scenario:

- you are flying and turn your radio off
- rssi warning stops the radio from shutting down
- you somehow acknowledge the turn off prevention and your radio really turns off
- you turn it back on, you will hit throttle warning since you where flying
- you will then hit switch warning (because at the very least cut will be in the wrong position)
Your pulse only start now in current setup
Now, if, and only if there are model notes, you need an extra button push to end the model notes display.

I hardly believe that one button extra push will make any material difference

LapinFou:
All your points are valid. I forgot that with OTX2.2.1 there is now a warning if the RX is sending a RSSI feedback.

radio/src/gui/480x272/menus.cpp
```diff
@@ -66,6 +66,32 @@ void pushMenu(MenuHandlerFunc newMenu)
   TRACE("pushMenu(%d, %p)", menuLevel, newMenu);
 }
 
+void readModelNotes() {
```
bsongis: code style issue here

radio/src/gui/480x272/menus.cpp
```diff
@@ -66,6 +66,32 @@ void pushMenu(MenuHandlerFunc newMenu)
...
+
+  clearKeyEvents();
+  event_t event = EVT_ENTRY;
+  while (event != EVT_KEY_FIRST(KEY_EXIT)) {
+    lcdRefreshWait();
+    lcdClear();
+    menuTextView(event);
+    event = getEvent();
+    lcdRefresh();
+    wdt_reset();
```
bsongis, line 89: did someone test it? there is no delay between 2 LCD refresh, I hope it's OK

radio/src/gui/common/stdlcd/view_text.cpp
```diff
@@ -102,6 +102,28 @@ void readTextFile(int & lines_count)
 #define EVT_KEY_PREVIOUS_LINE          EVT_KEY_FIRST(KEY_UP)
 #endif
 
+void readModelNotes() {
```
bsongis: code style as well

radio/src/storage/storage_common.cpp
```diff
@@ -100,7 +89,15 @@ void postModelLoad(bool alarms)
   LOAD_MODEL_CURVES();
 
   resumeMixerCalculations();
-  // TODO pulses should be started after mixer calculations ...
+  if (pulsesStarted()) {
```
bsongis, line 92: one less TODO, thanks!

bsongis (approved):

radio/src/gui/480x272/menus.cpp
3djc: I borrowed a x12 to test it, it seems to work fine, maybe a x10 test too would be nice.

image

bsongis: ok I will do it

3djc:
Checked ok on xlite too

bsongis:
Checked on X10

PR #5773, opened 2018-03-18
Different navigation through menus with the Rotary encoder on X10

lindongy:
1, fix BL_TIMER error.
2, Improve UI interactive throght rotary.


bsongis:
Please propose for 2.3 if needed, this is a too big change for 2.2

PR #5776, opened 2018-03-20
USB joystick support for 9xrpro

### LINKED ISSUE #3168: USB Joystick with 9XR Pro
[It seems like](https://github.com/opentx/opentx/wiki/OpenTX-USB-Joystick) this feature is supported on Taranis but not on 9XR Pro.

AntonEvmenenko:
Fixes #3168

This code is written based on the current joystick implementation for the Taranis and the example for HID mouse from [here.](https://github.com/lfwanner/var/tree/e9802710380a480fa91a6660e754f3a087be5484/sam3/at91lib/usb/device/hid-mouse)

Tested on Turnigy 9XR PRO. As I understand, it will work for all transmitters with SKY9X mainboard.

Also, masstorage for SKY9X removed, because it is no longer used (as i understand, it is used now only by the bootloader, which is downloaded from an [external repository](https://github.com/MikeBland/OpenRcBootloader)).


AntonEvmenenko:
Can someone say what is wrong, please? There is no need for this functionality? Or maybe I did something wrong (this is my first pull request in opentx).

3djc:
We are sorry, it is just time missing, It will be reviewed when time gets available

AntonEvmenenko:
Understood, thanks!

radio/src/targets/sky9x/joystick.cpp
```diff
@@ -0,0 +1,92 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ */
+
+#include "opentx.h"
+
+#ifndef SIMU
```
bsongis: if the whole file is not needed by simu it's better to put it inside FIRMWARE_TARGET_SRC in CMakeLists.txt and avoid this #ifdef

radio/src/targets/sky9x/joystick.cpp
```diff
@@ -0,0 +1,92 @@
...
+#include "usb/device/hid-joystick/HIDDJoystickDriver.h"
+extern void HIDDJoystickDriver_Initialize();
+unsigned char HIDDJoystickDriver_ChangeJoystickState(const HIDDJoystickInputReport *report);
+extern void USBD_Connect(void);
+}
+
+static void ConfigureUsbClock(void)
+{
+    /* Enable PLLB for USB */
+    PMC->CKGR_PLLBR = CKGR_PLLBR_DIVB(1)
```
bsongis: would you have a look at our indentation rules? we use 2 chars instead of 4 in OpenTX

bsongis (commented):
Do we *really* want to remove MASSSTORAGE from the firmware itself? We have it on all other radios, I don't see any major issue, but this needs to be discussed before we go this way.

On Taranis you have a dynamic choice of the driver used, perhaps it would be possible to have the same on SKY9X?

Unless we introduce a compilation option as we had in the past on Taranis: USB=MASSSTORAGE or USB=JOYSTICKS

radio/src/targets/sky9x/joystick.cpp
AntonEvmenenko: Thanks for the advice, fixed in 9a7e7f1

radio/src/targets/sky9x/joystick.cpp
AntonEvmenenko: Please explain, do I need to use these rules only for code written by me, or do standard libraries also need to be reformed in accordance with these rules (for example, radio/src/targets/sky9x/usb/common/hid/HIDIdleRequest.c in this pull request)?

AntonEvmenenko:
Yes, I would also like to discuss the masstorage. 

Despite the fact that the driver is formally present in the code, it does not work. This can be confirmed by the following methods:
- when you connect the USB cable to the 9xrpro transmitter, nothing happens
- void usbMassStorage() from opentx/radio/src/targets/sky9x/massstorage.cpp is not used anywhere  

Therefore, since this code is not used, I thought that it was not needed. 

So, before we add a dynamic selection between the masstorage and the joystick, we first need to add the masstorage :)

radio/src/targets/sky9x/usb/device/hid-joystick/HIDDJoystickDriver.c
```diff
@@ -0,0 +1,358 @@
...
+//         ConfigurationChanged() callback re-implementation
+//------------------------------------------------------------------------------
+
+//------------------------------------------------------------------------------
+/// Callback function when configureation changed
+/// \param cfgnum New configuration number
+//------------------------------------------------------------------------------
+void USBDDriverCallbacks_ConfigurationChanged(unsigned char cfgnum)
+{
+    if (cfgnum > 0) {
```
jhugard: Suspicious: empty if block and the test is meaningless because `cfgnum` is unsigned, so can never be negative.

Is this correct code? Should this have a comment on why it's ignored?  Does it need a trace?

radio/src/targets/sky9x/joystick.cpp
```diff
@@ -0,0 +1,88 @@
...
+            if ( channelOutputs[i+24] > 0 ) {
+                HID_Buffer[2] |= (1 << i);
+            }
+        }
+
+        //analog values
+        //uint8_t * p = HID_Buffer + 1;
+        for (int i = 0; i < 8; ++i) {
+            int16_t value = channelOutputs[i] / 8;
+            if ( value > 127 ) value = 127;
```
jhugard: channelOutputs is an int16_t, so with the above divide-by-8, there's no need to cap the value (can never be >abs(127)).  Can remove and save a few bytes.  (opinions on this will likely differ)

jhugard:
Thanks for taking this on!  I took a look at doing it a few years ago, but couldn't get the build chain working properly.

radio/src/targets/sky9x/usb/device/hid-joystick/HIDDJoystickDriver.c
AntonEvmenenko: You're right, this is a meaningless construction. It's just a part of the [standard example written by Atmel.](https://github.com/lfwanner/var/blob/e9802710380a480fa91a6660e754f3a087be5484/sam3/at91lib/usb/device/hid-mouse/HIDDMouseDriver.c) FYI, the same construction (the only difference is non-empty if block) [is already in the OpenTX](https://github.com/opentx/opentx/blob/2.2/radio/src/targets/sky9x/usb/device/massstorage/MSDDriver.c) :)

Fixed in 2efabc3

radio/src/targets/sky9x/joystick.cpp
AntonEvmenenko: If I understood correctly, you were a little mistaken:

max(int16_t) / 8 =  32767 / 8 = 4095, which is > 127

kilrah: It's a /8, not a >>8 (/256).

jhugard: Never mind.... yes, my bad.  Even wrote the comment with a >>8 then edited it to divide-by-eight.  Had the shift stuck in my head.  ty.

radio/src/targets/sky9x/joystick.cpp
jhugard: Looks like the other ATMEL files were all left with original formatting, which makes sense in case there is a patch or other update (would simplify reviewing the diff).

@bsongis The existing ATMEL files are all 4-space indent.  Should you update the style guide to mention integrated vendor files? https://github.com/opentx/opentx/wiki/OpenTX-Code-Style-Guide

bsongis:
usbMassStorage() worked before, it means it has been removed in one of our latest refactorings (between 2.1 and 2.2 for sure). Would you have a quick search at it?

radio/src/targets/sky9x/joystick.cpp
bsongis: I didn't say we do everything perfectly ;) I think this file (and massstorage.cpp file as well) should use our coding rules, as it is not exactly an ATMEL file. Otherwise yes, we try to change as little as possible fully imported code from other projects.

Wildnecromant:
Hello! Please, can you send me work bin of this firmware? I can't compile this, but very need it for my 9xr pro. My email is serious_dron@mail.ru

AntonEvmenenko:
Hello everyone again, i apologize for the long absence.

> usbMassStorage() worked before, it means it has been removed in one of our latest refactorings (between 2.1 and 2.2 for sure). Would you have a quick search at it?

I was able to discover some details related to this. @bsongis, you disabled masstorage for sky9x in cc38920. Also you left a todo "remove completely massstorage from sky9x firmware", which, by the way, is still present in the code. @bsongis, comment on this, please. Do we really need masstorage for sky9x now?

nicoschtein:
Any updates on this? @AntonEvmenenko @bsongis @jhugard

AntonEvmenenko:
Well, I'm still waiting for @bsongis answer to my last comment.

AntonEvmenenko:
Btw, if anyone needs this firmware and he has troubles with compiling it, here is the [bin](https://drive.google.com/open?id=1fI73SpqfDj1wqVIzuB10iFyYTr9Cm7PZ).

bsongis:
@AntonEvmenenko I am really annoyed ... I see that I removed the masssotage on SKY9X, and I also see the comment I wrote at that time in the same commit, but I can't remember why I did that. Today I think it would be nice that all radios behave the same and that the change I did at this time was bad.

AntonEvmenenko:
@bsongis, thanks for your answer.

Generally, i implemented "dynamic choice of the driver used". You can look at my experiments here: f7ed0cc and 41f6318. Unfortunately, there is a big problem: after using radio for some time in masstorage mode, if you unplug usb cable, the memory card doesn't mount back and the firmware doesn't start working normally. Unfortunately, I have not been able to deal with this problem.

So, i think, there is 2 ways:

1. Someone from the community will help fix this problem.  
2. I will revert deleting of the masstorage. After that i have to comment out USBDCallbacks_RequestReceived() and USBDDriverCallbacks_ConfigurationChanged() in [opentx/radio/src/targets/sky9x/usb/device/massstorage/MSDDriver.c](https://github.com/opentx/opentx/blob/2.2/radio/src/targets/sky9x/usb/device/massstorage/MSDDriver.c) because these callbacks should be implemented for joystick. Optionally, i can also create file opentx/radio/src/targets/sky9x/usb_driver.cpp with following content:

```
...

// TODO: implement choice between masstorage and joystick

extern "C" void USBDDriverCallbacks_ConfigurationChanged(unsigned char cfgnum)
{
//if (selectedUsbMode == USB_JOYSTICK_MODE) {
//} 
//else if (selectedUsbMode == USB_MASS_STORAGE_MODE){
//  MSDDriver_Reset();
//}
}

extern "C" void USBDCallbacks_RequestReceived(const USBGenericRequest *request)
{
//if (selectedUsbMode == USB_JOYSTICK_MODE) {
    HIDDJoystickDriver_RequestHandler(request);
//} 
//else if (selectedUsbMode == USB_MASS_STORAGE_MODE){
//  MSDDriver_RequestHandler(request);
//}
}

...

```

In this case, code related to masstorage will not be deleted. If someone will need this functionality (which, by the way, still has not happened, void usbMassStorage() from [opentx/radio/src/targets/sky9x/massstorage.cpp](https://github.com/opentx/opentx/blob/2.2/radio/src/targets/sky9x/massstorage.cpp) still is not used anywhere:smile:), he can add it.

AlexDaniel:
Any news? It's an amazing piece of work that is really needed, and it's kinda sad that it is just gathering dust…

raphaelcoeffic:
@AlexDaniel most probably! It seems however that it's not 100% finished, so that it has never been merged. Maybe @AntonEvmenenko bought himself another TX, or whatever happened. Looks like this feature is in search of a new maintainer... I don't have the TX, so not me ;-)

buzztiaan:
the only thing not '100%' seems to be that the tx requires a reboot after massstorage mode, doesnt seem like a big issue?

radio/src/targets/sky9x/joystick.cpp
AntonEvmenenko: Fixed in 777fff3

AntonEvmenenko:
Hello guys!

Cool to see that there is still interest in this PR!

Ok, I made changes that hopefully allow to merge. These are the changes that I wrote about above, namely:

1. I reverted deleting of the masstorage.
2. I created "dummy" usb_driver.cpp, which can be further developed.

So, at the moment, this PR **does not remove anything from the code, but only adds a usb joystick**. Masstorage remains in code, it is still not active (as now). Default mode of operation is usb joystick.

In the future, if someone will want to have a masstorage, he will have to activate it, and also add the ability to switch between masstorage and joystick. This is a **separate** big task that takes time and effort.

@bsongis, please comment on.

bsongis:
Why 2.2? We are not planning a new release of 2.2
We are missing testers for 2.3 on Sky9X / 9XR-PRO

kilrah:
Becasue it was set for 2.2 more than a year ago... but yeah it needs to be targeting 2.3 now

bsongis:
It seems there are not too many conflicts for such a rebase!

AntonEvmenenko:
Ok, I just tried to apply these changes to branch 2.3. I had to resolve conflicts in two files and also remove line 43 from radio/src/targets/sky9x/usb/device/hid-joystick/HIDDJoystickDriverDescriptors.c. After that I got successful build.

But.. The firmware just doesn't start. It just shows startup logo and nothing else happens. Current 2.3 branch without my modifications behaves the same (only for my 9xrpro, of course).

The only thing that confuses me is that I use the old version of the contents of the SD card (because [there is no actual version](https://downloads.open-tx.org/2.3/nightlies/sdcard/) for my tx). I don’t know if this can somehow affect.

bsongis:
OK then we broke the SKY9X board with all our changes :(

bsongis:
Please rebase against latest 2.3, 9XR-PRO seems working correctly now

AntonEvmenenko:
Ok, i just merged last changes and checked again. After turning on the inscription "Updating EEPROM" appeared (or something like this, did not remember the text for sure) for about 10-20 seconds, after that the tx started successfully!

image

Joystick works normally. Unfortunately, i can't test something else, because i have only tx (no transmitting or receiving modules).

@bsongis, please tell me, should i do something else for this PR? Should i somehow participate in merge conflicts resolving, or will you do everything yourself?

bsongis:
OK good, thanks for the test, would you push the result of the merge?

AntonEvmenenko:
Hmm.. Not sure. It's because of changed base branch.

I did this to change base branch locally:

```
git remote add upstream git@github.com:opentx/opentx.git
git fetch upstream
git checkout 9xrpro_joystick
git rebase --onto 2.3 2.2
```

To pull up the latest changes from 2.3:

```
git fetch upstream
git checkout 9xrpro_joystick
git merge upstream/2.3
```

Now i have this locally:

image

This was enough for tests, but could i push the result of the merge now? I doubt :(

If you can tell me how to do it right, I will be grateful.

I could just send you the content of the modified files (radio/src/targets/sky9x/CMakeLists.txt and radio/src/tasks_arm.cpp), maybe this is the easiest way?

AntonEvmenenko:
Done.

AlexDaniel:
Nevermind this comment, it's some issue on my side, and I was clearly looking at the wrong device. Sorry for the noise.

AlexDaniel:
OK, sorry for the noise again, it works perfectly! I'm so happy, thank you very much!

PR #5782, opened 2018-03-23
Improve RSSI shutdown display message on Horus

### LINKED ISSUE #5774: Problems with warning message on shutdown with RSSI active.
On the Horus, the message "Confirm Shutdown" does not format correctly. Part of the "w" and the "n" are missing off the right of the screen. 

Also, I suspect for many users, they will not realise why this message appears. Could you extend the message to read "Receiver still powered up, confirm shutdown" to make it perfectly clear. As this is displayed on a blank screen there should be enough room for all transmitters.

3djc:
Improve message displayed on Horus when powered off while receiver is still on.

image

This fixes #5774


kilrah:
And the translations? :P

radio/src/opentx.cpp
```diff
@@ -2827,7 +2827,11 @@ uint32_t pwrCheck()
 #endif
           lcdRefreshWait();
           lcdClear();
-          POPUP_CONFIRMATION("Confirm Shutdown");
+#if defined(PCBHORUS)
+          (warningText = "Shutdown ?", warningType = WARNING_TYPE_CONFIRM, warningInfoText = STR_MODEL_STILL_POWERED, popupFunc = runPopupWarning);
```
bsongis: rather use the `SET_WARNING_INFO` macro?

bsongis:
👍

PR #5789, opened 2018-03-29
Allow raw telem logs for ARM radio with SD.

3djc:
This change allows the raw recording of telemetry data without the need for a specially compiled binary. Recording is started from debug screen and only wroks while in that screen. Size of the recorded file is displayed dynamicaly.

image

image

image


bsongis:
It means that the mixer task will write on the SD card, this is something we wanted to avoid (write operation is blocking)

3djc:
Risk/benefits ratio not worth the risk !

Issue #3398, opened 2016-04-01
Wizard "throttle cut" shouldn't use Override SF


kilrah:
Otherwise people with the "nooverridech" option can't see nor remove it if they selected it during model creation.

fraternl:
What am I missing...???
Why is there an option "nooverridech" ???

jtaylor2:
>@fraternl
>What am I missing...???
>Why is there an option "nooverridech" ???

Because it is not needed and if used incorrectly can damage linkage and servos.
See issues 1428 and 609.
https://github.com/opentx/opentx/issues/1428
 https://github.com/opentx/opentx/issues/609

Jim

fraternl:
I read those tickets and it didn't convince me...
Someone already predicted a compatibility issue and here we have one now...
I don't have and will probably never have a helicopter. I checked the "noheli" option because it's there...
If they would take that option out I wouldn't care....

Half of these compiler options could be scratched imho.
Only if they affect memory usage they should be in there....
FAIMODE and FAICHOICE are good exceptions.

It also complicates programming and could be a source for bugs like this one...

lshems:
I think compiler options are nice for maintaining backwards compatibility. In case of the channel override, the only real issue is that override channel seems to force a ppm value, passing everything else. Then it should probably be defined as ppm value as well in the setup. That would make it very clear in its usage.

Perhaps take that into account with the general discussion about units: gvar (raw), %, ppm,....

fraternl:
These options are not there for compatibility... 
The nogvar, for instance, is created by request of users that don't want that extra page.
imho this could better be done at runtime....

It's only because they have this compile on demand that it's possible to do this way.
IMHO it's not appropriate to use it for this....
Only when it saves resources for the system...

timer3, noheli, nogvars, mixersmon, nooverridech   should (IMHO) not be a complier option. 
It leads to misunderstandings when someone is using foreign setups or advice and can also lead to bugs.
I am sure the developers are aware of this.
I wonder if the people asking for it are.

kilrah:
> These options are not there for compatibility... 

In a way yes, because when the nooverridech option was discussed it was otherwise going to be removed altogether and would have broken nearly ever model out there. I personally never understood that one, even less than the others. 

[...]

PR #5799, opened 2018-04-03
Make cut use mixer instead of override

3djc:
This fixes (long standing) #3398

### DIFF
--- a/companion/src/wizarddata.cpp
+++ b/companion/src/wizarddata.cpp
@@ -130,7 +130,6 @@ WizMix::operator ModelData()
     }
   }
 
-
   // Add the Flight Timer option
   if (options[FLIGHT_TIMER_OPTION] ){
     model.timers[timerIndex].mode.type = SWITCH_TYPE_TIMER_MODE;
--- a/companion/src/wizarddata.cpp
+++ b/companion/src/wizarddata.cpp
@@ -116,7 +116,7 @@ WizMix::operator ModelData()
       throttleChannel = i;
 
       // Add the Throttle Cut option
-      if( options[THROTTLE_CUT_OPTION] && throttleChannel >=0 ){
+      if( options[THROTTLE_CUT_OPTION]){
         MixData & mix = model.mixData[mixIndex++];
         mix.destCh = throttleChannel+1;
         mix.srcRaw = SOURCE_TYPE_MAX;
--- a/companion/src/wizarddata.cpp
+++ b/companion/src/wizarddata.cpp
@@ -112,21 +112,17 @@ WizMix::operator ModelData()
     addMix(model, ch.input1, ch.weight1, i, mixIndex);
     addMix(model, ch.input2, ch.weight2, i, mixIndex);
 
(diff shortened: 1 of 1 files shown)


bsongis:
Seems all good to me

companion/src/wizarddata.cpp
```diff
@@ -109,20 +108,26 @@ WizMix::operator ModelData()
...
-    model.customFn[switchIndex].swtch.type = SWITCH_TYPE_SWITCH;
-    model.customFn[switchIndex].swtch.index = IS_ARM(getCurrentBoard()) ? SWITCH_SF0 : SWITCH_THR;
-    model.customFn[switchIndex].enabled = 1;
-    model.customFn[switchIndex].func = (AssignFunc)throttleChannel;
-    model.customFn[switchIndex].param = -100;
+    if (ch.input1 == THROTTLE_INPUT || ch.input2 == THROTTLE_INPUT) {
+      throttleChannel = i;
+
+      // Add the Throttle Cut option
+      if( options[THROTTLE_CUT_OPTION] && throttleChannel >=0 ){
```
bsongis: throttleChannel is always >= 0, no?

3djc: It is IF you said your plane does have a motor. If you mistake and say  no motor but activate throttle cut, it would fail

companion/src/wizarddata.cpp
```diff
@@ -116,7 +116,7 @@ WizMix::operator ModelData()
       throttleChannel = i;
 
       // Add the Throttle Cut option
-      if( options[THROTTLE_CUT_OPTION] && throttleChannel >=0 ){
+      if( options[THROTTLE_CUT_OPTION]){
         MixData & mix = model.mixData[mixIndex++];
         mix.destCh = throttleChannel+1;
```
bsongis: why not `mix.destCh = i+1;` then?

companion/src/wizarddata.cpp
kilrah: Hmm shouldn't throttle cut be unsllectable instead in this case then?

companion/src/wizarddata.cpp
```diff
@@ -109,20 +108,26 @@ WizMix::operator ModelData()
...
 
-  // Add the Throttle Cut option
-  if( options[THROTTLE_CUT_OPTION] && throttleChannel >=0 ){
-    model.customFn[switchIndex].swtch.type = SWITCH_TYPE_SWITCH;
-    model.customFn[switchIndex].swtch.index = IS_ARM(getCurrentBoard()) ? SWITCH_SF0 : SWITCH_THR;
-    model.customFn[switchIndex].enabled = 1;
-    model.customFn[switchIndex].func = (AssignFunc)throttleChannel;
-    model.customFn[switchIndex].param = -100;
+    if (ch.input1 == THROTTLE_INPUT || ch.input2 == THROTTLE_INPUT) {
+      throttleChannel = i;
```
bsongis: `throttleChannel` is not anymore needed

PR #5806, opened 2018-04-07
XLite support

bsongis:
[X10] Serial on external module
[X10] Many fixes
[ARM] Backlight check moved to readKeysAndTrims() on all ARM boards. So that the backlight is again ON when pressing a trim or the SHIFT key


radio/src/opentx.cpp
```diff
@@ -285,8 +285,13 @@ void generalDefault()
...
   g_eeGeneral.vBatWarn = 66;
-  g_eeGeneral.vBatMin = -28; // 6.2V
-  g_eeGeneral.vBatMax = -38;   // 8.2V
+  g_eeGeneral.vBatMin = -28;  // 6.2V
+  g_eeGeneral.vBatMax = -38;  // 8.2V
+#elif defined(PCBXLITE)
+  // 2 x Li-Ion
+  g_eeGeneral.vBatWarn = 66;
+  g_eeGeneral.vBatMin = -23;  // 6.7V
+  g_eeGeneral.vBatMax = -37;  // 8.3V
```
kilrah: Hmm I'd change for X10 too, it's supposed to be the same after all...

radio/src/keys.cpp
```diff
@@ -106,7 +107,10 @@ void Key::input(bool val)
       if (m_cnt == KEY_LONG_DELAY) {
         // generate long key press
         // TRACE("key %d LONG", key());
-        putEvent(EVT_KEY_LONG(key()));
+#if !defined(LUA)
```
bsongis: It shouldn't be dependent of LUA / NO LUA

radio/src/keys.cpp
```diff
@@ -96,7 +96,8 @@ void Key::input(bool val)
       break;
     case KSTATE_START:
       // TRACE("key %d FIRST", key());
-      putEvent(EVT_KEY_FIRST(key()));
+      if (!IS_SHIFT_KEY(key()))
```
bsongis: Perhaps this test can be removed, I think the only problem we have with the SHIFT key is KEY_REPEAT

kilrah:
Current status:
- Xlite reboots fine from bootloader exit and a watchdog reset.
- Xlite recovery from USB mass storage is still random.
- X9E does not reboot from either bootloader exit or watchdog reset, but that is the case on the 2.2 branch as well so it got broken earlier.
- Page key and trims are broken on X7.

radio/src/gui/128x64/model_setup.cpp
```diff
@@ -700,7 +729,7 @@ void menuModelSetup(event_t event)
 #endif
         break;
       }
-#if defined(PCBTARANIS)
+  #if defined(PCBTARANIS)
```
bsongis: indentation

radio/src/gui/128x64/navigation.cpp
```diff
@@ -344,7 +346,7 @@ int checkIncDec(event_t event, int val, int i_min, int i_max, unsigned int i_fla
 #if defined(PCBXLITE)
   if (s_editMode > 0) {
 #endif
-    if (event==EVT_KEY_FIRST(KEY_RIGHT) || event==EVT_KEY_REPT(KEY_RIGHT) || (s_editMode>0 && (IS_ROTARY_RIGHT(event) || event==EVT_KEY_FIRST(KEY_UP) || event==EVT_KEY_REPT(KEY_UP)))) {
+    if (event==EVT_KEY_FIRST(KEY_RIGHT) || event==EVT_KEY_REPT(KEY_RIGHT)) {
```
bsongis: you break SKY9X here

radio/src/gui/128x64/navigation.cpp
```diff
@@ -355,7 +357,7 @@ int checkIncDec(event_t event, int val, int i_min, int i_max, unsigned int i_fla
         AUDIO_KEY_ERROR();
       }
     }
-    else if (event==EVT_KEY_FIRST(KEY_LEFT) || event==EVT_KEY_REPT(KEY_LEFT) || (s_editMode>0 && (IS_ROTARY_LEFT(event) || event==EVT_KEY_FIRST(KEY_DOWN) || event==EVT_KEY_REPT(KEY_DOWN)))) {
+    else if (event==EVT_KEY_FIRST(KEY_LEFT) || event==EVT_KEY_REPT(KEY_LEFT)) {
```
bsongis: same problem here

radio/src/gui/128x64/navigation.cpp
kilrah: Well it disables editing with up/down for it too, but we want that for consistency no?

kilrah:
Shall we merge and finish the last things in 2.2?

bsongis:
I have opened a new branch for next changes ...

PR #5812, opened 2018-04-09
Horus : Ignore pre-existing e_backlight_mode_off

3djc:
Since the BL changes, you cannot select e_backlight_mode_off anymore, but you can still have it in previous radio settings. 

With this change, e_backlight_mode_off behavior is similar to e_backlight_mode_keys on Horus.

We also need to change Companion to not allow it to set it to e_backlight_mode_off for Horus

### DIFF
--- a/radio/src/main_arm.cpp
+++ b/radio/src/main_arm.cpp
@@ -415,7 +415,12 @@ void perMain()
   }
 
   event_t evt = getEvent(false);
+
+#if defined(PCBHORUS)
+  if (evt && !(g_eeGeneral.backlightMode & e_backlight_mode_sticks)) {
+#else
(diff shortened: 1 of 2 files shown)


3djc:
I'm unsure what we should  be doing here:

1) Allow both e_backlight_mode_off to be choosen, but make it behave like e_backlight_mode_keys for Horus
Pro : allow migration between radios
        pre existing e_backlight_mode_off values are not an issue
Cons : two settings that behaves the same

2) Do not allow e_backlight_mode_off for Horus : if that value is seen, it is changed to e_backlight_mode_keys
Pro : you only see relevant choices
Cons :  the settings is lost when mlgrating from radio to radio

PR #5814, opened 2018-04-09
Add support for retrieving model ID over CRSF

fiam:
Add 2 frames types:

- REQUEST_MODEL_CONFIG_ID sent from the TX to the radio to request
the model configuration. Empty.
- MODEL_CONFIG_ID sent from the radio to the TX. Contains the number
of bytes of the model ID followed by the ID itself and the
null-terminated model name.

For retrieving the model ID, the TX sends REQUEST_MODEL_CONFIG_ID,
which makes the radio send MODEL_CONFIG_ID rather than CHANNELS_ID
in the next cycle. This allows the TX to only ask for the model ID
when it needs it without loosing a control cycle periodically.

### DIFF
--- a/radio/src/pulses/crossfire.cpp
+++ b/radio/src/pulses/crossfire.cpp
@@ -46,3 +46,30 @@ uint8_t createCrossfireChannelsFrame(uint8_t * frame, int16_t * pulses)
   *buf++ = crc8(crc_start, 23);
   return buf - frame;
 }
+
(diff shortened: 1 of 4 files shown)


fiam:
Main use case for this PR is allowing a TX module which can pair with multiple receivers (each RX with its own unique pairing key) to select the correct pairing key for the current model selected in the radio.

PR #5823, opened 2018-04-12
bootloader: use software reset instead of jump

### LINKED ISSUE #5627: Nightly bootloader reboot handling
2.2.2N bootloader does not reboot properly on exit on some models
Not working:
X7: looks like it is restarting for a sec and turns off
X9E : turns itself off directly

Working;
x12 : looks ok

I have not tested other models, all those work fine in 2.2.1

raphaelcoeffic:
fixes #5627


raphaelcoeffic:
tested so far on x7s, x9e and x12s.

PR #5825, opened 2018-04-12
Full 11 bit USB HID gamepad resolution and fixed offsets for compatibility

Ultrawipf:
This is a proposal to increase the usb gamepad resolution from 8 bit to the full 11 bit while improving compatibility for some games where you had to cut the range in half again.

The changes were tested on the taranis X9D+ and windows 10 with great results and should comply with the official USB HID standard.

### DIFF
--- a/radio/src/targets/common/arm/stm32/usb_driver.cpp
+++ b/radio/src/targets/common/arm/stm32/usb_driver.cpp
@@ -151,10 +151,13 @@ void usbJoystickUpdate()
     //analog values
     //uint8_t * p = HID_Buffer + 1;
     for (int i = 0; i < 8; ++i) {
-      int16_t value = channelOutputs[i] / 8;
-      if ( value > 127 ) value = 127;
-      else if ( value < -127 ) value = -127;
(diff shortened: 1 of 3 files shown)


raphaelcoeffic:
@Ultrawipf could you please squash your both commits together?

kilrah:
Will do that when merging, no prob...

This is interesting, great but also will cause some trouble with so many people having their simulator model now "broken"...

raphaelcoeffic:
Well, most simulators have a calibration function ;-)

Ultrawipf:
Thanks for the fast reply. 
The biggest issue with unity based simulators was that after calibrating away half the range from the previous 8 bit it definetly felt pretty choppy with only 128 values for the full range. This bugged me and probably thousands of other users for a long time.

Having the full resolution might be a big deal for opentx radios compared to others. It certainly feels smoother. I guess some simulator manuals have to be rewritten if this gets merged ;)

bsongis:
This seems good to me

schwabe:
@bsongis so ready to merge? Looks good to me too

Issue #5820, opened 2018-04-12
Custom failsafe: setting channel values to the current output with one button


raphaelcoeffic:
When setting the channel values on the "Custom failsafe" page, it would be way easier if one could just set all the channel custom failsafe values to the current output, similar to the "trim -> subtrim" function.

To have that function be intuitive enough, it seems that creating a new menu option just like "trim -> subtrim" seems to be the best option.

raphaelcoeffic:
@kilrah  suggested:
"could be long press on the "Set" button when custom failsafe is selected
i.e. short press open page, long press set all channels to custom and current position
someone actually posted a few days ago saying they tried that and expected it to work that way"

raphaelcoeffic:
Implemented in #5834

PR #5834, opened 2018-04-17
Set all custom failsafe channels at once

raphaelcoeffic:
Implements #5820.


raphaelcoeffic:
For now, it is only implementing the "Enter long-press" version as proposed by @kilrah.

While on the custom failsafe screen, each individual channel can already be set to current value by using enter long press while editing the value. The question is: should we have a special menu item on that screen, similar to "subtrim -> trim" menu in outputs? I think some space could be made for that menu item on Horus, and possibly X9D/E, but the X7 screen is really too small for that, and would require some special scrolling instead of paging (sorry, I'm too lazy for that one...).

Any other suggestion?

raphaelcoeffic:
Just added such a menu item for horus:
screen shot 2018-04-21 at 13 02 48

3djc:
output uses >>, not ->, but on x9, x7 it is => so we probably need to bring consistentcy Which also means the BW screens need it too :) (sorry it falls on you !)

raphaelcoeffic:
TODOs:
- [x] add "Outputs => failsafe" to X9.
- [x] add "Outputs => failsafe" to X7.

Screenshots:
screen shot 2018-04-21 at 13 09 51
screen shot 2018-04-21 at 12 37 47

radio/src/gui/128x64/model_setup.cpp
```diff
@@ -1200,9 +1200,18 @@ void menuModelSetup(event_t event)
             }
           } else if (menuHorizontalPosition == 1) {
             s_editMode = 0;
-            if (moduleData.failsafeMode == FAILSAFE_CUSTOM && event == EVT_KEY_FIRST(KEY_ENTER)) {
-              g_moduleIdx = moduleIdx;
-              pushMenu(menuModelFailsafe);
+            if (moduleData.failsafeMode == FAILSAFE_CUSTOM) {
+                if (event == EVT_KEY_LONG(KEY_ENTER)) {
```
bsongis: indentation issue

bsongis (approved):

radio/src/gui/128x64/model_setup.cpp
raphaelcoeffic: done!

bsongis:
Thanks!

PR #5841, opened 2018-04-23
Horus: add receiver number check

raphaelcoeffic:
Basically checks whether or not a receiver number is already used. In case it is, a warning is displayed.

Cases covered:
- check RX ID on any relevant change in module/proto/ID.
- assign a new free receiver ID on new model (internal RF).
- find the next free receiver ID while editing it (long press ENTER while editing).

The same features are now available as well on std LCD.

### DIFF
--- a/radio/src/CMakeLists.txt
+++ b/radio/src/CMakeLists.txt
@@ -157,7 +157,7 @@ endforeach()
 set(SRC ${SRC} debug.cpp)
 
 if(${EEPROM} STREQUAL SDCARD)
-  set(SRC ${SRC} storage/storage_common.cpp storage/sdcard_raw.cpp)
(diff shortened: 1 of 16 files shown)


bsongis:
Not related, what about changing this cli to a binary protocol + a python cmdloop script?

raphaelcoeffic:
@bsongis I like it to be able to connect just with a plain serial Terminal, without anything further. Do you have any special plans for the CLI?

bsongis:
Yes, I would like to see history, completion and more features without spending more flash, the python cmdloop script would connect to the serial port, send / receive binary data etc.

In short the same way adb works ...

raphaelcoeffic:
@bsongis I just rebased on top of 2.2, to avoid conflicts.

bsongis:
rebase needed :(

raphaelcoeffic:
@bsongis done, rebased on top of new `2.2`

PR #5845, opened 2018-04-25
Fr Translation

LapinFou:
Fixed few typos and add 1 missing translation.
Doesn't really need a review.

BTW, what does mean the abbreviation **CFN** in the _CustomFunctionData_?
It is in the file **companion\src\firmwares\customfunctiondata.cpp**, line n°248.
I can't translate it.

### DIFF
--- a/companion/src/translations/companion_fr.ts
+++ b/companion/src/translations/companion_fr.ts
@@ -11,7 +11,7 @@
     <message>
         <location filename="../wizarddialog.cpp" line="438"/>
         <source>Yes, controlled by a single channel</source>
-        <translation></translation>
+        <translation>Oui, contrôlés par 1 voie</translation>
     </message>
     <message>
         <location filename="../wizarddialog.cpp" line="439"/>
@@ -749,7 +749,7 @@ Peut être différente de la langue du firmware</translation>
     <message>
         <location filename="../firmwares/boards.cpp" line="311"/>
         <source>Ail</source>
-        <translation>Ail</translation>
+        <translation></translation>
     </message>
 </context>
 <context>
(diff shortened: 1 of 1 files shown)


kilrah:
CustomFunctionN... That was used in code usually, if it's a user-visible string it should be changed to SF

kilrah:
It's actually used to describe both SF and GF, which is detailed later in the name.

clipboard01

LapinFou:
Thanks Kilrah. 👍 

**SF** is translated into **FS**
**GF** is translated into **FG**
My suggestion would be "Custom Function N" translates in "Fonction Personnalisée N", so **CFN** → **FPN**

Any suggestion for the French translation?

kilrah:
GF should probably be FG...

OK for FPN...

LapinFou:
It is **FG**. I corrected my typo. 😃 
I will push the modification.

LapinFou:
CFN "Custom Function N" has been translated in "Fonction Personnalisée N".
**CFN** → **FPN**

companion/src/translations/companion_fr.ts
```diff
@@ -1503,12 +1503,12 @@ Voulez-vous les importer ?</translation>
...
         <location filename="../firmwares/customfunctiondata.cpp" line="94"/>
         <source>SetFailsafe Int. Module</source>
-        <translation>Déf. failsafe module interne</translation>
+        <translation>Définir failsafe module interne</translation>
     </message>
     <message>
         <location filename="../firmwares/customfunctiondata.cpp" line="96"/>
         <source>SetFailsafe Ext. Module</source>
-        <translation>Déf. failsafe module externe</translation>
+        <translation>Définir. failsafe module externe</translation>
```
kilrah: Pourquoi le '.' ?

LapinFou:
Nice catch. 😄

PR #5847, opened 2018-04-25
[X-Lite] Next round of changes

bsongis:
- PXX sent every 4ms (instead of 9ms)
- New option PXX_FREQUENCY (HIGH / LOW)

### DIFF
--- a/radio/src/pulses/pxx_arm.cpp
+++ b/radio/src/pulses/pxx_arm.cpp
@@ -163,7 +163,6 @@ void pxxPutPcmBit(uint8_t port, uint8_t bit)
     pxxPutPcmPart(port, 0);
     modulePulsesData[port].pxx.pcmOnesCount = 0;
   }
-
 }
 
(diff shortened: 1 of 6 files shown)


radio/src/pulses/pxx_arm.cpp
```diff
@@ -405,3 +422,22 @@ void setupPulsesPXX(uint8_t port)
...
 }
+
+void setupPulsesPXX(uint8_t port)
+{
+  initPcmArray(port);
+
+#if defined(PXX_FREQUENCY_HIGH)
+  setupFramePXX(port, 0);
+  if (NUM_CHANNELS(port) > 8) {
+    setupFramePXX(port, 8);
```
bsongis, line 432: the array size needs to be checked carefully, theorically the size would need to be 2 times bigger

bsongis: checked, it is OK

PR #5851, opened 2018-04-28
Review special function screen nav

3djc:
Special function screen changes:
- you cannot move right past last column
- you cannot move left of first column
- when on first column, long left will trigger copy/past/clear popup menu

### DIFF
--- a/radio/src/gui/128x64/model_special_functions.cpp
+++ b/radio/src/gui/128x64/model_special_functions.cpp
@@ -70,7 +70,7 @@ void onCustomFunctionsFileSelectionMenu(const char * result)
 }
 #endif // CPUARM && SDCARD
 
-#if defined(PCBX7)
+#if defined(PCBTARANIS)
 
 void onAdjustGvarSourceLongEnterPress(const char * result)
 {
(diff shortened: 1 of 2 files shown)


kilrah:
Hmm Right on an empty switch field makes cursor disappear...

radio/src/gui/128x64/model_special_functions.cpp
```diff
@@ -139,20 +139,27 @@ void onCustomFunctionsMenu(const char * result)
...
   }
 }
-#endif // CPUARM
+#endif // PCBTARANIS
 
 void menuSpecialFunctions(event_t event, CustomFunctionData * functions, CustomFunctionsContext * functionsContext)
 {
   int8_t sub = menuVerticalPosition - HEADER_LINE;
+#if defined(PCBXLITE)
+  if(menuHorizontalPosition > 4) menuHorizontalPosition = 4;
```
kilrah, line 148: Hmm not sure that's the best way, shouldn't it be handled by the menu system? @bsongis

3djc:
Replaced by a better version

PR #5855, opened 2018-04-29
3djc/xlite sf nav

3djc:
Fix SF nav:
- exiting the screen left or right make it appear on the other side
- Long Ent + SHIFT on first column bring copy/paste/clear menu
- no more cursor lost on right when cfn is empty

### DIFF
--- a/radio/src/gui/128x64/model_special_functions.cpp
+++ b/radio/src/gui/128x64/model_special_functions.cpp
@@ -139,7 +139,7 @@ void onCustomFunctionsMenu(const char * result)
     storageDirty(eeFlags);
   }
 }
-#endif // CPUARM
+#endif // PCBTARANIS
 
 void menuSpecialFunctions(event_t event, CustomFunctionData * functions, CustomFunctionsContext * functionsContext)
 {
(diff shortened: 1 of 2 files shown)


radio/src/gui/128x64/model_special_functions.cpp
```diff
@@ -171,6 +177,9 @@ void menuSpecialFunctions(event_t event, CustomFunctionData * functions, CustomF
     }
     POPUP_MENU_START(onCustomFunctionsMenu);
   }
+#if defined(PCBXLITE)
+  }
+#endif
 #endif // PCBX7
```
bsongis: wrong comment

radio/src/gui/128x64/model_special_functions.cpp
```diff
@@ -151,8 +151,14 @@ void menuSpecialFunctions(event_t event, CustomFunctionData * functions, CustomF
   uint8_t eeFlags = EE_MODEL;
 #endif
 
-#if defined(PCBX7)
+#if defined(PCBTARANIS)
+#if defined(PCBXLITE)
+  if (menuHorizontalPosition==0 && event==EVT_KEY_LONG(KEY_ENTER) && !READ_ONLY()) {
+    killEvents(KEY_ENTER);
+    if (IS_SHIFT_PRESSED()) {
```
bsongis: why? I think we need a comment here, it will be difficult to remember otherwise :)

PR #5866, opened 2018-05-02
Model print and compare user customisable style

### LINKED ISSUE #5750: Companion compare models enhancement suggestion
The 'compare models' feature is very useful but (at least on my iMac and for my eyesight) somewhat hindered by the too-subtle visual differences in use. Given the older demographic of model flyers I doubt my sight is drastically worse than the average!

Whilst the soft grey is nice for the 'plain' text, it blends too nearly into the green used for the changes in the left-side model details. The red is a bit easier to see when there is a substantial amount of it. The combination makes it quite  strain to work out the differences.

I'd like to urge some further cueing for the changes. A simple to implement option that might be enough would be adding underline to the text style. A 'change bar' at the outer edge of the two columns (as used in many word processors) would be another helpful cue. A more complex solution might be the option to only show the entires with changes; that really makes it clear!

A much more sophisticated approach - with obviously more work involved - would be a scheme typiifed by the text diif tool in TextWrangler. See https://i0.wp.com/subharanjan.com/wp-content/uploads/2014/09/Compared-Differences-TextWrangler.png for an illustration should you not be familiar with this. The ability to move versions from one side to the other (see the Apply buttons in the image) would be interesting too.

elecpower:
New feature to allow users the ability to customise the print style via a stylesheet. The main use is for model compare to more clearly highlight differences.
A reuseable simple file editor has been included to manage the customisable file.
The default is a resource file to ensure it is kept in sync with the code and any QT changes. The initial stylesheet version replicates the code.
The customised file is stored in the OS application data folder so it persists across Companion versions.
QT 5 has a number of limitations when using stylesheets.

Fixes #5750 

Cosmetic layout fixes.

### DIFF
--- a/companion/src/CMakeLists.txt
+++ b/companion/src/CMakeLists.txt
@@ -244,6 +244,7 @@ set(companion_SRCS
   radionotfound.cpp
   wizarddata.cpp
   wizarddialog.cpp
+  styleeditdialog.cpp
   )
 
 set(companion_MOC_HDRS
@@ -275,6 +276,8 @@ set(companion_MOC_HDRS
   modelprinter.h
   multimodelprinter.h
   modelslist.h
+  styleeditdialog.h
+  helpers_html.h
   )
 
 set(companion_UIS
@@ -295,6 +298,7 @@ set(companion_UIS
   flashfirmwaredialog.ui
   flasheepromdialog.ui
   radionotfound.ui
+  styleeditdialog.ui
   )
 
 if(WIN32)
--- a/companion/src/companion.qrc
+++ b/companion/src/companion.qrc
@@ -235,6 +235,7 @@
     <file>images/library/10802.png</file>
     <file>themes/default/style.css</file>
     <file>themes/default/style-osx.css</file>
+    <file>themes/default/modelprt.css</file>
     <file>themes/monochrome/16/paintbrush.png</file>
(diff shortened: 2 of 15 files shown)


elecpower:
QStandardPaths::AppDataLocation is only available from QT 5.4 and from the Travis build log looks like we compile against QT 5.3.
As QT 5.10 is available do you have a timeline for upping the supported version for Companion?
If you support the PR and the timeline is too far off I can rework the location code.

3djc:
I have bumped to 5.7 which is what we compile with, but you need to get the update

elecpower:
Is it me but are you sure Travis uses 5.7 as environment variable QT_BASE is set to 53 and the QT Version displays 5.3.2 in the log

kilrah:
He just changed it in the 2.2 branch. As he said you need to rebase, cherry-pick or such for the change to be carried into your branch.

elecpower:
Okay got it now. Thanks

elecpower:
Devs can one of you force this PR to rebuild?

bsongis:
Would you rebase against 2.2 to have Travis checks?

elecpower:
Rebased

bsongis:
my mistake ... would you rebase again?

bsongis:
Thanks!

PR #5867, opened 2018-05-02
[Xlite] Either ADC or PWM used for sticks (similar to X10)

bsongis:


3djc:
Tested ok on my xlite (well, doesn't break it, since I cannot test ADC sticks)

3djc:
Tested ok by FrSky

PR #5874, opened 2018-05-06
Reset backlight timer on rotary encoder (keys mode)

raphaelcoeffic:
I guess the rotary encoder can be consider a "key" as well.


3djc:
Sounds like a good idea to me !

raphaelcoeffic:
@3djc I just noticed the same could/should be added on X7 & X9E (commit has been added).

bsongis:
I think I have another change like this one in one of my local branches, I will add it to this review tomorrow

PR #5875, opened 2018-05-06
[X10] mute audio when fifo buffer is empty

### LINKED ISSUE #5561: X10S opentx v2.2.1 very loud noise at speaker, but no noise with FrOS
Its at a serie of the latest new X10S with openTx V2.2.1 
very very loud noise  at speaker with opentX V2.2.1
But no noise to hear with FrOS

Mybe FrOS switch the amplifer to Mute (off) if not in use
and opentx not?

could anyone test it, please
with an Oszi at the speaker, and  with an Sinus-Signal of 4kHz as wav-file

raphaelcoeffic:
Fixes #5561 by muting the X10 audio amplifier when no sound is being played. On targets with SOFTWARE_VOLUME, no audio is pushed to the queue when volume is muted.

This has been tested on X10 only.

### DIFF
--- a/radio/src/targets/common/arm/stm32/audio_dac_driver.cpp
+++ b/radio/src/targets/common/arm/stm32/audio_dac_driver.cpp
@@ -51,6 +51,17 @@ void dacInit()
   dacTimerInit();
 
   GPIO_InitTypeDef GPIO_InitStructure;
+
+#if defined(PCBX10)
+  GPIO_InitStructure.GPIO_Pin = AUDIO_MUTE_GPIO_PIN;
+  GPIO_InitStructure.GPIO_Mode = GPIO_Mode_OUT;
+  GPIO_InitStructure.GPIO_Speed = GPIO_Speed_2MHz;
+  GPIO_InitStructure.GPIO_OType = GPIO_OType_PP;
+  GPIO_InitStructure.GPIO_PuPd = GPIO_PuPd_UP;
+  GPIO_Init(AUDIO_MUTE_GPIO, &GPIO_InitStructure);
+  GPIO_SetBits(AUDIO_MUTE_GPIO, AUDIO_MUTE_GPIO_PIN);
+#endif
+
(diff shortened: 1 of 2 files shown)


radio/src/targets/common/arm/stm32/audio_dac_driver.cpp
```diff
@@ -51,6 +51,17 @@ void dacInit()
   dacTimerInit();
 
   GPIO_InitTypeDef GPIO_InitStructure;
+
+#if defined(PCBX10)
```
bsongis: why not `#if defined(AUDIO_MUTE_GPIO_PIN)` so that it could be easily reused?

radio/src/targets/common/arm/stm32/audio_dac_driver.cpp
```diff
@@ -83,6 +94,10 @@ void audioConsumeCurrentBuffer()
 
     nextBuffer = audioQueue.buffersFifo.getNextFilledBuffer();
     if (nextBuffer) {
+#if defined(PCBX10)
```
bsongis: same comment

radio/src/targets/common/arm/stm32/audio_dac_driver.cpp
```diff
@@ -91,6 +106,12 @@ void audioConsumeCurrentBuffer()
       DAC->SR = DAC_SR_DMAUDR1 ;                      // Write 1 to clear flag
       DAC->CR |= DAC_CR_EN1 | DAC_CR_DMAEN1 ;                 // Enable DAC
     }
+#if defined(PCBX10)
```
bsongis: same comment

bsongis (approved):
Excellent, thanks!

Ossyflop:
Hi all; I really like the work you are doing.  How do us mere mortals with an X12S implement this into our radios?

poge-st:
i have this problem on my X12s - can this be also fixed for the X12s ? is this the same reason ?

PR #5876, opened 2018-05-07
fixes #5870

### LINKED ISSUE #5870: X7 : leave BIND with enter not working anymore
Ok on 2.2.1, not ok on 2.2
It is ok on xlite both in 2.2.1 and 2.2

Step to reproduce : 
- on model page, launch bind, select bind option, binding start.
- push again enter, binding should stop, instead, it prompts for bind option again.

raphaelcoeffic:


raphaelcoeffic:
Adding a ref to #5870

radio/src/gui/128x64/model_setup.cpp
```diff
@@ -1139,7 +1139,7 @@ void menuModelSetup(event_t event)
               if (s_editMode > 0) {
                 if (l_posHorz == 1) {
                   if (IS_MODULE_R9M(moduleIdx) || (IS_MODULE_XJT(moduleIdx) && g_model.moduleData[moduleIdx].rfProtocol== RF_PROTO_X16)) {
-                    if (EVT_KEY_MASK(event) == KEY_ENTER) {
```
bsongis: @3djc we introduced this for the X-Lite, no?

3djc: Yes we did, xlite isn't going into bind with ` if (event == EVT_KEY_BREAK(KEY_ENTER))`, we should at least #ifdef it

raphaelcoeffic: Possibly some issue with the X-Lite navigation? Working differently from all others?

3djc: Yes, very different from all Taranis/Horus, more like 9x (but we made it more ‘taranis’ like than plain 9x)

PR #5879, opened 2018-05-09
Bsongis/flash 3 interfaces

bsongis:


3djc:
Tested OK for receiver flashing on xlite

bsongis:
Wording will be changed: "Flash s-port device"

PR #5881, opened 2018-05-09
Xlite PPM/Heartbeat trainer input

kilrah:
Will need review/comments...


radio/src/gui/gui_common_arm.cpp
```diff
@@ -597,7 +597,7 @@ bool isTrainerModeAvailable(int mode)
 #elif defined(PCBXLITE)
 bool isTrainerModeAvailable(int mode)
 {
-  if (g_eeGeneral.bluetoothMode == BLUETOOTH_TRAINER && (mode == TRAINER_MODE_MASTER_BLUETOOTH || mode == TRAINER_MODE_SLAVE_BLUETOOTH))
+  if ((g_eeGeneral.bluetoothMode == BLUETOOTH_TRAINER && (mode == TRAINER_MODE_MASTER_BLUETOOTH || mode == TRAINER_MODE_SLAVE_BLUETOOTH)) || mode == TRAINER_MODE_MASTER_CPPM_EXTERNAL_MODULE)
```
3djc, line 600: Are you sure you meant to condition that with BT mode been set to trainer ?

radio/src/opentx.cpp
```diff
@@ -469,7 +469,7 @@ void modelDefault(uint8_t id)
 #endif
 
 #if defined(PCBXLITE)
-  g_model.trainerMode = TRAINER_MODE_MASTER_BLUETOOTH;
+  g_model.trainerMode = TRAINER_MODE_MASTER_CPPM_EXTERNAL_MODULE;
```
3djc, line 472: Maybe better to check if BT is present, leave BT as default if yes, set this if not present ?

kilrah: It's just to select the first valid choice in the list (otherwise a new model would get master/jack which is invalid until it gets changed)

radio/src/gui/gui_common_arm.cpp
kilrah: Umm it's not?
That made me see some UI changes are needed too though...

radio/src/gui/128x64/model_setup.cpp
```diff
@@ -110,19 +110,14 @@ enum MenuModelSetupItems {
   ITEM_MODEL_PPM1_PROTOCOL,
   ITEM_MODEL_PPM1_PARAMS,
 #endif
-#if defined(PCBX7)
   ITEM_MODEL_TRAINER_LABEL,
```
bsongis: now available on Sky9X?

kilrah:
Needs changes, will rework...

Issue #5895, opened 2018-05-18
X-Lite N605 Bootloader Exit


MIKE-DAILEY:
N605. When exiting from bootloader (powerup with X held down) the system looks like it might not be selecting the correct gimbal type  When the system exits from the bootloader screen the gimbal inputs are jumping around.  A restart fixes the problem.

PR #5897, opened 2018-05-18
Potential fix for #5895

kilrah:
Reboot from bootloader or watchdog reset isn't subjected to PWR_PRESS_DURATION_MIN and so doesn't leave enough time to count enough transitions to switch to PWM gimbal mode.

Would be better to instead adjust the minimum number of transitions to the smallest possible reliable detection threshold to avoid an unnecessary delay in case of WDT reset. 

Needs to be checked on X7S, unless the PWM frequency is different in theory someone with PWM gimbals should currently have the same problem. Horus might get away with it due to longer boot time... or again it's just that Xlite PWM is much slower.

### DIFF
--- a/radio/src/targets/common/arm/stm32/sticks_pwm_driver.cpp
+++ b/radio/src/targets/common/arm/stm32/sticks_pwm_driver.cpp
@@ -22,14 +22,14 @@
 
 #define TIMESAMPLE_COUNT 6
 
-uint8_t analogs_pwm_disabled = 0;
+uint8_t sticks_pwm_disabled = 0;
 volatile uint32_t pwm_interrupt_count;
 volatile uint8_t  timer_capture_states[4];
 volatile uint32_t timer_capture_rising_time[4];
 volatile uint32_t timer_capture_values[4][TIMESAMPLE_COUNT];
 volatile uint8_t  timer_capture_indexes[4];
 
-void analogPwmInit()
+void sticksPwmInit()
 {
   GPIO_InitTypeDef GPIO_InitStructure;
   GPIO_InitStructure.GPIO_Pin = PWM_GPIOA_PINS;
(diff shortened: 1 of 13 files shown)


kilrah:
@bsongis how did you come up with the 1000 interrupts? Any reason not to reduce that instead?

radio/src/targets/taranis/board.cpp
```diff
@@ -252,6 +252,9 @@ void boardInit()
   else {
     pwrInit();
     backlightInit();
+#if NUM_PWMANALOGS > 0
+    delay_ms(730);
```
bsongis: wow are you sure?

bsongis:
1000 interrupts, it was perfect on X10, some tests are needed but it can be reduced a lot, perhaps even 100 would be enough

radio/src/targets/taranis/board.cpp
kilrah: Yup, tested... and kinda makes sense, PWM freq is 240Hz... so with 4 sticks about half a second is needed to get 1000 transitions! Hence why we would rather want to reduce that number...

bsongis: agreed, I am not sure it's mandatory to wait half a second here just to detect if we have PWM sticks!

radio/src/targets/common/arm/stm32/sticks_pwm_driver.cpp
```diff
@@ -101,10 +101,10 @@ extern "C" void PWM_IRQHandler(void)
   for (int i=0; i<4; i++) {
     if (PWM_TIMER->SR & (TIM_DIER_CC1IE << i)) {
       uint32_t capture = TIM_GetCapture(i);
+      pwm_interrupt_count++;
```
bsongis: overflow!

bsongis: unless, well ... we only use this variable at startup, it's ok then!
perhaps just add a comment?

kilrah: Was already the case before :)

bsongis: oh yes, true!
my bad ... I add the comment myself

PR #5902, opened 2018-05-21
Replace PyQt completely

raphaelcoeffic:
There is only `taranisicons.py` using PyQt, but it seems completely unused. That scripts might be removed all in all (not even the file paths are accurate).


raphaelcoeffic:
I checked that all generated files are 100% the same with the modified script. So basically, it should have absolutely no consequence on anything but getting rid of PyQt.

PR #5914, opened 2018-05-28
Bsongis/companion r9m fixes

bsongis:


3djc:
Tested ok for me.

Thought I should mention that thiis includes a design decision : when comming from/going to xlite, R9M settings; if any, are cleared, given it is impossible to match R9M and R9M mini settings

companion/src/modelprinter.cpp
```diff
@@ -228,13 +228,14 @@ QString ModelPrinter::printMultiSubType(unsigned rfProtocol, bool custom, unsign
 
 QString ModelPrinter::printR9MPowerValue(unsigned subType, unsigned val, bool telem)
 {
+  #warning "Here R9M power values are wrong";
```
kilrah: VC++ compilation fails on the #warning...

3djc: I will review the whole section as it is wrong for xlite

PR #5920, opened 2018-05-30
Companion : display 'mini' module in case of R9M and xlite

3djc:


3djc:
Merging with current R9M instead

PR #5939, opened 2018-06-03
Allow alternate locations for OpenSSL DLLs to be included with CPN installation

mpaperno:
The installer script now searches for `libeay32` and `ssleay32` DLLs in:
1) if MinGW build, then MinGW binary folder
2) in the `${WIN_EXTRA_LIBS_PATH}/OpenSSL` folder (WIN_EXTRA_LIBS_PATH is the existing user-configurable CMake variable)
3) `${PYTHON_DIRECTORY}/lib/site-packages/PyQt4/`  (fallback for current setups, no changes required)
4) Looks for an OpenSSL installation using CMake find_package() and if found uses DLLs from there.

Removes final PyQt4 dependency, I think.  Only other mention of PyQt I can find is in `radio\util\taranisicons.py` but I can't find anywhere this script is used.

### DIFF
--- a/companion/src/CMakeLists.txt
+++ b/companion/src/CMakeLists.txt
@@ -86,6 +86,7 @@ if(WIN32)
       set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /SUBSYSTEM:WINDOWS,5.01")  # is this even needed?
     endif()
   elseif(MINGW)
+    get_filename_component(MINGW_DIR ${CMAKE_CXX_COMPILER} DIRECTORY)
     # struct packing breaks on MinGW w/out -mno-ms-bitfields: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=52991 & http://stackoverflow.com/questions/24015852/struct-packing-and-alignment-with-mingw
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -mno-ms-bitfields")
   endif()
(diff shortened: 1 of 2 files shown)


mpaperno:
@kilrah After the `WIN_DO_FULL_INSTALL` build, all the required files are in one folder tree, ready to run. Would it simplify the NSIS installer script to just copy/install the whole tree?  I'm not really familiar with the scripting, but it seems like wildcards are definitely allowed (not sure if that works recursively, but still would simplify things).  Also I think this would make it independent of the compiler used (MSVC vs. MinGW).

raphaelcoeffic:
@mpaperno I think this script was a dead-end. Maybe we should remove it?

mpaperno:
The `taranisicons.py` script?  I don't know anything about it...  but +1  for dead-end removal.  :)

kilrah:
Sounds good, jsut need to see whether any of these alternatives are available on a "clean" setup or if build instructions need to be updated to install some source for those libs that will be found by these changes...

mpaperno:
The reason for the PR is that I was just going in to edit the build instructions for Windows (to include Pillow) and thought... can we get rid of PyQt?  :)

Nah, none of these sources for the DLLs are installed by default (except with the MinGW build).  I have 100s of GBs of various dev tools and libs installed, yet only a few copies of those DLLs.  So yes, instructions would need to be updated.  Another gotcha is that from OpenSSL they're only available in v1.x distros, not 2.x. Though it is not hard to find a d-load on the 'net for just those DLLs, so perhaps we find a good URL to include with the instructions.  I guess we could also consider including them in the repo.

Issue #3445, opened 2016-04-26
Companion Curve Editor, improve visualisation of curve points


Spainman:
Two suggestions that I personally think would improve usability and shouldn't require much changes code wise (assuming the plotting library doesn't suck) + a bug:

**Possible improvements:**
a) Let's get rid of the shadows under the curve points
- personally I think they fail as 'eye candy' 
- they hinder the visualization of the actual point position (I get an impression of double vision and no - there is no drunken OpenTx-ing involved ;-) )

b) Minimize the size of the curve points 
- personally I operate on the table values (works great with the keyboard cursor keys + tab, shift tab) and not with the mouse pointer, so a much smaller size would provide a better visualization of the transitions on the points
- maybe the graphing framework even provides the option to set the focus range different from the visual size, so that handling the points with the mouse pointer would not have to be impacted?
- if there is no such option, I still think that a point size that allows a good handling with the mouse but still is significantly smaller than it is now would still be an improvement

Opinions?

**Bug / missing functionality (?) :**
c) The smoothing option does not visually smooth the curve when editing the points manually (= not using the curve generator)
- I'm not sure it is intended. I usually edit the curves on the transmitter and there the smoothing obviously works when editing the points (which is lovely and a much appreciated functionality btw :+1: )

projectkk2glider:
Related: #2998, #1911 

This issue is actually a duplicate of #1330, I will close that one, since this has more info.

mpaperno:
Related: #4972

PR #5947, opened 2018-06-05
Companion curve enhancements

### LINKED ISSUE #4972: Allow add/remove points in curve editor
When editing a curve, if I decide I need more points to refine a curve I should be allowed to add a point via Ctrl-Click or some such thing.  The desired behavior would be that a Ctrl-Click would fit a point on the current curve at the current mouse X coordinate.

elecpower:
For Custom X curves add points using Ctrl+Left mouse button and delete points using Shift+Ctrl+Left mouse button. Fixes #4972.
For all types allow user selected point sizing and remove point shadow (looks fuzzy and does not work when point size is small Fixes parts a and b of #3445.

### DIFF
--- a/companion/src/modeledit/curves.cpp
+++ b/companion/src/modeledit/curves.cpp
@@ -122,8 +122,10 @@ Curves::Curves(QWidget * parent, ModelData & model, GeneralSettings & generalSet
     ui->curveNameLabel->hide();
   }
 
-  QGraphicsScene *scene = new QGraphicsScene(ui->curvePreview);
+  scene = new CustomScene(ui->curvePreview);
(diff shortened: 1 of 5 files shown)


companion/src/modeledit/curves.cpp
```diff
@@ -642,3 +651,78 @@ void Curves::ShowContextMenu(const QPoint& pos) // this is a slot
...
+    }
+    model->curves[currentCurve].points[newidx].x = x;
+    model->curves[currentCurve].points[newidx].y = y;
+    update();
+    emit modified();
+  }
+}
+
+CustomScene::CustomScene(QGraphicsView * view)
+{
```
mpaperno: Should call the QGraphicsScene() constructor here.

PR #5962, opened 2018-06-12
Do not display disabled trim on main views

3djc:
Based on a disucssion on rocket chat

Individual trims that are desactivated by flight mode settings are not displayed on main views

image

### DIFF
--- a/radio/src/gui/128x64/view_main.cpp
+++ b/radio/src/gui/128x64/view_main.cpp
@@ -105,8 +105,10 @@ void displayTrims(uint8_t phase)
     uint8_t att = ROUND;
     int16_t val = getTrimValue(phase, i);
 
+#if defined(CPUARM)
     if(getRawTrimValue(phase, i).mode == TRIM_MODE_NONE)
       continue;
+#endif
 
 #if !defined(CPUM64) || !defined(TELEMETRY_FRSKY)
     int16_t dir = val;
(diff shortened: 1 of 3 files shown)


LapinFou:
Maybe the same thing could be done on Companion's simulator?

3djc:
Not sure I understand your comment mon lapin!

LapinFou:
😄 

Here is an example;
- Taranis X9 
- Mode1 (Thr on the right)
- In the FMs, the throttle trim is disabled.

So, maybe, the trim should not be displayed on the simulator (same concept that you applied on the firmware).

image

3djc:
I see, but while trims are disabled, trims buttons are not and can be used for other purposes, so they need to stay

LapinFou:
You're correct, I forgot about that.
Only the scroll bar should be not display.

3djc:
There is a moment where "le mieux est l'ennemi du bien" !

LapinFou:
Oui. :)
Et puis, c'est purement cosmétique. Sans parler que si le trim est désactivé, l'ascenseur ne bouge pas.

bsongis:
ouch mon lapin va devoir réviser le bécherel :)

LapinFou:
??? Je ne voie pas de quoi tu parles... 😇

PR #5965, opened 2018-06-12
Turn backlight on for switch warning

### LINKED ISSUE #5959: Backlight turns off on a Switch Warning.
Taranis (non plus), on 2.2.2RC1
Backlight is turning off whenever a model is loaded and a Switch Warning occurs.
Clearing the warning turns the backlight back on.

To test, in Radio Settings have the Backlight Switch set to OFF.
Set a switch in Special Functions to turn ON the backlight. I use a Logical Switch based on the Right Slider position.
Have the backlight on and then load a model that will generate a switch warning. The light turns off.

In 2.2.1 the light is not turning off.
Jim

3djc:
SW do not turn backlight on when other do.
Also enables backlight for splash screen.

This (kinda) fixes #5959

### DIFF
--- a/radio/src/opentx.cpp
+++ b/radio/src/opentx.cpp
@@ -971,6 +971,7 @@ void doSplash()
 #endif
 
   if (SPLASH_NEEDED()) {
+    backlightOn();
     drawSplash();
 
 #if !defined(CPUARM)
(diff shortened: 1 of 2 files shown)


3djc:
Please do not merge yet, there seems to be a similar issue with splash screen

PR #5970, opened 2018-06-15
[Companion] Fix RF Output Power setting getting reset to default on every update

### LINKED ISSUE #5954: [Companion] Changing failsafe mode reset the RF power field
Hi folks,

- Companion 2.2.2N627 Win version.
- Select FrSky **R9M Module** for external RF.
- Select a RF power different of the default one.

Each time the failsafe mode is change, this will reset the RF power field to its default value.

mpaperno:
RF Output Power would go back to default (first choice in combobox) every time the `update()` routine got run, which was quite often, including when [re-]opening the model editor.  

Also prevents some redundant updates.

Fixes #5954

### DIFF
--- a/companion/src/modeledit/setup.cpp
+++ b/companion/src/modeledit/setup.cpp
@@ -483,6 +483,7 @@ void ModulePanel::update()
   ui->warning_r9mPower->setVisible((mask & MASK_R9M) && module.subType == R9M_LBT);
 
   if (mask & MASK_R9M) {
+    const QSignalBlocker blocker(ui->r9mPower);
     ui->r9mPower->clear();
     Board::Type board = firmware->getBoard();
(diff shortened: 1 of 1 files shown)


companion/src/modeledit/setup.cpp
```diff
@@ -483,6 +483,7 @@ void ModulePanel::update()
   ui->warning_r9mPower->setVisible((mask & MASK_R9M) && module.subType == R9M_LBT);
 
   if (mask & MASK_R9M) {
+    const QSignalBlocker blocker(ui->r9mPower);
```
mpaperno, line 486: This is the main fix for RF power getting reset.

3djc:
Tested ok on xlite and standard size r9m ! 👍

PR #5975, opened 2018-06-18
Trainer input on Xlite external module RX pin

kilrah:
"Trainer/jack" path seemed most logical since it's not the heartbeat pin


kilrah:
Rebased, ready

bsongis:
Rewritten in this PR:
https://github.com/opentx/opentx/pull/6497
More tests needed before it can be merged

PR #5985, opened 2018-06-22
Added X-Lite support to CROSSFIRE Lua

DieHertz:
Done similarly to https://github.com/betaflight/betaflight-tx-lua-scripts/pull/135
Should I create a new script directory for XLITE or it's okay to modify X7 scripts that way?


3djc:
- let's not create new dir unless strictly  necessary (which it shouldn't be)
- is that tested with  a module or purely theoritical ?

DieHertz:
It works and key mapping was customized to make it feel uniform with the rest of UI.
image

3djc:
👍

PR #5986, opened 2018-06-22
 Remove nofp option for a few board

3djc:
Remove no flight mode option often generating compile error on some boards

### DIFF
--- a/companion/src/firmwares/opentx/opentxinterface.cpp
+++ b/companion/src/firmwares/opentx/opentxinterface.cpp
@@ -1211,7 +1211,6 @@ void registerOpenTxFirmwares()
   /* 9XR-Pro */
   firmware = new OpenTxFirmware("opentx-9xrpro", QCoreApplication::translate("Firmware", "Turnigy 9XR-PRO"), BOARD_9XRPRO);
   firmware->addOption("heli", QCoreApplication::translate("Firmware", "Enable HELI menu and cyclic mix support"));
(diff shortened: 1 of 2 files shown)


companion/src/firmwares/opentx/opentxinterface.cpp
```diff
@@ -1211,7 +1211,6 @@ void registerOpenTxFirmwares()
   /* 9XR-Pro */
   firmware = new OpenTxFirmware("opentx-9xrpro", QCoreApplication::translate("Firmware", "Turnigy 9XR-PRO"), BOARD_9XRPRO);
   firmware->addOption("heli", QCoreApplication::translate("Firmware", "Enable HELI menu and cyclic mix support"));
-  firmware->addOption("nofp", QCoreApplication::translate("Firmware", "No flight modes"));
```
bsongis, line 1214: "nofp" was a bad name anyhow!

radio/util/fwoptions.py
```diff
@@ -125,8 +125,8 @@
 options_gruvin9x = {
     "heli": ("HELI", "YES", "NO"),
     "templates": ("TEMPLATES", "YES", "NO"),
-    "nofp": ("FLIGHT_MODES", "NO", "YES"),
     "nocurves": ("CURVES", "NO", "YES"),
+    "nofp": ("FLIGHT_MODES", "NO", "YES"),
```
bsongis, line 129: "nofm" would be better (for 2.3 ...)

3djc: no even, those are options for 'unsupported' boards :D

PR #5995, opened 2018-06-26
Clean up PWM stick reads, remove double averaging

schwabe:
We already do averaging in the normal anaIn/getADC().
Also according to Mike's and my own research we don't need the averaging.

Tested on two X-Lites with PMW sticks

### DIFF
--- a/radio/src/targets/common/arm/stm32/sticks_pwm_driver.cpp
+++ b/radio/src/targets/common/arm/stm32/sticks_pwm_driver.cpp
@@ -20,14 +20,9 @@
 
 #include "opentx.h"
 
-#define TIMESAMPLE_COUNT 6
-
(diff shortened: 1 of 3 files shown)


radio/src/targets/common/arm/stm32/sticks_pwm_driver.cpp
```diff
@@ -20,14 +20,9 @@
 
 #include "opentx.h"
 
-#define TIMESAMPLE_COUNT 6
-
-uint8_t sticks_pwm_disabled = 0;
+bool sticks_pwm_disabled = false;
```
bsongis, line 23: bool uses only one byte on arm gcc, right?

schwabe: Yes.

PR #6003, opened 2018-06-29
remove strstr() to prevent X-lite & X7 bootloader from overflowing

raphaelcoeffic:


raphaelcoeffic:
In case #6004 is merged, this one is useless, as `strstr` is just as cheap as `strncmp` with `nano.spec`.

PR #6006, opened 2018-06-30
Refactor Companion settings system

### LINKED ISSUE #4996: [Companion] Export-import companion configuration
Companion has different configurations: radio profile, simulator settings, application settings etc.
May be useful to provide a function that allow export and import all settings in a file,  in case of use on different PC or system reset a new installation may be simplified

mpaperno:
+ All stored properties are fully dynamic, using an extended version of the Qt Property System;
 + CompStoreObj is now QObject subclass;
 + All properties now have change notifier signals and reset (to default) handlers;
 + Do not store default values in settings;
 + Settings import now ignores all non-existent properties automatically;
 + Add settings export functions;
 + Add GUI and CLI methods for settings export and import to/from file (closes #4996);
 + On first start, offer option to import settings from file (as well as previous version if found);
 + Add more version macros to `version.h.in`;
 + Move `DownloadBranchType` and create `NewModelAction` enums and register with Qt meta system;
 + **Settings storage location determined by version number at build time** (from version.h / CMake config);
 + Offers to import settings from v2.2.x (if built as v2.3.x);
 + Backwards compatible with 2.2;

The best part is that you only need a one-line macro to add a new property/setting (or remove one).  Just about everything else is now automated.

To clarify the highlighted bullet point, if you build it eg. with VERSION_MAJOR=2 and VERSION_MINOR=3 (the defaults on the 2.3 branch), settings will be stored in `OpenTX/Companion 2.3`. The VERSION macros directly affect the numbered part, basically like before, but now it's automatic.

I documented the classes and macros pretty extensively, and also [generated some docs](http://max.wdg.us/OpenTX/doc/html/group___app_data.html).


mpaperno:
Please don't squash. (I already have an "un-squashed" version and it will be easier to reconcile if needed).

mpaperno:
image

image

image

First startup (with no settings), or with `--import` CLI option:
image

PR #6008, opened 2018-06-30
add LUA access to serial port (serial2 and USB)

raphaelcoeffic:
This should allow for implementing #5946 with just a LUA script.

For now it requires `-DUSB_SERIAL` to be passed to `cmake`, as USB CDC does use some buffers, but I think we could support it by default on the Horus targets, as memory is not an issue there. On Taranis targets, the serial port has easy access, so that it can be used instead.


raphaelcoeffic:
I used following function script for testing purposes:
```
local next_ts
local interval = 20

local function init()
   next_ts = getTime() + interval
end

local function run()

   local timenow = getTime() -- 10ms tick count

   if timenow >= next_ts then

      if timenow - next_ts > 2*interval then
         -- missed quite some ticks...
         next_ts = timenow + interval
      else
         next_ts = next_ts + interval
      end

      serialWrite("ticks: "..timenow.."\n")
   end
end

return { run=run, init=init }
```

radio/src/lua/api_general.cpp
```diff
@@ -1280,6 +1280,44 @@ static int luaResetGlobalTimer(lua_State * L)
   return 0;
 }
 
+/*luadoc
+@function serialWrite(str)
+@param str (string) String to be written to the serial port.
+
+Writes a string to the serial port. The string is allowed to contain any character, including 0.
+
+@status current Introducted in TODO
```
raphaelcoeffic, line 1289: I can edit this text once I know the exact release number.

Ryanf55:
edit: user error. BIN works for serial write.

Raphael sent me a BIN file with the above support for LUA on the serial port. I tried creating that function, but run into a syntax error when placed in the main script directory.

Thus, that's probably why I don't seeing anything out of the serial port in LUA mode with baud @ 115200 when I run it as a function script.
image
dsc01633

Github doesn't allow me to attach the LUA file... If you have any suggestions, please let me know. I'd love to get this feature working.

raphaelcoeffic:
@Ryanf55 "Custom Scripts" are for mixer script, which have another API (see doc). What you need is a "Special function". Just pick a switch or whatever should activate the script, then "Lua script" and then your script (script file should be located in `SCRIPTS/FUNCTIONS/` on your SD card).

bsongis:
What about 2.3 for this one?

raphaelcoeffic:
Why not, first app (SkyNav) needs some polishing anyway.

raphaelcoeffic:
@bsongis rebased on top of 2.3.

bsongis:
Thanks!

Ryanf55:
Thanks for the help. I put lser.lua into the SCRIPTS/FUNCTIONS and then set up a special function. Additionally, the taranis is set to "LUA" mode in the "Hardware/Serial Port", however I do not get anything out of the port. When it's in "S-Port Mirror" with no receiver, I saw this in realterm at 57.6k :
image


When in Lua mode, It sends a null char when first switched into LUA mode, but nothing when I activate the special function (reading at 115.2k. I can check on the scope if needed.

*Attached lser.LUA zipped
[lser.zip](https://github.com/opentx/opentx/files/2174793/lser.zip)

kilrah:
Did you consider the fact that the port has RS232 polarity?
For S.Port mirror you need to view as hex, there is no meaningful ASCII content.

Ryanf55:
Thanks Kilrah. Using an actual RS232 port on my desktop instead of the usb-serial adapter worked. Apparently the serial adapters are TTL instead of RS232.
image

Can't wait to implement some serial control of some peripherals!

Ryanf55:
Any update to LUA access of the RX line? Even a small buffer at a lower baud rate would be awesome. 

Thanks!

Ryanf55:
@raphaelcoeffic  In reference to my last request, I would be happy with a function that returns the serial received data register. At this point, I can deal with slowing the input stream so that only one character is sent per 50ms. For example, a function like serialRead() that returns an int that has a value 0-511 if there is received data, or -1 if no data has been received. 

Thanks in advance!

raphaelcoeffic:
@Ryanf55 I'm indeed considering that read function, but it needs to be done carefully to not eat more memory than we are already doing right now. I should be able to come back to that shortly.

Ryanf55:
@raphaelcoeffic Hope you are doing well. Have you had a chance looking at the read function and associated memory issues?

Ryanf55:
Just a heads up on this..  Companion simulator throws an error if I run any Lua code with serialWrite() in it.

radio/src/targets/common/arm/stm32/serial2_driver.cpp
```diff
@@ -101,6 +101,8 @@ void serial2Init(unsigned int mode, unsigned int protocol)
         uart3Setup(FRSKY_D_BAUDRATE, true);
       }
       break;
+    case UART_MODE_LUA:
+      uart3Setup(DEBUG_BAUDRATE, false);
```
Ryanf55, line 105: Can this not be DEBUG_BAUDRATE because 115k inverted requires special hardware to read. Softwareserial works well on pretty much all Arduino hardware at 57k and is able to read inverted logic. 

If you can, change it to 57k and I don't think anyone would complain or make it some config option.

Thanks!

Issue #5735, opened 2018-03-01
X10S sync overwrites model files


ufoDziner:
Hello,
  I normally make model changes in the field on the radio, and in companion when at home.  If I make changes at the field, I import the models, then save them to the local .otx file.  If I make them in the companion, I save to the .otx file, then write the models and settings to the radio.  Both of these work well.  The problem comes with sync.  It seems that no matter which way I have the SD card sync, the model files on the radio are overwritten by old files.  So, after a sync, I have to write the models and settings to the radio again.  If I want to have the current files that I just wrote to the radio synced with my local SD mirror, I have to copy them manually.

It seems like there needs to be a way to separate/specify which set of model files to use in the sync vs. the rest of the files.  This way, it would be easy to use the model.bin files off of the radio to sync with the local SD, but send the new local SD files to the radio.

Thanks

raphaelcoeffic:
@schwabe could this be related to the file time stamps not being updated properly on FAT?

raphaelcoeffic:
@ufoDziner could you please tell me which options you use for the SD card sync?

ufoDziner:
@raphaelcoeffic I'm attaching some screenshots to show the settings and results.  The otx1.png shows the settings and and my saved .otx file.  Just before taking that screenshot, I imported the model settings from the radio and saved them to my .otx file to ensure my local file is the most current.  I then imported them again (which shows as "document2.otx" in the third image).  Next Icompleted a sync using the sittings in the first image.  The second image shows the summary after I completed the sync.  Finally I imported the model settings from the radio after the sync ("document3.otx" in the last image).  As you can see, my radio settings are now different than what I had before.  I had made no recent changes locally, only on the radio.  It seems that no matter what I do on the radio, it will always get overwritten with the local settings during a sync.  

![otx1](https://user-images.githubusercontent.com/7434287/38698570-609ea872-3e4a-11e8-9e35-24f55a6e79cd.png)
![otx2](https://user-images.githubusercontent.com/7434287/38698576-652b4efe-3e4a-11e8-8d05-41aa507fb022.png)
![otx3](https://user-images.githubusercontent.com/7434287/38698586-6aaac99a-3e4a-11e8-8512-7fc52925ba1a.png)

raphaelcoeffic:
@ufoDziner there must be an issue with the timestamps. Could you please check the timestamps for the files on the SD card and in your local directory after each of the steps you took? At some place, something must be screwed there.

[...]

Issue #5213, opened 2017-09-13
sdcard sync and hiden files


s0lstice:
Hello,

I'm under linux and during a synchronisation from local to sdcard I see the copy of ".trash/*" files.

Can we ignor the hiden files ?

You make a good work, thank you ;)

Mickaël

schwabe:
The Remote should not show them. Or do you mean during synchronisation?

s0lstice:
I can see hiden files on the remote (Horus). In "system/Card SD-HC" I have [.fseventsd], [.Spotlight-V100], [.Trashes], ... the other files are vilisble and needed.

As I deleted all local sdcard directory to use a new version without unsused files. The ".Trashes" dirctory contain all files of last version. In companion, the synchronization have teack a while.

I can't remove unecessary files because the sdcard is mount as read only.

3djc:
Why is it mounted read only ???

s0lstice:
I don't know why. The horus firmware have the "mass storage" option activied. When I plug the horus I can't change the content of the sd card with the files browser. I need to use companion.

It is the same when I directly insert the sd card with an adapter.
I have a full access on another sd card mounted with the same adapter.

When I look what have done the automount with ":$mount" cli the answer seems to be good :

`:$ mount`
`/dev/sdc1 on /media/s0lstice/9016-4EF8 type vfat (rw,nosuid,nodev,relatime,uid=1000,gid=1000,fmask=0022,dmask=0022,codepage=437,iocharset=iso8859-1,shortname=mixed,showexec,utf8,flush,errors=remount-ro,uhelper=udisks2)`

3djc:
"It is the same when I directly insert the sd card with an adapter.
I have a full access on another sd card mounted with the same adapter."

Doesn't this says this as nothing to do with opentx ?

s0lstice:
I don't know enough opentx. This is my first radio with this software.

I can just say that I can not write on the sd card formatted for opentx from the computer. When I'm using an adapter or the USB from the radio. I need to use companion.

If my memory is good, in december, I have write on sd card to do the first install and the update few days after. This is the only SD card on which I can not write.

kilrah:
Then that SD is either faulty or needs reformatting. There's nothing specific to OpenTX with it, and companion uses nothing different from what your system would.

Also, what firmware version on the Horus? Pretty sure hidden file display was disabled at some point.

s0lstice:
I will format the sdcard and give you the resulat.

[...]

Issue #5106, opened 2017-07-23
SD synchronization bag.


kinshas:
Windows 10, was checked 64bit and 32bit the same problem, SD synchronization bag. Companion and OpenTX 2.2! OpenTX and Comp 2.19 worked perfectly.
During synchronization the companion showed that message:

Starting synchronization: E:\RC Plane\SD TARANIS 2_2 -> D:\

Finished synchronizing E:\RC Plane\SD TARANIS 2_2 -> D:\ :
     New: 0; Updated: 0; Skipped: 1022; Errors: 0;
================================================================================
Starting synchronization: D:\ -> E:\RC Plane\SD TARANIS 2_2

Creating destination file: E:\RC Plane\SD TARANIS 2_2\System Volume Information\IndexerVolumeGuid
Copy failed: 'D:\System Volume Information\IndexerVolumeGuid' to 'E:\RC Plane\SD TARANIS 2_2\System Volume Information\IndexerVolumeGuid': Cannot create E:\RC Plane\SD TARANIS 2_2\System Volume Information\IndexerVolumeGuid for output
Creating destination file: E:\RC Plane\SD TARANIS 2_2\System Volume Information\WPSettings.dat
Copy failed: 'D:\System Volume Information\WPSettings.dat' to 'E:\RC Plane\SD TARANIS 2_2\System Volume Information\WPSettings.dat': Cannot create E:\RC Plane\SD TARANIS 2_2\System Volume Information\WPSettings.dat for output

Finished synchronizing D:\ -> E:\RC Plane\SD TARANIS 2_2 :
     New: 0; Updated: 0; Skipped: 1022; Errors: 2;

3djc:
https://answers.microsoft.com/en-us/insider/forum/insider_wintp-insider_files/cannot-delete-system-volume-information-folder-in/0b80793f-4d9a-4df2-808c-59a24835709e

and

https://answers.microsoft.com/en-us/insider/forum/insider_wintp-insider_files/whats-wpsettingsdat-generated-by/e11bca97-8c76-4662-8897-774ea3d5691a

Not a companion issue

kilrah:
The error is you having a "System Volume Information" on your SD card. You likely have formatted it NTFS instead of FAT32.

kinshas:
to kilrah SD formatted in FAT32! And I remind 2.19 worked well!
to 3djc looks like you're right but still cannot fix!

3djc:
[...]

PR #6009, opened 2018-06-30
SD Sync "version 3"

### LINKED ISSUE #5005: [Companion] - SD Sync window doesn't resize during operation.
While this issue isn't a show stopper by any means, the window can't be resized while the operation is occurring.

I've included a screenshot to show the issue.

Ubuntu 16.04, 2.2.0 released on May 31st, 2017.

screenshot_2017-06-04_18-59-20

mpaperno:
+ Move UI to new FileSyncDialog class;
  + Improve UI responsiveness during operation (closes #5005);
  + Add log verbosity option (lower verbosity helps with UI response and overall speed);
  + Add several filtering settings in new "extra options" panel (include/exlude patterns, symlinks, recursive, empty folders) (overkills #5213 :-) );
  + Exclude patterns `.*` and `System Volume*` by default (addresses #5106);
  + Sanity check file timestamps when using date comparison, future dates are invalid (addresses #5735);
  + Folders are no longer included in file counts (total/created/etc);
  + Ignore unreadable files/dirs (also enables NTFS permissions check);
  + Integrates the progress/log widget, options form remains open (but disabled) during sync;

With "extra options" panel visible, which is everything starting with "Filters" line:
image

### DIFF
--- a/companion/src/CMakeLists.txt
+++ b/companion/src/CMakeLists.txt
@@ -246,6 +246,7 @@ set(companion_SRCS
   wizarddata.cpp
   wizarddialog.cpp
   styleeditdialog.cpp
+  dialogs/filesyncdialog.cpp
   )
 
 set(companion_MOC_HDRS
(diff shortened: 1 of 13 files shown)


mpaperno:
Would prefer not to squash this either, at this point. The commits are atomic.

bsongis:
Thanks!

PR #6010, opened 2018-07-01
Prevent leftover network handles/threads after update checks.

mpaperno:
Noticed this while checking the sound file handles issue.  From QNetworkManager [docs](http://doc.qt.io/qt-5/qnetworkaccessmanager.html#details):

>**Note**: After the request has finished, it is the responsibility of the user to delete the QNetworkReply object at an appropriate time.

Also cleans up multiple instances of QNetworkManager being created.

I think it's safe for 2.2.2 (and PR is against 2.2 branch), but should definitely go into 2.3 either way.

### DIFF
--- a/companion/src/mainwindow.cpp
+++ b/companion/src/mainwindow.cpp
@@ -80,8 +80,9 @@ const char * const OPENTX_COMPANION_DOWNLOAD_URL[] = {
 };
 
 MainWindow::MainWindow():
-  downloadDialog_forWait(NULL),
+  downloadDialog_forWait(nullptr),
   checkForUpdatesState(0),
+  networkManager(nullptr),
   windowsListActions(new QActionGroup(this))
 {
   // setUnifiedTitleAndToolBarOnMac(true);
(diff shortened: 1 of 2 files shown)


companion/src/mainwindow.cpp
```diff
@@ -80,8 +80,9 @@ const char * const OPENTX_COMPANION_DOWNLOAD_URL[] = {
 };
 
 MainWindow::MainWindow():
-  downloadDialog_forWait(NULL),
+  downloadDialog_forWait(nullptr),
```
bsongis, line 83: why not give the value in the declaration?

companion/src/releasenotesfirmwaredialog.cpp
```diff
@@ -48,5 +48,6 @@ void ReleaseNotesFirmwareDialog::replyFinished(QNetworkReply * reply)
 {
   ui->textEditor->setHtml(reply->readAll());
   ui->textEditor->setOpenExternalLinks(true);
+  reply->deleteLater();
```
bsongis, line 51: oh a deleteLater function :)

bsongis:
needs to be merged into 2.3

companion/src/mainwindow.cpp
mpaperno: For the dialog? Because we don't want to create the dialog here... ?  Or do you mean give it a default `nullptr` value in the header?  I'm not sure on the latter honestly... it's a style thing I guess and I haven't really transitioned (yet)... either way is OK, I think, as long as it's consistent per class.  But since complex values can't be initialized in declaration, I tend to stick to constructor for consistency.

3djc:
I have done it :)

PR #6013, opened 2018-07-01
telemetry simulator fixes

raphaelcoeffic:
- Trigger "telemetry lost" properly when "simulate" is ticked out or window is closed (RSSI=0).
- GPS precision fixed when read from CSV log file.

### DIFF
--- a/companion/src/simulation/telemetrysimu.cpp
+++ b/companion/src/simulation/telemetrysimu.cpp
@@ -73,7 +73,7 @@ TelemetrySimulator::TelemetrySimulator(QWidget * parent, SimulatorInterface * si
 
 TelemetrySimulator::~TelemetrySimulator()
 {
-  timer.stop();
+  stopTelemetry();
   logTimer.stop();
   delete logPlayback;
   delete ui;
@@ -93,10 +93,10 @@ void TelemetrySimulator::onSimulatorStopped()
 void TelemetrySimulator::onSimulateToggled(bool isChecked)
 {
   if (isChecked) {
-    timer.start();
+    startTelemetry();
   }
   else {
-    timer.stop();
+    stopTelemetry();
   }
 }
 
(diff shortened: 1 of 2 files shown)


mpaperno:
Setting RSSI automatically to zero when telemetry is disabled is a bad idea IMHO. This will have other effects on the simulated radio and will be new/unexpected behavior. Only way to send valid RSSI again after this is to enable telemetry again.  And telemetry loss can already be simulated now by setting RSSI value to zero manually.

raphaelcoeffic:
The funny thing is: when I just close the window (without this patch), all the value get red (stale), making you believe it is disabled, but it’s not. Setting RSSI to 0 is just making it consistent with the radio. Otherwise, you have a situation in the sim that does not exist in the radio, which is weird at best.

raphaelcoeffic:
Or how do you call it when no value are received at all, but telem is not lost?

raphaelcoeffic:
As it seems, the simulator always starts with telemetry disabled (RSSI=0).

- When the telemetry simulator is activated (press the knob, and then check "simulate"), it starts sending values including RSSI>0.

- When "simulate" is unchecked again (or window closed), all telemetry value are stalled (after a while), but it does not return to the initial state, which is not quite ok in my opinion. This state is practically never encountered on a real radio.

With this patch, the simulator is basically returned to the initial state as freshly started (RSSI=0), which corresponds to what happens with the radio (telemetry lost, as if you would unplugged the RX, which is the state the simu starts with).

mpaperno:
OK, perhaps I'm working with outdated info.  IIRC there was a time when w/out a valid RSSI other things in the radio were affected (don't remember what).  And there used to be a default RSSI sent at telemetry reset, but looks like that got (rightly) removed in 8eea072.

So I guess an automatic reset to zero on stop/close won't be that big a deal, and yes one could argue that it's "correct" behavior.  OTOH it changes current/established behavior and will raise the "telemetry lost" and/or low rssi warning(s), which may be annoying.  I'm not entirely sure I see a net benefit of this change (since zero RSSI can already be simulated), but I could live with it.

Would be good to get the GPS coordinates fix in though.  BTW, why not 7 decimals?  Or are none of the log formats that precise?

raphaelcoeffic:
I believe 6 digits is all we will get. Crossfire does not have it better than FrSky:
```
  int32_t value;
  switch(id) {
    case GPS_ID:
      if (getCrossfireTelemetryValue<4>(3, value))
        processCrossfireTelemetryValue(GPS_LATITUDE_INDEX, value/10);
      if (getCrossfireTelemetryValue<4>(7, value))
        processCrossfireTelemetryValue(GPS_LONGITUDE_INDEX, value/10);
```

`32bit / 10` is more something like 28.5 bits, right?

mpaperno:
Yea actually the precision will be dictated by the SPort telemetry value being sent, not the log format, and that's only good to 6 decimals at best... so never mind that.  :)

bsongis:
to be ported to 2.3

raphaelcoeffic:
Ok, will be done!

PR #6014, opened 2018-07-01
WIP: add “delay” attribute to LUA GPS data

raphaelcoeffic:
This allows for checking how old the GPS position is (when it has been received) which is quite helpful in scripts which rely on fresh GPS positions.

A new attribute `delay` has been added to the GPS table:
- If the GPS position is "fresh enough", it's value corresponds to the position delay (1=160ms).
- If the position is older (sensor lost), `nil` is used.

related to #5946.

### DIFF
--- a/radio/src/lua/api_general.cpp
+++ b/radio/src/lua/api_general.cpp
@@ -108,7 +108,7 @@ Return the time since the radio was started in multiple of 10ms
 @retval number Number of 10ms ticks since the radio was started Example:
 run time: 12.54 seconds, return value: 1254
 
-The timer internally uses a 32-bit counter which is enough for 30 years so
+The timer internally uses a 32-bit counter which is enough for 497 days so
 overflows will not happen.
 
 @status current Introduced in 2.0.0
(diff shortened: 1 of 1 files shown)


mpaperno:
The new attribute needs to be documented in `getValue()` docs.  The name, and unit it represents. A note about it being `nil` when "too old."

Perhaps "age" is a better name? The term "delay" can be used in GPS lingo to describe delay between satellite and Rx or a signal being received vs. reported.  "Age" usually means time since last report or signal.

Also, as we were discussing, I'm trying to think of "the best" way to expose this age attribute for all telem. values.

* `getFieldInfo` is probably too "heavy" (even if you could query by ID as well as name) with the chars and such.  But tempting because it's already there and overloaded (eg. with Unit).
* A dedicated `getTelemetryAge([field])` would be quickest and to the point, and when used w/out a parameter could return the global last-received age and/or TELEMETRY_STREAMING() result (or something else useful).
* It may be useful or more efficient in some applications to get all the current data for a sensor with one call, eg. current value, min, max, age/validity, unit, precision. So maybe `getTelemetryItem(id)` would be nice and still pretty fast.
* Ironically, sensor management in general is under-represented in Lua (there's API in models lib for a bunch of things like mixes and curves, but not sensors). Only the somewhat-awkward `setTelemetryValue` which serves 2 separate functions.  So, a set of functions to manage sensors properly, including something like `getSensorInfo()` could satisfy multiple needs.  But for fast checks inside a loop probably not as efficient as the 2nd and 3rd options.

raphaelcoeffic:
I was thinking about `age`as well and have no strong preference here.

What about having both `getTelemetryAge()` and `getTelemetryItem()`? I think they serve 2 different purposes:
- gather general information about a specific sensor (`getTelemetryItem()`, with a big table, called once per sensor / script)
- get the "validity" of value parallel to fetching it (`getTelemetryAge()`&`getValue()` will probably used together in some "inner" loop, every x milliseconds)

mpaperno:
Yea, none of the options I listed are mutually exclusive.

This is probably just semantics, but I think of `getTelemetryItem()` as all "transient" info (current value, age, min/max), even unit and precision could reasonably be changed by user mid-session.  I think `getSensorInfo()` (mostly for data stored in `TelemetrySensor`) would make more sense for the "run once at init()" usage scenario (would be quite useful actually). 

I agree with your 2nd point entirely though.  It's too bad we can't just change return type of `getValue()` for all Units :)  So maybe a new function like `getDetailedValue()` is in order.  Not sure what other data is should return besides value and age, but at least with a table return format it will be easily extensible in the future.   It could also optionally cover my "other" issue with `getValue()` with it returning zeros instead of actual (last known) value as shown in sensors list.

bsongis:
Still WIP? Seems good

OpenUAS:
@bsongis seems, but is not ;) hinted improvements/fixes to not implement so still WIP. Would be great if you could add the hinted improvments, than renew the pull. TIA, it is likely then indeed good :)

raphaelcoeffic:
@bsongis @mpaperno Sorry guys, I don't remember what the consensus was :-D Was it to have `getDetailedValue()` that would return the value and the extras (age and so)? So that `getValue()` stays as it is, right?

raphaelcoeffic:
@bsongis @mpaperno ping! need your input guys!

PR #6015, opened 2018-07-02
generate_datacopy.py fixes

raphaelcoeffic:
It seems there were quite some changes to the structures since the last time this script has run (2016). Not sure if this is correct or not, but it looks better to me now.

This fixes basically three issues:
- use the proper libclang.dylib (OSX specific)
- fix the empty copy functions (related to PACK() macro not defined properly)
- remove all the unrelevant structs coming from include files (llvm 6.0? should happend as well somewhere else).


bsongis:
thanks! would you port it to 2.3 ?

raphaelcoeffic:
Yes, no problem. I can do it in the next days.

3djc:
I have merged it :)

PR #6018, opened 2018-07-06
[2.3] telemetry simulator fixes/improvement

raphaelcoeffic:
This is basically the rebased version of #6013 on top of 2.3.

**Fix:**
- fixes GPS precision (6 digits now).

**Improvement:**
- send RSSI=0 when the telemetry simulator is stopped.

This changes the behaviour of the telemetry simulator a bit, and makes it a tad more consistent with the behaviour observed in the radio simulator with respect to telemetry.

When the radio simulator starts, no telemetry is received, which is equivalent to the physical radio when no receiver is plugged. When the telemetry simulator is started by checking the "Simulate" checkbox, telemetry packets are sent, and the radio simulator now behaves as if the receiver had been plugged.

When however the telemetry simulator is disabled by unchecking "Simulate", it would previously just stop sending telemetry packets, without setting RSSI to 0, so that the radio would think that the receiver was still plugged, whereby no RSSI telemetry packets were sent nor received. This does not really happen for real, as the FrSky modules do send RSSI packets even if the receiver is not plugged (however with 0 in case no receiver is present).

This patch basically mirrors the behaviour of the FrSky module by sending one last RSSI=0 packet when:
- telemetry simulator window is closed.
- "Simulate" is unchecked.

Please note that it will effectively trigger a "Telemetry lost" message, just like it would when you unplug the receiver. Previously, you would only hear "Sensor lost" after a while, but the telemetry was still official "ON", albeit no packets were received.

### DIFF
--- a/companion/src/simulation/telemetrysimu.cpp
+++ b/companion/src/simulation/telemetrysimu.cpp
@@ -447,6 +447,32 @@ void TelemetrySimulator::generateTelemetryFrame()
   }
 }
 
+void TelemetrySimulator::startTelemetry()
+{
+  timer.start();
+}
+
+void TelemetrySimulator::stopTelemetry()
+{
+  timer.stop();
+
+  bool ok = (ui && ui->rssi_inst);
+
+  int id = 0;
+  if (ok) {
+    id = ui->rssi_inst->text().toInt(&ok, 0);
+  }
+
+  if (!ok)
+    return;
+
+  uint8_t buffer[FRSKY_SPORT_PACKET_SIZE] = {0};
+  generateSportPacket(buffer, id - 1, DATA_FRAME, RSSI_ID, 0);
+
+  QByteArray ba((char *)buffer, FRSKY_SPORT_PACKET_SIZE);
+  emit telemetryDataChanged(ba);
+}
+
 uint32_t TelemetrySimulator::FlvssEmulator::encodeCellPair(uint8_t cellNum, uint8_t firstCellNo, double cell1, double cell2)
 {
   uint16_t cell1Data = cell1 * 500.0;
@@ -1055,24 +1081,5 @@ void TelemetrySimulator::LogPlaybackController::setUiDataValues()
     else {
       // file is corrupt - shut down with open logs, or log format changed mid-day
     }
-  }
-}
-
-void TelemetrySimulator::startTelemetry()
-{
-  timer.start();
-}
-
-void TelemetrySimulator::stopTelemetry()
-{
-  timer.stop();
-
-  bool ok;
-  uint8_t buffer[FRSKY_SPORT_PACKET_SIZE] = {0};
-  generateSportPacket(buffer, ui->rssi_inst->text().toInt(&ok, 0) - 1, DATA_FRAME, RSSI_ID, 0);
-
-  if (ok) {
-    QByteArray ba((char *)buffer, FRSKY_SPORT_PACKET_SIZE);  
-    emit telemetryDataChanged(ba);
   }
 }
--- a/companion/src/simulation/telemetrysimu.cpp
+++ b/companion/src/simulation/telemetrysimu.cpp
@@ -73,7 +73,7 @@ TelemetrySimulator::TelemetrySimulator(QWidget * parent, SimulatorInterface * si
 
 TelemetrySimulator::~TelemetrySimulator()
 {
-  timer.stop();
+  stopTelemetry();
   logTimer.stop();
   delete logPlayback;
   delete ui;
@@ -93,10 +93,10 @@ void TelemetrySimulator::onSimulatorStopped()
 void TelemetrySimulator::onSimulateToggled(bool isChecked)
 {
   if (isChecked) {
-    timer.start();
+    startTelemetry();
   }
   else {
-    timer.stop();
+    stopTelemetry();
   }
 }
 
@@ -1055,5 +1055,24 @@ void TelemetrySimulator::LogPlaybackController::setUiDataValues()
     else {
       // file is corrupt - shut down with open logs, or log format changed mid-day
     }
+  }
+}
+
+void TelemetrySimulator::startTelemetry()
+{
+  timer.start();
+}
+
+void TelemetrySimulator::stopTelemetry()
+{
+  timer.stop();
+
+  bool ok;
+  uint8_t buffer[FRSKY_SPORT_PACKET_SIZE] = {0};
+  generateSportPacket(buffer, ui->rssi_inst->text().toInt(&ok, 0) - 1, DATA_FRAME, RSSI_ID, 0);
+
+  if (ok) {
+    QByteArray ba((char *)buffer, FRSKY_SPORT_PACKET_SIZE);  
+    emit telemetryDataChanged(ba);
   }
 }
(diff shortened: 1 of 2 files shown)


companion/src/simulation/telemetrysimu.cpp
```diff
@@ -1057,3 +1057,22 @@ void TelemetrySimulator::LogPlaybackController::setUiDataValues()
     }
   }
 }
+
+void TelemetrySimulator::startTelemetry()
```
mpaperno: These should be positioned up with the other `TelemetrySimulator` class functions.

companion/src/simulation/telemetrysimu.cpp
```diff
@@ -1057,3 +1057,22 @@ void TelemetrySimulator::LogPlaybackController::setUiDataValues()
...
+
+void TelemetrySimulator::stopTelemetry()
+{
+  timer.stop();
+
+  bool ok;
+  uint8_t buffer[FRSKY_SPORT_PACKET_SIZE] = {0};
+  generateSportPacket(buffer, ui->rssi_inst->text().toInt(&ok, 0) - 1, DATA_FRAME, RSSI_ID, 0);
+
+  if (ok) {
```
mpaperno: Perhaps 
```
bool ok = (ui && ui->rssi_inst);
const int id = (ok ? ui->rssi_inst->text().toInt(&ok, 0) : 0);
if (!ok)
  return;
```
The pointer checks are a bit overly paranoid, but why not since it's called in destructor. But main point being the return before buffer creation/etc.

mpaperno: My comment is actually more applicable to line 1070.

companion/src/simulation/telemetrysimu.h
```diff
@@ -205,6 +205,8 @@ class TelemetrySimulator : public QWidget
         uint32_t encodeDateTime(uint8_t yearOrHour, uint8_t monthOrMinute, uint8_t dayOrSecond, bool isDate);
     };  // GPSEmulator
 
+    void startTelemetry();
```
mpaperno: These should also be moved up, into `TelemetrySimulator protected slots:`

mpaperno:
Seems good to me, just some cosmetics.  Assuming this doesn't cause any other "adverse" side effects, but we should have enough time with 2.3 to find out  :)  I think it's reasonable that telemetry is "lost" when telem simu is stopped/closed.

3djc:
" I think it's reasonable that telemetry is "lost" when telem simu is stopped/closed."
Hum, then we probably need to find a way to start it automaticaly with default values when simu is started, if not model with telemetry alarms will require telem been started manualy each time to stop the screeming

mpaperno:
It's been like this for quite some time (no telem until telem simu started). There used to be some code in the firmware that always sent RSSI and a few other values, but that's been removed for some months.  The RSSI thing I haven't validated 100%, but on a quick check I didn't see where/how it would be set to > 0 w/out telem simu active.  And I'm pretty sure telemetry alarms won't start screaming until some telemetry has been received, and then lost.

But yes, they will start "screaming" when one closes or disables the telem simu, vs. manually setting RSSI to zero... which is new.  And if there are any other side-effects to RSSI going to zero, this will bring them out.

raphaelcoeffic:
Having no telemetry on start is basically what happens when started without a receiver connected, which is the normal way.

Having RSSI set to 0 is the normal way when the receiver is disconnected. So I don’t see anything bad will happen that’s not already happening on the radio when you unplug the receiver. Try it out and find out for yourself ;-)

3djc:
i have my radio set to beep on !telem because there is not reason for the radio to be on without a receiver been on (I turn radio on, then receiver on obviously)? that's fine on the radio because that's why I turn my radio on for.

Simu on the other end, I do not use to fly, obviously, but to do différent things

mpaperno:
Exactly what @3djc said... "bad" is relative, if the simulator is annoying you with warnings while you're trying to set up something, that can be "bad" also.  Having to have the telem simu window open and active to prevent it could be rather annoying, I think... and a waste of resources. The telem simu runs in the GUI thread.

3djc:
I can see why some might want that (not sure we ever got a request for that, but anyway), maybe we could default it OFF (so that most users won't see a diference), with a way to enable that behavior either in CPN settings or on the telem simu window itself ?

raphaelcoeffic:
@3djc could you please explain me your proposal a bit more? I think I don't really understand what this is about.

3djc:
I mean leave it unchanged (like today) by default, and allow a  companion program settings to change it to this way

raphaelcoeffic:
@3djc I think it would be even better to have the "Simulate" checkbox be persistent (backed by some setting), and even continue generating the telemetry when the window is closed, as long as "Simulate" is checked. Otherwise I'm afraid it gets too complex, apart from the fact that I would have no idea how to explain the setting you're proposing.

3djc:
That would also be different from today, and result in other annoyance (ie announcing alt avery xx secs, ...)

raphaelcoeffic:
Ok, so for your particular use-case, you want/need to have telemetry on with no values being sent, otherwise you’ll get some announcement of some sort? Don’t you think this is a fairly uncommon case? Apart from that, you would have it by just clearing all the values in the dialog box.

mpaperno:
You could simply offer another checkbox right in telem simu window -- "disable telemetry/RSSI on stop/exit."  There's lots of room under the "Simulate" one. The checkbox could (now or later) be backed by a new persistent setting (in AppData or even per radio Profile).  It's now one a one-line macro to create new settings... ;)   Or at least it could be static for duration of session for now.

> and even continue generating the telemetry when the window is closed

The window IS the simulator (like I said, runs in GUI thread), plus this should be up to the user anyway IMHO.

>Don’t you think this is a fairly uncommon case?

Not at all.  And we have definitely had complaints before when lost telemetry alarms have gone off unexpectedly/annoyingly.

> you would have it by just clearing all the values in the dialog box.

That's awkward at best.

companion/src/simulation/telemetrysimu.cpp
```diff
@@ -447,6 +447,27 @@ void TelemetrySimulator::generateTelemetryFrame()
...
+void TelemetrySimulator::startTelemetry()
+{
+  timer.start();
+}
+
+void TelemetrySimulator::stopTelemetry()
+{
+  timer.stop();
+
+  bool ok = (ui && ui->rssi_inst);
```
mpaperno: Well there's not much point in setting a value here if you're not going to use it or actually care if the pointers are valid...  :)

raphaelcoeffic: Sorry, just took your code without checking, stupid me!

raphaelcoeffic: done

bsongis: why not write something like this here
```
if (!ui || !ui->rssi_inst)
  return;
```

it avoids the next 2 tests, or I forget something?

bsongis:
I think the other PR closes this one? If I am wrong please tell me!

mpaperno:
I think the [first commit](https://github.com/opentx/opentx/pull/6018/commits/41f4e45495b2a19434277a70837e0696c0d8461d) in this branch (the fix part) can/should be cherry-picked into 2.2 branch.  It won't break anything and it's definitely a good fix.  And certainly into 2.3 if nothing else.

mpaperno:
Cherry-picked the fix into 2.2, and merged into 2.3.

PR #6019, opened 2018-07-06
Allow minimizing the telemetry simulator window when un-docked.

mpaperno:
**Needs testing on Mac and Linux.**  Incomplete tag until tested.

### DIFF
--- a/companion/src/simulation/simulatormainwindow.cpp
+++ b/companion/src/simulation/simulatormainwindow.cpp
@@ -250,6 +250,8 @@ void SimulatorMainWindow::restoreUiState()
   toggleMenuBar(m_showMenubar);
   restoreGeometry(g.profile[m_radioProfileId].simulatorOptions().windowGeometry);
   restoreState(windowState, m_savedUiStateVersion);
+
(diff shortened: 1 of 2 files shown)


raphaelcoeffic:
Minimising works on Mac, but then I cannot resize the dialog anymore, which is quite a pain. Docking does not work anymore either.

mpaperno:
Can't resize which dialog?

Are you sure docking doesn't work?  You can dock by dbl-click on the title bar.  If you drag-to-dock then there needs to be enough space available in the main window, otherwise it won't show any available drop locations.

But maybe indeed this breaks it on Mac...

raphaelcoeffic:
@mpaperno the telemetry simulator dialog (or whatever this is). Concerning docking, you're right, still works, I just have to use the right thing! (double click)

mpaperno:
You mean it's no longer resizable when non-minimized?  At all, or until it's minimized the first time?

raphaelcoeffic:
neither nor.

mpaperno:
I think this is no longer necessary now that the telem. sim window can just be hidden w/out interrupting telemetry.

PR #6020, opened 2018-07-06
Add start/pause/RSSI-disable options to Telemetry Simulator

mpaperno:
image

There are tooltips to help explain the new options.  All 3 are saved to settings.  Default behavior is like before these changes (telem disabled, pause on hide, don't set rssi = 0).

The pause/no-pause option includes log replay.  So log can be playing with window hidden.

Also now makes sure simulation is enabled when log replay is started.

Ref #6018  (this replaces the "send RSSI=0 when the telemetry simulator is stopped" part).

### DIFF
--- a/companion/src/simulation/telemetrysimu.cpp
+++ b/companion/src/simulation/telemetrysimu.cpp
@@ -97,6 +97,9 @@ void TelemetrySimulator::showEvent(QShowEvent * event)
 
 void TelemetrySimulator::startTelemetry()
 {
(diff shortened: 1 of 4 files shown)


mpaperno:
I should mention that telemetry simulation will not start until window is shown for the first time, even if "Simulate" is enabled in persistent settings. So it should not be running in the background w/out user being aware.

Also the 3 checkbox options are per radio profile, not global.

PR #6021, opened 2018-07-08
2.2

Tiefflieger68:
Companion:
Display receiver number in the model overview of Companion.
It would be great to see the receiver number in the  model overview of Companion, behind the model name.
This may help to find a free Receiver ID before creating a new model.


schwabe:
This pull request seems to have gone completely wring

PR #6022, opened 2018-07-08
Full size R9M support

kilrah:


radio/src/pulses/pulses_arm.h
```diff
@@ -105,7 +105,7 @@ PACK(struct PxxUartPulsesData {
 #define MULTIMODULE_BAUDRATE         100000
 #define MULTIMODULE_PERIOD           7 /*ms*/
 
-#if !defined(INTMODULE_USART) || !defined(EXTMODULE_USART)
+#if !defined(INTMODULE_USART) || !defined(EXTMODULE_USART)|| defined(MODULE_R9M_FULLSIZE)
```
bsongis: one space missing

bsongis:
All good. With this option we only allow FULL SIZE modules.
It means a new option in Companion + update the scripts.
My vote goes for 2.3

kilrah:
It's actually incompatible with the PXX_FREQUENCY=HIGH needed for prod radios, still has to be fixed... :(

kilrah:
Rebased on 2.3 and completed.

sevet:
This option will only allow FULL SIZE module?
Can't it have an option for both modules? The modules are different and have different uses:
R9M lite being low power and low latency is good for Racing
R9M Full being higher power is good for Long range.
It will be a problem to have to choose on the firmware install level.

kilrah:
Yes becasue a ton of stuff changes based on module type, different protocol, different settings etc. Too much of a pain to maintain with a runtime setting.

You can have both firmwares on the SD card and reflash as needed if you really use both regularly.

SHPL00P:
Id be happy with JUST full module support as long as I have telemetry Ill be happy since I have already put together my module adapter. How does one go about testing?

3djc:
It will be widely available in 2.3 only

kilrah:
Until then you'd have to build this branch yourself.

DieHertz:
Looks like this is the answer to my question @kilrah

NickBozos:
When the 2.2.3 will be out? We will be able to use r9m on our x-lite tx?

3djc:
It probably won't be in 2.2.3, but you still have the option of PPM and switches to make it work if you absolutely want to.

NickBozos:
Thank you for the response. I think i will purchase the r9m lite . If it is true that the latency is better with the r9m lite , maybe for good.

3djc:
We are sending PXX for R9 mini at twice the rate of 'full size' R9 module. According to test others have made, it should have a positive impact on latency. Given we have no hard data I cannot say more than that.

NickBozos:
Thanks!

PR #6023, opened 2018-07-08
WIP: Improve latency for generated pulses

schwabe:
We run the mixer 2ms before setupPules and setupPules 2ms before
actually starting to send the pulses. The last 2ms are very pessimistic.
This commit changes that time to a more realistic but still more
than comfortable 150µs.

In my own tests the setup of pulses was always < 70 µs but be still
conservative here.

Fix multi sync rate to be always 7ms when no sync signal is present and
adjust mixer interval to sync rate.


schwabe:
http://plai.de/opentx/test-hb-sync.bin binary is built from this branch

This allows you to set the delay between setupPulses and hetbeat (in the model setup for the internal module). You probably need to experiment a bit to get a good value.

Syncing works fine and modifying the sync value is live: https://youtu.be/UHgvvqQkPLY

MikeBland:
"We run the mixer 2ms before setupPulses"
Please check this as, from my measurements, I saw about 7mS between starting the mixer and sending the pulses. The mixer, when I looked at the code, was run from the real time scheduler, which, I think, doesn't provide sufficient accuracy for the timing relative to sending the pulses.

schwabe:
@MikeBland the real time scheduler should give a 2ms resolution. It is not great but not introduce more than 2ms latency. It might be a tick off. Ersky9x runs the timer in a fixed 2ms interval, right?

From my test with debug timers we have:

Mix2IntPuls: 3.210ms - 4.537ms

together with mixer run time and those 2ms that are reduced to 150µs in this commit that adds up to 5 to 6,3 ms, so almost the 7ms that you are seeing.

bsongis:
@schwabe I think CooCos provides timers which have a far better precision than 500us. Also for PXX there is no reason to wait after the channels array is filled. Instead of having a mixer driven by pulses we could have pulses driven by mixer. The same for crossfire.

MikeBland:
I'm still using the same CoOs i have since I started using it and this using the systick 2mS time for software timers so only a 2mS resolution.
I think that if you want to synchronise the PXX frame to the heartbeat pulse, you will need to synchronise the mixer to the PXX frame.

@schwabe Yes, ersky9x runs the mixer every 2mS

maverikou:
How does one test and verify this without an oscilloscope?

bsongis:
2.3 now improves a lot the latency, using HEARTBEAT to be in sync with the internal module

PR #6036, opened 2018-07-11
Fixes #6032

### LINKED ISSUE #6032: OpenTx 2.2.2 RC4 - "Trim Idle Only" affects Elevator channel instead of Throttle channel
Tested 2.2.2 RC4 in Companion and on a X9D with the same results.

When the Trim Idle Only function is activated, it is affecting the operation of the assigned Elevator trim on the Elevator channel instead of the operation of the assigned Throttle trim on the Throttle channel.  Crossing the elevator and throttle trims still results in the Trim Idle Only function affecting only the operation of the assigned Elevator trim (TrmT) on the Elevator channel instead of the throttle trim and channel.

bsongis:


radio/src/tests/mixer.cpp
```diff
@@ -75,6 +75,12 @@ TEST_F(TrimsTest, throttleTrim)
   setTrimValue(0, THR_STICK, TRIM_MIN);
   evalMixes(1);
   EXPECT_EQ(channelOutputs[2], -1024);
+  // checks ELE sticks are not affected by throttleTrim
+  // stick max + trim min
+  anaInValues[ELE_STICK] = +1024;
+  setTrimValue(0, ELE_STICK, TRIM_MIN);
+  evalMixes(1);
+  EXPECT_EQ(channelOutputs[1], 1024-250);
```
bsongis: does this check fail without the fix?

3djc: It now does with the fix (applied this to 2.2 and it failled)

bsongis:
must be pushed to 2.3

3djc:
Done

PR #6037, opened 2018-07-12
Enable ~ 9 ms frame interval for the lower 8 out of 16 channels in D16

mikeller:
This changes the frame interval for the lower 8 channels of D16 with 16 channels from 18 ms to ~ 9 ms. It does so by taking advantage of the fact that in FrSky's over the air protocols, every channel out of the 8 channels in a frame is sent with a bit to indicate if it is one of the lower or upper 8 channels.

Making the assumption that most users will be using 8 or less channels for often changing stick / analog inputs, and the other channels for only rarely changing switches, in every PXX frame for upper channels, the output values for the upper 8 channels are compared to their last value, and the upper channel value is only sent if the upper channel has changed, with the lower channel value being sent instead if not. All upper channels will be sent once every second, to enable synchronisation on startup, and after events like RX resets.

Best case this will result in the lower 8 channels being sent with a 9 ms interval all of the time (bar one 18 ms interval per second). Worst case this reverts to the same behaviour that it had before this change, with channels 1 - 8 and 9 - 16 being sent intermittently, with an interval of 18 ms.

Before this change:

image 2

After this change:

image 1

Tested with XJT and XSR, and R9M and R9mini, and confirmed working on both.

### DIFF
--- a/radio/src/pulses/pxx_arm.cpp
+++ b/radio/src/pulses/pxx_arm.cpp
@@ -24,6 +24,8 @@
 #define PXX_SEND_FAILSAFE                  (1 << 4)
 #define PXX_SEND_RANGECHECK                (1 << 5)
 
+#define PXX_UPPER_CHANNEL_MAX_INTERVAL_FRAMES 110
+
 const uint16_t CRCTable[]=
 {
   0x0000,0x1189,0x2312,0x329b,0x4624,0x57ad,0x6536,0x74bf,
@@ -303,6 +305,10 @@ inline void putPcmTail(uint8_t port)
 
 inline void setupFramePXX(uint8_t port, uint8_t sendUpperChannels)
 {
+  static int lastUpperChannelValues[8];
+  static uint32_t frameCounter = 0;
+  static bool triggerSendUpperChannels = true;
+
   uint16_t pulseValue=0, pulseValueLow=0;
 
   initPcmCrc(port);
@@ -315,8 +321,10 @@ inline void setupFramePXX(uint8_t port, uint8_t sendUpperChannels)
 
   /* FLAG1 */
   uint8_t flag1 = (g_model.moduleData[port].rfProtocol << 6);
+  bool sendAllUpperChannels = false;
   if (moduleFlag[port] == MODULE_BIND) {
     flag1 |= (g_eeGeneral.countryCode << 1) | PXX_SEND_BIND;
+    sendAllUpperChannels = true;
   }
   else if (moduleFlag[port] == MODULE_RANGECHECK) {
     flag1 |= PXX_SEND_RANGECHECK;
@@ -335,6 +343,9 @@ inline void setupFramePXX(uint8_t port, uint8_t sendUpperChannels)
 
   /* FLAG2 */
   putPcmByte(port, 0);
+
+  triggerSendUpperChannels |= (frameCounter % PXX_UPPER_CHANNEL_MAX_INTERVAL_FRAMES) == 0;
+  frameCounter++;
 
   /* CHANNELS */
   for (int i=0; i<8; i++) {
@@ -375,18 +386,31 @@ inline void setupFramePXX(uint8_t port, uint8_t sendUpperChannels)
       }
     }
     else {
+      bool channelSent = false;
       if (i < sendUpperChannels) {
+        if (triggerSendUpperChannels) {
+          sendAllUpperChannels = true;
+          triggerSendUpperChannels = false;
+        }
+
         int channel = 8 + g_model.moduleData[port].channelsStart + i;
         int value = channelOutputs[channel] + 2*PPM_CH_CENTER(channel) - 2*PPM_CENTER;
-        pulseValue = limit(2049, (value * 512 / 682) + 3072, 4094);
+        if (sendAllUpperChannels || value != lastUpperChannelValues[i]) {
+          pulseValue = limit(2049, (value * 512 / 682) + 3072, 4094);
+          channelSent = true;
+        }
+        lastUpperChannelValues[i] = value;
       }
-      else if (i < NUM_CHANNELS(port)) {
-        int channel = g_model.moduleData[port].channelsStart + i;
-        int value = channelOutputs[channel] + 2*PPM_CH_CENTER(channel) - 2*PPM_CENTER;
-        pulseValue = limit(1, (value * 512 / 682) + 1024, 2046);
-      }
-      else {
-        pulseValue = 1024;
+
+      if (!channelSent) {
+        if (i < NUM_CHANNELS(port)) {
+          int channel = g_model.moduleData[port].channelsStart + i;
+          int value = channelOutputs[channel] + 2*PPM_CH_CENTER(channel) - 2*PPM_CENTER;
+          pulseValue = limit(1, (value * 512 / 682) + 1024, 2046);
+        }
+        else {
+          pulseValue = 1024;
+        }
       }
     }
 


kilrah:
> Best case this will result in the lower 8 channels being sent with a 9 ms interval all of the time (bar one 18 ms interval per second)

This is not acceptable. You're not considering the fact that that this single packet in a second with the upper channels may be lost.

schwabe:
What raises also a very worrying concern about Frskys own 
RF implementation. Unless there's some bit in the telemetry that tells the sender that a packet has been successfully received, the RF link might take very long for a channel update if frames are lost. And btw. hardcoding that channels 1to8 are more important than 9to16 feels just wrong

schwabe:
To really tweak our code we need some better documentation from Frsky. All this stabbing in the dark what their modules do becomes increasingly frustrating

3djc:
And I think @MikeBland has established that the module is already doing channel priority sorting : https://github.com/opentx/opentx/issues/4640#issuecomment-403535481

kilrah:
> What raises also a very worrying concern about Frskys own
RF implementation. Unless there's some bit in the telemetry that tells the sender that a packet has been successfully received, the RF link might take very long for a channel update if frames are lost. 

Why? it's just not expected or needed that people start doing stupid things like this. 

Not seeing where we're "stabbing in the dark". 
Indeed as Mike has documented just send all 16 channels to the module within the 9ms frame and it takes care of it.

MikeBland:
The assumption that channels 9-16 are only used rarely, and don't matter if they take a "long" time to update is wrong. If I'm flying with a S6R/S8R, and I select auto-level mode or knife edge mode or hover mode or change the gain (controlled by channels 9-11), then I want that to happen NOW, not a second later!
Simply send all 16 channels to the module every 9mS and let the module take care of things.
In passing, the RF protocol includes sequence numbers, with acknowledgement for the two-way SPort telemetry data, so the module and receiver do know if packets have been missed. Missing telemetry data is re-transmitted, I don't know if the channel control data is re-transmitted as a result of missed RF frames.

kilrah:
Retransmitting control data is pointless since it's outdated by the time it would be due to the very nature of a periodic transmission. But that's the whole point - for that to work it has to stay periodic.

mpaperno:
> Retransmitting control data is pointless

Potentially even dangerous or at least counter-productive... a frame out of place could easily be worse than a missed one.  Even re-transmitting stale telemetry data is questionable IMHO.  Unless it's done before the next packet gets through.  Otherwise the receiving side really needs some way to determine the original packet order.

mikeller:
@kilrah:

> This is not acceptable. You're not considering the fact that that this single packet in a second with the upper channels may be lost.

I assume that this is covered by the sequence id that is fed back to the TX module in 

https://github.com/betaflight/betaflight/blob/86b36b948861072306af419cd1ec3203c6d83c36/src/main/rx/cc2500_frsky_x.c#L385

If this is not the case, then your reasoning in

> Indeed as Mike has documented just send all 16 channels to the module within the 9ms frame and it takes care of it.

is just faulty and broken, as this will be just as susceptible to frames being lost, and updates never happening as my proposal.

@schwabe:

> And btw. hardcoding that channels 1to8 are more important than 9to16 feels just wrong

Can you explain why?

The hard reality is that in the protocol used on the CC2500 hardware, it is only possible to transmit 8 channels in a 9 ms slot. Assuming that the users want to have control over 16 channels, using the convention that 8 of them are higher priority, and allowing the users to allocate the functions with higher priority to these channels, certainly seems to be a better option compared to randomly selecting which channels are slowed down, and not giving the user the advantage of access to faster channels.

> To really tweak our code we need some better documentation from Frsky. All this stabbing in the dark what their modules do becomes increasingly frustrating

Totally agreed. Feel free to ask FrSky for a spec and post here.

@3djc:

> And I think @MikeBland has established that the module is already doing channel priority sorting : #4640 

This at least solves the problem of introducing an additional 9 ms of latency for no reason at all, like OpenTx is doing at the moment. I still prefer my approach of allowing the user to choose which channels get priority.

@MikeBland:

> The assumption that channels 9-16 are only used rarely, and don't matter if they take a "long" time to update is wrong.

Yes, agreed. But I don't see why you are making this utterly out of context comment here. As the code shows (and I am explaining in the pull request), changes to channels 9 - 16 are transmitted every 18 ms, i.e. with just as little latency as the 'dumb' approach that OpenTx is taking at the moment.

kilrah:
> I assume that this is covered by the sequence id that is fed back to the TX module in

That's only about telemetry, not control.

> is just faulty and broken, as this will be just as susceptible to frames being lost, and updates never happening as my proposal.

Nope, it's just doing a better "middle grounds" approach. It prioritises active channels but only so much, maybe 3-4 times in a row. Then it sends the other anyway to make sure it's sent often enough.

AKA like you but if it had a more sensible timeout of say 50-100ms rather than 1 second, and not treating channels unevenly.

MikeBland:
@mpaperno The telemetry transmission is a reliable connection. The telemetry data is sent in packets of 6 bytes with a sequence number. The sequence number is used to detect lost packets and a re-transmission of lost packets is requested. The packets are re-assembled into the correct order using the sequence number before onward transmission to the radio. A SPort packet is more than 6 bytes long.

Just give the module all 16 channels in a 9mS interval and let the module handle the sending. If you are only using the first 8 channels then these are sent as soon as they change. If more than 8, then, if possible, use the channels by careful choice. e.g. if one of channels 1-8 is used for "Gear", which won't change often use the corresponding channel in the 9-16 range for the function that is most critical.

schwabe:
I am closing this pull request as it is not a good idea. I will prepare something to send both pxx back to back every 9ms for the next version. That should give the same 9ms refresh and if there is some strangness involved with it, it is frsky to blame and not opentx.

mikeller:
@schwabe: Can you provide some explanation as to why this is a 'not a good idea'?

> I will prepare something to send both pxx back to back every 9ms for the next version.

This will result in upper / lower channels being given priority at random, based on timing of the frame being sent over the radio relative to the two PXX frames being sent. I do not think that this is what users want in most situations, they will want the sticks to _always_ have higher priority than some of the switches.

Can I suggest that you re-open this pull request, and we can continue the discussion here?

@kilrah:

> Nope, it's just doing a better "middle grounds" approach. It prioritises active channels but only so much, maybe 3-4 times in a row. Then it sends the other anyway to make sure it's sent often enough.

Good point. Do you have documentation from FrSky on this?

I have changed the max interval to account for this.

kilrah:
Search on RCGroups, Mike did a post on it from what he got from FrSky. 

> This will result in upper / lower channels being given priority at random, based on timing of the frame being sent over the radio relative to the two PXX frames being sent.

No, it is unrelated to timing, and does not prioritise the entire group of 8 but the one of each pair (1/9, 2/10...) that moves more. Which is the best and more versatile approach and easy to make the best of in the way Mike mentioned above.

mikeller:
@kilrah:

> No, it is unrelated to timing, and does not prioritise the entire group of 8 but the one of each pair (1/9, 2/10...) that moves more.

So let's assume that you have changes on channels 1 and 9, with channel 1 having a change of 10, and channel 9 having a change of 1000 (because it's a switch). You send the lower 8 channels, and then the TX module generates the next RF frame (because this timing is arbitrary), before the upper 8 channels are sent. So how should it work that the change of 1000 that according to you should be sent is actually sent, considering the TX module does not know about it yet?

3djc:
No, you miss the point. We will be sending the whole 16 chan more frequently to the module, so when the module is preparing to send data, it would have the data in your scenario

kilrah:
There are 2 groups 1-8 and 9-16. 8 channels are sent over RF, and each of those 8 can be from any group.
So RF might send 9, 2, 3, 12, 5, 14, 15, 8.

The module will have gotten all 16 from the radio every 9ms so it can make its choice.

mikeller:
@kilrah:

> The module will have gotten all 16 from the radio so it can make its choice.

Can you please explain how you make sure of that, assuming that the module can start sending the frame at any time? I.e. the module is not going to know a priori which of the two frames you are sending is the 'first' or 'second' one of two frames that you generated from the same data, it will just start sending after receiving an arbitrary frame.

kilrah:
The module provides syncing information that is going to be made use of. We'll send all 16 channels once per cycle in time for the module to send them.

mikeller:
@kilrah: Do you have a link to the specification for this synching information?

kilrah:
No, that is from direct discussion with FrSky. But again Mike has documented all his experimentations about it.

mikeller:
So is the SmartPort serial port used for synchronisation, or is there an extra wire needed for this? Do you have a link to what Mike has documented?

kilrah:
Extra wire, the heartbeat pulse on JR pin 2. 
No link, search for his posts on RCG.

mikeller:
Nice, thanks.

I still think taking control of what data is actually updated, and sending only this data in one frame is a better approach than sending redundant data and leaving it up to the TX.

If this is combined with synchronisation with the TX module, an even lower latency is achievable.

kilrah:
If you recreated the same thing but where the parameters could be adjusted and tweaked then OK. But certainly not your current PR.

schwabe:
@mikeller We don't have enough information from the module to that. In the worst case the module always sends the channels that we are currently not sending. (getting a 18ms instead 9ms delay for 16 channels). Also we prefer an approach that should work for everyone.

And if the module really does some selecting of the channels to send, it is better to have only the module to do the selection. Having both OpenTX and the module select channels to send might get you unpredictable results. 

In terms of latency, with 16ch you will have an extra 1-2 ms from sending 16ch instead of 8ch to the module. But if you want minimal latency you use only 8channels anyway.

Not that for the x-lite internal module, we already use that approach. 

Check out my schwabe/sync_heartbeat_pulses branch that contains the current work in progress.

mikeller:
@kilrah:

I am definitely happy to add more features to this - the initial work was just to show that it can be done and works.

What parameters do you have in mind for tweaking?

I would certainly want the interval to be adjustable (probably with a 1 - 110 range).

schwabe:
@mikeller I see no good advantage of your approach to always sending 16ch that would make me prefer your approach. Sending always 16ch is simple and robust and does leave any doubt of channel update rate. And if you really care about the 2ms introduced by sending 16ch instead of 8ch to the XJT module you should seriously consider using 8ch only or a better latency optimised system like Crossfire or AHDS2A

kilrah:
> What parameters do you have in mind for tweaking?

Max timeout, amunt of extra movement that triggers prioritizing,...

But as @schwabe says the case is so remote that it makes no sense us having to validate and maintain such a beast when we can just let the module do its job.

mikeller:
@kilrah: I have added the parameters that you asked for - hardcoded for now, will add them to the GUI once you're ok with them.

I can see a number of reasons why the firmware taking control over this instead of just relying on the underlying hardware to do it makes sense:
- configurability;
- less unneeded latency.

schwabe:
Configurability is a downside. If it not needed it is better. And trading 2ms for a lot of uncertainty and complexity is not worth it imo

MikeBland:
In my opinion doing something similar to that which I have in ersky9x now provides excellent results.
In one 4.5mS interval:
Run mixer
Send channels 9-16 to the module
In the next 4.5mS interval:
Run mixer
Send channels 1-8 to the module, synchronised to the heartbeat.

This sends all 16 channels to the module before it sends a RF frame, with channels 1-8 sent for minimum latency (closest to the heartbeat). The module then makes the decision whether a low channel or a high channel needs to be sent.

Where the hardware permits (serial XJT) the two PXX frames are sent closer together just before the heartbeat.

mikeller:
@schwabe:

> And trading 2ms for a lot of uncertainty and complexity is not worth it imo

Don't you think that that is a bit ironic, coming from you, @schwabe, considering what you are trying to do in https://github.com/opentx/opentx/tree/schwabe/sync_heartbeat_pulses has already a lot more complexity and risk for regressions than my proposal, and it's not even finished yet?

If you were to apply your statement objectively, and if it was only the points that you criticised that were relevant, then you'd have to say that https://github.com/opentx/opentx/tree/schwabe/sync_heartbeat_pulses needs to be abandoned, and my proposal used instead.

@MikeBland:

> In one 4.5mS interval:
> Run mixer
> Send channels 9-16 to the module
> In the next 4.5mS interval:
> Run mixer
> Send channels 1-8 to the module, synchronised to the heartbeat.

This still seems to be introducing 4.5 ms unneeded latency for the upper channels, plus run the mixer twice where only once is needed / desired. Why don't you change this to run the mixer only once, and send all channels to be updated in the second 4.5ms?

kilrah:
Because you can't send 16 channels to the module in 4.5ms. An 8-channel frame takes about 3,5ms.

Besides that, don't you realize that @schwabe 's branch you criticise is actually a prerequisite for all the options discussed here to be meaningful? If there is no sync it's completely pointless even bothering.

schwabe:
> This still seems to be introducing 4.5 ms unneeded latency for the upper channels, plus run the mixer twice where only once is needed / desired. Why don't you change this to run the mixer only once, and send all channels to be updated in the second 4.5ms?

Just sending one frame per 9ms interval is also known for introducing sometimes 9ms extra latency since the moule decides to always send the other frame.

mikeller:
@kilrah:

> Because you can't send 16 channels to the module in 4.5ms. An 8-channel frame takes about 3,5ms.

Since the radio can't send 16 channels in a 9 ms frame, this doesn't matter. All we need to do is send the 8 channels that will actually be sent.

> Besides that, don't you realize that @schwabe 's branch you criticise is actually a prerequisite for all the options discussed here to be meaningful? If there is no sync it's completely pointless even bothering.

I am not criticising his branch, I am just stating that his proposed changes introduce a lot more complexity than mine - he is the one who stated that he thinks that added complexity for reduced latency is a bad thing.

Also, I disagree. Both changes have the effect of reducing latency independent of each other. If both are done then we are getting close to the minimum possible latency for this part of the signal pathway.

schwabe:
> Since the radio can't send 16 channels in a 9 ms frame, this doesn't matter. All we need to do is send the 8 channels that will actually be sent.

Which we cannot predict or know since the XJT is a blackbox to us. Or can you tell me what 8ch they are going to be in a lab setup like your where only one channel is changing? As long as I don't have this information from Frsky I am very conservative about it. And beside FrskyOS also sends 16ch to the module on the Horus.

> I am not criticising his branch, I am just stating that his proposed changes introduce a lot more complexity than mine - he is the one who stated that he thinks that added complexity for reduced latency is a bad thing.

You either have not really looked at my branch or  you ignore that everything is neatly in [different commit for the different aspects of latency reduction](https://github.com/opentx/opentx/commits/schwabe/sync_heartbeat_pulses). And my approach is [actually reducing complexity](https://github.com/opentx/opentx/commit/1399770c87d818b3a275c91657865609a52281e9)

MikeBland:
"Since the radio can't send 16 channels in a 9 ms frame, this doesn't matter. "
Yes you can send 16 channels from the radio to the module in a 9mS frame, I have this working in ersky9x, and synchronised to the heartbeat.
For a single channel changing I have a latency from change on the radio to change on the SBUS output of between 14.4 and 23.4 mS. The 9mS variation is solely due to the time within the 9mS frame that the change actually happens.
We can't be certain of what the module does, but it definitely sends the same channel twice in consecutive RF frames. I set up a test where both channels 1 and 9 were changing slowly, but with channel 1 changing slightly quicker than channel 9.
The RF sent the channels as 19119191919119, so when necessary channel 1 pre-empted channel 9.

mikeller:
@MikeBland:

> "Since the radio can't send 16 channels in a 9 ms frame, this doesn't matter. "
> Yes you can send 16 channels from the radio to the module in a 9mS frame,

I think you misread my statement. When I said 'the radio can't send 16 channels in a 9 ms frame', the 'radio' I referred to is the CC2500 RF chip inside the TX module. At the settings that are used for FrSky, this hardware is only capable of sending one RF frame containing 8 channels inside a 9 ms slot. So my approach is to, instead of oversending data to the TX module, and incurring unneeded latency for that, only send the data that can effectively be sent over the radio link.

Issue #5144, opened 2017-08-09
Refactoring ComboBoxes in Companion


bsongis:
- have only one SwitchModel per model instead of one per tab
- only do the update work if something has changed

mpaperno:
Proposed new UI with collapsible grouping:

Sources:
![image](https://user-images.githubusercontent.com/1366615/41016780-9a91f192-691f-11e8-8dbb-92536921e6ba.png)

Switches:
![image](https://user-images.githubusercontent.com/1366615/41016820-cf20c3fc-691f-11e8-8ac1-7bb742af8b31.png)

:+1:  / :-1:  :question:    

I had this working quite a while ago already, but... versioning got out of hand at the time. But I can port it.  

The tree view combo box is a [custom widget](https://github.com/mpaperno/maxLibQt/blob/master/src/widgets/TreeComboBox.h) I made.  It also uses a custom [proxy model](https://github.com/mpaperno/maxLibQt/blob/master/src/itemmodels/GroupedItemsProxyModel.h) which automatically groups items based on criteria (like switch/source type in this case), to create a tree from flat data.  So it all fits in perfectly with Qt item models.

mpaperno:
> have only one SwitchModel per model instead of one per tab

Definitely +1 on this, the question I always had though was who is to own it.  The logical choice would be the ModelData, but we're using memcpy (and such) to copy and move it around... which is not really going to work with pointers and child object ownership (and definitely not as a QObject with signals/slots/etc which would be very handy).  

So perhaps we need something like a "ModelManager" class which would be responsible for handling ModelData and all associated metadata for a model.  I think this would be easier (and probably better) than refactoring ModelData into a QObject.  Especially if one goal is to keep ModelData parallel/reusable with the radio version.

PR #6038, opened 2018-07-12
Conitnued RawSource/RawSwitch data models refactoring

mpaperno:
* Create dedicated RawSource data model class, same as was already done for RawSwitches;
* Abstract the data model classes with a base class;
* Same filtered data model class now handles filtering for both RawSource and RawSwitch;
* Filter model made more efficient by creating an "is available" data role as the primary filter condition, allows dynamic filtering and no need to invalidate whole model;
* Prevent source/switch combo boxes from changing the saved value when data models change and an item is removed, eg. if a line was using a LS which is removed (the currentIndexChanged() event would fire and new, incorrect, value would be saved to ModelData);
* Each FlightModePanel, TimerPanel and TelemetryCustomScreen were using own copies of data models, now they share copies created in the parent window.
* Removes all AVR-specific exceptions when building the switch/source data models.

There's one subtle but potentially important usability change.  Say you had a function which uses a LS.  Then remove that LS.  Previously, the function which used that LS would either be reset to no switch, or, in some cases, be set to a different switch.  Now, the function data (the LS index) is NOT cleared -- the LS index remains as the configured switch for that function.  It does still show as a blank field (combobox) in Companion, as before.

The (possible) advantage is that if you then re-create that LS you deleted, the function definition is still valid, and will automatically use the new LS (and the CPN UI will update automatically to reflect this).  So if you eg. inadvertently reset an LS during editing, all the items which used that LS are still valid when you re-create it.

OTOH if you delete the LS and then create a new one in the same slot with an entirely different meaning, any functions which used that LS will still use it... which could of course be incorrect.  It's user-error in any case, just that now the consequence might be different.

Of course what we really need is a warning system for the user which would alert them about deleting items which are used in other places, and an overall check to make sure everything configured is valid (before saving or whatever).  But that is a different scope entirely.

Related task: #5144 

Next step is to only have one copy of the base data models instead of creating one in each edit window.  It would make most sense for these to be owned by ModelData (and GeneralSettings for CFs).  But they need some refactoring to make it safe to store pointers and manage the data model objects properly (the current "memset 0" and memcpy techniques don't cut it).  Another idea would be to create the data models in the editor dialog and pass them to each editor window along with all the other parameters already passed to them.  Latter might be simpler/safer option.


mpaperno:
Forgot to mention... each commit in this PR is atomic and tested separately.  I refactored the refactoring a couple times...  :)   Check individual commits for the evolution.  It might be easier to chase down any possible bugs if this is not squashed.

PR #6041, opened 2018-07-13
Fix setting x values < 0 for custom curvs (fixes #6026)

### LINKED ISSUE #6026: Please fix model.setCurve()
There are two issues with model.setCurve(), one with creating a Custom X curve and the other with creating a Fixed X curve.

The issue with creating a Custom X curve is that an "X" value of zero results in a value of -100%, but any  value above that, from 1 to 100, plots a point to the right of the Y axis, and any negative value results in a return of "5", meaning "x value not monotonically increasing." There appears to be no way to plot negative X values other than X = -100%.

The issue with creating a Fixed X curve is that Y values returned from model.getCurve() are indexed starting with zero, yet model.setCurve() requires Y values indexed starting with one. That means that copying a curve from one slot ("oldcurve") to another ("newcurve") requires adjusting the index like this:

for i = 1, curvepoints do
	newcurve.y[i] = oldcurve.y[i - 1]
end

This second issue is interfering with a project I'm working on because I don't know how you'll decide to fix it. Whether you fix model.getCurve() or model.setCurve(), so that they match each other, some scripts will be broken. 

For further details please see my post on RCCGroups:

https://www.rcgroups.com/forums/showthread.php?p=39907354&postcount=2656

schwabe:


radio/src/lua/api_model.cpp
```diff
@@ -860,9 +860,9 @@ Set Curve parameters
 @param curve (unsigned number) curve number (use 0 for Curve1)
 
 @param params see model.getCurve return format for table format. setCurve uses standard
- lua array indexing and array start at index 1
+ lua array indexing and arrays start at index 1
 
-The first and last x value must 0 and 100 and x values must be monotonically increasing
+The first and last x value must -100 and 100 and x values must be monotonically increasing
```
bsongis, line 865: must **be**?

bsongis: unless it's first and last value**s**

PR #6084, opened 2018-08-03
All macros for modules removed

bsongis:


bsongis:
Almost all automatic rename, only the pulses/modules.h file is new

Issue #5372, opened 2017-11-15
Flexible storage model


raphaelcoeffic:
The idea here is to make the config storage flexible, so that new options can be added without having to convert everything.

Some requirements out of the top of my head:
 - should stay binary as the storage space is tight.
 - should allow new features without eeprom format change

Anything else?

Some folks think that JSON would be good, @bsongis even has a draft prototype. Some other think that we should stick to binary to be able to support all the TXs. Maybe we could have something that can be serialized in different ways, depending on the TX (different codes for the same thing makes it harder to maintain however).

raphaelcoeffic:
@tadango @kilrah @bsongis @3djc  please feel free to comment and add your ideas. The sooner we have a consensus, the sooner I/we can start implementing it.

RealTadango:
What is the difference between binary and JSON? It's all just a bunch of bits. A small size is important so don't store values matching default. JSON is great for that but long and repeated variable names take up a lot of space. I don't see a simple solution for that right away.....

bsongis:
I already have 2 branches with this work in mind, one with JSON, the other with YAML
All the libs I found use too many malloc and I would like to find time to rewrite something from scratch, without one malloc. This would allow us to use this storage method on Taranis as well.
Do you want to give me an hand with this work?

raphaelcoeffic:
@bsongis I can try to! To do it properly without malloc will basically mean calling some functions while parsing that will deal with the objects, so that you do not need to store anything. So basically some kind of streaming parser? I happen to use JSON a lot for my work, but the other one would be fine as well.

schwabe:
Or just generate a generate a parser that does not need memory since it directly sotres the information inside the structs.

raphaelcoeffic:
@schwabe you mean inside the memory structs for settings? Well, basically means the parser only streams the payload and calls function pointers instead of storing any data itself and passing them for later analysis.

3djc:
Unless most of the code is rewritten, the data needs to be written in the current data structs, have a look in https://github.com/opentx/opentx/blob/2.2/radio/src/datastructs.h.

So in essence, data need to be read from JSON/YAML and put in the struct for model loading, and the other way arround for saving. But all the radio are different, so the receipees will be complex

raphaelcoeffic:
Ok, YAML is being worked on currently, please stay tuned for more.

raphaelcoeffic:
I meant dices, of course ;-)

raphaelcoeffic:
[...]

PR #6099, opened 2018-08-14
[WiP] YAML storage

raphaelcoeffic:
**Description**

This is a first draft implementation (in progress) of #5372.

It is basically based on a very thin parser which supports a small subset of the YAML language (see `radio/src/storage/yaml/yaml_parser.{h,cpp}`).

ATM the parser is used in the radio firmware to read and save the models list, radio settings and model data. The implementation works in two different ways for these different data type:
- models list:
   - here a very small handwritten layer (see `radio/src/storage/yaml/yaml_modelslist.cpp`) is used to read/write the `modelslist`.
- radio settings and model data:
  - a tree walker (`radio/src/storage/yaml/yaml_tree_walker.{cpp,h}`) is used travel through a statically allocated tree.
  - the statical tree itself to generated by a python script (`radio/util/generate_yaml.py`) which uses `clang` and it's python binding to parse and iterate through the structures abstract syntax tree, and then,
  - with the help of a template (`radio/util/yaml_parser.tmpl`), a set of cpp files are generated (`radio/src/storage/yaml/yaml_datastructs_{x10,x12s}.cpp`).

Now the question is: how will companion handle YAML?

~~One possibility is to use the same generator used for radio and model data, but on the corresponding structs in companion and with a different template, so that classes or methods are generated which can set the data structs in companion (mainly `GeneralSettings` and `ModelData`)~~.

Companion will use a normal YAML parser (probably `yaml-cpp`) and feed its structures from handwritten code. This allows far more flexibility where resources don't come at a premium.

**Notes**

So far, a data model change was required to be able to select the proper enum member to be output:
- the `theme` and `widget` `ZoneOptionValue` has been replaced by `ZoneOptionValueTyped`, which now includes the type (string, unsigned, etc...) of the value contained in the enum. This shifts quite some bits and will require a conversion.

**Memory**

It looks like there is lots of space left in the menu stack. My X10 reports something like ~2800 bytes untouched. I believe this is the same on Taranis targets, as the stack has the same size. That would mean that the YAML RAM needs fit without an issue in the stack (around 400 bytes, all in all). Should I still try to relocate some of it into the `reusableBuffer`?

I'm just not sure the reusableBuffer is always unused when models are loaded / saved (especially when the previous one is saved from the menu before the new one is loaded). The reusableBuffer usage is not quite documented (some reverse-engineering needed from my part)...


mpaperno:
Are you saying the data structure definitions are now dynamic, generated (for some radios) before the firmware is compiled?  Only for ModelData and GeneralData, or for all the sub-types within those as well?

And the other part of this change is the actual loading of dynamic data values to/from YAML files?

And they're combined into one large change because the YAML parser needs to know the structure of the data anyway?

If the data structs are still radio-specific then no, it will not work in CPN as-is since that needs to store all possible data for all radio types at the same time.  But if it's possible to have a "combined" template which produces every possible member, then perhaps yes.  

OTOH, if I'm understanding the first part correctly, not having valid definitions for all that data before building something (running the generator or whatever) is going to be potentially awkward.  But maybe they could be generated and committed with the rest of the code instead of doing it at build time?

raphaelcoeffic:
>Are you saying the data structure definitions are now dynamic, generated (for some radios) before the firmware is compiled? Only for ModelData and GeneralData, or for all the sub-types within those as well?

No, it's only how the parser reads/writes the data which is dynamically generated. The structures themselves stay as they are. The parser gets a "tree walker" and some call backs passed. As it iterates through the YAML document, it calls a couple of methods:
- `to_parent`: go up one level
- `to_child`: go down one level
- `to_next_element`: go to next array element (if existing).
- `find_node`: find an attribute within the current level.
- `set_attr`: sets the previously found attribute to some value.

Each of these call back methods get a context passed, which stores the walker's state. The walker and it's state depend on the document parsed. A very simple one is use for modelslist (nothing generated here) whereby a rather more complex one is used for the radio and model data struct on the radio (counts bit offset for each field and sets it based on bit offset, bit width and type).

>And they're combined into one large change because the YAML parser needs to know the structure of the data anyway?

see previous answer. I hope it answers that question as well.

>If the data structs are still radio-specific then no, it will not work in CPN as-is since that needs to store all possible data for all radio types at the same time. But if it's possible to have a "combined" template which produces every possible member, then perhaps yes.

I guess there will be a different generated file for CPN parsing based on the target radio. This can be achieved by using some ifdefs which are used only by the generator, whereby the generator is called once for every different data set. It is probably possible to reduce these further by generating some `if (board_type == [whatever]` within the generated part. Let's see if I can come up with something like that. ATM, it's only X10 and X12S, and they differ only by 1 or 2 fields.

>OTOH, if I'm understanding the first part correctly, not having valid definitions for all that data before building something (running the generator or whatever) is going to be potentially awkward. But maybe they could be generated and committed with the rest of the code instead of doing it at build time?

Yes, this is the way it is done right now. The files are generated by a separate make target, which is called manually. The generated files can/should be checked-in. Every time something changes in the data structures (or enums), these files should re-generated for the new values to be taken into account.

radio/src/storage/yaml/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
...
+
+get_property(dirs DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY INCLUDE_DIRECTORIES)
+foreach(dir ${dirs})
+  #message(STATUS "dir='${dir}'")
+  set(YAML_GEN_ARGS ${YAML_GEN_ARGS} -I${dir})
+endforeach()
+
+#message(STATUS "YAML_GEN_ARGS=${YAML_GEN_ARGS}")
+
+add_custom_target(yaml_data
```
raphaelcoeffic, line 37: @mpaperno This is a manual target. We might need a script that calls that target for each and every board which uses YAML. ATM it would be X10 and X12S.

raphaelcoeffic:
@bsongis I rebased on 2.3 and added/fixed a couple of stuff in case you want to have a look at it.

raphaelcoeffic:
here is how `radio.yml` looks like for now:

```yaml
version: 219
variant: 0
chkSum: 65535
vBatWarn: 66
txVoltageCalibration: 0
backlightMode: 3
trainer: 
   mix: 
    - idx: 0
      srcChn: 0
      mode: 2
      studWeight: 100
    - idx: 1
      srcChn: 1
      mode: 2
      studWeight: 100
    - idx: 2
      srcChn: 2
      mode: 2
      studWeight: 100
    - idx: 3
      srcChn: 3
      mode: 2
      studWeight: 100
view: 0
spareRadio: 0
fai: 0
beepMode: 0
alarmsFlash: 0
disableMemoryWarning: 0
disableAlarmWarning: 0
stickMode: 0
timezone: 0
adjustRTC: 0
inactivityTimer: 10
telemetryBaudrate: 0
splashSpares: 0
hapticMode: 0
switchesDelay: 0
lightAutoOff: 2
templateSetup: 0
PPM_Multiplier: 0
hapticLength: 0
beepLength: 0
hapticStrength: 0
gpsFormat: 0
unexpectedShutdown: 1
speakerPitch: 0
speakerVolume: 0
vBatMin: 67
vBatMax: 83
backlightBright: 0
globalTimer: 0
bluetoothBaudrate: 0
bluetoothMode: 0
countryCode: 0
imperial: 0
jitterFilter: 0
disableRssiPoweroffAlarm: 0
USBMode: 0
jackMode: 0
spareExtraArm: 0
ttsLanguage: "en"
beepVolume: 0
wavVolume: 2
varioVolume: 0
backgroundVolume: 1
varioPitch: 0
varioRange: 0
varioRepeat: 0
auxSerialMode: 0
switchConfig: 
 - sw: SA
   type: 3pos
 - sw: SB
   type: 3pos
 - sw: SC
   type: 3pos
 - sw: SD
   type: 3pos
 - sw: SE
   type: 3pos
 - sw: SF
   type: 2pos
 - sw: SG
   type: 3pos
 - sw: SH
   type: toggle
potsConfig: 
 - pot: S1
   type: without_detent
 - pot: 6POS
   type: multipos_switch
 - pot: S2
   type: with_detent
slidersConfig: 
 - sl: S3
   type: with_detent
 - sl: S4
   type: with_detent
 - sl: EXT1
   type: with_detent
 - sl: EXT2
   type: with_detent
currModelFilename: "model1.yml"
blOffBright: 20
bluetoothName: ""
themeName: "Default"
themeData: 
   options: 
    - idx: 0
      type: Unsigned
      value: 
         unsignedValue: 65535
    - idx: 1
      type: Unsigned
      value: 
         unsignedValue: 57603
ownerRegistrationID: ""
gyroMax: 0
gyroOffset: 0
```

raphaelcoeffic:
and here is a basic model:
```yaml
header: 
 - name: "\x0D\xF1\xFC\xFB\xF4\x1B\x1C"
   bitmap: ""
telemetryProtocol: 0
thrTrim: 0
noGlobalFunctions: 0
displayTrims: 0
ignoreSensorIds: 0
trimInc: 0
disableThrottleWarning: 0
displayChecklist: 0
extendedLimits: 0
extendedTrims: 0
throttleReversed: 0
beepANACenter: 0
mixData: 
 - idx: 0
   weight: 100
   destCh: 0
   srcRaw: I0
   carryTrim: 0
   mixWarn: 0
   mltpx: 0
   offset: 0
   swtch: NONE
   flightModes: 0
   delayUp: 0
   delayDown: 0
   speedUp: 0
   speedDown: 0
   name: ""
 - idx: 1
   weight: 100
   destCh: 1
   srcRaw: I1
   carryTrim: 0
   mixWarn: 0
   mltpx: 0
   offset: 0
   swtch: NONE
   flightModes: 0
   delayUp: 0
   delayDown: 0
   speedUp: 0
   speedDown: 0
   name: ""
 - idx: 2
   weight: 100
   destCh: 2
   srcRaw: I2
   carryTrim: 0
   mixWarn: 0
   mltpx: 0
   offset: 0
   swtch: NONE
   flightModes: 0
   delayUp: 0
   delayDown: 0
   speedUp: 0
   speedDown: 0
   name: ""
 - idx: 3
   weight: 100
   destCh: 3
   srcRaw: I3
   carryTrim: 0
   mixWarn: 0
   mltpx: 0
   offset: 0
   swtch: NONE
   flightModes: 0
   delayUp: 0
   delayDown: 0
   speedUp: 0
   speedDown: 0
   name: ""
expoData: 
 - idx: 0
   mode: 3
   scale: 0
   srcRaw: Rud
   carryTrim: 0
   chn: 0
   swtch: NONE
   flightModes: 0
   weight: 100
   name: ""
   offset: 0
   curve: 
    - type: 1
      value: 0
 - idx: 1
   mode: 3
   scale: 0
   srcRaw: Ele
   carryTrim: 0
   chn: 1
   swtch: NONE
   flightModes: 0
   weight: 100
   name: ""
   offset: 0
   curve: 
    - type: 1
      value: 0
 - idx: 2
   mode: 3
   scale: 0
   srcRaw: Thr
   carryTrim: 0
   chn: 2
   swtch: NONE
   flightModes: 0
   weight: 100
   name: ""
   offset: 0
   curve: 
    - type: 1
      value: 0
 - idx: 3
   mode: 3
   scale: 0
   srcRaw: Ail
   carryTrim: 0
   chn: 3
   swtch: NONE
   flightModes: 0
   weight: 100
   name: ""
   offset: 0
   curve: 
    - type: 1
      value: 0
thrTraceSrc: 0
switchWarningState: 153391689
rssiSource: 0
potsWarnMode: 0
moduleData: 
 - idx: 0
   type: TYPE_XJT_PXX1
   rfProtocol: 0
   channelsStart: 0
   channelsCount: 8
   failsafeMode: 0
   subType: 0
   invertedSerial: 0
   mod: 
      pxx: 
         power: 0
         receiver_telem_off: 0
         receiver_channel_9_16: 0
         external_antenna: 0
         fast: 0
inputNames: 
 - idx: 0
   val: "\x12\xEB\xFC"
 - idx: 1
   val: "\x05\xF4\xFB"
 - idx: 2
   val: "\x14\xF8\xEE"
 - idx: 3
   val: "\x01\xF7\xF4"
potsWarnEnabled: 0
screenData: 
 - idx: 0
   layoutName: "Layout2P1"
   layoutData: 
      zones: 
       - idx: 0
         widgetName: "ModelBmp"
      options: 
       - idx: 0
         type: Bool
         value: 
            boolValue: 1
       - idx: 1
         type: Bool
         value: 
            boolValue: 1
       - idx: 2
         type: Bool
         value: 
            boolValue: 1
       - idx: 3
         type: Bool
         value: 
            boolValue: 1
view: 0
modelRegistrationID: ""
```

raphaelcoeffic:
> I added in other places macro to only silence the warnings in the headers and not in the rest of opentx source code.

@schwabe And how does it look like? It seems something is missing for the `simu` target.

schwabe:
E.g. in mavlink.h:

```
#if __clang__
// clang does not like packed member access at all. Since mavlink is a 3rd party library, ignore the errors
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Waddress-of-packed-member"
#endif
#include "GCS_MAVLink/include_v1.0/ardupilotmega/mavlink.h"
#if __clang__
// Restore warnings about packed member access
#pragma clang diagnostic pop
#endif
```

raphaelcoeffic:
This is already merged into 2.4

PR #6107, opened 2018-08-17
Support for new R9M firmwares

3djc:


3djc:
Not happy with the status, pulling back

PR #6118, opened 2018-08-22
Support for non certified firmwares

3djc:
Companion firmware build option done, but companion handling for new types needs to be added


kilrah:
👍

3djc:
Many thx for your help Max :)

radio/src/CMakeLists.txt
```diff
@@ -46,6 +46,7 @@ option(TRACE_LUA_INTERNALS "Turn on traces for Lua internals" OFF)
 option(FRSKY_STICKS "Reverse sticks for FrSky sticks" OFF)
 option(NANO "Use nano newlib and binalloc")
 option(NIGHTLY_BUILD_WARNING "Warn this is a nightly build" OFF)
+option(FLEX_R9M "Add R9M options for non certified firmwwares" OFF)
```
mpaperno: Any objection to changing this to `MODULE_R9M_FLEX` so it groups nicely with `MODULE_R9M_FULLSIZE`?

3djc: Hum, full-size refers to an actual module type while flex to a firmware, so it would more something like FIRMWARE_R9M_FLEX if anything imho

mpaperno: True.  It's firmware of the module, don't want to confuse that with OTX firmware though (the other FIRMWARE_* options)...  maybe one of: `MODULE_[FW_]R9M_[FW_]FLEX[FW]`?  :)   I think it relates to "MODULE" at least, the most out of any other current prefixes.  When using a GUI (or any alpha-sorted list view) it's nice to have groups/organization.

bsongis: I would also vote for MODULE_R9M_FLEX or MODULE_R9M_FLEX_FW

radio/src/CMakeLists.txt
```diff
@@ -352,6 +353,10 @@ if(NIGHTLY_BUILD_WARNING)
   add_definitions(-DNIGHTLY_BUILD_WARNING)
 endif(NIGHTLY_BUILD_WARNING)
 
+if(FLEX_R9M)
+    add_definitions(-DFLEX_R9M)
```
bsongis: same here for the definition

bsongis (approved):
just the option / definition names to sort, but it's a nice job!

radio/src/CMakeLists.txt
3djc: So MODULE_R9M_FLEX_FW it is :)

Daniel-1276:
Hi developers, what is the best approach to get a stable opentx firmware including #6118 ? Do I have to manually build the firmware somehow? I understand this commit was addressed to 2.2 branch on 30.08.2018 , meaning after 2.2.2 release ... is there any "stable" nightly build for the 2.2.3 release that can be recommended? Many thanks for your support guys !

3djc:
This will be available as stable when we release 2.2.3 (no ETA on this). Anything in between (2.2.3 nightlies) should be fairly safe to fly however.

lukaszsuski:
Hi there, I also came up with the same idea as Daniel-1276 mentioned.
I wrote about this here: https://www.rcgroups.com/forums/showpost.php?p=40350801&postcount=33
The reason is that because the latest nightly that I used (2.2.3N681) was pretty unflyable for me - had moments where the aircraft didnt reacted to the inputs and one where it completly freaked out giving full right roll till it crashed. Going back to stock 2.2.2 seems ok till now. 
Do You think that it could be due to nightly opentx version? Or more r9m lite FW is to blame? (FS set to drop)

3djc:
There is no reason we can think off that would create that behaviour in current 2.2.3 nightlies. We purposefully have done only very small changes in 2.2 branch, putting all the bigger stuff in 2.3.

The changes to support flex r9m are purely cosmetics, ie putting the right description to flags sent to module, but in essence, the very same code runs in 2.2.2

We have been hearing ‘noise’ about various, not opentx related, issues when flex is used at high power 866mhz, but it remains unclear if it is a flex module firmware issue, hardware related, or some interference issues.

lukaszsuski:
Thanks for the reply. My case is exactly r9m lite 868MHz@100mw_telem_on.  
So I will try test more with 2.2.2+flex FW@100mw_telem_off and 2.2.2+flexr9m_patch+flex FW@100mw_telem_on. 
If I will have any meaningful results I'll update here :)
In the meantime can You only confirm that these compile flags are enough for xlite?
MODULE_R9M_FLEX_FW=ON HELI=OFF PCB=XLITE
I ask because my local bin file built with these is roughly 20kb larger than stock xlite fw from frsky.

3djc:
Well if you compile yourself, you need at least to ensure you are using the same compiler (gcc-arm-none-eabi-4_7-2013q3-20130916)

PR #6122, opened 2018-08-25
Fix mixer outputs monitor

### LINKED ISSUE #6064: Mixer outputs display clipping at about +-100%
Companion 2.2.2
Displaying  Mixer values in the new Mixerwindow is buggy...
eg. Output from the ElevMixer: +-200%
Moving the stick slow, the Mixer does not show  values more than about +- 102%.
Moving the stick very fast, you can see bigger values up to the expected +-200%
The values in the radios display are  correct...

mpaperno:
Wasn't updating properly unless channel value had also changed (fixes #6064).

(PR against 2.2, assuming there will be a 2.2.3. Will pick/merge to 2.3 afterwards/instead.)


kilrah:
Thanks!

PR #6124, opened 2018-08-25
Prevent page changes during value editing

### LINKED ISSUE #6123: It is possible to change UI page while range check is ON
On all radios

kilrah:
To be discussed. Fixes #6123 but also affects all other value changes... which I think is actually a good thing, I have pressed PAGE by mistake during editing more than once...

### DIFF
--- a/radio/src/gui/128x64/navigation.cpp
+++ b/radio/src/gui/128x64/navigation.cpp
@@ -682,6 +682,9 @@ void check(event_t event, uint8_t curr, const MenuHandlerFunc * menuTab, uint8_t
 #endif
 
       case EVT_KEY_LONG(KEY_PAGE):
+        if (s_editMode>0)
+          break;
+          
         if (curr > 0)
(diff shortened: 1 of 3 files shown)


mpaperno:
> I have pressed PAGE by mistake during editing more than once.

Me too.  :+1:  on change.

Can probably remove [this block](https://github.com/opentx/opentx/blob/2.2/radio/src/gui/128x64/view_main.cpp#L665) also.

kilrah:
Hmm this block should still be used for the hack module which if I remember well can only be bound at power up, which was done by holding the momentary switch when turning radio ON... but I'd be very rusty on that. Need to check a bit more in detail before throwing it out.

mpaperno:
Yes, you're right about the code block. It's used in pulses_avr and also for SKY9X
https://github.com/opentx/opentx/blob/2.2/radio/src/pulses/dsm2_arm.cpp#L143

PR #6131, opened 2018-08-28
[Lua] Optimize luaLoadScriptFileToState() a bit…

### LINKED ISSUE #6130: Seems OpenTX is doing more IO than needed with `luac` switch
**Describe the bug**
Loading pre-compiled lua script with `loadScript(FILE_PATH .. "config.luac", "b")` and getting the following debug output:

```
convertToSimuPath(): /SCRIPTS/TELEMETRY/iNav/config.luac -> C:/Dev/SD Card/SCRIPTS/TELEMETRY/iNav/config.luac
findTrueFileName(C:/Dev/SD Card/SCRIPTS/TELEMETRY/iNav/config.luac)
	found in map: C:/Dev/SD Card/SCRIPTS/TELEMETRY/iNav/config.luac
f_stat(C:/Dev/SD Card/SCRIPTS/TELEMETRY/iNav/config.luac) = OK
convertToSimuPath(): /SCRIPTS/TELEMETRY/iNav/config.lua -> C:/Dev/SD Card/SCRIPTS/TELEMETRY/iNav/config.lua
findTrueFileName(C:/Dev/SD Card/SCRIPTS/TELEMETRY/iNav/config.lua)
	found in map: C:/Dev/SD Card/SCRIPTS/TELEMETRY/iNav/config.lua
f_stat(C:/Dev/SD Card/SCRIPTS/TELEMETRY/iNav/config.lua) = OK
luaLoadScriptFileToState(/SCRIPTS/TELEMETRY/iNav/config.luac, b): loading /SCRIPTS/TELEMETRY/iNav/config.luac
convertToSimuPath(): /SCRIPTS/TELEMETRY/iNav/config.luac -> C:/Dev/SD Card/SCRIPTS/TELEMETRY/iNav/config.luac
findTrueFileName(C:/Dev/SD Card/SCRIPTS/TELEMETRY/iNav/config.luac)
	found in map: C:/Dev/SD Card/SCRIPTS/TELEMETRY/iNav/config.luac
f_open(C:/Dev/SD Card/SCRIPTS/TELEMETRY/iNav/config.luac, 1) = 06AC3E28 (FIL 0E6BF1C4)
f_close(06AC3E28) (FIL:0E6BF1C4)
```

**Expected behavior**
As the "b" specifies binary only and the file name is correct, I would expect only the following to happen:
```
f_stat(C:/Dev/SD Card/SCRIPTS/TELEMETRY/iNav/config.luac) = OK
f_open(C:/Dev/SD Card/SCRIPTS/TELEMETRY/iNav/config.luac, 1) = 06AC3E28 (FIL 0B65EDE4)
f_close(06AC3E28) (FIL:0B65EDE4)
```
Doing all this other IO seems unneeded and would seem to cause a huge slowdown which isn't at all necessary.  Maybe the `convertToSimuPath()` and `findTrueFileName()` is required and doesn't cause file IO, but it would seem that `f_stat()` is IO and the extra IO wouldn't be required.

It seems that when the `luac` build option is included in the firmware, additional IO is being done even when all the scripts are compiled, there's no source code files, and the `loadScript` "b" option is selected.  This results in a 30% slow-down when the firmware includes the `luac` build option compared to when the `luac` build option is not included.

The 30% slow-down is calculated by measuring fps.  With the `luac` build option unchecked, 15.0 fps was recorded.  When firmware was used with the `luac` option was checked, only 11.5 fps was recorded.

Compiling on the transmitter is usually a futile exercise as the limited memory can many times cause it to not be able to compile at all.  So, the `luac` option is of little to no use in my experience (instead, compiling with the companion or Travis are better options).  But, it also seems there's some significant overhead from the `luac` option which appears to be caused by additional and unneeded IO activities.

mpaperno:
… to avoid unnecessary f_stat() when certain mode flags are set (closes #6130).

If anyone sees more/better way to optimize this, I'm all ears/eyes.  The logic to decide which file(s) to stat/load can be a bit boggling.

Bottom line is that it is always going to be a bit slower when loading with `luac` enabled just due to the extra code involved in any case. In normal operation (eg. not inside a 30Hz loop) it should still be faster, at least for larger scripts, than loading plain-text source (this was demonstrated in my [perf. tests](https://github.com/opentx/opentx/issues/3318#issuecomment-264688157) when writing the `luac` feature, though of course things may have changed in the rest of OTX since then).  With the memory savings being the main attraction of course.

It would be really really nice to get some _real_ performance statistics between the different loading techniques... something with micros resolution.  I might try to dig up the old code I used for previous benchmarks and see if it's still relevant.... but there's probably an easier way using current debug timers (the high-res one should work).

Attached file has a simple test script which runs through all the possible mode combinations using `loadScript()` to load a small external function. Normal/default debug output can be checked to make sure  `luaLoadScriptFileToState` is doing the right thing.  Both files should go into `/SCRIPTS/TELEMETRY/` and `tLdScr.lua` should be configured as a telem. page. 

[tLdScr.zip](https://github.com/opentx/opentx/files/2326265/tLdScr.zip)

### DIFF
--- a/radio/src/lua/interface.cpp
+++ b/radio/src/lua/interface.cpp
@@ -401,11 +401,10 @@ static void luaDumpState(lua_State * L, const char * filename, const FILINFO * f
 */
 int luaLoadScriptFileToState(lua_State * L, const char * filename, const char * mode)
 {
-  if (luaState == INTERPRETER_PANIC) {
+  if (luaState == INTERPRETER_PANIC)
     return SCRIPT_PANIC;
-  } else if (filename == NULL) {
+  else if (filename == NULL)
     return SCRIPT_NOFILE;
-  }
 
   int lstatus;
   char lmode[6] = "bt";
@@ -421,13 +420,17 @@ int luaLoadScriptFileToState(lua_State * L, const char * filename, const char * 
   uint8_t extlen;
   char filenameFull[LEN_FILE_PATH_MAX + _MAX_LFN + 1] = "\0";
   FILINFO fnoLuaS, fnoLuaC;
-  FRESULT frLuaS, frLuaC;
-
+  FRESULT frLuaS = FR_NO_FILE;
+  FRESULT frLuaC = FR_NO_FILE;
   bool scriptNeedsCompile = false;
+  bool needRename = false;
   uint8_t loadFileType = 0;  // 1=text, 2=binary
 
-  memset(&fnoLuaS, 0, sizeof(FILINFO));
-  memset(&fnoLuaC, 0, sizeof(FILINFO));
+  // compile mode flag: 0=default, 1=force, 2=skip
+  const uint8_t cflag = (lmode[1] == '\0' ? 0 : strchr(lmode, 'c') ? 1 : strchr(lmode, 'x') ? 2 : 0);
+  const bool forceText = (cflag == 1 || lmode[0] == 't');
+  const bool forceBinary = (!forceText && !strncmp(lmode, "b", 2));
+  const bool skipBinary = (!forceBinary && !strchr(lmode, 'b'));
 
   fnamelen = strlen(filename);
   // check if file extension is already in the file name and strip it
@@ -439,19 +442,28 @@ int luaLoadScriptFileToState(lua_State * L, const char * filename, const char * 
   }
   strncat(filenameFull, filename, fnamelen);
 
-  // check if binary version exists
-  strcpy(filenameFull + fnamelen, SCRIPT_BIN_EXT);
-  frLuaC = f_stat(filenameFull, &fnoLuaC);
+  // can skip bin file chck if: compile is forced or compile is skipped and we're not loading a binary.
+  if (cflag != 1 && !(cflag == 2 && skipBinary)) {
+    // check if binary version exists
+    memset(&fnoLuaC, 0, sizeof(FILINFO));
+    strcpy(filenameFull + fnamelen, SCRIPT_BIN_EXT);
+    frLuaC = f_stat(filenameFull, &fnoLuaC);
+  }
 
-  // check if text version exists
-  strcpy(filenameFull + fnamelen, SCRIPT_EXT);
-  frLuaS = f_stat(filenameFull, &fnoLuaS);
+  // can skip src file check if: mode is "b" or compile is skipped and we're not loading the src version.
+  if (!forceBinary && !(cflag == 2 && !skipBinary)) {
+    // check if text version exists
+    memset(&fnoLuaS, 0, sizeof(FILINFO));
+    strcpy(filenameFull + fnamelen, SCRIPT_EXT);
+    frLuaS = f_stat(filenameFull, &fnoLuaS);
+    needRename = true;
+  }
 
   // decide which version to load, text or binary
   if (frLuaC != FR_OK && frLuaS == FR_OK) {
     // only text version exists
     loadFileType = 1;
-    scriptNeedsCompile = true;
+    scriptNeedsCompile = (cflag != 2);
   }
   else if (frLuaC == FR_OK && frLuaS != FR_OK) {
     // only binary version exists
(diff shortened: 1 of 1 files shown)


teckel12:
Nice work!  I'll look at creating a local development environment for OpenTX to test this PR.  I think in another time and place, we'd make a great team.  Maybe not today, but I have a rule on forgiving all overnight, so maybe tomorrow 😉

schwabe:
Another idea to fix the lua scripts causing audio problems might also to rate limit loadscript to 10/s. Good side effect it would also teach people to fix their scripts

teckel12:
@schwabe I would consider that a bad idea as a script may need to load 20 external scripts at init to reduce the memory used and keep within the memory limitations.

schwabe:
Let me phrase it different the effect would be that half of I/o is reserved for system functions like logging and audio and lua gets the other half. But it is just an idea.

teckel12:
@schwabe Or instead of being specific to just the loadScript function, throttle lua (both background and UI) if it uses over a certain percentage of I/O or CPU resources.

In another branch I had written code to limit the UI refresh rate to be 5fps (unless there was user input).  And to be quite honest, I couldn't tell the difference between 5fps and 20fps (the max) when it comes to telemetry.  It almost seems that OpenTX is allowing a lua script to run at 20fps, but telemetry doesn't get updated that frequently anyway.  So it's kinda a waste of CPU resources to be refreshing at 20fps when the data can't be any quicker than 5fps.

This could have to do with the speed of S.Port telemetry transmission, as it seems realistic that the protocol would be much slower than the loop speed.

Which actually gives me a good idea...  We should have a trigger in OpenTx that's exposed in lua when the transmitter receives one "cycle" of telemetry values.  This would allow a lua telemetry script to sync with this trigger and only refresh the screen when a new set of values are available (otherwise, it's displaying the same data maybe several times before it changes).

kilrah:
CPU and memory are already limited, only IO isn't.

> I couldn't tell the difference between 5fps and 20fps (the max) when it comes to telemetry.

We had told you that such high refresh rate was pointless already...

> We should have a trigger in OpenTx that's exposed in lua when the transmitter receives one "cycle" of telemetry values.

There is no such thing as a "cycle".  Individual values come as they come and their update frequency is both variable and determined by the sender and the available bandwidth. Most come between 1 and 5Hz.
It's the script developer's job to optimize his scripts in the most appropriate way for his use case by checking for availability/change as best.

teckel12:
@kilrah If you believe I was trying to get it to paint at 20fps, that's not at all the case.  It was only noticed when it was so slow that it caused lag.  I'm actually trying to get it to paint as slowly as possible, so it's quite the contrary.

There's more than one way to skin a cat.  I put the word "cycle" in quotes for exactly that reason.  Instead, you would look at the RSSI telemetry sensor and trigger based on that.

mpaperno:
> rate limit loadscript to 10/s. ... fix their scripts

Or they'll just use Lua `loadfile()` (et. al) instead and still blame OTX for being too slow...  :)   I think we'd have to limit it on some lower level to really be effective. And of course that would add more overhead either way.

OTOH if a script is causing delays and stuttering, people just shouldn't use it... it's a simple enough test... just don't run the script and see if problem is solved.  Lua has its limits (esp. low ones on the F2 MCUs), people need to realize that up front.

teckel12:
@mpaperno FYI, the script wasn't causing delays or stuttering for almost a month with several beta testers.  It wasn't till someone tried it with the `luac` option enabled that this problem surfaced.  Thanks for looking into it BTW.

mpaperno:
@teckel12 As explained to you several times, your script is already stressing the disk I/O (and several other factors) to their limits.  As explained AGAIN in the top post, **of course** using `luac` will slow down this process further, to where your SD thrashing if finally affecting things enough to notice.  To be very clear, this code change will NOT (and is not designed to) fix your scripts/issues.

The above is not up for discussion here, I will delete any replies.

teckel12:
@mpaperno As explained to you several times, my script is **no longer** stressing the disk I/O.  That was now two versions ago and I had already created a fix before I even posted an issue here.  Let me say this again, my script has been adjusted to **NOT** stress the I/O **BEFORE** posting an issue here.  The issue here was created to show an example to illustrate the bug in OpenTX.

Much of the stress came from the `luac` switch trashing the SD card's I/O when it wasn't needed as it wasn't respecting loadScript switches.  This is the reason I opened this issue here, so you all could be aware of a problem on your end.  Again, I had already fixed things on my end and was taking steps to accommodate the bugs with the `luac` option till such time a bug fix is released.

These are the facts, which you refuse to accept and keep insisting what was in the past is still in the present.  Delete the truth if you wish, but we both know it's the truth as the timeline doesn't lie.

teckel12:
1. Issue reported with Lua Telemetry
2. Problem isolated to OpenTX's `luac` option
3. Fixed Lua Telemetry to get around OpenTX `luac` bug
4. Posted issue with OpenTX reporting `luac` option bug
5. OpenTX contributors attack because their code is bug free
6. Finger pointing that it's not a bug in OpenTX, but lazy programming with Lua Telemetry
7. Pull request to fix OpenTX even though it's bug free and there's no way it could be a bug in OpenTX
8. In denial and history revisionist OpenTX contributors still point blame even through there's a pull request that fixes the bug in OpenTX

mpaperno:
Hold the door on this one... found a couple bugs (`Tx` doesn't try binary if source is missing, and `t` still loads binary if source is missing).  And may be able to optimize I/O a little more when the timestamps do not actually need comparing.

mpaperno:
Replaced by #6149.

PR #6145, opened 2018-09-02
Turn on backlight on power press (but not on startup)

### LINKED ISSUE #6089: XLite 2.2.2 power off during warnings - low light backlight and flickering
small cosmetic issue
start up, leave on throttle warning for a while 
the screen backlight goes off when is throttle warning still on display
press off button and keep to power off like normal
screen light to some slow backlight and flickering until shitched off

3djc:
This makes power button to be considerend as a key as far as backlight is concerned.
On startup, the screen stay off, as per design choice (prevent battery drain in backpack for exemple)
This fixes #6089

### DIFF
--- a/radio/src/targets/taranis/keys_driver.cpp
+++ b/radio/src/targets/taranis/keys_driver.cpp
@@ -149,7 +149,7 @@ void readKeysAndTrims()
     keys[index++].input(trims_input & i);
   }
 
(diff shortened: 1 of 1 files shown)


pitts-mo:
Is this commit also seeing Taranis x9d+ power switch state ON as an active key press?

more info:
I have recently updated from 2.2.2 to nightly 2.2.3N683, I prefer backlight on keys only for 5 seconds but have found backlight will no longer turn off unless backlight is distinctly set to off.

PR #6148, opened 2018-09-03
Add Lua file system operations

mpaperno:
Add `fileSystem()` function allowing various operations: File/folder status (exists/info), create folder, copy/rename/move/delete, and get a dir listing.

Had to fix up simufatfs a little to have it return more file info data (only returned name for dir listing), which also consolidates a little code.

Attached are example for the ref. guide (I'll commit that if this is OK'd) and a testing run-once-type script.

[fileSystem-example.lua.txt](https://github.com/opentx/opentx/files/2346418/fileSystem-example.lua.txt)

[tFlSys.lua.txt](https://github.com/opentx/opentx/files/2346422/tFlSys.lua.txt)

### DIFF
--- a/radio/src/lua/api_general.cpp
+++ b/radio/src/lua/api_general.cpp
@@ -1280,6 +1280,190 @@ static int luaResetGlobalTimer(lua_State * L)
   return 0;
 }
 
+/*luadoc
+@function fileSystem(operation, path [, to_path | get_file_info])
+
+Performs various file system operations.
+
+@param operation (string) (required) The action to perform. Valid operations are:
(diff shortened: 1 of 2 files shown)


mpaperno:
Refactored the new `luaFileOp()` a little (use switch/case) and figured out why `simufatfs ::f_unlink()` wasn't working for directories... so fixed that also.  Squashed on my end and force-pushed :smiling_imp:  to keep the merge clean.

0815Creeper:
I would also love to see this feature. It would open completely new ways for lua on opentx. Eg I missed a Texteditor for checklists and scripts on the radio itself, but with the option of listing files in a directory it would be possible to implement this using only a lua script.

PR #6149, opened 2018-09-03
Optimize luaLoadScriptFileToState()

### LINKED ISSUE #6130: Seems OpenTX is doing more IO than needed with `luac` switch

mpaperno:
Refactor and optimize `luaLoadScriptFileToState()` to avoid unnecessary f_stat() when certain mode flags are set (closes #6130).  Replaces #6131.

On this round I ran more tests and a bunch of statistics.  Many stats.  I have prepared a [gist with all the results](https://gist.github.com/mpaperno/6a981df2ec0c825c2b6fd2e80c169de0).

Attached file has a simple test script which runs through all the possible mode combinations using loadScript() (this one runs as as a FUNCTION).  BUT you'll need #6148 to have it work.  (This is used to delete/rename files during the tests.)  The results of the test are posted in the gist above.

[tLdScr.zip](https://github.com/opentx/opentx/files/2346435/tLdScr.zip)

### DIFF
--- a/radio/src/lua/api_general.cpp
+++ b/radio/src/lua/api_general.cpp
@@ -1200,15 +1200,24 @@ then the loaded script (or "chunk") is returned as a function. Otherwise, return
    * Add `d` to keep extra debug info in the compiled binary.
        Eg: "td", "btd", or "tcd" (no effect with just "b" or with "x").
 
-@notice
-  Note that you will get an error if you specify `mode` as "b" or "t" and that specific version of the file does not exist (eg. no .luac file when "b" is used).
-  Also note that `mode` is NOT passed on to Lua's loader function, so unlike with loadfile() the actual file content is not checked (as if no mode or "bt" were passed to loadfile()).
-
 @param env (integer) See documentation for Lua function loadfile().
 
 @retval function The loaded script, or `nil` if there was an error (e.g. file not found or syntax error).
 
 @retval string Error message(s), if any. Blank if no error occurred.
+
+@notice
+  Note that you will get an **error** if you specify `mode` as `b` or `t` and that specific version of the file does not exist (eg. no .luac file when `b` is used).
+
+@notice
+  Mode `b` is **optimized** for the fastest loading time by skipping any file checks and will _not_ check for, or compile, the source version at all.
+  Similarly, `tx` mode is also the fastest way to load the source version.  Both of these are comparible in terms of performance to using `loadfile()` directly.
+
+@notice
+  The `mode` parameter is ignored when `luac` is disabled by firmware build options. Whatever is specified in `file` is loaded directly.
+
+@notice
+  Also note that `mode` is NOT passed on to Lua's loader function, so unlike with `loadfile()` the actual file content is not checked (as if no mode or "bt" were passed to `loadfile()`).
 
 @status current Introduced in 2.2.0
 
--- a/radio/src/lua/interface.cpp
+++ b/radio/src/lua/interface.cpp
@@ -368,7 +368,24 @@ static void luaDumpState(lua_State * L, const char * filename, const FILINFO * f
   } else
     TRACE_ERROR("luaDumpState(%s): Error: Could not open output file.", filename);
 }
-#endif  // LUA_COMPILER
+#endif  // defined(LUA_COMPILER)
+
+#define LDSCRPT_MODE_b        0x01
+#define LDSCRPT_MODE_t        0x02
+#define LDSCRPT_MODE_T        0x04
+//#define LDSCRPT_MODE_s      0x08  // reserved
+#define LDSCRPT_MODE_x        0x10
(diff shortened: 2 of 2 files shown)


teckel12:
@mpaperno I hope this makes it into OpenTX in 2.2.4 as these are solid improvements in certain cases. 

I do have a question on your results in the section that compares loadScript vs. loadfile. It seems that loadScript "b" is slightly faster than loadfile? I'm using loadfile as it would seem loading a binary with loadfile would be the fastest. But maybe I'm mistaken in how it works.

Basically, for the fastest possible execution (and lowest memory usage) with binary files, should I be using loadScript(filename, "bx") or loadfile(filename)?

mpaperno:
The results do show a tiny (couple µs) performance gain of this new `loadScript('b')` vs. `loadfile('b')` with binary files, and they're consistent across all tests so I believe that's correct. I had noticed it also and tried it different ways to be sure. My best guess (w/out looking at the Lua library code in several years) is that `loadfile()` has some additional overhead, perhaps verifying the file type, vs. `loadScript()` doesn't check anything with `b`. That might also explain why the difference on a larger file is even less significant.

The docs could be clearer on this (the example for `x` is misleading), but `bx` is same as just `b` since that never checks for the `.lua` file at all (in both old and new versions). You can see how in the first set of results none of `b, b-nb, bx` will look for ("stat") the `.lua` version (even when binary doesn't exists, as with `b-nb` test).

teckel12:
@mpaperno Thanks for the confirmation. And for the effort on this.

PR #6155, opened 2018-09-05
Refactor print center beep

elecpower:
Replace hard coded per radio mappings with mapping based on each radio's capabilities. Should allow settings for new radios to be printed without needing to change this code block.


mpaperno (approved):
:+1:

bsongis:
Excellent

PR #6158, opened 2018-09-07
Allow power-off while model checklist is displayed during startup.

mpaperno:
I've got no way to test it with `PWR_BUTTON_PRESS`.  Seems to work fine otherwise.

This does nothing for Simu[lator] as of yet (it will still hang), but I do have another patch coming up for that.


3djc:
👍

bsongis:
Yes, good catch

PR #6159, opened 2018-09-07
Refactor simu startup and shutdown handling

mpaperno:
* Use `pwrCheck()` as primary control point for shutdown, centralizes thread control in `simpgmspace` module;
 * Remove `main_thread_running` in favor of separate variables for startup mode and shutdown flag;
 * Don't start a thread for ARM startup, just run `simuMain()` directly (the thread just exits anyway);
 * Refactor `SIMU_SLEEP()` macro, and remove all unnecessary uses of it, also remove `SIMU_SLEEP_NORET`;
 * Add `startType` param to `opentxStart()` to control splash and startup checks;
 * Fixes backlight never turning off in simulator with "SWITCH" power button type (`pwrPressed()` was always `true`);
 * Fixes simu shutdown when model checklist is displayed at startup;
 * Adds stubs to possibly simulate a "soft" power button in the future.

There's a bunch of annoying AVR exceptions... I will make a cleaner one for 2.3.

### DIFF
--- a/radio/src/definitions.h
+++ b/radio/src/definitions.h
@@ -72,8 +72,7 @@ typedef __int24 int24_t;
 #else
   #define FORCEINLINE inline __attribute__ ((always_inline))
   #define NOINLINE    __attribute__ ((noinline))
-  #define SIMU_SLEEP(x)
-  #define SIMU_SLEEP_NORET(x)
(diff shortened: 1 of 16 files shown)


bsongis:
Only pushed to 2.3

PR #6174, opened 2018-09-18
Allow shutdown while reading model notes

3djc:
This fixes #6158

### DIFF
--- a/radio/src/gui/480x272/menus.cpp
+++ b/radio/src/gui/480x272/menus.cpp
@@ -80,7 +80,7 @@ void readModelNotes()
 
   clearKeyEvents();
   event_t event = EVT_ENTRY;
-  while (event != EVT_KEY_FIRST(KEY_EXIT)) {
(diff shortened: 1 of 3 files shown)


mpaperno:
:confused:    #6158 is a PR.  Or does it not work?

Regarding `pwrPressed()`, as mentioned, we also have this macro in opentx.h
```
#if defined(PWR_BUTTON_PRESS)
  #define pwrOffPressed()              pwrPressed()
#else
  #define pwrOffPressed()              (!pwrPressed())
#endif
```
So seems like at least `pwrOffPressed()` should be used?  IMHO the whole thing is a bit confusing, esp. with that macro well hidden in the huge header.

Other similar situations seem to use `pwrCheck()`, including the alerts routine.  It would be good to standardize this IMHO.

<s>BTW model notes are not shown when switching models, not sure if that's on purpose for some reason or a bug.</s> EDIT: now I can't recreate it... sorry.

3djc:
Duplicate

PR #6180, opened 2018-09-19
Companion AVR removal - stage 1.

mpaperno:
So far the main change is removal of AVR radio types from the selectable radios list in Profiles.  The base AVR firmwares are kept to allow import/conversions from those radios.

This adds a popup warning msg if selected profile uses an unavailable radio type (both at startup and when switching profiles). The warning simply prompts the user to update the profile... nothing is changed until user edits/deletes it.  At this point user could technically still open/edit an AVR radio file, though this will presumably need to change going forward.

Otherwise mostly refactoring stuff.


bsongis:
I didn't review everything in detail, but it seems good

Issue #5679, opened 2018-02-09
In Companion - Radio Protocol setting for Multiprotocol Module has no Custom option  


jtaylor2:
Companion 2.2.1 downloaded Windows and self compiled 2.2.2 on openSUSE 42.3, radio type x9d

External Radio Module - DIY Multiprotocol Module - Multi Radio Protocol has no "Custom" option in Companion.
Radio and Simulator are OK.

If a model with the Custom protocol is opened in Companion the Multi Radio Protocol shows "FlySky" and the Sub Type list displays Subtype 0 to 7.

I didn't try writing back to the radio to see what happens, but on the Simulator if you read  a model with Mode Multi Custom 0 0 into Companion and write it to a bin it changes it to FlySky Std.

jtaylor2:
This issue still exists in Companion 2.2.4

PR #6183, opened 2018-09-22
Multiprotocol listing updates for 2.2

MRC3742:
Multiprotocol listing updates for both Radio UI and Companion.
Changes, updates to bring OpenTX Multi listings in line to current Multiprotocol Module version 1.2.1.0

Compiled, tested and working using v2.2.3 on X9D+

This PR does not address problems with saving "Custom" settings into Companion and no "Custom" option in Companion listed in issue #5679


LapinFou:
Since the development of the Multi project will never be align with the timeline of the OpenTX development, I got a suggestion:
Would it be possible to store the protocol information on the SD card?
Thus, updating the multi module with new protocols will not require a new OpenTX version. Only updating a text file will be require.

schwabe:
Yes would be possible but I never got around writing the code. And if I implement something like that, the module should send the protocol information and not the SD card

3djc:
> the module should send the protocol information and not the SD card

Indeed !

LapinFou:
Fair point.
I will post an enhancement request on Pascal's GitHub.

LapinFou:
After a 2nd thought, maybe it would be difficult to do it now without breaking the ascendant compatibility...

Here is the ticket I opened → https://github.com/pascallanger/DIY-Multiprotocol-TX-Module/issues/193

LapinFou:
Some news on https://github.com/pascallanger/DIY-Multiprotocol-TX-Module/issues/193
Looks like Pascal is not against the idea. 😄

schwabe:
@LapinFou Yeah. But I have currently not much time for OpenTx development. The Multimodule side is the easy one.

LapinFou:
Well, as long as everyone like the idea, there is some good chance that something will be done one day.
That is the most important. 😄

schwabe (approved):

MRC3742:
Thank You for the approval as I was confident this was complete and tested when PR was made.

For anyone with a better understanding of commits,
Is there a way to change multiple files using one commit ?
This PR looks like a lot of commits but each translation/*.txt file was just a copy, paste, commit.
Using the web interface I had to commit on every file change or file update.
I believe that these can still be squashed into one commit before it is merged by maintainers ?
Thanks - Rich

kilrah:
You cannot edit multiple files using the web interface, you need to clone locally and use the "normal" git workflow.
We will squash, no issue.

MRC3742:
Thanks - I'll try that approach in the future.

Does another PR need to be submitted for this to be added to the 2.3 branch? or possibly are you hoping to have that version updated from the MPM via serial as is being suggested.

radio/src/translations/de.h.txt
```diff
@@ -115,7 +115,7 @@
 #define TR_DSM_PROTOCOLS       "LP45""DSM2""DSMX"  //LP45 =Low Power 4 und 5 Kanal
 
 #define LEN_MULTI_PROTOCOLS    "\006"
-#define TR_MULTI_PROTOCOLS     "FlySky""Hubsan""FrSky\0""Hisky\0""V2x2\0 ""DSM\0  ""Devo\0 ""YD717\0""KN\0   ""SymaX\0""SLT\0  ""CX10\0 ""CG023\0""Bayang""ESky\0 ""MT99XX""MJXQ\0 ""Shenqi""FY326\0""SFHSS\0""J6 PRO""FQ777\0""Assan\0""Hontai""OLRS\0 ""FS 2A\0""Q2x2\0 ""Walk.\0""Q303\0 ""GW008\0""DM002\0""Cabell""Esky15""H8 3D\0""Corona""CFlie\0"
+#define TR_MULTI_PROTOCOLS     "FlySky""Hubsan""FrSky\0""Hisky\0""V2x2\0 ""DSM\0  ""Devo\0 ""YD717\0""KN\0   ""SymaX\0""SLT\0  ""CX10\0 ""CG023\0""Bayang""ESky\0 ""MT99XX""MJXq\0 ""Shenqi""FY326\0""SFHSS\0""J6 PRO""FQ777\0""Assan\0""Hontai""OpnLrs""FS 2A\0""Q2x2\0 ""Walk.\0""Q303\0 ""GW008\0""DM002\0""Cabell""Esy150""H8 3D\0""Corona""CFlie\0""Hitec\0""WFly\0 ""Bugs\0 "
```
bsongis, line 118: I wonder if we really need to translate the protocols list, are there any differences? If no I propose to move them in translations.h

bsongis:
Needed in 2.3

radio/src/translations/de.h.txt
schwabe: No probably not.

PR #6221, opened 2018-10-13
Jumper T12 board

schumixmd:
Added Jumper T12 as a spearate board.
The definition in many cases is isolated from the Taranis X7 or Taranis Xlite so that it doesn't interfere with original board code. 
The navigation is implemented with Taranis Xlite code without Shift key.
The simulator is in progress.


companion/src/firmwares/boards.cpp
```diff
@@ -63,6 +63,8 @@ uint32_t Boards::getFourCC(Type board)
     case BOARD_TARANIS_XLITE:
       return 0x3978746F;
     case BOARD_TARANIS_X7:
+      return 0x3678746F;
```
bsongis, line 66: no

schumixmd:
pull request will not be merged

schwabe:
@schumixmd I think @bsongis was mainly refering to the fact that each board should have its unique id.

schumixmd:
@schwabe @bsongis @kilrah I may reopen the PR any time, but I see no reason to keep it open if the PR will not be pulled in 2.2. 
If not - please close! If yes - let's discuss what do make sense and what not.

companion/src/firmwares/boards.cpp
schumixmd: @bsongis Not sure about the identified format. 0x**3A**78746F fits?

PR #6229, opened 2018-10-28
Horus Widget BattCheck: updated look & feel

offer-shmuely:
bigger values of the important cell-value/percentage on every widget size
View is visible even while selecting the widget (instead of warn message)
move the values inside the battery picture, to enable larger font

### DIFF
--- a/radio/sdcard/horus/WIDGETS/BattCheck/main.lua
+++ b/radio/sdcard/horus/WIDGETS/BattCheck/main.lua
@@ -19,7 +19,7 @@
 local shadowed  = 0
 
 local options = {
-  { "Sensor", SOURCE, 1 },
+  { "Sensor", SOURCE, 0 },
   { "Color", COLOR, WHITE },
   { "Shadow", BOOL, 0 }
 }
@@ -28,12 +28,24 @@ local options = {
 local function create(zone, options)
   local myZone  = { zone=zone, options=options, counter=0 }
   histCellData = {}
+
+  -- use default if user did not set, So widget is operational when selecting the widget
+  if myZone.options.Sensor == 0 then
+    myZone.options.Sensor = "Cels"
+  end
+
   return myZone
 end
 
 -- This function allow updates when you change widgets settings
 local function update(myZone, options)
   myZone.options = options
+
+  -- use default if user did not set, So widget is operational when selecting the widget
+  if myZone.options.Sensor == 0 then
+    myZone.options.Sensor = "Cels"
+  end
+
 end
 
 -- A quick and dirty check for empty table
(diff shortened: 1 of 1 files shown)


offer-shmuely:
Allow to see the value while selecting a widget type:

battcheck2-show-value-on-widget-select

offer-shmuely:
Comparison of original and new look on all widget size:

battcheck-vs-battcheck2

3djc:
It is a good proposal, but I have two issues :
- you  change the 'main focus' from total lipo voltage to lowest cell. The two are valid ways, and surely we could debate forever the pro and cons of each,the fact remain that this will push a change on users that might not want it if we replace default battcheck with this

- you default to cells. For a user that is familiar with the whole OpenTX/telemetry system, it might sound good, but for them it is no big deal to setup the widget. For one that is barely understanding radio basic, he/she may think it is the actual transmiter bat, there is nothing telling what the default did pick up. Also, this defaulted entry does not follow sensor renaming. If you rename your cels sensor, it does not follow like it should

radio/sdcard/horus/WIDGETS/BattCheck/main.lua
```diff
@@ -177,40 +195,46 @@ end
...
+  lcd.setColor(CUSTOM_COLOR, getPercentColor(percent))
+  lcd.drawGauge(zone.zone.x+2, zone.zone.y+2, myBatt.w-4, zone.zone.h, percent, 100, CUSTOM_COLOR)
+  -- write text
+  if isHaveData then
+    lcd.setColor(CUSTOM_COLOR, zone.options.Color)
+    local topLine = string.format("%2.1fV      %2.0f%%", mainValue, percent)
+    lcd.drawText(zone.zone.x + 20, zone.zone.y +2, topLine , MIDSIZE + CUSTOM_COLOR + shadowed)
   else
-    lcd.drawText(zone.zone.x, zone.zone.y+10, "No FLVSS sensor data", LEFT + SMLSIZE + INVERS + CUSTOM_COLOR + shadowed)
+    lcd.drawText(zone.zone.x + 10, zone.zone.y+10, "No Voltage Data", LEFT + SMLSIZE + INVERS + CUSTOM_COLOR + shadowed)
```
3djc: If you point the widget to RxBt, there is voltage info, just not the right type, maybe no cells data is better ?

radio/sdcard/horus/WIDGETS/BattCheck/main.lua
```diff
@@ -177,40 +195,46 @@ end
 
 --- Size is 160x30 1/8th
 local function zoneSmall(zone)
-  local myBatt = {["x"]=0, ["y"]=0, ["w"]=75, ["h"]=28, ["segments_w"]=15, ["color"]=WHITE, ["cath_w"]=6, ["cath_h"]=20}
+  local myBatt = {["x"]=0, ["y"]=0, ["w"]=155, ["h"]=35, ["segments_w"]=25, ["color"]=WHITE, ["cath_w"]=6, ["cath_h"]=20}
   local mySensor = getCels(zone.options.Sensor)
+  local isHaveData = (type(mySensor) == "table")
```
3djc: isHaveDate is not a nice wording, maybe isDataAvailable ?

radio/sdcard/horus/WIDGETS/BattCheck/main.lua
offer-shmuely: good comment, will be changed

radio/sdcard/horus/WIDGETS/BattCheck/main.lua
offer-shmuely: correct, will be changed

offer-shmuely:
image

offer-shmuely:
all 3 comments are fixed.
additional suggestion?

my next goal:
1. I like to try and add indication to the lowest cell value (on other view than the full screen)m but we need to consider a way to handle the screen real-estate
2. keep the view visible when the receiver (and telemetry turned off), so I can inspect the info later, this will need a proper way to indicate that the voltage widget is not operative

3djc:
If you look at value widget, it shows 'non live data' in another color. Maybe not only change the color of the text , but also the batt fill (I would keep the cells color since it allows to detect early battery issues) ?

3djc:
Very nice indeed

PR #6233, opened 2018-10-30
Add string type to lua widget inputs

### LINKED ISSUE #5545: Add user customizable text that can be entered by the user in the widget options in a custom lua widget.
Usefull for labels ect .
The option structure must be made flexible .

local options = {
  { "Source", SOURCE, 1 },
  **{ "Text", STRING, "Self Level On" },**
  { "Color", COLOR, WHITE },
  { "Shadow", COLOR, BLACK }
}

3djc:
Allow thinghs like

```
local options = {
  { "Option1", SOURCE, 1 },
  { "Option2", VALUE, 1000 },
  { "Option3", COLOR, RED },
  { "Option4", STRING, "NAME" },
  { "Shadow", BOOL, 0 }
}
```

I'm unsure if STRING or TEXT should be used tho

This closes #5545

### DIFF
--- a/radio/src/lua/api_general.cpp
+++ b/radio/src/lua/api_general.cpp
@@ -1360,6 +1360,7 @@ const luaR_value_entry opentxConstants[] = {
   { "SHADOWED", SHADOWED },
   { "COLOR", ZoneOption::Color },
   { "BOOL", ZoneOption::Bool },
+  { "STRING", ZoneOption::String },
   { "CUSTOM_COLOR", CUSTOM_COLOR },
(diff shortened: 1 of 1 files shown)


offer-shmuely:
Cool, that will help also for widget like BattCheck,
So the default source can be “Cels” (that is usually the common telem name)

fabioeh:
Thank you for implementing. My original request #5545 was the same.

3djc:
Done in 2.3

wimalopaan:
Doesn't work in companion/simulator 2.3.9

PR #6242, opened 2018-11-11
quad wizadr images for @3djc

mhotar:


3djc:
Got them and using them !

PR #6244, opened 2018-11-14
Per model themes

mortehoe:
I have been running 'yaapu frsky telemetry script' which requires a background theme, however ,according to the developer running the background from within the script causes it to run too slow.The developer states the only way is to wait for opentx to implement per model themes.


kilrah:
Invalid pull request, use an issue

Issue #6246, opened 2018-11-16
FlySky Telemetry for VOLTS type is actually signed


CapnBry:
**Describe the bug**
I have connected FlySky receivers through an iRangeX 4-in-1 module. Normally the values look good (RX voltage, RSSI, etc) but when I added telemetry from iNAV, negative values show as their twos-compliment unsigned versions. "Altitude" from iNAV which is calibrated to 0m on startup shows 655.3m if the model is lowered slightly. "GPS Lat MSB" and "GPS Lon MSB" show similarly wonky numbers for South latitudes and West longitudes, respectively (e.g. 572.72 for -82.64).

Standard AFHDS2A protocol transmits all values as 2 raw bytes and a stock FS-i6 transmitter shows voltage type values properly as signed. I traced it back through the whole chain and looks like the raw data is just forwarded from the multiprotocol module for interpretation in opentx. I believe that all VOLTS type values are actually 16 bit signed, and all RPM type values are 16 bit unsigned.

I should reiterate that this isn't iNAV specific, stock FlySky transmitters properly show negative telemetry values for voltage telemetry types. The multiprotocol module doesn't interpret the data either, it just forwards it, so fixing the interpretation is in opentx's court.

The fix is simple, to just cast the int32 value as int16 without altering the bits.
https://github.com/opentx/opentx/blob/2.2/radio/src/telemetry/flysky_ibus.cpp#L106
~~~
diff --git a/radio/src/telemetry/flysky_ibus.cpp b/radio/src/telemetry/flysky_ibus.cpp
index 7f54c6594..980b158ce 100644
--- a/radio/src/telemetry/flysky_ibus.cpp
+++ b/radio/src/telemetry/flysky_ibus.cpp
@@ -103,6 +103,9 @@ static void processFlySkySensor(const uint8_t *packet)
       else if (id == FS_ID_TEMP)
         // Temperature sensors have 40 degree offset
         value -= 400;
+      else if (sensor->unit == UNIT_VOLTS)
+        // Voltage types are signed 16bit integers
+        value = (int16_t)value;
       setTelemetryValue(TELEM_PROTO_FLYSKY_IBUS, id, 0, instance, value, sensor->unit, sensor->precision);
       return;
     }
~~~

[...]

PR #6252, opened 2018-11-20
Display FlySky VOLTS values as signed integers

CapnBry:
As discussed in #6246 

Standard AFHDS2A protocol transmits all values as 2 raw bytes
and a stock FS-i6 transmitter shows voltage type values properly
as signed. This change interprets all UNIT_VOLTS FlySky telemetry
items as being signed 16-bit values so their display matches the
values displayed on stock FS-i6 and FS-i6X receivers.

Signed-off-by: Bryan Mayland <bmayland@capnbry.net>


schwabe (approved):

schwabe:
This should also be cherry picked in 2.3 when merged.

PR #6256, opened 2018-11-28
PXX1.5

bsongis:
Preparation for upcomming PXX changes


radio/src/targets/taranis/telemetry_driver.cpp
```diff
@@ -127,7 +127,7 @@ extern "C" void TELEMETRY_DMA_TX_IRQHandler(void)
   if (DMA_GetITStatus(TELEMETRY_DMA_Stream_TX, TELEMETRY_DMA_TX_FLAG_TC)) {
     DMA_ClearITPendingBit(TELEMETRY_DMA_Stream_TX, TELEMETRY_DMA_TX_FLAG_TC);
     TELEMETRY_USART->CR1 |= USART_CR1_TCIE;
-    if (telemetryProtocol == PROTOCOL_TELEMETRY_FRSKY_SPORT) {
+    if (telemetryProtocol == PROTOCOL_TELEMETRY_FRSKY_SPORT || telemetryProtocol == PROTOCOL_TELEMETRY_PXX2) {
```
bsongis, line 130: hm. no, this is not anymore the right place for this

OpenUAS:
Nice, but what about squashing and add at least a PR description, would make it a bit easier to honor PR IMHO

3djc:
In this particular case, we use the PR mechanism more as a colaboration tool than anything else, but yes, I'll add some comments

PR #6286, opened 2019-01-06
Reflow CRSF script splash for X7

nmaggioni:
The Crossfire LUA script has a really bad UI on the X7, most of the devices' settings go off screen to the right. This PR fixes the splash screen (_"Waiting for Crossfire devices..."_) only.

screen-2019-01-06-160542


3djc (approved):
Thanks !

kilrah:
Thanks, good start. Will merge when the rest is fixed

PR #6289, opened 2019-01-07
Fr Translation

LapinFou:
If it is not too late for the release 2.2.3, here is the latest French update (only 1 translation missing).


LapinFou:
I didn't see that 2.2.3 was already out.
So, I'm definitely too late. 😭

3djc:
yes sir, désolé :(

Issue #5088, opened 2017-07-12
Add RX number to the model list in Companion


pointu:
Can u add the rx number in the model list ?

agibson2:
That would be a nice feature to quickly see what is available.  Another idea might be to somehow indicate which numbers are in use by models with a * next to them or something when changing a rx number for a model.

pointu:
Bonjour,

En congès jusqu'au 28 aout 2017, Laissez votre message en cas d'urgence extrême je vous joindrais.

Jean-Jacques Pointel
Directeur
Phone : +33 (0)1.79.97.24.00

PR #6296, opened 2019-01-11
Add RX numbers to CPN models list

elecpower:
Enhancement #5088
Lists RX # even if not displayed/editable in model edit. This allows residual configuration to be visible which may be the cause of issues. Residual configuration possible from changing RX type and copying models.

\+ means # referenced by more than one model
\* means # references non-existent model

screenshot from 2019-01-11 21-19-20

screenshot from 2019-01-11 21-19-52

### DIFF
--- a/companion/src/mdichild.cpp
+++ b/companion/src/mdichild.cpp
@@ -534,7 +534,6 @@ void MdiChild::initModelsList()
   connect(modelsListModel, &QAbstractItemModel::dataChanged, this, &MdiChild::onDataChanged);
 
   ui->modelsList->setModel(modelsListModel);
-  ui->modelsList->header()->setVisible(!firmware->getCapability(Capability::HasModelCategories));
   if (IS_HORUS(board)) {
     ui->modelsList->setIndentation(20);
     // ui->modelsList->resetIndentation(); // introduced in next Qt versions ...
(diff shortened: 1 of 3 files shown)


kilrah:
Nice... what do you mean by

> * means # references non-existent model

elecpower:
@kilrah eg X9D+ Index 04 has RX# 06 but that slot is empty so no bound receiver

kilrah:
That's kind of irrelevant, and doesn't say anything about whether a receiver is bound or not

notlebowski:
I welcome this feature a lot, but can't you use red for a double reference and leave it at that?
Thanks for making this

elecpower:
@kilrah agree pointing to a valid model does not confirm an RX is bound to it. Pointing to an empty model is worth highlighting IMHO as for non-categorised radios you may have to scroll to to find the linked model. Also for some users certain digits can look the same at first glance. 
@notlebowski see next commit for red.

elecpower:
screenshot from 2019-01-13 08-31-22

bsongis:
Do we really want this one in 2.2? It's in 2.3, no?

elecpower:
If 2.2 is now bug fix only then I have not problem with closing this PR

bsongis:
Yes, thanks!

PR #6297, opened 2019-01-11
Check receiver ID only in D16 and R9 modes

### LINKED ISSUE #6194: Model ID in use message in D8 mode
OTX 2.2.2 x9d+

Message "model id already in use" (sorry for bad translation using german version) appears after successful bind of a D8 receiver.

- Copy a model with D16 mode
- Switch mode to D8 and bind
- After bind messagebox with "model id already in use" appears but there is no possibility to change or remove the receiver number.
- Also reproducible after powercycle of TX and rebind.

3djc:
Fixes #6194


3djc:
Getting a much better fix

PR #6308, opened 2019-01-22
Enhanced Italian language sounds

mzambrini:
I modified some Italian language sounds that were not consistent.

I also added three absent translations


3djc:
Grazzie mille !

PR #6321, opened 2019-02-05
Fix negative gvars precision

### LINKED ISSUE #6320: (minus)GV's with x.x precision returns 10 times bigger value in mixer
(OpenTX 2.2.3)
Example:
add mixer to CH1 as MAX and weight of -GV1, where GV1 has set x.x precision and value 10 and (for example) conditioned by SA^.
By swiching SA to ^ channel value becomes -100% (not -10%)
(The same example with +GV1 works fine.)

Bug can be found in Companion simulator X9D+, Horus, and bug is on real X9D+ too.

elecpower:
Fixes #6320


bsongis:
Thanks it is perfect now

PR #6326, opened 2019-02-11
Mirror crossfire telem on serial

### LINKED ISSUE #5357: Option to mirror CRSF telemetry to the Serial port in the battery tray on Taranis X9D
Smartport telemetry can be mirrored without a problem. Is it possible to implement the mirroring of CRSF protocol? I might be able to implement the functionality. Where should I start?

3djc:
Output fixed at 115k (but fifo buffer is protected in case of overflow)

Users tested ok

This fixes #5357

### DIFF
--- a/radio/src/telemetry/crossfire.cpp
+++ b/radio/src/telemetry/crossfire.cpp
@@ -181,8 +181,7 @@ bool isCrossfireOutputBufferAvailable()
 
 void processCrossfireTelemetryData(uint8_t data)
 {
-
-  #if defined(SERIAL2)
+#if defined(SERIAL2)
   if (g_eeGeneral.serial2Mode == UART_MODE_TELEMETRY_MIRROR) {
     serial2Putc(data);
   }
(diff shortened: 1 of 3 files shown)


ropalet:
Is this new function has been implemented in the official version of opentx?

3djc:
No, not yet. It should be part of 2.2.4 or 2.3

radio/src/telemetry/crossfire.cpp
```diff
@@ -181,6 +181,13 @@ bool isCrossfireOutputBufferAvailable()
 
 void processCrossfireTelemetryData(uint8_t data)
 {
+
+  #if defined(SERIAL2)
```
bsongis: cosmetics: too many spaces at the left

radio/src/telemetry/crossfire.cpp
```diff
@@ -181,6 +181,13 @@ bool isCrossfireOutputBufferAvailable()
 
 void processCrossfireTelemetryData(uint8_t data)
 {
+
```
bsongis: no newline here

bsongis (approved):
I think it's ok to merge it in 2.2 once the cosmetics comments are solved
I would suggest you make another PR for 2.3

Lance0724:
How do I use it with Jumper T16?
When I open AUX_SERIAL define in CMakelists then compile code(Version2.3.13) got comlipe error.

PR #6331, opened 2019-02-15
Bug fix #6324

### LINKED ISSUE #6324: 9XR pro ppm2 not working
Hi i m trying to use second PPM output from 9xr pro but not able to output any signal from this port.
How output is configured in openTX? Do i need to add pull up resistor or some kind of stuff?

regards

BF

blairfancy:
This will solve issue #6324 PPM2 was not properly initialized for SKY9X based radio.
I m not aware of all rev of PCB especially REVA perhaps it is necessary to adjust some IFDEF for that

### DIFF
--- a/radio/src/pulses/modules.h
+++ b/radio/src/pulses/modules.h
@@ -71,10 +71,10 @@ inline bool isModuleCrossfire(uint8_t idx)
 }
 #endif
 
-#if defined(EXTRA_MODULE)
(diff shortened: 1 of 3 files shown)


bsongis:
Needs to be rebased, sorry!

bsongis:
Merged, thanks (the whole Sky9x board would need some tests if you have time) !

PR #6343, opened 2019-02-27
Fix Companion model edit SF changing adjust GV# not resetting mode

### LINKED ISSUE #6342: Companion model edit special functions changing function Adjust GV# mode not reset
Changing the adjusted GV# which had a mode other than Value leaves the old mode unchanged but the parameter is a value. eg Adjust GV1 mode Source param SC after modifying to Adjust GV2 mode Source param 0 which is an invalid combination.

sf adjust gv bug

**General information**
 - OpenTX version 2.2.3
 - Hardware Frsky X9D+
 - OpenTX Companion 2.2.3 on Ubuntu 16.04 and Win 10

elecpower:
Fixes #6342


bsongis:
Needed in 2.3, right?

elecpower:
Yes needed in 2.3

PR #6349, opened 2019-03-01
Model print and compare enhancements

elecpower:
Add global functions if enabled for model
Add checklist contents if display enabled


3djc:
Good to see you active 👍 , much much appreciated
We are not pushing PR at this time because we have a huge merge coming, but we will tread the PR list as soon as it is done

elecpower:
Overlap with newer PR #6359

PR #6368, opened 2019-03-13
Improve LUA  system timers handling

3djc:
Allows get and reset of system timers (globalTimer, throttle timer, and throttle percent timer)

### DIFF
--- a/radio/src/lua/api_general.cpp
+++ b/radio/src/lua/api_general.cpp
@@ -1061,7 +1061,7 @@ static int luaGetGeneralSettings(lua_State * L)
 }
 
 /*luadoc
-@function getSystemTimers()
+@function getGlobalTimers()
 
 Returns radio timers
 
@@ -1073,7 +1073,7 @@ Returns radio timers
 @status current Introduced added in 2.3.0.
 
 */
-static int luaGetSystemTimers(lua_State * L)
+static int luaGetGlobalTimers(lua_State * L)
 {
   lua_newtable(L);
   lua_pushtableinteger(L, "gtimer", g_eeGeneral.globalTimer + sessionTimer);
@@ -1436,7 +1436,7 @@ static int luaResetGlobalTimer(lua_State * L)
   g_eeGeneral.globalTimer = 0;
   size_t length;
   const char *option = luaL_optlstring(L, 1, "", &length);
-  if(!strcmp(option, "ALL")) {
+  if (!strcmp(option, "ALL")) {
     s_timeCumThr = 0;
     s_timeCum16ThrP = 0;
   }
@@ -1490,7 +1490,7 @@ const luaL_Reg opentxLib[] = {
 #endif
   { "getVersion", luaGetVersion },
   { "getGeneralSettings", luaGetGeneralSettings },
-  { "getSystemTimers", luaGetSystemTimers },
+  { "getGlobalTimers", luaGetGlobalTimers },
   { "getValue", luaGetValue },
   { "getRAS", luaGetRAS },
   { "getTxGPS", luaGetTxGPS },
--- a/radio/src/lua/api_general.cpp
+++ b/radio/src/lua/api_general.cpp
@@ -1434,8 +1434,8 @@ Resets the radio global timer to 0.
 static int luaResetGlobalTimer(lua_State * L)
 {
   g_eeGeneral.globalTimer = 0;
-  size_t lenght;
-  const char *option = luaL_optlstring(L, 1, "", &lenght);
+  size_t length;
+  const char *option = luaL_optlstring(L, 1, "", &length);
   if(!strcmp(option, "ALL")) {
     s_timeCumThr = 0;
     s_timeCum16ThrP = 0;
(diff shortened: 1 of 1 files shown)


CapnBry:
Argh, you beat me by 13 minutes with your pull request. I knew I shouldn't have gone to get a cup of coffee! This is redundant with #6369 so only one of the two should be applied.

If this pull request is used, I would think that the table returned by `luaGetSystemTimers()` both include global timer (global+session) as well as just the session timer alone. It also should have better names for the members, as the table of timers returned by getSystemTimers shouldn't have the redundant word "timer" in every table entry. Possible replacements: sess, total, throt, thpct?

The same names should be used for reset, if the reset-by-string implementation is used although I feel like the case-sensitivity of the implementation's comparison makes usage error-prone.

What are your thoughts on this implementation versus my own (#6369)?

radio/src/lua/api_general.cpp
```diff
@@ -1270,15 +1292,23 @@ static int luaGetUsage(lua_State * L)
...
 
 Resets the radio global timer to 0.
 
-@status current Introduced in 2.3.0
+@status current Introduced in 2.2.2
 */
 static int luaResetGlobalTimer(lua_State * L)
 {
   g_eeGeneral.globalTimer = 0;
+  size_t lenght;
```
bsongis: typo: length

radio/src/lua/api_general.cpp
```diff
@@ -1270,15 +1292,23 @@ static int luaGetUsage(lua_State * L)
...
 
-@status current Introduced in 2.3.0
+@status current Introduced in 2.2.2
 */
 static int luaResetGlobalTimer(lua_State * L)
 {
   g_eeGeneral.globalTimer = 0;
+  size_t length;
+  const char *option = luaL_optlstring(L, 1, "", &length);
+  if(!strcmp(option, "ALL")) {
```
bsongis: space missing

radio/src/lua/api_general.cpp
```diff
@@ -931,6 +931,28 @@ static int luaGetGeneralSettings(lua_State * L)
   return 1;
 }
 
+/*luadoc
+@function getSystemTimers()
```
bsongis: rather `getGlobalTimers`

CapnBry:
I don't want to nitpick about this, but getGlobalTimer()'s [return table is pretty bonkers](https://github.com/opentx/opentx/commit/a1e146bd63ce917afd3afcd9b0f9e123661e65b6#diff-9daa9cafd95d28bdda59a1330b862adaR1068) from a labeling standpoint. Of the four timer values returned, three are named "timer" and the other isn't. None of them should include the word "timer" at all, since the return value is all timers. The key for each timer should be more descriptive than a single letter prepended to the word timer. global, session, throttle (or thrott), and throttlepct (or throttpct) are self-documenting instead of gtimer, ttimer, and tptimer. Also, the function always returns all the timers so that should be getGlobalTimer**s**, but that's even more minor.

resetGlobalTimer() should use the same names, and also does nothing if the optional string parameter is blank. I would think this should be resetGlobalTimer**s** as well, and make "all" and no argument reset all the timers, so you only need to know the name of the timer you want if you only want to reset one. I believe the former will be the most used case and there's no need to pass "all" to get the most desired result.

Both are very minor changes, but I feel it would save someone from having to reference the documentation for the most general case (for reset) an remove the ambiguity/redundancy of the global timer table keys.

bsongis:
Please keep in mind that we want to keep as much as possible compatibility with existing scripts

CapnBry:
Absolutely I agree. `resetGlobalTimer()` is the only function currently in the wild. Both our pull requests to extend its functionality and add a `get` were within 24 hours of that code being added, which I realize now there's been a release since then. I misread the code in this pull's resetGlobalTimer() and see that it always resets the global timer regardless of if you want to or not. This is probably not the intended functionality for a LUA script user. You want to either reset a single timer, or all the timers, but never a single timer and also have the side effect of losing the global timer in the process. Maybe change the change there can be **only when there is no parameter**, reset the global timer, which matches the existing functionality.

The names of the strings added by this pull request are specific to this pull request so I believe they should be changed to be consistent with themselves (i.e. it would be stimer to match the others, not "session") but if they're going to be changed, drop "timer" from all of them and label them what they are: total, session, throttle, thottlepct. I would also argue that having the user pass arbitrary strings instead of constants into functions to decide what the function does is creating a potential pitfall for API users ("What is the word I need to pass to reset the throttle timer? thrtime? timerthrottle? ttimer? Not the const TIMER_THROTTLE for sure") but I think that ship has sailed already since this PR has been accepted. At the very least make the keys unambiguous and remove the substring they all have in common,

bsongis:
Why not, I will discuss this with 3djc tomorrow morning

bsongis:
Please check the latest commits around this function and let us know your comments!

CapnBry:
Looks great! I know the behavior is slightly different: resetGlobalTimer() in 2.2 only resets the global timer and not the session timer, but I feel like that's more of a bug/oversight since it just lowers what the user sees as the global timer to just be the current session (in 2.2). The way you've got it in that commit returns a consistent value between get and set and I think that's the better way to do it.

100% approve of this.

Issue #5668, opened 2018-02-05
Feature request: add globalTimer to getGeneralSettings()


Storchy1:
Is it possible to add one more Element to the table of the Lua getGeneralSettings().
The Element is g_eeGeneral.globalTimer.
It would be perfect to have an access to g_eeGeneral.globalTimer over Lua scripts to reset the value
A LUA script can then automatically reset the GlobalTimer, if the battary was charged.

3djc:
You have access to timers using model.getTimer and model.setTimer, so sure what more you need

3djc:
OK, I see what you want to do, sorry

3djc:
Addiong it to getGeneralSettings() would allow you to read the value, but you still need a way to reset it, which I do not think LUA allows currently, so you would need a new function for that

Storchy1:
Yes, you are right, the first step is to get the value over Lua Script function getGeneralsettings(). I think this will be possible without to many effort. Bit the second step (reset the value) will be a complex process.

Storchy1:
Thank you very much

CapnBry:
I realize this is closed due to the pull request being accepted, but the current elapsed time is not a "setting" as you'd expect as part of the return of `getGeneralSettings()`. I'm looking for the Throttle Time and Throttle Percent Time in LUA and we don't have those either. Wouldn't this be better served by a new function `getSystemTimer(which)` or `getGlobalTimer(which)` that can return:
```
0: return g_eeGeneral.globalTimer; break; // global timer
1: return s_timeCumThr; break;  // throttle timer
2: return s_timeCum16ThrP/16; break; //throttle percent timer
```
(obviously it would return lua_pushinteger() for them but you get the idea). `resetGlobalTimer()` can remain the same or change to `resetGlobalTimer(which)` to allow for resetting of the throttle timers as well with similar code.

Storchy1:
Of course, it would be better to have this two functions to get all timer values to LUA. Please ask kilrah for these function. Perhaps he has time to develop it.

CapnBry:
Glad you approve! I've gone ahead and made the changes myself (#6369). Tested out working on Taranis Q X7.

PR #6369, opened 2019-03-13
LUA getGlobalTimer() and extend resetGlobalTimer() to support all global timers

CapnBry:
As introduced in #5668, this change extends pull request #5669, which was to add LUA functions to get and reset the global total timer. This change reverts #5669's addition of the 'gtimer' field in the return value of getGeneralSettings(), since the global timer value is not really a setting even if it is persistent.

The new method `getGlobalTimer([which])` returns any of the global timers:
  * 0 - Session (DEFAULT)
  * 1 - Global(TOT) 
  * 2 Throttle time (THs)
  * 3 Throttle Percent time (TH%)

Also `resetGlobalTimer()` is now extended to `resetGlobalTimer([which])` to support resetting any of those timers.

I was on the fence about if which=0 would mean "All", which would make `getGlobalTimer` need to return either a table or just all 4 values, but I think most usages will just be retrieving one value. I also don't particularly like the magic number of 0 to 3 for the parameter but I didn't want to clutter up the LUA namespace with more constants.

Tested set/reset working on Taranis Q X7.

### DIFF
--- a/radio/src/lua/api_general.cpp
+++ b/radio/src/lua/api_general.cpp
@@ -912,7 +912,6 @@ Returns (some of) the general radio settings
  IMPERIAL units
  * `language` (string) radio language (used for menus)
  * `voice` (string) voice language (used for speech)
- * `gtimer` (number) radio global timer in seconds (does not include current session)
 
 @status current Introduced in 2.0.6, `imperial` added in TODO,
 `language` and `voice` added in 2.2.0, gtimer added in 2.2.2.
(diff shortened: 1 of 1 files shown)


bsongis (approved):
Seems OK to me

bsongis:
Will need to be ported to 2.3 if accepted

bsongis:
I didn't see the #6368 which does the same
I will close this one

PR #6372, opened 2019-03-17
Fixes ModelID unduly checked in D8

### LINKED ISSUE #6194: Model ID in use message in D8 mode

3djc:
Thix fixes #6194


bsongis:
@3djc needs to be done in 2.3, right?

3djc:
I have done it aleady in 2.3

PR #6383, opened 2019-04-04
Use unique "usage" for reporting each analog value in usb HID code.

karliss:
Before my changes HID description used Slider report twice for two different values. On Linux this resulted in joystick with 7 instead of 8 analog channels and the last channel was somewhat affected by both pots that were mapped to channels 7 and 8.  Different HID/joystick drivers might have handled it differently. It somewhat makes sense, keyboard page has two different usage id for reporting left shift and right shift.

Tested on ArchLinux with XD9+.

### DIFF
--- a/radio/src/targets/common/arm/stm32/usbd_hid_joystick.c
+++ b/radio/src/targets/common/arm/stm32/usbd_hid_joystick.c
@@ -99,14 +99,14 @@ static uint8_t  USBD_HID_DataIn (void  *pdev, uint8_t epnum);
     uint8_t buttons1; //bit 0 - button 1, bit 1 - button 2, ..., mapped to channels 9-16, on if channel > 0
     uint8_t buttons2; // mapped to channels 17-24, on if channel > 0
     uint8_t buttons3; // mapped to channels 25-32, on if channel > 0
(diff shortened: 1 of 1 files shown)


kilrah:
Hm this has the potential to break existing setups on other OSs...

karliss:
Did some testing on Windows
* Windows game controller configuration - worked both ways showed different name for one of axis
* Random joystick tester found on internet (http://www.planetpointy.co.uk/joystick-test-application/) worked both ways and didn't detect a difference
* Firefox+Html5 gamepad test https://html5gamepad.com/ - before changes got confused recognized 8 axis but the last two of them were mapped to the same input, after my changes worked correctly.

I guess it depends on which windows API and how gets used. In the worst some people will have to adjust the input mapping for one of the axis if they are using all 8 of them depending on the software they use.

I would say the changes are less breaking breaking than #5825 .

karliss:
What other testing would be require to move forward with this?

AlexDaniel:
Uh, this should not be closed. It is still an issue, and there is a similar one for [sky9x too](https://github.com/opentx/opentx/issues/7052).

PR #6389, opened 2019-04-14
Really new to this but think I have found and fixed a bug.

zmeusmith:
Really new to this as I have only ever writen code for myself and dont use github so Im probably doing all this wrong but I found this:

in:
JumperTX/radio/src/targets/simu/simpgmspace.cpp

Line 205 reads #elif defined(PCBX10)

Should read #elif defined(PCBT16) for the firmware to compile.


kilrah:
You seem to be submitting to the wrong repo. We have nothing Jumper related here.

zmeusmith:
Oops my bad. Sorry about that.

PR #6411, opened 2019-05-05
Horus widget model name

offer-shmuely:
image


kilrah:
Similarly to the other issue, you are overstepping the zone boundary:

capture_x12s_19-05-13_18-30-08

offer-shmuely:
Is that a problem on the top bar?
There is no other option, since we can’t select different layouts of top bar.

How do you suggest to write (widget or not) the model name in big letters?

PR #6412, opened 2019-05-05
Horus widget batt check updates

offer-shmuely:
1. calculate all values only once per 'refresh' to reduce CPU utilization
2. on battery disconnect from plane: 
        keep the data presented on screen (min/max/average)   
        non-active data is marked as blink


offer-shmuely:
image

offer-shmuely:
lshems commented that it is bad practice to use the global parameters,
it interrupt when using more than one instance of the widget (and I use 3-4 on any of my models)
working on updating the code....

offer-shmuely:
This fix also issue:6418

offer-shmuely:
Updated layout of the "1/4 screen size"
Now it use better the real estate, so additional data can be displayed even at this 1/4 size
:-)

I find it useful, since mostly I use the 1/4 screen with 2 BattCheck (one for each battery) + timer + LowestCellV

Hope other will find this usefull as well

image

offer-shmuely:
additional changes fro cleaner look

image

radio/src/translations/cz.h.txt
```diff
@@ -833,15 +833,15 @@
...
 #define TR_WAITING_FOR_TX              "čekám na TX..."
 #define TR_NO_TOOLS                    "Žádné nástroje k dispozici"
-#define TR_NORMAL                      "Normal"
-#define TR_NOT_INVERTED                "Not inv"
-#define TR_NOT_CONNECTED               "!Connected"
-#define TR_CONNECTED                   "Connected"
+#define TR_NORMAL                      "Normální"
+#define TR_NOT_INVERTED                "Neinvert"
+#define TR_NOT_CONNECTED               "Nepřipojen"
+#define TR_CONNECTED                   "P5ipojen"
```
Eldenroot: it should be "Připojen"

Eldenroot (changes requested):

offer-shmuely (commented):
.

radio/src/translations/cz.h.txt
offer-shmuely: looks that your changes are not relevant to this pull-request
please check them

offer-shmuely:
waiting for comments (or a merge)

3djc:
I get an error when trying to use it on top bar
image

offer-shmuely:
thanks!!
fixed the error.
more comments please...

3djc:
image

In top bar, vertical allignement doesnt seem to be ok

offer-shmuely:
view in top-bar updated
thank you for your comments!!
Can I get more of them please?

image

dmyers7:
Thanks for working this.  This has driven me nuts for years.  The widget would more often than not crash when a full battery is plugged into the FrSky Batt Cell sensor and I could never figure out why and would fly with just the overall voltage on another widget.  Just grabbed the script from the Git as I wait for the SD Card contents to be updated on OpenTX.  Will give it a go in a bit with a fully charged battery.  Scripted compiled and started fine.  Topping off a battery now.  BTW, the RC1 notes for 2.3.3 mentioned that this has been released at part of the new SD Card package but all the packages are still dated from Nov 16 and the BattCheck Main.lua does not reflect the changes. Interestingly, even with a slight over charge to insure a little over 4.2V / cell the FrSky Cell Sensor (SP-FLVS) clamps the reading at 4.2V until the voltage drops below 4.2V.  Anyhow, all good and not crashing now; thanks again.

3djc:
It should be included in next RC. As you can see, it is still opened

offer-shmuely:
I am happy that it’s help
I will be happy to get more feedback and suggestions for new additional updates

For example, I am not happy with meaning of the color of each cell

3djc:
We have merged it for inclusion in 2.3.3RC2

salarian:
Is it possible to make this widget work with the VFAS (i.e. the battery voltage from FrSky telemetry)?

3djc:
Vfas does not contain individual cell data, which this plugin is all about

salarian:
Ah, I see. It makes sense. Thanks for quick response. The thing is this plugin is included in 2.3.3 for Jumper T16 as a standard widget. So as an end-user, once I saw it there I assumed it would work with any telemetry system...
Probably the description of the widget should be updated to mention it would only work with FLVS.

offer-shmuely:
I believe that the sensor supply single voltage? (i don’t have one)
Can you what you expect to see? 
Maybe it can be adapted

salarian:
Given that **VFAS** is a single number, probably the desired outcome would be to show the battery's charge percentage. As far as cell voltage goes, since individual cell voltages are not available, only an average should be shown. There are two ways to get the average cell voltage: 
- to use another telemetry variable (for example, FrSky telemetry typically has **VFAS** for total and **A4** for cell voltage).
- to estimate it from the total voltage. Betaflight e.g. uses a configurable divisor and simply divides the total battery voltage by it.

offer-shmuely:
Hi salarian
I prepare a test script that support single-voltage telemetry
do you  like to test it? (I do not have any VFAS)
how can I send you?

3djc:
Just as a comment, we should not compute cells value based on total voltage and estimated cell count. Either we have an accurate data and we can use it, or we dont. There is no issue to make a simpler vfas widget that translate total voltage to %,  without the cells portion

offer-shmuely:
Why?
we can work on "perfect" mode for the FVLSS
and for VFAS/A1/A2 we can work on simpler mode that asume a lipo, i.e. per cell voltage is 3.0 to 4.2
and calculate the percent.

I can create a different Widget for the A1/A2/VFAS, but I afraid it will harder for the user to choose the correct widget

What do you say?

3djc:
Because it is plain wrong to lead the user to belive all its cells are perfectly match and none is low. We either have accurate data, and we can do things with it, or we don't, and it's better not to display it. It can be same widget and you just skip the whole individual cell display

offer-shmuely:
ok, I will remove the individual cells display.

asalarian:
> 
> 
> Hi salarian
> I prepare a test script that support single-voltage telemetry
> do you like to test it? (I do not have any VFAS)
> how can I send you?

I am on vacation and traveling this week. I'll be back by the weekend though. I'll be happy to try the patch then.

offer-shmuely:
https://github.com/offer-shmuely/opentx/tree/shmuely/horus_batt_widget_support_single_cell/radio/sdcard/horus/WIDGETS/BattCheck

asalarian:
> https://github.com/offer-shmuely/opentx/tree/shmuely/horus_batt_widget_support_single_cell/radio/sdcard/horus/WIDGETS/BattCheck

Hello @offer-shmuely 

Sorry that I "disappeared" for a long time!
I tested the script with a few quads and few battery packs and it works great! 

The only thing I can see is a bit of a text that is appearing just besides the widget. Is it intentional (e.g. for your debug purposes?)

screen

Thanks a lot!

offer-shmuely:
yes, it's a debugging print of the reset-event-detections
I fixed it.
however:
Since I accepted @3djc suggestion, we are NOT going to embedded this functionality into the official BattCheck on the official versions

So I changed the name of the widget to BattVFAS
Please use the official one for multi-cell sensors
and the BattVFAS only for single cell sensors

Enjoy


https://github.com/offer-shmuely/opentx/blob/shmuely/horus_batt_widget_support_single_cell/radio/sdcard/horus/WIDGETS/BattVFAS/main.lua

PR #6416, opened 2019-05-09
Fix value widget font size

offer-shmuely:
horus 'value' widget: 

suggestion to fix font size to use all the available area (current font are too small, and have too match space)

Not tested, I can't compile the CPP code. but I have lua scripts that do the same functionality with the larger fonts, and there is enough place for them.


3djc:
I'm afraid it is not right
image

offer-shmuely:
well, yes I see
this is because I use on my lua script less spaces in the following are

image


Mine look like this:
image

is it possible to shrink a little?

offer-shmuely:
and if we are discussing this waste space issue,
I saw many spaces that waste screen real estate
like:
image

image


do you think it better on a separate topic to discuss it?
I think that if we minimize the spaces on the theme, 
then each widget can choose for itself how match space it need, 
so it can use bigger fonts, or more data
and all of that on the same screen size (very important for widget like graph, complex gauge, more info on cell and so on)

kilrah:
The zones are defined by the layout, you must not go outside of them. If you want different ones you can create custom layouts.

offer-shmuely:
@kilrah I know, my suggestion is to change a little the layout zones, so bigger font can be use without go outside of the zones, especially if the "trims/slider" not needed on this tab
custom-layouts? how?

lshems:
Actually the current way of passing the widget location and size is rather basic.

It doesn't check if you draw outside.

So you can just check the left corner position and size, and adapt the widget actual frame by using your own boundaries.

Just don't draw outside the main LCD coordinates. It can cause a crash of ....

offer-shmuely:
Back to the main subject
Can we enlarge a little the layout zones (shrinking the useless spaces),
so more data and/or bigger font can be used?

bsongis:
We won't change the layout zones in 2.3
Instead we have a new UI system coming in 2.4 (which will prevent from drawing out of the zones)
Perhaps it would be better to provide a Lua script for this widget, no?

offer-shmuely:
Cool.
Please keep spaces between widget to minimum 
So each widget can decide if need the space, or need as match area as possible

kilrah:
We will leave the margins since they are needed to ensure that by default different widgets from different sources "cooperate" (each widget can draw to the last pixel of its zone, if there's no margin and the widget you place next to it draws to the first pixel then they "touch" each other and it's ugly).
As said you can do custom layouts if you want less margin.

But yeah having the ability to use custom layouts for the top bar too would be nice to add.

Note that you can also disable the top bar and redo that yourself to match what you want, even today.

lshems:
The mention that it will no longer be allowed to draw outside a widget predefined zone is alarming to me, if there is no way to define the widget zone layout in a feasible way, meaning no 'own-built' opentx.

It would limit the use of widgets immensely, and without reason I can think of.

Perhaps I'm missing something?

lshems:
It is good practice to ENFORCE layout zones to be exactly adjacent, so that the widget developer can choose to have adjacent widgets touching with unnoticeable separation or not.

Why would you do so?

Ugly hardly seems an argument.

kilrah:
> Perhaps I'm missing something?

That as mentioned twice already you can write your own layouts in lua and make it just the way you want.

lshems:
Any hints to where I can find info on how that works?

I never understood the layout to be Lua definable.

Good it is lua, but wonder if people will learn how to properly size widget content with complete freedom of size.

Without fixed size fonts, and some functions to get that info, autoscaling is difficult.

Sorry again for not seeing Lua as way to define the layout.

bsongis:
@offer-shmuely perhaps the margin could be a layout parameter, would it work?

lshems:
No link to the solution suggested.

PR #6441, opened 2019-05-22
Companion checklist edit save fails if missing or invalid SD path

### LINKED ISSUE #6435: Companion model checklist edit fails saving file
Occurs when profile settings SD folder path is blank or invalid.

OpenTX Companion (all versions on Ubuntu 14.04)

elecpower:
Fixes #6435 
Error is not clear why the save failed.
Test before launching editor and display a more meaningful message so user knows where to fix.


bsongis:
Thanks for your help!

PR #6446, opened 2019-05-24
Fix negative gvars precision (2.3)

elecpower:


bsongis:
I have just pushed a fix slightly different from your PR, thanks for the help!

PR #6447, opened 2019-05-24
Add RX numbers to CPN models list (2.3)

elecpower:


companion/src/modelslist.cpp
```diff
@@ -710,10 +720,46 @@ void TreeModel::refresh()
...
 }
+
+bool TreeModel::isModelIdUnique(unsigned modelIdx)
+{
+  int cnt = 0;
+  for (unsigned i=0; i<radioData->models.size(); i++) {
+    ModelData & model = radioData->models[i];
+    for (unsigned j=0; j<CPN_MAX_MODULES; j++) {
+      if (!model.moduleData[j].protocol == PULSES_OFF && model.moduleData[j].modelId == modelIdx) {
+        cnt++;
```
bsongis: you may return false from here to avoid looping on all models

elecpower: @bsongis yes will speed up. In testing the change I also found a bug and will fix both in the next commit.

elecpower:
@bsongis after retesting there isn't a bug, well I hope not. So just exit as soon as 1st duplicate found.

companion/src/mdichild.cpp
```diff
@@ -543,6 +542,14 @@ void MdiChild::initModelsList()
     ui->modelsList->setIndentation(0);
   }
   refresh();
+
+  if (firmware->getCapability(Capability::HasModelCategories)) {
+    ui->modelsList->header()->resizeSection(0, ui->modelsList->header()->sectionSize(0) * 2);   // pad out categories and model names
+    }
```
bsongis: indentation issue there!

companion/src/modelslist.cpp
```diff
@@ -710,10 +720,53 @@ void TreeModel::refresh()
...
 }
+
+bool TreeModel::isModelIdUnique(unsigned modelIdx)
+{
+  int cnt = 0;
+  for (unsigned i=0; i<radioData->models.size(); i++) {
+    ModelData & model = radioData->models[i];
+    if (!model.isEmpty()) {
+      for (unsigned j=0; j<CPN_MAX_MODULES; j++) {
+        if (!model.moduleData[j].protocol == PULSES_OFF && model.moduleData[j].modelId == modelIdx) {
```
bsongis: `model.moduleData[j].protocol != PULSES_OFF`

companion/src/modelslist.cpp
```diff
@@ -710,10 +720,53 @@ void TreeModel::refresh()
...
+          if (cnt < 1) {
```
bsongis:
```
if (cnt++ > 0) {
  return false;
}
```

bsongis:
Thanks!

PR #6428, opened 2019-05-18
Update Multi protocols

pascallanger:
Added:
 - BAYANG DHD_D4
 - MJXQ PHOENIX and Freq tuning for E010
 - E01X E016H
 - GD00X V1 and V2
 - BUGSMINI BUGS3H

PR #6462, opened 2019-05-29
Multi protocol updates for 2.2.4 to align with MPM v1.2.1.51

MRC3742:
Multiprotocol listing updates for both Radio UI and Companion.
Updates to bring OpenTX Multi listings in line to current Multiprotocol Module version 1.2.1.51

Compiled, tested and working using v2.2.4 on X9D+


3djc:
This looks good, thanks ! Does similar work need to be done against 2.3 (we are still unsure which will be out first)

MRC3742:
pascallanger has already done an update #6428 for the Radio GUI which has been merged into 2.3  but I will do an additional PR to update Companion  for 2.3 later today.

3djc:
Thanks a lot, merging then !

PR #6465, opened 2019-05-29
Updated some French translation.

LapinFou:
There is some inconsistency about the naming convention (in Eng) int the SD synchronizer.
Sometimes, they are named as **local/radio**, sometimes as **source/destination**.
I think this is confusing.
In this PR, I used only the naming convention **local/radio**.

Here is a screenshot to highlight my explanation:
In English:
tmp2

In French (before this PR):
tmp

### DIFF
--- a/companion/src/translations/companion_fr.ts
+++ b/companion/src/translations/companion_fr.ts
@@ -2550,32 +2550,32 @@ Si vide, la ligne est toujours active.</translation>
     <message>
         <location filename="../dialogs/filesyncdialog.cpp" line="131"/>
         <source>Source Folder:</source>
-        <translation>Dosssier Source:</translation>
+        <translation>Dosssier Local:</translation>
     </message>
     <message>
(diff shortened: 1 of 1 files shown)


bsongis:
I also thing it would be better to fix the original english strings. There are also some discrepancies with first uppercase letter.

bsongis:
I would wait for @3djc and @kilrah comment on the right strings to use in english

3djc:
I'm unsure about that. I like the local/radio as it helps understand what directories you are expected to enter here, but also source/destination as it helps figure what is going to happen. Could we consider soemthing like Local Folder (source): and Radio Folder (destination): ?

bsongis:
Any solution will be ok for me as soon as it is consistent

LapinFou:
I think Local/Radio only is better, because there is no ambiguity.
If you go for local/radio, I don't see the advantage to also mention source/destination.

bsongis:
It seems you both agree, I am ok with local / radio as well, it is very clear

PR #6467, opened 2019-05-29
Travis build error fix for PR #6466 Update MPM for 2.3

MRC3742:
Looks like a small copy/paste error when updating file from my 2.2 branch!!!


3djc:
Eh ! Copy paste error are my specialty !! Travis seems to be happier with this

PR #6482, opened 2019-06-06
PXX synchro using Hearbeat

bsongis:


radio/src/switches.cpp
```diff
@@ -431,7 +431,7 @@ bool getSwitch(swsrc_t swtch, uint8_t flags)
   else if (cs_idx == SWSRC_ON) {
     result = true;
   }
-#if defined(DEBUG_LATENCY)
+#if defined(DEBUG_LATENCY_MIXER_RF)
```
bsongis: change not needed (or we have to remove the switch completely)

bsongis: change in next commit

PR #6514, opened 2019-06-18
Default switchs / pots / sliders configuration moved to the HAL

bsongis:


lshems:
Question: is there any change in the assignments of the id for switches and other sources, as well as for the SF 'sources'?

kilrah:
This commit has zero user influence, it's just moving code that is only used when formatting eeprom from a file to another.

lshems:
Thanks for clarifying

PR #6561, opened 2019-07-12
Trainer function can now replace the channels values instead of to th…

bsongis:
…e sticks positions


raphaelcoeffic:
What‘s up? Not up for merge? It seems quite useful for having own settings on the student radio (less agressive controls?) without heavy programming on the mixers.

bsongis:
Nobody tested it, even me :)

LapinFou:
This new feature looks very very interesting. 👍

bsongis:
Someone tests it?

LapinFou:
No I didn't. Unfortunately, I won't have anytime this WE. Maybe next week...
Sorry.

robustini:
Nobody tested it? I feel... :p
At the moment I use a screen where I've all the PR of the Trainer, but in fact the replacement would be more immediate.

LapinFou:
I didn't have time to test it. I need to compile the firmware for my X9D+, but I never done it before...

robustini:
@bsongis , I did some tests and it works really well, no problems found.

bsongis:
Thanks for your tests, PR merged!

LapinFou:
@bsongis 
Hi Bertrand,
I would like to test this new feature, however I don't understand how it is working. Moreover, it seems this is not implemented in Companion.
Any help would be very welcome.

Seb

LapinFou:
Merry Christmas to everyone. 😃 
Anyone to explain me how does work this new feature?

Seb

kilrah:
Select Chans in Trainer special function, channels sent out will be the ones from the trainer box unmodified when enabled.

LapinFou:
Thanks Kilrah.
I realized this new feature has not been implemented in Companion. 😅 
I will try on my radio, but I don't understand what is replaced by what... Sorry to be dump on this one.
Could you clarify how it must work?

bsongis:
PR opened for companion adaptation

kilrah:
> Could you clarify how it must work?

When the SF is on the master radio doesn't do anything anymore but send out what it receives on the trainer port straight out. Normally it goes through the whole stick processing.

LapinFou:
> When the SF is on the master radio doesn't do anything anymore but send out what it receives on the trainer port straight out. Normally it goes through the whole stick processing.

Sorry for not understanding it correctly. I'm trying hard. 😭
Let me rephrase my question:
- The classic trainer mode works this way; the trainer inputs are going through the trainer menu (calibration + multiplier + mapping). The sources (on the master radio) "Ail/Prf/Thr/Ele" will be replace with the student/trainer inputs according the chosen mapping in the trainer menu.
- For user looking for a more complex master/slave setup, the sources TRxx can be used. However the calibration + multiplier are applied on the fourth first channels (TR01-TR04).

My understanding of this new feature is when the trainer "channel" mode is enable, the sources "Ail/Prf/Thr/Ele" are not replaced anymore with the inputs coming from the student radio. Meaning the sources "Ail/Prf/Thr/Ele" are always raw data coming from the master gimbals.
Is it correct?
If yes, then my question is where are the trainer sources and how to use it?
Shall we used the sources TRxx?
Or, the mixes (on the master radio) will be replace with the student raw data according the chosen mapping in the trainer menu? If yes, how can we choose the mapping order? Shall it be done on the slave radio?

kilrah:
As said when that mode is used the master radio does nothing at all. It takes the trainer signal values and sends them straight to the module, all processing bypassed.

3djc:
Using channel takes channel 1 to 4 values directly, so if your slave radio has the mixes and all, just send it

LapinFou:
To both of you, thanks for helping.
If I understand it correctly, it means when this trainer mode is active:
- TR1 is directly sent to the RF module (CH1)
- TR2 is directly sent to the RF module (CH2)
...
- TR4 is directly sent to the RF module (CH4)
If yes, it means the **inputs** (for eg DR) / **mixes** / **outputs** (limits, direction, etc...) settings must be duplicated on the slave radio??
If yes, then it means this mode cannot be used with a model requiring more than 4 channels, correct?

3djc:
It works exactly as normal (stick) trainer, but instead of the slave sending raw stick value, it sends ch1 - c4 instead, everything else is 100% the same as regular trainer

3djc:
I think we have 16 TR chans

LapinFou:
Oh! OK.
I think a finally got it.
So, it means when this trainer mode is enabled, the following datapath is used;
TR1 sent directly to CH1 on the RF module,
TR2 sent directly to CH2 on the RF module,
...
TR16 sent directly to CH16 on the RF module,
Off course the number of transmitted channels depend of the setting chosen in the "trainer port".

It also mean than all INPUTS, MIXES and OUTPUTS must be duplicated on the slave radio (since they are bypassed as mentioned @kilrah ).
Did I got it correctly this time?

LapinFou:
Off course, I guess the "calibration" and the "multiplier" in the global trainer menu are also bypassed, correct?

kilrah:
Yes correct. EVERYTHING comes from the slave directly, so you need whatever configuration you want done on it.

LapinFou:
Thanks!
Finally I got it. 😄 
What confused me is the description "channels values". To me, channels values are coming out of MIXES and going in the OUTPUTS (or somewhere else if CHxx is used). In others words, to me, the channels are somewhere in the middle of the OpenTX datapath, not at the final outputs.
I think I would have got the concept faster with a description like "Trainer function can now replace the **outputs values** instead of to the sticks positions" or "Trainer function can now replace the **outputs values sent to the master RF module** instead of to the sticks positions".

PR #6567, opened 2019-07-13
Fix HW patch feature

### LINKED ISSUE #6564: Saving models and settings to Radio can delete all models if Hardware Patch is enabled
I could not find this bug listed nor could I find a Companion specific issues area so I apologize if it's been reported somewhere else before as a Companion issue on GitHub. 

This is my 1st post. The problem is serious in that it will consistently delete all models and settings in I believe any Taranis radio whenever the option to "Patch HW settings from profile" is pre-enabled just before a "Write Models and Settings to Radio" operation is requested..

This write step will consistently delete all models and settings in the Radio IF another Settings option is also pre-enabled.  To set this other option needed to trigger this issue, read in any .otx model file and in the upper right of the dialogue box listing the models select "Edit Radio Settings” and then at the bottom right of that page enable the “ Store calib. and selected settings in Profile” option.. the combination of both options bieng pre-set will cause this model delete issue on both companion 2.2.3 and Companion 2.2.4.

I’m using the Windows version of Companion on Win 10 64 bit with a Taranis Q 7XS and as a new user found it very frustrating in that it was initially intermittent since I did not recognize the correlation with the two options as indicated above.. for reference with screen grabs..  It’s also posted on RCGroups.com under a thread with this title: “Anyone else had a problem saving .otx files with OpenTx and Companion 2.2.3? to a Tar(anis) by jims123

**Expected behavior**
under no circumstance would I expect to lose all models and settings unless I intentionally deleted them  

**Screenshots**
there are up on RCG

**General information**
 - OpenTX version (e.g. 2..2.3 and 2.2.4)
 - Hardware (e.g. Frsky 7X or 7XS)
 - OpenTX Companion (e.g. 2.2.3 and 2.2.4)

3djc:
HW patch was not working at all, this fixes the basic functionnaly and also fixes #6564

### DIFF
--- a/companion/src/firmwares/er9x/er9xeeprom.cpp
+++ b/companion/src/firmwares/er9x/er9xeeprom.cpp
@@ -59,7 +59,7 @@ t_Er9xTrainerMix::operator TrainerMix()
 
 t_Er9xTrainerData::t_Er9xTrainerData()
 {
-  memset(this, 0, sizeof(t_Er9xTrainerData));
+  memset(reinterpret_cast<void *>(this), 0, sizeof(t_Er9xTrainerData));
 }
 
 t_Er9xTrainerData::operator TrainerData ()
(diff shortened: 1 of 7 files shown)


jims1234:
Great ,

PR #6586, opened 2019-07-24
CZ translations -radio

mhotar:


radio/util/translate.py
```diff
@@ -18,7 +18,9 @@
                        ('\\207', u'š'),
                        ('\\207', u'Š'),
                        ('\\210', u'ú'),
+                       ('\\210', u'Ú'),
                        ('\\211', u'ů'),
+                       ('\\211', u'Ů'),
```
kilrah, line 23: These are not in the font!

mhotar: Right, but for monoscreens they are translated to lower case characters now. See the same code for both "ú" and "Ú"

kilrah: Ah yes, OK...

bsongis:
Thanks!

PR #6591, opened 2019-07-28
Flighmode needs switch conversion too

### LINKED ISSUE #6590: X-Lite conversion from 2.2.4N706 to 2.3.0N54 changed setup
FYI

Yesterday I tested an upgrade from 2.2.4N706 to the 2.3.0N54.
Upon startup the X-Lite started to convert the setup.
I then noticed that of at least my default model the setup has changed.

The flight modes were assigned totally different switches than on the original model.
I have not investigated other changes yet.

Please close this ticket if conversion errors should be expected at this stage of development.
I will close this ticket if there is no further interest.

I have an otx of before and after the conversion.
I can post those upon request.

2 2 4N706
2 3 0N54

3djc:
This fixes #6590


bsongis:
Thanks!

PR #6616, opened 2019-08-12
Allow rxnum > max models

### LINKED ISSUE #6608: Companion 2.3.0N59 / X9 Lite Receiver No >61
Bug in Companion 2.3.0N59 and before

for a model internal Radio system, one can chose a receiver No. up to 63

up to 61 everything is fine. if one choses 62 or 63 as a receiver No., Companion crashes when closing the modell edit window.

if the modification is done in the radio, The radio binds the receiver fine, but when reading the modells from the radio via USB, companion crashes.

took me a while to find out, I have 2 identical x9-Lite, one was always good and the other unreadable. duplicating the good one to the bad one solved the problem, until I changed back the receiver No. to 62.
repeated this 4 times until I found the Bug.
changing back the receiver No. to 61 or below fixes every thing

Hope this is clear enough
and Thanks for the Work
Lucifer

2x X9-Lite
1x X9D first version (800 mAh)
1x X9D second version (2000 mAh)
1x X9E

3djc:
This fixes #6608

@elecpower : I have remove this strange lines, could you have a look ? Do you recall their intended purposes ?

### DIFF
--- a/companion/src/modelslist.cpp
+++ b/companion/src/modelslist.cpp
@@ -726,17 +726,13 @@ void TreeModel::refresh()
         protocol = model.moduleData[j].protocol;
         // These are the only RXs that allow nominating RX # but changing RX or copying models can leave residual configuration which can cause issues
         // if (protocol == PULSES_PXX_XJT_X16 || protocol == PULSES_PXX_XJT_LR12 || protocol == PULSES_PXX_R9M || protocol == PULSES_DSMX || protocol == PULSES_MULTIMODULE) {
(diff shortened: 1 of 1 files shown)


bsongis:
Yes, the code removed seems very wrong:
- it mixes model index and RX Num
- it misses a protection against array overflow

bsongis (approved):

elecpower (commented):
@3djc purpose was to highlight module receiver # pointing to a none existent model but my error in the coding. Should have used an existing function or written something like isModelIdUnique which avoids array out of bounds and gets the logic correct.

3djc:
Thx @elecpower.

PR #6623, opened 2019-08-12
X10: fixed optional inputs

raphaelcoeffic:
~~`GMBL` & `GMBL` are still shown as `SI` and `SJ`, but I'm missing a good idea, as there is not enough space on the screen for more than 2 characters~~.

~~On the hardware page, there are shown as `GMBL` & `GMBL`~~.

The switches are now displayed as `SI` and `SJ` everywhere.


radio/src/gui/gui_common.cpp
```diff
@@ -180,7 +180,7 @@ bool isSourceAvailable(int source)
   }
 
 #if defined(PCBX10)
-  if ((source>=MIXSRC_S3 && source<=MIXSRC_S4) || (source>=MIXSRC_MOUSE1 && source<=MIXSRC_MOUSE2))
+  if (source>=MIXSRC_MOUSE1 && source<=MIXSRC_MOUSE2)
```
bsongis, line 183: cosmetics: spaces missing (as done below)

radio/src/storage/conversions/conversions_218_219.cpp
```diff
@@ -54,7 +54,7 @@ int convertSource_218_to_219(int source)
 #endif
 
 #if defined(PCBHORUS)
-  if (source >= MIXSRC_GMBL)
+  if (source >= MIXSRC_SI)
```
bsongis, line 57: Would you add a non regression test for X10 translations? As we did for X7 / X9D+ / X9E

bsongis: OK the non regression test is already there, perhaps even done by me, I cannot remember

PR #6626, opened 2019-08-14
Companion ACCESS support

raphaelcoeffic:
TODOs:
General settings:
- [x] registrationId

Model settings:
- [x] registrationId
- [x] receiver list (`read-only` + `delete`)
- [x] telemetry sensor receiver
- [x] show receiver in various "sources" (vario, RSSI, etc)


bsongis:
Yes, I think the receivers list should be readonly. Perhaps later add the possibility to remove a receiver?

raphaelcoeffic:
> Yes, I think the receivers list should be readonly. Perhaps later add the possibility to remove a receiver?

Later? I can add just now, why not. Just not sure what it means on the TX. Is there anything to be done physically with the receiver after deleting it from the model?

kilrah:
On the radio we have 2 options, Delete and Reset - Delete just clears the struct on the radio, but Reset also sends an "unbind" command to the receiver. Obviously if we do it in companion only Delete is possible

raphaelcoeffic:
> Delete just clears the struct on the radio

Ok, then let's have it!

Allerhopp:
Here is my EEPROM backup (carbo from chat) with the broken telemetry (companion).
Names, sensor id ... also sensor instance in radio x9d is wrong with 3 digits instead of 2. Or is it a new feature, i do not understand .....
[eeprom-2019-08-18-190908.zip](https://github.com/opentx/opentx/files/3513432/eeprom-2019-08-18-190908.zip)

N064 X9D I did a new sensor search in the default model R9M to be sure.

raphaelcoeffic:
@Allerhopp there is really a lot of problems with Companion and X9D+ at the moment! Thx for pointing at it! It will be fixed shortly. I've found already a couple of issues, but there are more...

companion/src/firmwares/opentx/opentxeeprom.h
```diff
@@ -83,7 +83,8 @@ class ProtocolsConversionTable: public ConversionTable
       addConversion(PULSES_PXX_XJT_D8, val);
       addConversion(PULSES_PXX_XJT_LR12, val++);
 
-      addConversion(PULSES_ACCESS_ISRM, val++);
+      addConversion(PULSES_ACCESS_ISRM, val);
```
bsongis, line 86: I wonder if my previous code didn't already break the module type conversion (i.e. a DSM2 module)

PR #6627, opened 2019-08-14
stale-bot: stale after 180 days, close after another 90 days

raphaelcoeffic:


bsongis:
Seems a nice plugin, I would like to give it a try

bsongis (approved):

PR #6628, opened 2019-08-14
X9D+ 2019 / R9M Bind Fix (+ more code shared between 212 and 128 to a…

bsongis:
…void this again)


bsongis:
Confirmed tested on X9D+ 2019

PR #6629, opened 2019-08-14
Add support for pwr-on and power-off speed

### LINKED ISSUE #4639: OFF- switch without delay for Horus
The OFF- switch must be hold for very long time before the TX goes down.
It is not important it is annoying. No one can switch off unconsciously, the lid over the switch reliably…
Please make the push for “OFF” shorter...

raphaelcoeffic:
closes #4639 

TODOs:
- [x] radio setup for other screen sizes.
- [x] translation strings for other languages (only english for now).
- [x] companion support.


raphaelcoeffic:
On both settings, using the maximum speed basically means disabling the power-on / shutdown delay completely.

~~Turning the power-on delay off does have some funny side effects: my X9 lite boots to OTX every time I flash it via STM32 bootloader (no clue yet how to fix it, in case it's possible).~~ (fixed)

companion/src/firmwares/opentx/opentxeeprom.cpp
```diff
@@ -2405,6 +2405,15 @@ OpenTxGeneralData::OpenTxGeneralData(GeneralSettings & generalData, Board::Type
 
   if (version < 218) internalField.Append(new UnsignedField<8>(this, generalData.rotarySteps));
   internalField.Append(new UnsignedField<8>(this, generalData.countryCode));
```
bsongis: 2 bits here instead of 8

radio/src/translations/en.h.txt
```diff
@@ -593,6 +593,10 @@
 #define TR_BLOFFBRIGHTNESS             INDENT "OFF brightness"
 #define TR_BLCOLOR                     INDENT "Color"
 #define TR_SPLASHSCREEN                "Splash screen"
+#define TR_PWR_ON_SPEED                "Pwr On speed"
```
bsongis, line 596: duplicated lines

radio/src/translations.cpp
```diff
@@ -236,6 +236,10 @@ const char STR_MENULOGICALSWITCH[] = TR_MENULOGICALSWITCH;
 const char STR_MENULOGICALSWITCHES[] = TR_MENULOGICALSWITCHES;
 const char STR_MENUCUSTOMFUNC[] = TR_MENUCUSTOMFUNC;
 const char STR_SPLASHSCREEN[] = TR_SPLASHSCREEN;
+#if defined(PWR_BUTTON_PRESS)
```
bsongis: I am not sure it's needed to add this `#if defined`, the optimizer should do his job

radio/src/targets/taranis/board.cpp
```diff
@@ -189,9 +199,14 @@ void boardInit()
   DBGMCU_APB1PeriphConfig(DBGMCU_IWDG_STOP|DBGMCU_TIM1_STOP|DBGMCU_TIM2_STOP|DBGMCU_TIM3_STOP|DBGMCU_TIM6_STOP|DBGMCU_TIM8_STOP|DBGMCU_TIM10_STOP|DBGMCU_TIM13_STOP|DBGMCU_TIM14_STOP, ENABLE);
 #endif
 
+#if defined(EEPROM) && defined(PWR_BUTTON_PRESS)
+  storageReadRadioSettings();
```
bsongis: then we read the radio settings many times on startup?

bsongis: I think having 2 different paths for reading radio settings is not good from maintenance point of view. We should avoid this if possible

bsongis (changes requested):
After reading this PR here are my questions / suggestions (some are not related to the PR itself, but ...):
- [x] run the animation in `opentxInit()` - after `storageReadRadioSettings()` has been called - instead of `boardInit()`
- [x] create a function `runStartupAnimation()` to sort things
- [x] move the `rtcInit()` call inside `boardInit()`
- [x] question: why not the same granularity for delays at power off than power off (3 bits each instead of 2bits + 4bits) ?
- [x] question: why we don't have such an animation on X10 / X12S ?

bsongis:
- [x] Move `backlightInit()` to `boardInit()` and make it not start the backlight, just init

bsongis:
- [x] Call `drawSplash()` in `main()` only when there is no animation ...

bsongis:
- [x] Put the X9E animation back

bsongis:
- [x] While we are in PWR code, I see `pwrResetHandler()` doesn't use the correct define for AHB1ENR!

raphaelcoeffic:
> question: why not the same granularity for delays at power off than power off (3 bits each instead of 2bits + 4bits) ?

This is related to the delay used currently (1s vs. 4s). During the discussions, it seems there are a wide range of preferences for the shutdown delay (anything from 0s to 5s), while the power on delay is basically on/off for most people with respect to preferences.

> question: why we don't have such an animation on X10 / X12S ?

Don't know for the X12S, but my X10 already takes a few seconds when I press to really boot (same can be observed with the bootloader). I guess this is because the delay is "built-in" the electronic.

notlebowski:
Related to power on/off...
When I received my X-Lite I thought it was broken.
I didn't keep the button pressed long enough and thought it immediately turned off again.
If the backlight would have turned on immediately (and stayed on) I would have been less confused.

I still think the backlight should go on directly after power-on. It gives you a confirmation of your action. Even if the power-on was accidental this would be better as it would be more like you would get aware of it.

bsongis:
Everybody agrees about @notlebowski comment?

3djc:
I think backlight should be on as soon as power button is pushed, and stay on while maintained. When on, then settings should be applied imho

bsongis:
- [x] call `backlightOn()` - which uses radio settings - as soon as the PWR button is pressed

raphaelcoeffic:
> 1. While we are in PWR code, I see `pwrResetHandler()` doesn't use the correct define for AHB1ENR!

which one should it be? I could not find any `define` using that register.

bsongis:
I am doing all modifications I suggested

companion/src/firmwares/opentx/opentxeeprom.cpp
```diff
@@ -2404,7 +2404,16 @@ OpenTxGeneralData::OpenTxGeneralData(GeneralSettings & generalData, Board::Type
 
 
   if (version < 218) internalField.Append(new UnsignedField<8>(this, generalData.rotarySteps));
-  internalField.Append(new UnsignedField<8>(this, generalData.countryCode));
+  internalField.Append(new UnsignedField<2>(this, generalData.countryCode));
+
+  if (version >= 219 && IS_HORUS_OR_TARANIS(board)) {
+    internalField.Append(new UnsignedField<2>(this, generalData.pwrOnSpeed, "Power On Speed"));
+    internalField.Append(new UnsignedField<4>(this, generalData.pwrOffSpeed, "Power Off Speed"));
```
bsongis: 2 + 2 spare (won't change anything, but ...)

radio/src/datastructs.h
```diff
@@ -771,7 +771,14 @@ PACK(struct RadioData {
   NOBACKUP(uint32_t globalTimer);
   NOBACKUP(uint8_t  bluetoothBaudrate:4);
   NOBACKUP(uint8_t  bluetoothMode:4);
-  NOBACKUP(uint8_t  countryCode);
+  NOBACKUP(uint8_t  countryCode:2);
+#if defined(PCBHORUS) || defined(PCBTARANIS)
```
bsongis: no if defined here (VC++ compilation)

radio/src/datastructs.h
```diff
@@ -771,7 +771,14 @@ PACK(struct RadioData {
...
+  NOBACKUP(uint8_t  pwrOnSpeed:2);
+  NOBACKUP(uint8_t  pwrOffSpeed:2);
+  NOBACKUP(uint8_t  spareCC:2);
```
bsongis: you would you name it spareX (where X = N+1)

radio/src/datastructs.h
```diff
@@ -771,7 +771,14 @@ PACK(struct RadioData {
...
+#else
+  NOBACKUP(uint8_t  spareCC:6);
```
bsongis: remove this, not needed

radio/src/gui/128x64/startup_shutdown.cpp
```diff
@@ -30,9 +28,11 @@ const unsigned char bmp_sleep[]  = {
 #include "sleep.lbm"
 };
 
-void drawStartupAnimation(uint32_t duration)
+void drawStartupAnimation(uint32_t duration, uint32_t total_duration)
```
bsongis: totalDuration

radio/src/opentx.cpp
```diff
@@ -1728,8 +1728,17 @@ void copyTrimsToOffset(uint8_t ch)
   storageDirty(EE_MODEL);
 }
 
+
+const uint8_t _pwrOnOffDuration_lookup[] = {
```
bsongis: why divide by 2 ?

radio/src/targets/common/arm/stm32/bootloader/boot.cpp
```diff
@@ -213,6 +213,17 @@ int main()
   pwrInit();
   pwrOff();
 
+  // wait a bit for the inputs to stabilize...
```
bsongis, line 216: It has been removed on purpose and now it will break other boards. please add if defined(HORUS) here

raphaelcoeffic: It was for the X9 lite, not for Horus. I could not start the bootloader after setting the power-up delay to 0.

radio/src/opentx.cpp
raphaelcoeffic: Well, linearly it makes: 2, 1.333, 0.666, 0. Feels a bit weird.

raphaelcoeffic: Or really 3, 2, 1, 0 ? But that would make 3s the default value, which is quite long.

radio/src/targets/common/arm/stm32/bootloader/boot.cpp
raphaelcoeffic: anyways, it seems it works for me now without that busy loop (X9 lite).

radio/src/gui/128x64/radio_setup.cpp
```diff
@@ -506,6 +509,20 @@ void menuRadioSetup(event_t event)
       }
 #endif
 
+#if defined(PWR_BUTTON_PRESS)
+    case ITEM_RADIO_SETUP_PWR_ON_SPEED:
+        lcdDrawTextAlignedLeft(y, STR_PWR_ON_SPEED);
```
bsongis, line 514: wrong indentation

radio/src/gui/128x64/radio_setup.cpp
```diff
@@ -506,6 +509,20 @@ void menuRadioSetup(event_t event)
...
+        lcdDrawNumber(RADIO_SETUP_2ND_COLUMN, y, g_eeGeneral.pwrOnSpeed, attr|LEFT);
+        if (attr) CHECK_INCDEC_GENVAR(event, g_eeGeneral.pwrOnSpeed, 0, 3);
```
bsongis: -1 => 2 or -2 =>1

radio/src/opentx.cpp
```diff
@@ -1728,8 +1728,17 @@ void copyTrimsToOffset(uint8_t ch)
...
+  200, // 2s
+  100, // 1s
+  50,  // 0.5s
+  0,   // none
+};
+
 #if defined(STARTUP_ANIMATION)
-#define PWR_PRESS_DURATION_MIN        100 // 1s
+
+#define PWR_PRESS_DURATION_MIN() (_pwrOnOffDuration_lookup[g_eeGeneral.pwrOnSpeed & 3])
```
bsongis: pwrOnSpeed + 1

bsongis: or pwrOnSpeed + 2

radio/src/datastructs.h
```diff
@@ -771,7 +771,10 @@ PACK(struct RadioData {
   NOBACKUP(uint32_t globalTimer);
   NOBACKUP(uint8_t  bluetoothBaudrate:4);
   NOBACKUP(uint8_t  bluetoothMode:4);
-  NOBACKUP(uint8_t  countryCode);
+  NOBACKUP(uint8_t  countryCode:2);
+  NOBACKUP(uint8_t  pwrOnSpeed:2);
```
bsongis: int8_t

radio/src/opentx.cpp
```diff
@@ -1728,9 +1728,11 @@ void copyTrimsToOffset(uint8_t ch)
   storageDirty(EE_MODEL);
 }
 
+
 #if defined(STARTUP_ANIMATION)
-#define PWR_PRESS_DURATION_MIN        100 // 1s
-#define PWR_PRESS_DURATION_MAX        500 // 5s
+
+#define PWR_PRESS_DURATION_MIN() (2 - g_eeGeneral.pwrOnSpeed)
```
bsongis: avoid this define, could be inline function

radio/src/opentx.cpp
```diff
@@ -2025,6 +2027,9 @@ int main()
 
 #if !defined(SIMU)
 #if defined(PWR_BUTTON_PRESS)
+
+#define PWR_PRESS_SHUTDOWN_DELAY() (2 - g_eeGeneral.pwrOffSpeed)
```
bsongis: also inline here, it will avoid the cast<uint32_t> later

radio/src/gui/212x64/startup_shutdown.cpp
```diff
@@ -57,9 +57,11 @@ void drawStartupAnimation(uint32_t duration)
   lcdRefresh();
 }
 
-void drawShutdownAnimation(uint32_t duration, const char * message)
+void drawShutdownAnimation(uint32_t duration, uint32_t totalDuration, const char * message)
 {
-  uint8_t index = limit<uint8_t>(0, duration / (PWR_PRESS_SHUTDOWN_DELAY / 4), 3);
+  if (totalDuration == 0) return;
```
bsongis: cosmetics: newline before `return` here (and in some other places)

radio/src/gui/212x64/radio_setup.cpp
```diff
@@ -466,6 +470,20 @@ void menuRadioSetup(event_t event)
         break;
 #endif
 
+#if defined(PWR_BUTTON_PRESS)
+    case ITEM_RADIO_SETUP_PWR_ON_SPEED:
+        lcdDrawTextAlignedLeft(y, STR_PWR_ON_SPEED);
```
bsongis, line 475: cosmetics: indentation = 2 chars

radio/src/gui/212x64/radio_setup.cpp
```diff
@@ -466,6 +470,20 @@ void menuRadioSetup(event_t event)
...
+        lcdDrawNumber(RADIO_SETUP_2ND_COLUMN, y, g_eeGeneral.pwrOnSpeed + 1, attr|LEFT);
+        if (attr) CHECK_INCDEC_GENVAR(event, g_eeGeneral.pwrOnSpeed, -1, 2);
+        break;
+
+    case ITEM_RADIO_SETUP_PWR_OFF_SPEED:
+        lcdDrawTextAlignedLeft(y, STR_PWR_OFF_SPEED);
```
bsongis, line 481: cosmetics: indentation = 2 chars

radio/src/gui/128x64/startup_shutdown.cpp
```diff
@@ -30,9 +28,11 @@ const unsigned char bmp_sleep[]  = {
 #include "sleep.lbm"
 };
 
-void drawStartupAnimation(uint32_t duration)
+void drawStartupAnimation(uint32_t duration, uint32_t totalDuration)
 {
-  uint8_t index = limit<uint8_t>(0, duration / (PWR_PRESS_DURATION_MIN / 5), 4);
+  if (totalDuration == 0) return;
```
bsongis: newline

radio/src/gui/128x64/startup_shutdown.cpp
```diff
@@ -47,9 +47,11 @@ void drawStartupAnimation(uint32_t duration)
   lcdRefreshWait();
 }
 
-void drawShutdownAnimation(uint32_t duration, const char * message)
+void drawShutdownAnimation(uint32_t duration, uint32_t totalDuration, const char * message)
 {
-  uint8_t index = limit<uint8_t>(0, duration / (PWR_PRESS_SHUTDOWN_DELAY / 5), 4);
+  if (totalDuration == 0) return;
```
bsongis: newline

radio/src/opentx.cpp
```diff
@@ -1728,9 +1728,11 @@ void copyTrimsToOffset(uint8_t ch)
   storageDirty(EE_MODEL);
 }
 
+
 #if defined(STARTUP_ANIMATION)
-#define PWR_PRESS_DURATION_MIN        100 // 1s
-#define PWR_PRESS_DURATION_MAX        500 // 5s
+
+#define PWR_PRESS_DURATION_MIN() (2 - g_eeGeneral.pwrOnSpeed)
+#define PWR_PRESS_DURATION_MAX   500 // 5s
```
bsongis: `constexpr` if possible instead of `#define`

radio/src/targets/common/arm/stm32/bootloader/boot.cpp
bsongis: I think the code should be there for HORUS

radio/src/gui/480x272/widgets.h
```diff
@@ -69,7 +69,7 @@ int16_t editGVarFieldValue(coord_t x, coord_t y, int16_t value, int16_t min, int
 void drawMenuTemplate(const char * title, uint8_t icon, const uint8_t * icons=nullptr, uint32_t options=0);
 void drawSplash();
 void drawSleepBitmap();
-void drawShutdownAnimation(uint32_t duration, const char * message);
+void drawShutdownAnimation(uint32_t duration, uint32_t total_duration, const char * message);
```
bsongis: totalDuration

radio/src/opentx.cpp
```diff
@@ -1728,9 +1728,15 @@ void copyTrimsToOffset(uint8_t ch)
   storageDirty(EE_MODEL);
 }
 
+
 #if defined(STARTUP_ANIMATION)
-#define PWR_PRESS_DURATION_MIN        100 // 1s
-#define PWR_PRESS_DURATION_MAX        500 // 5s
+
+inline uint32_t PWR_PRESS_DURATION_MIN()
```
bsongis: perhaps uint8_t ?

bsongis: it seems the unit is not the same than PWR_PRESS_DURATION_MIN, it's prone to mistakes!

radio/src/opentx.cpp
```diff
@@ -2070,7 +2082,7 @@ uint32_t pwrCheck()
       if (g_eeGeneral.backlightMode != e_backlight_mode_off) {
         BACKLIGHT_ENABLE();
       }
-      if (get_tmr10ms() - pwr_press_time > PWR_PRESS_SHUTDOWN_DELAY) {
+      if (get_tmr10ms() - pwr_press_time > (uint32_t)PWR_PRESS_SHUTDOWN_DELAY()) {
```
bsongis: no cast here

bsongis (approved):

PR #6630, opened 2019-08-15
TOOLS scripts : chdir in script path 

3djc:
This allows TOOLS scripts that rely on several lua files, like Crossfire or BF, to work correctly


bsongis:
perfect, thanks!

PR #6637, opened 2019-08-18
revert changes in taranis ext module driver done while changing X10 e…

bsongis:
…xpress pulses


radio/src/targets/taranis/extmodule_driver.cpp
```diff
@@ -64,14 +64,14 @@ void extmodulePpmStart()
 
   EXTMODULE_TIMER->CR1 &= ~TIM_CR1_CEN;
   EXTMODULE_TIMER->PSC = EXTMODULE_TIMER_FREQ / 2000000 - 1; // 0.5uS from 30MHz
+  EXTMODULE_TIMER->ARR = 45000;
```
bsongis, line 67: it shouldn't change anything, yes?

bsongis:
Thanks!

PR #6640, opened 2019-08-18
New Multi protocols

pascallanger:
Protocols: ZSX and Flyzone


3djc:
Merci Pascal

raphaelcoeffic:
@3djc isn't missing something like:

```
  {MODULE_SUBTYPE_MULTI_ZSX,        0, false,      NO_SUBTYPE,            nullptr},
  {MODULE_SUBTYPE_MULTI_FLYZONE,    0, false,      NO_SUBTYPE,            nullptr},
```

in the corresponding tables?

pascallanger:
You don't need it since there is a catch all at the end:
```
  // Sentinel and default for protocols not listed above (MM_RF_CUSTOM is 0xff)
  {0xfe,                            0, false,      NO_SUBTYPE,            nullptr}
```

PR #6641, opened 2019-08-19
added missing pieces for MPM new protocols

raphaelcoeffic:
related to #6640


3djc:
True, suggesting minor string changes too (don't forget to prefix your branch !)

pascallanger:
You don't need it since there is a catch all at the end:
```
  // Sentinel and default for protocols not listed above (MM_RF_CUSTOM is 0xff)
  {0xfe,                            0, false,      NO_SUBTYPE,            nullptr}
```
You only need to add a line if there is a subtype or option is needed.
Pascal

3djc:
yes,  but if you add more later, it may come back and bite you :)

pascallanger:
Then you have some work to do to define all of them... None of the protocols which does not have subtype or option are defined.

pascallanger:
So for me you do them all or none but not half...

3djc:
You have a very valid point Pascal

raphaelcoeffic:
Right, seems the changes in the table are not needed: it’s using a search based on protocol ID, so no problem here. Let’s remove those without options.

3djc:
So we can close this and i'll  cherry pick the translation screen

raphaelcoeffic:
Companion changes are still needed ;-)

MRC3742:
I have all the changes done for Companion and will submit a PR later today.
Also I have updates for 2.2.4 if you still want to update the 2.2 branch?

Rich

bsongis:
@MRC3742 We have said there we wouldn't release any new 2.2 version. I hope we are not lying, it's difficult to know in advance...

radio/src/translations/untranslated.h
```diff
@@ -129,5 +129,5 @@
 #define LEN_DSM_PROTOCOLS              "\004"
 #define TR_DSM_PROTOCOLS               "LP45""DSM2""DSMX"
 
-#define LEN_MULTI_PROTOCOLS            "\006"
-#define TR_MULTI_PROTOCOLS             "FlySky""Hubsan""FrSky\0""Hisky\0""V2x2\0 ""DSM\0  ""Devo\0 ""YD717\0""KN\0   ""SymaX\0""SLT\0  ""CX10\0 ""CG023\0""Bayang""ESky\0 ""MT99XX""MJXq\0 ""Shenqi""FY326\0""SFHSS\0""J6 PRO""FQ777\0""Assan\0""Hontai""OpnLrs""FS 2A\0""Q2x2\0 ""Walk.\0""Q303\0 ""GW008\0""DM002\0""Cabell""Esy150""H8 3D\0""Corona""CFlie\0""Hitec\0""WFly\0 ""Bugs\0 ""BugMin""Traxas""NC1701""E01X\0 ""V911S\0""GD00X\0""V761\0 ""KF606\0""Redpin""Potens""ZSX\0  ""FLYZON"
+#define LEN_MULTI_PROTOCOLS            "\007"
```
MRC3742, line 132: I'm not sure you want to expand the Protocol to seven characters as the space in the X9D+ screen is limited to six? Not sure on the other Radios if this is also limited? Please review before changing.
Thanks - Rich

MRC3742:
@bsongis  Okay on 2.2.4
Are the nightlies still being built in this branch?
I don't mind updating it for the benefit of someone using this option.

radio/src/translations/untranslated.h
bsongis: I see this comment too late. It needs to be checked on 128x64 board, perhaps I pressed the button too quickly

bsongis:
I don't think we have 2.2 nightlies today

3djc:
We do if changes are pushed in 2.2 branch

bsongis:
If we keep nightlies alive, it means we know we are lying :) 

Either we want to release a new 2.2 (and we keep nightlies), either we stop nightlies (and pull requests). 

Difficult choice. I would vote for keeping 2.2 open, with nightlies and pull requests. And we just say that if 2.3 is ready soon, perhaps there will be no new 2.2 release

bsongis:
ok. we keep 2.2 open. but only for critical issues, so definitely not this one :)

PR #6655, opened 2019-08-22
Make crossfire, multi and luac baseline

3djc:
Sky9x do not get yet crossfire, as it needs fixing first


radio/src/targets/common/arm/CMakeLists.txt
```diff
@@ -156,18 +155,6 @@ if(MULTI_SPORT)
   add_definitions(-DMULTI_SPORT)
 endif()
 
-if(CROSSFIRE)
```
bsongis: why?

PR #6659, opened 2019-08-23
Telem as mainviews

### LINKED ISSUE #5156: Option to boot directly to telemetry screen
Seems there is no way to boot directly to a telemetry screen?
If we are able to customize a telemetry screen for our needs I think it should be possible to power the radio directly to that screen instead of booting to a main screen that doesn't have any of the info we need. There are times that I find myself pressing the page button before even powering the radio.
Looking on rcgroups for this option a few users showed interest in this.
Thanks

3djc:
- Telem (1 to 4) are now part of  mainViews
- Model setup includes a new settings that allow to select a specific screen at model load, or remember the last one used

This fixes #5156

Remains to be done
[ ] companion support


mcuprojects:
Hey so this hasn't been merged? Why? seems like a useful feature and it would actually simplify the UX

PR #6671, opened 2019-08-27
Change instance only for FrSky telemetry

3djc:
This fixes conversion of Crossfire sensors for example

### DIFF
--- a/radio/src/storage/conversions/conversions_218_219.cpp
+++ b/radio/src/storage/conversions/conversions_218_219.cpp
@@ -271,7 +271,7 @@ void convertModelData_218_to_219(ModelData &model)
 
   for (uint8_t i=0; i<MAX_TELEMETRY_SENSORS_218; i++) {
     newModel.telemetrySensors[i].id = oldModel.telemetrySensors[i].id;
-    if (oldModel.telemetrySensors[i].type == 0)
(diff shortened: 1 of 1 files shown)


raphaelcoeffic:
I’m still wondering why we are adding `0xE0`. I don’t have the feeling this is even necessary.

radio/src/storage/conversions/conversions_218_219.cpp
```diff
@@ -271,7 +271,7 @@ void convertModelData_218_to_219(ModelData &model)
 
   for (uint8_t i=0; i<MAX_TELEMETRY_SENSORS_218; i++) {
     newModel.telemetrySensors[i].id = oldModel.telemetrySensors[i].id;
-    if (oldModel.telemetrySensors[i].type == 0)
+    if (oldModel.telemetrySensors[i].type == 0 && (oldModel.moduleData[0].type ==  MODULE_TYPE_XJT_PXX1 || oldModel.moduleData[1].type == MODULE_TYPE_XJT_PXX1))
```
bsongis, line 274: this test could be done before entering into the loop

bsongis (changes requested):

bsongis (approved):
The PR has the rest of the function hidden, I didn't see that it also updated all other fields. Sorry for my negative comment, this PR is perfect

PR #6675, opened 2019-08-28
2.3 Fix BF scripts

### LINKED ISSUE #6580: Betaflight LUA - no telemetry
I use the X-Lite Pro with the full crossfire module and opentx 2.3 nightly, telemetry works well, but the BF LUA doesn't show values and says no telemetry.

With my X10S on opened 2.2 all was fine and the LUA worked with the same quad.

3djc:
This fix #6580

### DIFF
--- a/radio/src/lua/api_general.cpp
+++ b/radio/src/lua/api_general.cpp
@@ -432,8 +432,13 @@ When called without parameters, it will only return the status of the output buf
 
 static int luaSportTelemetryPush(lua_State * L)
 {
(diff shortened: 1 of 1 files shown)


bsongis (commented):
I would rather add a new test at the beginning of the function like this:
```
if (telemetryProtocol != PROTOCOL_TELEMETRY_FRSKY_SPORT) {
  lua_pushboolean(L, false);
  return 1;
}
```

PR #6687, opened 2019-08-31
Default modules to OFF for new models

3djc:
As per RCG discussion

### DIFF
--- a/radio/src/opentx.cpp
+++ b/radio/src/opentx.cpp
@@ -491,14 +491,16 @@ void modelDefault(uint8_t id)
   }
 #endif
 
-#if defined(HARDWARE_INTERNAL_MODULE)
+#if defined(HARDWARE_INTERNAL_MODULE) && defined(FRSKY_RELEASE)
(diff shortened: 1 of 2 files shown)


radio/src/opentx.cpp
```diff
@@ -491,14 +491,16 @@ void modelDefault(uint8_t id)
   }
 #endif
 
-#if defined(HARDWARE_INTERNAL_MODULE)
+#if defined(HARDWARE_INTERNAL_MODULE) && defined(FRSKY_RELEASE)
   g_model.moduleData[INTERNAL_MODULE].type = IS_PXX2_INTERNAL_ENABLED() ? MODULE_TYPE_ISRM_PXX2 : MODULE_TYPE_XJT_PXX1;
   g_model.moduleData[INTERNAL_MODULE].channelsCount = defaultModuleChannels_M8(INTERNAL_MODULE);
+#elif defined(HARDWARE_INTERNAL_MODULE)
```
bsongis: this code can be removed, the model structure is filled with 0 at the beginning of the function

bsongis (changes requested):

bsongis (approved):

PR #6688, opened 2019-08-31
Forward key events to fullscreen LUA widgets

### LINKED ISSUE #3821: Horus X12S N357 - LUA widget - key event
Is there no key event on widget lua scripting on horus?

Or do i miss something? How do you work with keys in a widget?

raphaelcoeffic:
closes #3821

### DIFF
--- a/radio/src/gui/480x272/layouts/layout1x1.cpp
+++ b/radio/src/gui/480x272/layouts/layout1x1.cpp
@@ -65,10 +65,13 @@ class Layout1x1: public Layout
       return zone;
     }
 
-    virtual void refresh();
+    virtual void refresh(event_t event);
+
+  protected:
+    void drawScreen();
 };
 
-void Layout1x1::refresh()
+void Layout1x1::refresh(event_t event)
 {
   theme->drawBackground();
 
@@ -85,8 +88,7 @@ void Layout1x1::refresh()
     drawMainPots();
     drawTrims(mixerCurrentFlightMode);
   }
-
-  Layout::refresh();
+  Layout::refresh(event);
 }
 
 BaseLayoutFactory<Layout1x1> layout1x1("Layout1x1", LBM_LAYOUT_1x1, OPTIONS_LAYOUT_1x1);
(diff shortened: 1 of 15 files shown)


raphaelcoeffic:
@lshems in case you'd like to test: now is the time :-)

raphaelcoeffic:
Here is what I used to test:
```lua
local function createWidget()
end

local function refreshWidget(widget, ev)
   local ev_txt = ""
   if ev == EVT_VIRTUAL_EXIT then
      ev_txt = "EXIT"
   elseif ev == EVT_VIRTUAL_ENTER then
      ev_txt = "ENTER"
   elseif ev == EVT_TELEM_BREAK then
      ev_txt = "TELEM BREAK"
   elseif ev == EVT_SYS_BREAK then
      ev_txt = "SYS BREAK"
   elseif ev == EVT_MODEL_BREAK then
      ev_txt = "MODEL BREAK"
   elseif ev == EVT_VIRTUAL_NEXT then
      ev_txt = "NEXT"
   elseif ev == EVT_VIRTUAL_PREVIOUS then
      ev_txt = "PREVIOUS"
   end
      
   lcd.drawText(100, 100, ev_txt, DBLSIZE);
end

return { name="TestEv", options={}, refresh=refreshWidget, create=createWidget }
```

lshems:
Nice. This night nightlies?

Still having a hard time compiling myself

raphaelcoeffic:
> Nice. This night nightlies?
> 
> Still having a hard time compiling myself

nope, at least not as long as it's not merged.

lshems:
Ok, so full screen only.

Could this be changed to first widget in main screen only?

Would add a lot more flexibility without adding complexity.

raphaelcoeffic:
Possibly, this here was the broad consensus, as far as I remember. It will probably be improved further in 2.4 to support focusing on a specific widget, which would then have the controls (thus losing the keys used to control focus change).

raphaelcoeffic:
@jfrickmann Would this work for your app?

jfrickmann:
> @jfrickmann Would this work for your app?

I think so. Thanks!

teckel12:
This would be a very powerful addition to full-screen widgets on the Horus!  Limiting it to full-screen widgets only removes any possible conflicts.  Can't think of a good reason why this OpenTX enhancement would cause any issues.

Nice work @raphaelcoeffic !

lshems:
Whats happening? No updates apparently?

kilrah:
Complete UI redesign in progress for 2.4.

lshems:
What's up?

PR #6711, opened 2019-09-05
Bsongis/rssi from 2 different modules

bsongis:
WIP


robustini:
I hope this solves the RSSI display on the X12S, it remains at 0 on both antennas with the night build of today.

kilrah:
Since that was merged 3 days ago it obviously doesn't, but that's not something we know about.

bsongis:
please open issue if anything wrong

PR #6733, opened 2019-09-10
Deactivate aux_serial on Horus targets

### LINKED ISSUE #4664: [Horus] Serial port options not implemented / Emergancy mode
The serial port options under the Hardware section is not working / implemented. When set to SBUS trainer the internal serial port header does not accept any S.Bus input. When set to Telemetry the Horus goes into Emergancy mode.

This might be related to https://github.com/opentx/opentx/issues/3954 if the internal UART header RX pin is the same as the heartbeat pin.

raphaelcoeffic:
closes #4664 

Please note that this still allows for using USB serial (CDC) for LUA or debuggging.


radio/src/targets/common/arm/stm32/aux_serial_driver.cpp
```diff
@@ -20,6 +20,8 @@
 
 #include "opentx.h"
 
+#if defined(AUX_SERIAL)
```
bsongis: may be removed?

brainbubblersbest:
Can anyone from the Team deeper involved in this, give a technical description. 
Why this Was never be as functional as in other newer Horus Based Radios?
-electrical issue
-processing issue related to timers/channels/streams or interrupts?
NO one from the User side understand why this Function Was never be fixed but Disabled in a Daily OTX Build.
Please.

kilrah:
Nobody motivated to do the work.

brainbubblersbest:
This is, by far, the best and honest answer i heard this Year 🤣
Would this be so much work, if it's working already on other Radios? For Auxiliary Serial Connection are coming more and more usecase scenarios these Days. But maybe this Radio is already too old and a bit rare, to take that much effort.

PR #6739, opened 2019-09-11
Bsongis/xlite ixjt flash broken

bsongis:


3djc:
Neeed in 2.3.0 too :)

Issue #6721, opened 2019-09-07
X10s Express - Text viewer crashes and causes a freeze


Quadpilotman:
When scrolling through the readme.txt in the firmware folder scrolling down causes a screen freeze

To Reproduce

Download the access firmware zip from the frsky site.
Unpack the files and copy over to the firmware directory on the sdcard.
Select radio setup menu with long sys key press.
Select the  SD-HC card with one short press.
Scroll down with the wheel to the FIRMWARE folder and select.
Scroll down the readme.txt file and select view text.
Scroll all the waydown the text viewer freezes the last line displayed is 

1. Further reduce the system latency

General information

OpenTX version 2.3.0N81
Hardware FrSky X10S Express

PR #6777, opened 2019-09-14
do not invalidate filenames buffer on EVT_REFRESH

raphaelcoeffic:
otherwise the filenames are reset to the end of list, whereby the indexes stay the same on [LONG ENTER] (pop-up menu).

This causes then funny things like that when scrolling further down the menu:
IMG_0122


raphaelcoeffic:
I just added another fix for the vertical offset when leaving the text viewer (while retesting that #6721 is still fixed).

bsongis:
to be ported to 2.3, thanks!

PR #6779, opened 2019-09-14
fixed gvar display when precision is used.

raphaelcoeffic:
Issue:
Screen Shot 2019-09-14 at 19 12 55
Screen Shot 2019-09-14 at 19 12 42

Now, as we don't have a small font that can be made bold, it is just small in this case.


bsongis:
to be ported to 2.3

PR #6781, opened 2019-09-14
check unique RX# based on protocol

### LINKED ISSUE #6765: RX #: false positive "same number" detection
Looks like the RX number detection doesn't make any difference of which module is used.

In the example below (red & blue rectangles), Companion doesn't make the difference between the internal XJT module and the external multi module.
tmp

raphaelcoeffic:
fixes #6765 

This should be enough now. We could of course start comparing the subType if multi-proto module is used but well...

### DIFF
--- a/companion/src/modelslist.cpp
+++ b/companion/src/modelslist.cpp
@@ -722,17 +722,17 @@ void TreeModel::refresh()
       }
       int protocol;
       QString rxs;
-      for (unsigned j=0; j<CPN_MAX_MODULES; j++) {
-        protocol = model.moduleData[j].protocol;
+      for (auto const& moduleData: model.moduleData) {
+        protocol = moduleData.protocol;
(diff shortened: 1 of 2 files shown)


companion/src/modelslist.cpp
```diff
@@ -746,15 +746,17 @@ void TreeModel::refresh()
...
+bool TreeModel::isModelIdUnique(unsigned modelIdx, unsigned protocol)
 {
   int cnt = 0;
   for (unsigned i=0; i<radioData->models.size(); i++) {
     ModelData & model = radioData->models[i];
     if (!model.isEmpty()) {
-      if (model.moduleData[module].protocol != PULSES_OFF && model.moduleData[module].modelId == modelIdx) {
-        if (++cnt > 1) {
-          return false;
+      for (unsigned j=0; j<CPN_MAX_MODULES; j++) {
```
bsongis:
```
for(auto moduleData: model.moduleData) {
...
}
```

PR #6792, opened 2019-09-15
Fix telemetry sensor module/rx source data being lost in some cases

kilrah:


3djc:
Tested ok on x7 where it would fail previously 👍

companion/src/firmwares/opentx/opentxeeprom.cpp
```diff
@@ -1941,7 +1940,7 @@ class SensorField: public TransformedField {
         if (model.moduleData[0].isPxx1Module() || model.moduleData[1].isPxx1Module())
           sensor.instance = (_instance & 0x1F) + (version <= 218 ? -1 : 0); // 5 bits instance
         else
-          sensor.instance = _instance;
+          sensor.instance = _instance & 0x1F;
```
bsongis: the sensors will move from the external module to the internal module, no?
I would say the mask is 0x9F

PR #6801, opened 2019-09-18
Flysky telem quick fix

### LINKED ISSUE #6800: 2.3 multi telemetry issue
For RCG

> I've put opentx 2.3.0 RC4 on my Taranis Q X7 and now with the multimodule, I get "RF signal critical" every time my receiver establishes a connection. I'm not sure what's changed internally in opentx but it seems what's happening is this:
Receiver connects, transmitter shows TRSS and A1 values on the model telemetry page
opentx sets the telemetry streaming flag
opentx reports "RF signal critical"
Receiver sends RSNR telemetry (<1s after connect), multiprotocol forwards it
opentx RSNR shows up, signal is fine

>

3djc:
A more major (and much better) rework is been prepared by @pascallanger , but in the meantime, this brings improves handling of some critical Flysky telem values

This fixes #6800

### DIFF
--- a/radio/src/telemetry/flysky_ibus.cpp
+++ b/radio/src/telemetry/flysky_ibus.cpp
@@ -91,7 +91,10 @@ static void processFlySkySensor(const uint8_t *packet)
   }
 
   if (id == FS_ID_SNR) {
-    if(value>0) value+=20;
+    if(value>0) {
+      value+=20;
+      telemetryStreaming = TELEMETRY_TIMEOUT10ms;
+    }
     telemetryData.rssi.set(value);
   }
 
@@ -125,7 +128,6 @@ void processFlySkyPacket(const uint8_t *packet)
     int index = 1 + (4 * sensor);
     processFlySkySensor(packet+index);
   }
-  telemetryStreaming = TELEMETRY_TIMEOUT10ms;
 }
 
 void processFlySkyTelemetryData(uint8_t data)
--- a/radio/src/telemetry/flysky_ibus.cpp
+++ b/radio/src/telemetry/flysky_ibus.cpp
@@ -91,6 +91,7 @@ static void processFlySkySensor(const uint8_t *packet)
   }
 
   if (id == FS_ID_SNR) {
+    if(value>0) value+=20;
     telemetryData.rssi.set(value);
   }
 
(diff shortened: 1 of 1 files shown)


radio/src/telemetry/flysky_ibus.cpp
```diff
@@ -99,7 +100,9 @@ static void processFlySkySensor(const uint8_t *packet)
     if (sensor->id == id) {
       // The Noise and Signal sensors that are specified in dB send the absolute value
       if (id == FS_ID_NOISE || id == FS_ID_RSSI)
-        value = -value;
+        value = 135 - value;
+      else if (id == 0xfe) // ERR RATE
```
bsongis: perhaps an ID instead of 0xfe (please use upper case for hexa)

CapnBry:
I don't know if it is too late to comment on this, but while setting the `telemetryStreaming` only when RSNR is received is a good fix, I don't agree with altering the telemetry values being received. The way it is now, they match up with what's being sent, and what you'd see on a FlySky receiver. I believe it should stay like the original:
```
if (id == FS_ID_NOISE || id == FS_ID_RSSI)
        value = -value;
```
And change the `telemetryStreaming` flag like this:
```
diff --git a/radio/src/telemetry/flysky_ibus.cpp b/radio/src/telemetry/flysky_ibus.cpp
index d4b00c995..f140e852a 100644
--- a/radio/src/telemetry/flysky_ibus.cpp
+++ b/radio/src/telemetry/flysky_ibus.cpp
@@ -90,8 +90,9 @@ static void processFlySkySensor(const uint8_t *packet)
     id = 0x100;
   }

-  if (id == FS_ID_SNR) {
-    telemetryData.rssi.set(value);
+  if (id == FS_ID_ERR) {
+    telemetryStreaming = TELEMETRY_TIMEOUT10ms;
+    telemetryData.rssi.set(100-value);
   }

   for (const FlySkySensor * sensor = flySkySensors; sensor->id; sensor++) {
@@ -122,7 +123,6 @@ void processFlySkyPacket(const uint8_t *packet)
     int index = 1 + (4 * sensor);
     processFlySkySensor(packet+index);
   }
-  telemetryStreaming = TELEMETRY_TIMEOUT10ms;
 }

 void processFlySkyTelemetryData(uint8_t data)
```

pascallanger:
Have a look at my branch for comments instead since this is what will be pushed in the end when everything will be tested: https://github.com/pascallanger/opentx/blob/sweet-16-squash/radio/src/telemetry/flysky_ibus.cpp
Pascal

PR #6817, opened 2019-09-23
Fix new switches

3djc:


bsongis:
Fix confirmed on X9D+ 2019

Issue #6821, opened 2019-09-24
Skin of the X9-Lite Companion simulator


Ceeb182:
Currently the skin of the `X9-Lite` Companion simulator is the same as the one of the `X7`. 
I designed a new skin for the `X9-Lite` simulator.
![X9-Lite Image for Companion](https://zupimages.net/up/19/39/jz20.png "X9-Lite")
I'm not a designer, I wouldn't be angry if the skin isn't chosen.

As this is the first time I'm proposing a contribution, I hope I'm not too wrong about OpenTx GitHub Workflow. I created a branch `ceeb182/simuimages` with all the image files on my local OpenTx repository.

Ceeb182:
Thank you for using the skin of the X9-Lite. 
Martin Hotar's skins are much better made. So feel free to use his version if he offers one.

There is a cosmetic change to be made.
The simulator screen is not quite vertically centered, there is a one-pixel shift. 
![X9-Lite Image for Companion](https://zupimages.net/up/19/40/p39y.png "X9-Lite with one pixel offset")

Ceeb182:
I close this issue because I open a new one which is #6961

PR #6822, opened 2019-09-24
#6821 - Skin of the X9-Lite Companion simulator

Ceeb182:
I think the graphic part of the of the `X9-Lite` skin is ready to be used.


bsongis:
Thanks, would you open a PR for 2.3.0 which is almost released?

Ceeb182:
Excuse-me but I didn't understand what I had to do now ? A new PR (Pull Request) ?

3djc:
Maybe leave it with us.  @bsongis maybe it is better if we fix the related coding before merging into 2.3.0 ?

PR #6832, opened 2019-09-26
Check multi signature before flashing

3djc:
Prevent flashing a wrong file to multi. If wrong file selected, bring an info window on required file type

#6888 is a pre-requisite since it handles the new v2 signatures


3djc:
Not needed anymore, covered by already merged PR

Issue #6856, opened 2019-09-29
X9D Systemmenue/Hardware: serial port function can not be accessed


Allerhopp:
X9D 2013
Highlighting and accessing serial port function is not possible in the radio system menue. It is blocked like RTC voltage above, where it makes sense.

Changing the setting with companion works.

kilrah:
The 2 debug "buttons" can't be selected either

kilrah:
X9D only, X9D+ is OK

PR #6857, opened 2019-09-29
Fix shift in sources on X9DP pre-2019 (fix #6860 and #6856)

### LINKED ISSUE #6860: From Mike: Shift in X9D/X9D+/X-lite sources
Telemetry Display Values X9D Plus - 2.3.0.
“Time” displays TX battery value
“GPS” displays TX time. Not sure where "GPS" is coming from because there is no "GPS" sensor showing on the Telemetry page. 
Not sure if there are other examples.

kilrah:
Issue #6860 is related to SI. Solved by declaring SI for all X9D and relying on SWITCH_EXISTS() to hide it on non-2019 (default to none, and is not editable).


kilrah:
This breaks the radio_hardware page.
Not sure how to go about it, knowing it's already broken for X9D (not plus) before the fix...

kilrah:
OK this version doesn't break hardware page, fixes #6856 at the same time, and fixes some mistakes in the values used for array sizes in datastructs.

kilrah:
Original Xlite was actually affected too, and is now also corrected

PR #6865, opened 2019-09-30
Shift sensor display by +1 to match old display value

3djc:


companion/src/firmwares/opentx/opentxeeprom.cpp
```diff
@@ -1914,7 +1914,7 @@ class SensorField: public TransformedField {
       if (sensor.type == SensorData::TELEM_TYPE_CUSTOM) {
         _id = sensor.id;
         _subid = sensor.subid;
-        _instance = (sensor.instance & 0x1F) | (sensor.rxIdx << 5) | (sensor.moduleIdx << 7);
+        _instance = ((sensor.instance & 0x1F) - 1) | (sensor.rxIdx << 5) | (sensor.moduleIdx << 7);
```
kilrah: ((sensor.instance -1) & 0x1F) or you truncate!

PR #6888, opened 2019-10-03
Added support for flashing AVR multi on horus targets

### LINKED ISSUE #6859: Multi : use new binary signature

raphaelcoeffic:
~~Taranis-like targets not yet implemented:~~
~~- need to share EXTI9_5_IRQn with rotary encoder & heartbeat detection~~

Tested on:
- [x] T16
- [x] X9 lite

This needs to be tested on other Taranis-like hardware, especially:
- [x] X9D(+)
- [x] X7

Both these targets use shared interrupts between combinations of:
- rotary encoder,
- heartbeat detection
- and now the software serial for module update.

This closes #6859


raphaelcoeffic:
This needs to be tested on other Taranis-like hardware, especially:
- [ ] X9D(+)
- [ ] X7

Both these targets use shared interrupts between combinations of:
- rotary encoder,
- heartbeat detection
- and now the software serial for module update.

raphaelcoeffic:
now includes the changes contained in #6832

Issue #6873, opened 2019-09-30
2.3.0 on Taranis  QX7 Switches Si and SJ for p400 and p401. Switch changes are not detected


pelle17b:
**Describe the bug**
I have a Taranis X7 that is modified with 2 position toggle switches connected to p400 and p401. I have configured Si and SJ as 2POS in the HARDWARE menu.

1. No matter what position the switches are in I get a switch warning for Si and SJ when i start the radio.
2. If I configure any of the Si and SJ switches as the SOURCE in the INPUTS menu for a model, there is no reaction when the configured switch is toggled. 
 
**To Reproduce**
As above

**Expected behavior**
To be able to detect changes for Si and SJ switches similar to other switches. 

**General information**
 - OpenTX version 2.3.0
 - Hardware Taranis X7

bsongis:
Would you please check that this issue is fixed in the next nightly build? Thanks!

pelle17b:
Will do. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

pelle17b:
Tested with Nightly 2.3 1N3.
Still both Si and SJ are configured as 2POS in the HARDWARE menu.

Si works as expected as far as I can see. 
- Switch changes are detected in INPUTS menu as expected, and it is working when used in LOGICAL SWITCHES menu.
- Si is present and working in HARDWARE - Debug - [KEYS] menu
- Switch warning is working as expected for Si when radio is started and models changed (tested both ways and both with warning triggered and without).

SJ does not work.
- It is not present in the HARDWARE - Debug - [KEYS] menu
- Switch changes for SJ are not detected in INPUTS menu or if it is used in LOGICAL SWITCHES menu.
- Switch warning is gone when radio is started (i.e. not as before), but it does not appear if SJ is flipped.

3djc:
Yes, sorry about that. SJ was looking for the wrong pin, fixed it, will be in next nightly

PR #6889, opened 2019-10-03
Fix SI and SJ switches on Taranis Q X7 #6873

alexeystn:
Switches SI and SJ connected to P400 and P401 in Taranis Q X7 now operate as expected.
Switch SJ did not appeared in Modes/Setup S-Warning list, so startup warning for it couldn't be configured.

### DIFF
--- a/radio/src/gui/128x64/model_setup.cpp
+++ b/radio/src/gui/128x64/model_setup.cpp
@@ -612,7 +612,7 @@ void menuModelSetup(event_t event)
           }
 
           int current = 0;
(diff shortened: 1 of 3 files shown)


3djc:
That is not the right fix, as we also added to switches to other size screens, please open an issue so that a fix for eveything can be done

3djc:
Note to self, never review PR before you had your SECOND cup of coffee. You fix is perfect, so much so that it should be appied to 2.3.1 directly

PR #6909, opened 2019-10-07
Bsongis/x9d p3 fix

bsongis:


3djc:
Tested ok on simu, don't have x9d hardware

PR #6913, opened 2019-10-09
Fix companion D8 import/export

### LINKED ISSUE #6912: Under X9D-2.3.1 in companion, models with XJTD8 in .otx becomes XJTD16
**Describe the bug**
In 2.3.1  after a transfer of .otx file in a X9D, all models using a XJTD8  internal are converted to XJTD16

**To Reproduce**
1-From companion open a model with internal RF XJTD8
2. run the simu you will see D16 

D8-D16

3djc:
This fixes #6912 

@raphaelcoeffic Could you review that since that area is your code, just want to make sure I did not miss something


3djc:
Closing as the issue appears bigger than anticipated

PR #6916, opened 2019-10-09
Use vertical axis for field/value changes and horizontal for Nav

3djc:
In previous PR relating to virtual event, navigatoin was defined as such

image

it appears that this is found not natural and an uneeded change. The new proposed defined navigation is bellow

image


Matze-Jung:
Maybe try this [fork](https://github.com/Matze-Jung/betaflight-tx-lua-scripts#mapping-navigation-controls). It maps the navigation controls on a per platform base and extends configurability that way.

PR #6937, opened 2019-10-12
Add a sinus curve (1 Hz) as source for automatic testing

raphaelcoeffic:
This is available when compiling with DEBUG_LATENCY=MIXER_RF.


radio/src/dataconstants.h
```diff
@@ -670,6 +670,10 @@ enum MixSources {
   MIXSRC_CYC2,                         LUA_EXPORT("cyc2", "Cyclic 2")
   MIXSRC_CYC3,                         LUA_EXPORT("cyc3", "Cyclic 3")
 
+#if defined(DEBUG_LATENCY_MIXER_RF)
+  MIXSRC_DEBUG_SINUS,
```
bsongis, line 674: Why not at the end of the list?

raphaelcoeffic: I had troubles to have it included in the sources accepted by the mixer screen. Otherwise the patch would have been much bigger (changes everywhere in UI, etc).

kilrah: There should be only one place, isSourceAvailable()

raphaelcoeffic: Should, but it does not help, as this only works for the correct categories. So, placing this at the end requires including more categories of inputs, or even “creating” one.

raphaelcoeffic:
keepalive ping!

raphaelcoeffic:
Ping

PR #6950, opened 2019-10-14
Companion : fix channel numbers

### LINKED ISSUE #6927: Companion : wrong channel number limit for LR12
Actually, probably wrong for many protocols

3djc:
This fixes #6927

### DIFF
--- a/companion/src/modeledit/setup.cpp
+++ b/companion/src/modeledit/setup.cpp
@@ -366,6 +366,48 @@ void ModulePanel::setupFailsafes()
   lock = false;
 }
 
+ int ModulePanel::getMaxChannelCount()
+{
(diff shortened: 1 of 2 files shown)


companion/src/modeledit/setup.cpp
```diff
@@ -366,6 +366,48 @@ void ModulePanel::setupFailsafes()
   lock = false;
 }
 
+int ModulePanel::getMaxChannelCount()
+{
+  const PulsesProtocol protocol = (PulsesProtocol)module.protocol;
+  switch (protocol) {
+    case PULSES_PXX_R9M:
```
kilrah, line 375: Sure of that one?

PR #6951, opened 2019-10-14
Display rxnum in modellist only when needed

### LINKED ISSUE #6926: Companion : PPM module show rxnum in modellist

3djc:
This fixes #6926

### DIFF
--- a/companion/src/firmwares/moduledata.cpp
+++ b/companion/src/firmwares/moduledata.cpp
@@ -59,7 +59,7 @@ bool ModuleData::isPxx1Module() const
 
 bool ModuleData::supportRxNum() const
 {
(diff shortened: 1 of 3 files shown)


MRC3742:
@3djc 
Shouldn't this include the External DSM Hack Module Protocols of LP45, DSM2 and DSMX.
These had model match capability with RX numbers from 0 to 20.
Building Companion from the current 2.3 branch (10-16-2019) these RX model numbers are all blank?

3djc:
Yes they should be added to my changes which I'm doing right now.

MRC3742:
Thanks - your fix makes these RX numbers now visible in Companion.

PR #6958, opened 2019-10-16
IT audio updated

robustini:
Added audio for lost and recovered trainer signal.


robustini:
I close it, there are still other things missing, I try to complete it.

PR #6964, opened 2019-10-18
Fixed the issue that the power option changes when R9M (without acces…

WismyYao:
Fixed the issue that the power option changes when R9M (without access) bind with 'ch1-8 tele ON' option.


3djc:
Tested on x7 (where the bug does not exist with R9M) : still ok
Tested on xlite where the issue did exist with r9m lite : fixed

bsongis:
Thanks to you both

PR #6976, opened 2019-10-20
French translation update

LapinFou:
Here are the latest translations.
One question, what is `Act` ??


kilrah:
Activity. Goes false after backlight timeout

LapinFou:
Oh ! OK.
So,
Act is TRUE when the backlight is ON
Act is FALSE when the backlight is OFF

Correct ?

kilrah:
Yes

LapinFou:
Thanks for the clarification.
👍

PR #6981, opened 2019-10-21
Wismy/x10(s) upgrade kit

WismyYao:
x10(s) upgrade kit


radio/src/thirdparty/FrSky/libAccess.h
```diff
@@ -0,0 +1,16 @@
+/*
+ * Copyright (C) FrSky RC, Inc - All Rights Reserved
+ *
+ * Unauthorized copying of this file, via any medium is strictly prohibited
+ * Proprietary and confidential
```
raphaelcoeffic: What is the license under which OpenTx is granted permission to use this file and the corresponding library? We need something more permissive. It must be at least perpetual and authorize any use OpenTx are fit.

raphaelcoeffic (changes requested):

radio/src/thirdparty/FrSky/libAccess.h
bsongis: Here is the new header:
```
 /* 
  * Copyright (C) FrSky Electronic Co., Ltd. - All Rights Reserved 
  * 
  * This library is only allowed to run on FrSky products.
  * Any other product needing this library file must obtain FrSky's permission first.
  *
  */
```
To make it clear, FrSky wants that ACCESS runs only on their system

PR #6984, opened 2019-10-22
[2.3.1] IT TTS update

robustini:
Update IT TTS voice files, I ask for verification.


robustini:
Requires an update, I close.

PR #6991, opened 2019-10-23
IT language and TTS update for OpenTX V2.3.x

robustini:
Various additions and fix, IT should be complete, i'm sorry for yet another editing...


robustini:
@bsongis can you put your eyes here?
Fails checks on ARM9X, dunno why...

3djc:
> @bsongis can you put your eyes here?
> Fails checks on ARM9X, dunno why...

Travis glitch, relaunched the process

robustini:
Wait for any merge, I'm doing checks in each section.

robustini:
Thanks @kilrah, I'm arguing with the spaces in the SYS/HARDWARE menu, already corrupted in 2.3.1, when I finish all the check I write it.

robustini:
Everything now seems OK, but I can't check on X9D since I don't own it.
A verification with tx without the color graphic display would be welcome.
@bsongis it is better to delete my changes to the 2.3 branch yesterday because the HARDWARE section was corrupt.

3djc:
You can check with either simu or libsimulator ;)

robustini:
> 
> 
> You can check with either simu or libsimulator ;)

True, now I do it.

robustini:
Also verified on X9D and X7, the latter unfortunately having such a small display does not allow too long writings, there's some overlap of some character, but before it was the same or worse.
To adapt it perfectly to this display too, the various wordings would be reduced too, unfortunately the Italian in many terms is much longer than English.
It would be useful to have two sets of separate languages, or create other conditions within the existing one to overcome this limit.
For the rest I can't find any issues, fingers crossed...

robustini:
From that I understand these conditions exist already in the DEFINE, correct me if I'm wrong, I don't know if they are applicable everywhere.
Example:

// For all the tx (unified)
#define TR_BATTERYWARNING      INDENT"Avviso batteria"

// For low and full graphics (separate)
#define TR_BATTERYWARNING      TR(INDENT "Avv. Batt.", INDENT "Avviso Batteria")

Is correct and applicable everywhere?
If the syntax is correct I've to update some phrases for the small displays, like the one above.

kilrah:
Yes that's correct.

robustini:
Ok, now it should be okay...

robustini:
I find nothing else anomalous, everything is translated.
I will work on it further after the release candidate of 2.3.2 and users feedback.

robustini:
I close it and open another one with global changes.

PR #6996, opened 2019-10-24
 WIP: Joystick input

annerajb:
Hello,

I am updating this old branch from https://github.com/pickardjoe/opentx my intention is to have it functioning again. 

I have a Taranis xlite pro on the way.

In the meantime can I get some feedback since I just took the original code and put it where it was in the first place (which may not be the correct location nowadays).
How can I add a test so that nightly?/on commit the USB_CONTROL flag is compiled (to make sure it at least compile in the future)

Outstanding todos on this PR:

- Some sort of CRC/Checksum
- Documentation of the protocol
- Add support on a few open-source tools to send the input from a pc input device to the radio. 

Let me know of recommendations or suggestions.


kilrah:
This will have to target the next major version (2.4) due to data changes, with proper conversion and companion support. 
I think this mode should be a selection in the "serial port mode" setting in radio/hardware.

bsongis:
Right, please open a PR against the branch 2.4 (or bsongis/X10_new_UI)

PR #7013, opened 2019-10-27
Update it.h.txt

robustini:
1) Better global alignment with the EN version (removed INDENT not needed)
2) Fixed some items
3) Fixed some spaces

@bsongis, please can you do a merge? Thanks!


bsongis:
Right in 2.4 INDENT will be removed completely

PR #7020, opened 2019-10-29
Add support for BT compilation on T16

### LINKED ISSUE #7010: Telemetry mirror and bluetooh for Jumper T16
I am following the development for Jumper T16. My request is if you can have "mirror telemetry", either by physical port, or by a bluetooh module. Right now I have a bluetooh module installed for Horus 10 on my Jumper T16, the visibility of the bluetooh works, it is even possible to change the name but the telemetry does not come out. Greetings

3djc:
This fixes #7010


Guillesan:
I have tried the latest version 2.3.2.33, already with the possibility of using bluetooh. In opentx configuration in the station the module (name, etc) is effectively configured, in the telemetry option. However, the bluetooh's visibility does not work if it is recognized by my android tablet, but using a bluetooh connection review app informs me that the jumper module does not have the serial option activated ....

kilrah:
It's a BLE module, no serial.

Guillesan:
Of course it is a BLE, but to obtain telemetry the module must necessarily have the active serial service.  but you don't have it, so you can't extract telemetry data to mission Planner or another program

kilrah:
The software needs to support using the BT GATT APIs to connect to these modules, not serial.

Issue #1536, opened 2014-07-19
Easy reorder of Logical Switch and Special Functions in Companion


shadowjig:
Can a feature be added to allow the user to reorder Logical Switch and Special Functions in Companion?  Maybe a 1,2,3,4,etc order or a drag and drop reorder?

Thanks.

elecpower:
Great idea.

To address the resultant broken/errant LS references raised by @nigelsheffield you could take the spreadsheet approach where the app moves and updates all affected references. A not so small code change especially when it would need to check and update throughout a model's settings but it would at least not require adding other references.

PR #7025, opened 2019-10-30
Companion new actions for logical switches context menu

elecpower:
New functions: 
- Insert - insert a blank row at current row and move all following rows down one row (last row must be empty otherwise menu item disabled
- Delete - delete current row and move all following rows up one row
- Move Up - swap current and previous rows
- Move Down - swap current and next rows
- Clear - clear current row (was Delete)
- Clear All

Code housekeeping:
- change custom switch to logical switch references
- change dialog field prefixes
- cosmetics

Further enhancements:
- used existing icon png files for Insert Row and Delete Row but these are not consistent with all themes (@mhotar maybe able to help here)

Related Issues #1536

### DIFF
--- a/companion/src/modeledit/logicalswitches.cpp
+++ b/companion/src/modeledit/logicalswitches.cpp
@@ -35,19 +35,22 @@ LogicalSwitchesPanel::LogicalSwitchesPanel(QWidget * parent, ModelData & model, 
   const int srcGroups = firmware->getCapability(GvarsInCS) ? 0 : (RawSource::AllSourceGroups & ~RawSource::GVarsGroup);
   rawSourceItemModel = new RawSourceFilterItemModel(&generalSettings, &model, srcGroups, this);
 
+  lsCapability = firmware->getCapability(LogicalSwitches);
+  lsCapabilityExt = firmware->getCapability(LogicalSwitchesExt);
+
   QStringList headerLabels;
   headerLabels << "#" << tr("Function") << tr("V1") << tr("V2") << tr("AND Switch");
-  if (firmware->getCapability(LogicalSwitchesExt)) {
+  if (lsCapabilityExt) {
     headerLabels << tr("Duration") << tr("Delay");
   }
-  TableLayout * tableLayout = new TableLayout(this, firmware->getCapability(LogicalSwitches), headerLabels);
+  TableLayout * tableLayout = new TableLayout(this, lsCapability, headerLabels);
 
   s1.report("header");
 
   const int channelsMax = model.getChannelsMax(true);
 
   lock = true;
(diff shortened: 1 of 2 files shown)


kilrah:
Don't think we want to do that if all LS references aren't being updated, becasue that's what people will expect would happen. And we've always considered that going down this rabbit hole might be risky maintenance wise...

elecpower:
But there exists the issue of broken references across the App now. So should the functionality that creates this situation be removed until there is an ideal solution? I'm reasonably certain there would be many complaints.
My thought in submitting the PR (and was planning to add to other data sets) is to add the base functionality with the current caveat of breaking links.
Agree it would be preferable to maintain referential integrity but when will someone have time to go down the rabbit hole...

kilrah:
> But there exists the issue of broken references across the App now

Which ones?

elecpower:
Inputs and Mixes support the extra actions proposed in this PR plus anywhere clear/delete and cut are permitted actions (Flight Modes, Outputs, Curves, Logical Switches). None of these existing actions, that I can find, chase and update references. The results of these broken references can vary considerably when operating a model.
Even manually re-keying/reordering settings can result in broken references.
There have been Issues logged for the ability to reorder and clear/delete telemetry sensors (currently this can be a painful re-keying exercise).
Converting between radios can break references so there is a warning. Major upgrades come with warnings.
There is a general caveat that users should confirm their models react as expected on the ground. Simulator to the rescue plus physical pre-operation checks.
A major attraction of OpenTX is the ability to get under the hood and home bake your model settings and most users know this comes with risks.

kilrah:
> Inputs and Mixes 

Editing these does not require changing references in other places. If you cut/paste a mixer line you're not going to have an LS have to change from using CHx to CHy (unless that's specifically what you want). Shifting an LS does pretty much always require updating all references to it.

elecpower:
> Editing these does not require changing references in other places

Well not always. When all rows of a virtual input are moved/deleted any reference in for example a mix line to the empty input is "broken" as the mix is not updated. Same applies if all a channel's mixes are moved/deleted. Any references to the now empty channel are "broken". The receiver mapped channels are less prone to this issue but "calculation" channels can have lots of edits and rework.

kilrah:
Yes but that's a very rare thing to do, and you know you're doing it and will have to change things - you're doing it precisely because you want to change things.

Shifting an LS is the opposite, it's only cosmetic and it will be expected that no behavior changes.

elecpower:
So looks like I will have to venture down the formidable rabbit hole...

elecpower:
In the process of finalising a new PR. Closing this PR due to the amount of code changes and conflicts since submitted.

PR #7031, opened 2019-11-01
Joystick input based on 2.4 branch. 

annerajb:
Moved this to 2.4 branch.
Also added a docker for running the travis test locally on the same or close to equivalent environment.


annerajb:
Ok so after some learning of the code I am leaning for adding a new button on the companion simulator in the left side. 

This would reuse the existing joystick code in that menu/widget for setup of the joystick channels. 

Then add a new class which receives the signal for axis changed emitted by the existing joystick.cpp class.
This new class would handle (if toggled on) the sending of the data to the radio thru uart. 
For selecting which uart port to use to transmit will reuse the existing setting (if it makes sense) that already exist in the UI for the avrdude menu (if I recall correctly)
Any suggestions or feedback for this approach before I start it?

PR #7046, opened 2019-11-04
IT: fixed some incorrect layouts

robustini:
1) Cosmetics
2) Fixed some incorrect layouts in the antennas selection (X12S)


robustini:
Need rework...

PR #7049, opened 2019-11-05
Wismy/x10(s) upgrade kit

WismyYao:
add a popup for auth-failure


radio/src/telemetry/frsky_pxx2.cpp
```diff
@@ -227,6 +227,11 @@ void processAuthenticationFrame(uint8_t module, uint8_t * frame)
   uint8_t cryptoType = frame[3];
   uint8_t messageDigest[16] = {0};
 
+  if (0x04 == frame[0] && 0xA5 == frame[4]) {
```
raphaelcoeffic: Could you please add “defines” for your constants?

WismyYao: Which file should I put the 'defines' in ? How about the strings ?

raphaelcoeffic: `pulses/pxx2.h` would be a good place, next to the other frame defs. Thx!

bsongis: I will take it

radio/src/telemetry/frsky_pxx2.cpp
```diff
@@ -227,6 +227,11 @@ void processAuthenticationFrame(uint8_t module, uint8_t * frame)
...
+    POPUP_INFORMATION("Auth-failure");
```
bsongis: Missing translations

radio/src/translations/de.h.txt
```diff
@@ -1157,6 +1157,11 @@
 #define TR_MENU_INVERT                 "Invertieren<!>"
 #define TR_JITTER_FILTER               "ADC Filter"
 #define TR_RTC_CHECK                   TR("Check RTC", "Check RTC voltage")
+
+#if defined(ACCESS_LIB)
```
bsongis: We don't add so many ifdef here

radio/src/translations.h
```diff
@@ -910,6 +910,10 @@ extern const char STR_ABOUT_PARENTS_2[];
 extern const char STR_ABOUT_PARENTS_3[];
 extern const char STR_ABOUT_PARENTS_4[];
 
+#if defined(ACCESS_LIB)
```
bsongis: The compiler can optimize this if not needed

radio/src/translations.cpp
```diff
@@ -770,3 +770,7 @@ const char STR_ABOUT_PARENTS_1[] = TR_ABOUT_PARENTS_1;
 const char STR_ABOUT_PARENTS_2[] = TR_ABOUT_PARENTS_2;
 const char STR_ABOUT_PARENTS_3[] = TR_ABOUT_PARENTS_3;
 const char STR_ABOUT_PARENTS_4[] = TR_ABOUT_PARENTS_4;
+
+#if defined(ACCESS_LIB)
```
bsongis: Here as well, no if defined

radio/src/telemetry/frsky_pxx2.cpp
bsongis: Now I read it more carefully, the 0x04 is not a flag, it's just a sanity check, I have changed it to `frame[0] >= 4`

PR #7050, opened 2019-11-05
Mixer scheduler

raphaelcoeffic:
Re-work of the mixer scheduling. It uses a separate timer (TIM13) to trigger a CoOS flag, which is waited for in the mixer loop.

TODO:
- [x] fix/test on X9D(+)
- [ ] deal with `SKY9X`

### DIFF
--- a/radio/src/mixer_scheduler.h
+++ b/radio/src/mixer_scheduler.h
@@ -18,10 +18,16 @@
  * GNU General Public License for more details.
  */
 
-#ifndef _MIXER_SCHEDULER_DRIVER_H_
-#define _MIXER_SCHEDULER_DRIVER_H_
+#ifndef _MIXER_SCHEDULER_H_
+#define _MIXER_SCHEDULER_H_
+
+#define MIXER_SCHEDULER_DEFAULT_PERIOD_US 4000u // 4ms
+
+#define MIN_REFRESH_RATE      4000 /* us */
+#define MAX_REFRESH_RATE     25000 /* us */
 
 #if !defined(SIMU)
+
 // Call once to initialize the mixer scheduler
 void mixerSchedulerInit();
 
@@ -44,6 +50,12 @@ void mixerSchedulerEnableTrigger();
 // Disable the timer trigger
 void mixerSchedulerDisableTrigger();
 
+// Fetch the current scheduling period
+uint16_t getMixerSchedulerPeriod();
+
+// Trigger mixer from an ISR
+void mixerSchedulerISRTrigger();
+
 #else
 
 #define mixerSchedulerInit()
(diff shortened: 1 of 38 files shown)


ctzsnooze:
Hi Raphael - this is great work - any chance for a QX7 binary for testing?

JyeSmith:
Hey @raphaelcoeffic I'm also keen to try it out when you're ready for testers (QX7).

raphaelcoeffic:
@ctzsnooze @JyeSmith Guys, I just rebased on top of current 2.3 branch. Seems I still have to fix the lag/refresh rate display for crossfire. After that we can have a test version, but please take into account that this feature has been refused for 2.3 and will be integrated first into 2.4.

AlessandroAU:
Keen to get this working in order to improve latency!

JyeSmith:
@raphaelcoeffic what is the best way implement the heartbeat on a module?  Say you want to request packets at 200Hz?

How exactly does it work 😅

raphaelcoeffic:
@JyeSmith the module needs to report two things: the requested period (and thus frequency), and the current lag.

After that, the frequency is adapted to reduce lag and be equal the requested frequency.
The telemetry packets sent by multimodule and crossfire can be used as an example (crossfire is probably best as it contains only these data).

AlessandroAU:
What is the current lag? The time difference between when the TX module wants to send the packet and when it is received from opentx?

raphaelcoeffic:
@AlessandroAU for you module, as it implements the crossfire protocol, all you need is to send the proper telemetry packets (should be easy to find in this MR). Let me know if you need any help.

raphaelcoeffic:
> What is the current lag? The time difference between when the TX module wants to send the packet and when it is received from opentx?

Yes, exactly.

AlessandroAU:
I'm a bit draft and still can't seem to figure out it,

can you give me an example of what the tlm packet would look like?

Edit: This format for the 'sync packet' seems to work 
image

AlessandroAU:
@raphaelcoeffic Any recommendations of how often the sync update packet should be sent to the TX?

radio/src/mixer_scheduler.h
```diff
@@ -0,0 +1,80 @@
...
+ * GNU General Public License for more details.
+ */
+
+#ifndef _MIXER_SCHEDULER_H_
+#define _MIXER_SCHEDULER_H_
+
+#define MIXER_SCHEDULER_DEFAULT_PERIOD_US 4000u // 4ms
+
+#define MIN_REFRESH_RATE      4000 /* us */
+#define MAX_REFRESH_RATE     25000 /* us */
```
JyeSmith, line 27: Is it possible to have a longer MAX_REFRESH_RATE?  ExLRS goes as low as 25Hz and I believe Ghost goes down to 15Hz.

robertlacroix:
@raphaelcoeffic anything we can help with to get this mergeable?

3djc:
Outdated, so closing it. Merge is currently planed for 2.4

PR #7070, opened 2019-11-09
Internal multimodule support added for T12.

qba667:
Internal multiprotocol module implemented for T12.
Mostly affected in model setup, I made i a little bit more flexible - approach same as for T16.


3djc:
Hello Jakub. We have our own implementation on the way, finalysing some details with JumperRC

PR #7092, opened 2019-11-17
Try to contact MPM with inverted and non-inverted RX.

raphaelcoeffic:
This an attempt at fixing the FW flashing of some older MPMs that seem to work differently from those I have.


raphaelcoeffic:
This has been tested with success with the module that were failing before.

bsongis:
Some traces cleaning needed ;)

raphaelcoeffic:
> Some traces cleaning needed ;)

done!

PR #7093, opened 2019-11-17
Disable interrupts before jump to application in horus bootloader

raphaelcoeffic:
As far as I could see, it seems SysTicks interrupt is not enabled at this point. Did I forget anything else?

### DIFF
--- a/radio/src/targets/common/arm/stm32/bootloader/boot.cpp
+++ b/radio/src/targets/common/arm/stm32/bootloader/boot.cpp
@@ -105,6 +105,11 @@ void init10msTimer()
   NVIC_EnableIRQ(INTERRUPT_xMS_IRQn);
 }
 
+void deinit10msTimer()
+{
+  NVIC_DisableIRQ(INTERRUPT_xMS_IRQn);
+}
+
 extern "C" void INTERRUPT_xMS_IRQHandler()
 {
   INTERRUPT_xMS_TIMER->SR &= ~TIM_SR_UIF;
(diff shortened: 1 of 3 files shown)


bsongis:
Today we have those variables, we need to think if they are all 3 needed:
- `shutdownRequest`
- `shutdownReason`
- `powerupReason` (NOINIT)

bsongis:
Cases to take care / test:
1) Normal start
2) Exit from bootloader
3) Soft reset (from the application)
4) Watchdog reset

raphaelcoeffic:
> Soft reset (from the application)

Do we have such? What happens then?

And what about `NVIC->ICER`? Should I take care of that?

bsongis:
We have a soft reset when "Reset from factory"

MikeBland:
Soft Reset from the application.
When using the erskyTx bootloader, I have an option to run an "app". To exit such an app (and the bootloader really), it is best to use the NVIC_SystemReset() function as this guarantees the main openTx application is started with a completely clean slate, no interrupts or DMA left possibly enabled and active. and all I/O registers reset to default values, so exactly the same as a power on.
Since there is a dedicated flag in the RCC_CSR register that is set when this occurs, I cannot see why openTx cannot just test for this flag being set to determine a controlled restart has been requested. I believe only this flag is set in the CSR. Flags in the CSR should, of course, all be cleared after they have been read by the main application.
If required, a special value could be written to one of the RTC backup registers to confirm the soft reset is valid, with the RTC backup register being cleared immediately after being read. Different values could be used to indicate different reasons for the soft reset if necessary.

bsongis:
Mike we are working on another approach, taking into account your concern.

bsongis:
In fact it's even exactly what you are proposing: "Different values could be used to indicate different reasons for the soft reset if necessary."

bsongis:
Replaced by #7096

PR #7096, opened 2019-11-18
MAGIC value for powerReason should be stored in RTC backup on Horus

bsongis:
- [X] Allows the radio main application to start correctly when coming from an app in RAM
- [X] Refactoring of the `UNEXPECTED_SHUTDOWN()` process
- [X] Saves the RTC battery as BACKUP RAM is not needed on normal shutdown
- [X] On PWR OFF on an ALERT, the radio won't show the next alert before shutting down
- [X] Bootloader flashable through the SD Manager


MikeBland:
I'm suggesting using one of the 20 RTC Backup registers, not a location in the backup RAM.
If the backup RAM is enabled when power is off, the life of the RTC battery is reduced to about half.

radio/src/storage/rtc_backup.h
```diff
@@ -23,15 +23,9 @@
...
-enum PowerReason {
-  SHUTDOWN_REQUEST = 0xDEADBEEF,
-  SOFTRESET_REQUEST = 0xCAFEDEAD,
-};
-
 PACK(struct RamBackup {
-  uint32_t shutdownRequest;
   uint16_t size;
-  uint8_t data[4090];
+  uint8_t data[4096];
```
raphaelcoeffic: `4K - sizeof(uint16_t) == 4094`?

bsongis: ouch

radio/src/functions.cpp
```diff
@@ -35,7 +35,7 @@ void testFunc()
 #endif
 
   // for testing the WD reset uncomment the following line
-  // while (1);
+  while (1);
```
bsongis: ATTENTION !

radio/src/targets/common/arm/stm32/rtc_driver.cpp
```diff
@@ -67,27 +67,28 @@ void rtcInit()
...
+
   struct gtm utm;
   rtcGetTime(&utm);
   g_rtcTime = gmktime(&utm);
+#endif
 
-#if defined(RAMBACKUP)
+#if defined(RTC_BACKUP_RAM) && !defined(BOOT)
   RCC_AHB1PeriphClockCmd(RCC_AHB1Periph_BKPSRAM, ENABLE);
   PWR_BackupRegulatorCmd(ENABLE);
```
raphaelcoeffic, line 93: The manual says that `PWR_GetFlagStatus(PWR_FLAG_BRR)` should be checked after the regulator is turned on to make sure the regulator is ready. Is this something we should care about?

bsongis: well, we have not need any issue since day 1 on this, I propose to keep it unchanged

PR #7100, opened 2019-11-19
Better handling of read only model files by Companion

### LINKED ISSUE #7056: Companion in Windows 10: Handling of read-only otx files could be improved.
In Windows 10, when you make changes to an otx file that has its _read only_ attribute set, Companion correctly refuses to close unless you "Save As..." with another filename, remove the _read only_ attribute, or discard the changes. Of course this is what one would expect. However, the problem is that there's no explanation provided to the user. This could be causing some frustration, particularly if someone finally resorts to discarding and losing the work they did just to get the program to close.

One improvement for cases where a _read only_ otx file is opened would be to add "(read only)" to the title at the top of the otx window, and if a modification is made, another improvement would be to show "Save As..." in place of "Save" alongside "Discard" and "Cancel" in the window  that pops up when you try to close a modified otx file or close Companion.

elecpower:
Fixes #7056

### DIFF
--- a/companion/src/mdichild.cpp
+++ b/companion/src/mdichild.cpp
@@ -701,6 +701,10 @@ void MdiChild::updateTitle()
   if (availableEEpromSize >= 0) {
     title += QString(" - %1 ").arg(availableEEpromSize) + tr("free bytes");
   }
+  QFileInfo fi(curFile);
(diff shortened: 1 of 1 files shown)


bsongis:
Thanks!

PR #7102, opened 2019-11-20
Fix ar9x compile/boot

sreichholf:
When removing REVA in 0ec34fd72c5d1ab356c9129aac52985466005654 boardInit() for AR9X got lost completely due to some #if/#endif mixup.
I've also readded rtcInit() to boardInit() for AR9X so the (optional) rtc is working again.

Additionally there's a small compile fix in view_main.cpp for datatype issues when compling for ATSAM

### DIFF
--- a/radio/src/gui/128x64/view_main.cpp
+++ b/radio/src/gui/128x64/view_main.cpp
@@ -186,7 +186,7 @@ void drawTimerWithMode(coord_t x, coord_t y, uint8_t index)
     const uint8_t negative = (timerState.val<0 ? BLINK | INVERS : 0);
     LcdFlags att = RIGHT | DBLSIZE;
     if (timerState.val < 60 * 60) { // display MM:SS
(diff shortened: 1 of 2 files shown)


radio/src/gui/128x64/view_main.cpp
```diff
@@ -186,15 +186,15 @@ void drawTimerWithMode(coord_t x, coord_t y, uint8_t index)
     const uint8_t negative = (timerState.val<0 ? BLINK | INVERS : 0);
     LcdFlags att = RIGHT | DBLSIZE;
     if (timerState.val < 60 * 60) { // display MM:SS
-      div_t qr = div(abs(timerState.val), 60);
+      div_t qr = div((int)abs(timerState.val), 60);
```
bsongis, line 189: good, this will fix an issue with recent compilers as well

bsongis:
Thanks!

PR #7128, opened 2019-11-28
Add TXRSSI and TXLQI sensors to FrSky D8/D16/LBT when using MULTI

pascallanger:
- Add TXRSSI and TXLQI sensors to FrSky D8/D16/LBT when using MULTI
- Clean up Flysky telemetry
- Clean up Hitec telemetry
- Clean up HoTT telemetry
- Add missing Multi protocols


3djc:
I think this feature should be disabled by default in multi then, because introducing this will add two 'low value' mandatory sensors where folks are already struggling for sensors slot.

Also,  this allocates address on the sensor pool that FrSky manages that may clash in the future if FrSky also decides to use those values

So  if someone reaaaaaaly wants that, maybe, but i do not belive it should be for everybody

pascallanger:
You can change the sensors IDs to what you want. They are totally virtual and have no meaning from a decoding standpoint. They are just there for the sensor display...
By the way, there is nothing added to Multi. This values are being sent since the FrSkyD protocol was introduced. Er9x and ErskyTX are displaying these values, it's missing in OpenTX...

pascallanger:
I've changed the IDs to 0xFFFE..FFFC, better now?

Allerhopp:
Seems I was the one asking for this feature. I'm totally happy if it is disabled by default in _Config.h or somewhere else in multi (just tell me where ;))

pascallanger:
It's not in Multi but OpenTX.

Allerhopp:
I understand 3djc's comment different - but misunderstanding is my middle name ;-)

3djc:
You are both right. I have a concern with two more sensors been populated out of a limited sensors slots. But since we increased it in 2.3.x, maybe we can afford two more

PR #7135, opened 2019-11-30
Multi fixes

3djc:
- fix failsafe line sometimes not  displayed
- display subprotocol even when only one exits
- change handling for unknow protocols (only allow to set type and subtype)


3djc:
Redoing

PR #7136, opened 2019-11-30
Improve multi

3djc:
fix failsafe line sometimes not displayed
display subprotocol even when only one exits
change handling for unknow protocols (only allow to set type and subtype)
do not display RSSI alarm disabled when multi DSM protocol are selected

### DIFF
--- a/radio/src/gui/gui_common.h
+++ b/radio/src/gui/gui_common.h
@@ -161,43 +161,6 @@ inline uint8_t MODULE_CHANNELS_ROWS(int moduleIdx)
 
 
 #if defined(MULTIMODULE)
-// When using packed, the pointer in here end up not being aligned, which clang and gcc complain about
-// Keep the order of the fields that the so that the size stays small
-struct mm_options_strings {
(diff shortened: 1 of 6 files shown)


radio/src/gui/gui_common.h
```diff
@@ -223,6 +186,23 @@ inline bool isMultiProtocolSelectable(int protocol)
   return protocol != MODULE_SUBTYPE_MULTI_SCANNER;
 }
 
+inline uint8_t MULTIMODULE_PROTOCOL_KNOWN(uint8_t moduleIdx)
```
bsongis: why not `bool` return code?

radio/src/gui/gui_common.h
```diff
@@ -223,6 +186,23 @@ inline bool isMultiProtocolSelectable(int protocol)
...
+{
+  if (!isModuleMultimodule(moduleIdx))
```
bsongis: `{` to be consistent with the next tests within the same function?

PR #7140, opened 2019-12-01
Add diagnostic screens to colorlcd radios

3djc:
image

image


robustini:
@3djc, I'm trying now but in HARDWARE under DEBUG I can only choose ANALOGS and not SWITCHES, bad italian translation or there's other issue?
I checked the IT language file with EN and at that point they seem to me aligned.

3djc:
Should be fixed

PR #7143, opened 2019-12-03
Range test fix (it sent only the upper or lower channels, but not both)

### LINKED ISSUE #7139: Control failure in range test mode
Since the upgrade to OpenTx 2.3 I have experienced a number of control failures when range testing. RSSI readings are OK, no warnings, but there is no movement of aircraft control surfaces. This has happened to me with a X9D transmitter and a G-RX8/XM+ receiver combination. After several switches in an out of RT mode I eventually got control surface movement.

On RC Groups two other reports are documented, with X9D+ and X8R and X9D+ and X10 with several receivers. (See OpenTx discussion thread https://www.rcgroups.com/forums/showpost.php?p=43266039&postcount=1352 and following posts).

bsongis:
This fixes #7139

### DIFF
--- a/radio/src/pulses/pxx1.cpp
+++ b/radio/src/pulses/pxx1.cpp
@@ -22,7 +22,7 @@
 #include "pulses/pxx1.h"
 
 template <class PxxTransport>
-uint8_t Pxx1Pulses<PxxTransport>::addFlag1(uint8_t module)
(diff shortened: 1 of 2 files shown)


3djc:
Tested ok on x7, range test and failsafe both tested ok

PR #7154, opened 2019-12-05
MULTI: Bring selectable option values of 128x64 and 212x64 in line with 480x272

pascallanger:
Consistency between the display sizes.


bsongis:
Better implementation done in #7159

Issue #7165, opened 2019-12-07
COLORLCD SD card  + USB pluging issue


3djc:
2.3.3RC1 or RC1 + USB/SD bug fix

Radio set to STORAGE on USB plug
Go to SD card, IMAGES for example
Plug USB
Radio goes to main view
Unplug USB 
Radio goes back to SD Card , but with corrupted display

Issue #7164, opened 2019-12-07
COLORLCD radios, bug in SD Manager


3djc:
OTX 2.3.3 RC1:
- press SYS
- press PAGE
- scroll to IMAGES, press ENTER
- scroll to readme.txt, press ENTER, view text

The text is not completely readable / is truncated.
- press RTN
The cursor is no longer on readme.txt,
- 2 clicks down
Special characters are displayed.

I seem to recall a fix about that, possible 2.2, need to dig closed issues

PR #7172, opened 2019-12-09
Bsongis/sd manager pause resume

bsongis:


bsongis:
Fixes #7165 
Fixes #7164

PR #7186, opened 2019-12-12
Use target compiler for generating lua exported definitions

fiam:
Include the lua subdirectory only after CMAKE_C_COMPILER has been
set to the cross compiler. Otherwise, cmake will use the build machine
compiler to preprocess dataconstants.h, which breaks on some systems
and can cause problems due to different machine dependent types.

### DIFF
--- a/radio/src/CMakeLists.txt
+++ b/radio/src/CMakeLists.txt
@@ -146,7 +146,6 @@ include_directories(${CMAKE_CURRENT_SOURCE_DIR})
 include_directories(${CMAKE_CURRENT_BINARY_DIR})
 include_directories(${CMAKE_CURRENT_BINARY_DIR}/bitmaps/${GUI_DIR})
 
-add_subdirectory(lua)
 include(gui/CMakeLists.txt)
 
 if(RTC_BACKUP_RAM)
@@ -551,3 +550,7 @@ if(NOT MSVC)
   PrintTargetReport("firmware")
 
 endif(NOT MSVC)
+# Include the lua subdirectory after CMAKE_C_COMPILER has been
+# set. Otherwise, the custom commands it includes will use the
+# default CMAKE_C_COMPILER instead of the one we set manually
+add_subdirectory(lua)


bsongis:
Any MacOS expert for this one? @raphaelcoeffic ?

bsongis:
@schwabe would you ensure it won't break the MacOS compilation system?

kilrah:
Why is this required? We've been building on macOS for years without it...

raphaelcoeffic:
@fiam could you please describe the original error you were getting? Maybe copy/paste it so we can see it? I never had any issue with this.

Btw. you might have to add something like `-DCMAKE_OSX_DEPLOYMENT_TARGET=10.10` to your cmake command line to set a particular SDK version to be used.

schwabe:
I also feel this change is masking some real problem/other problem.  Also that this is just for lua feels extraordinary strange.

fiam:
@raphaelcoeffic Running cmake as `cmake -DPCB=X7 -DDISABLE_COMPANION=ON -DCMAKE_BUILD_TYPE=Debug -DLUA=ON -DLUA_COMPILER=ON -DPYTHON_EXECUTABLE=/usr/local/bin/python3 -DCMAKE_VERBOSE_MAKEFILE=ON ../ && make firmware` it stops at:

```
[ 33%] Generating lua_exports_x7.inc
cd /tmp/opentx/build/radio/src/lua && /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc -E -DCPUARM -DLUA -DLUA_INPUTS -DLUA_EXPORT_GENERATION -DSIMU -I/tmp/opentx/radio/src/targets/taranis -I/tmp/opentx/radio/src/thirdparty -I/tmp/opentx/radio/src/targets/common/arm/stm32 -DPCBTARANIS -DPCBX7 -DRADIO_X7 -DEXPORT /tmp/opentx/radio/src/dataconstants.h > lua_exports_x7.txt
In file included from /tmp/opentx/radio/src/dataconstants.h:24:
In file included from /tmp/opentx/radio/src/targets/taranis/board.h:24:
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/11.0.0/include/inttypes.h:30:15: fatal error: 'inttypes.h' file not found
#include_next <inttypes.h>
              ^~~~~~~~~~~~
1 error generated.
make[3]: *** [radio/src/lua/lua_exports_x7.inc] Error 1
make[2]: *** [radio/src/lua/CMakeFiles/lua_export_x7.dir/all] Error 2
make[1]: *** [radio/src/CMakeFiles/firmware.dir/rule] Error 2
make: *** [firmware] Error 2
```

Adding `-DCMAKE_OSX_DEPLOYMENT_TARGET=10.10` results in the same error.

What happens here without this patch is:

- The host C compiler is called in preprocessor only mode (not sure how safe is that for generating the lua exports and why you're not calling the target compiler, but I didn't dive into that)
- The host C compiler autodetected by cmake with Xcode and several SDKs installed is `/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc`
- When you invoke the compiler using that command, it has no system includes, just some libc/libc++/swift headers. Hence it fails to find system's `inttypes.h`

The solution in this PR works as:

- cmake already detects `CMAKE_OSX_SYSROOT` automatically and points to the most recent SDK. It's empty on non-macOS systems
- Pass the autodetected system root to the compiler when generating the exports file, so it can find the system headers.

Now, since we're already discussing this... could someone clarify why the compiler targeting the host is being used to run the preprocessor for generating the lua exports file?

bsongis:
Why not?

fiam:
The `add_subdirectory(lua)` call happens before the compiler is overridden, so the macro referencing the C compiler in that subdirectory will use the default one. Same strategy is used for tests and the simulator target, so it seems intentional to me. However, I don't see why you would want to use the default compiler (instead of the cross compiler) to preprocess those files, since they will be consumed by the cross compiler later (via include in `lua/api_general.cpp`).

fiam:
@bsongis Not sure if you're asking why I think using the host compiler to preprocess those files is a bad idea. If you're asking for something else, please clarify.

The reason you should use the target compiler for preprocessing files that are compiled for the target is that there are architecture/register size/endianess builtin defines that might affect the resulting generated code. e.g. for amd64 the compiler will `#define __x86_64__ 1` while for ARM it will `#define __arm__ 1`. There's a lot of them and it's not rare to have code like `#if __BYTE_ORDER__...`. You can view the whole list by calling gcc like `gcc -dM -E - < /dev/null`. Maybe the headers that generate the lua exports don't depend on any of those predefined symbols for now, but to me it seems like a time bomb waiting to explode into very difficult to diagnose bugs.

Again, there might be some valid reason for calling the host compiler, but I'm just not seeing any. If anyone can clarify, that'd be great.

bsongis:
The result should be exactly the same on all compilers. That's why I ask.

fiam:
@bsongis It might be generating equivalent output right now on your machine, but it's merely by chance. `dataconstants.h` includes `board.h`, which includes `inttypes.h`. At that point you're already on machine dependent territory.

Unless there are specific reasons to call the build machine compiler (which, given by the absence of replies to my question about that, I'd assume there are none), the best solution is to call the cross compiler to generate this file.

bsongis:
Yes, far better

PR #7193, opened 2019-12-15
T16 mute improve

### LINKED ISSUE #7190: OTX 2.3.x: trim beeps skipped when trim held
Hardware: Jumper T16 (OTX 2.3.3 RC1), Horus X12S.  

**Expected behaviour**: transmitter issues beep with each trim increment. 
**Actual behaviour**: when the trim is held, some beeps are skipped. Applies to all trims, regardless of direction.

I've seen this on both the T16 and X12S on various 2.3 nightlies and 2.3.3 RC1. Video below illustrates the issue in a new model on the T16:

https://youtu.be/ujpX2_9XQYU

(Works as expected on X9D running 2.2.3).

3djc:
This fixes #7190

### DIFF
--- a/radio/src/targets/common/arm/stm32/audio_dac_driver.cpp
+++ b/radio/src/targets/common/arm/stm32/audio_dac_driver.cpp
@@ -115,7 +115,7 @@ void audioConsumeCurrentBuffer()
     else {
 #if defined(AUDIO_UNMUTE_DELAY)
       uint32_t now = RTOS_GET_MS();
-      if((RTOS_GET_MS() - audioQueue.lastAudioPlayTime > AUDIO_MUTE_DELAY) && !readTrims()) {
+      if ((RTOS_GET_MS() - audioQueue.lastAudioPlayTime > AUDIO_MUTE_DELAY) && !readTrims()) {
         // mute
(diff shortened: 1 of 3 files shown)


radio/src/targets/common/arm/stm32/audio_dac_driver.cpp
```diff
@@ -113,8 +113,17 @@ void audioConsumeCurrentBuffer()
     }
 #if defined(AUDIO_MUTE_GPIO_PIN)
     else {
+#if defined(AUDIO_UNMUTE_DELAY)
+      uint32_t now = RTOS_GET_MS();
+      if ((RTOS_GET_MS() - audioQueue.lastAudioPlayTime > AUDIO_MUTE_DELAY) && !readTrims()) {
```
bsongis: why `!readTrims`?

radio/src/targets/common/arm/stm32/audio_dac_driver.cpp
```diff
@@ -113,8 +113,17 @@ void audioConsumeCurrentBuffer()
...
+        // mute
+        GPIO_SetBits(AUDIO_MUTE_GPIO, AUDIO_MUTE_GPIO_PIN);
+        audioQueue.lastAudioPlayTime = now;
```
bsongis: I think the initialisation is not done properly

PR #7196, opened 2019-12-16
I don't see why <= instead of < here. Untested

bsongis:


3djc:
Strange indeed, especially when 128 is different. tested ok on colorlcd

bsongis:
Bug introduced on 2019-06-24
Copied then for colorlcd

PR #7198, opened 2019-12-16
Bsongis/bluetooth new protocol

bsongis:


robustini:
@bsongis is there any news on bluetooth advanced support?
This PR is very interesting, it would be nice to be able to transfer a model via BT.

brainbubblersbest:
Yeah, what is on this?
Maybe the Otx Team can allow the users to connect Hm-10 modules or hc05 Modules on that place and feed this direct with Serial. This would been a simple workaround since this issue is stale almost a year now. This would t bring back the Free link compability. But we can use after Marked modules with that. Programmed with 115200 baud and we can use bt2.0 Standard or bt4.2 modules for reliable connections. Or wifi with esp32.
#8073 this where worth me a donation for the project

mpjf01:
Can we leave this oine open please

PR #7202, opened 2019-12-16
[taranis] Allow passing the HW revision at compile time

fiam:
This allows flashing/debugging with stlink. Otherwise PA14 is (SWCLK)
configured as an input and the connection is lost.


bsongis:
I am afraid it breaks MSVC compilation
Is it possible you nose at horus/board.h, the implementation seems better

kilrah:
Yes it does break simu compilation in VC++

fiam:
@kilrah Oops, my bad. I don’t have a properly set up VS installation to compile with MSVC. Could you paste the error you get so I can try to figure out exactly what breaks the build? Thanks!

bsongis:
I will take it, this implementation is not good. I will copy Horus one

fiam:
@bsongis This PR already had it implemented like the horus implementation with the pcbrev field in HardwareOptions. Note that this is for development purposes only. Released firmwares need the runtime PCB revision detection. I didn't use the PCBREV option because that one is used for compiling firmwares that are intended for different radios.

kilrah:
> Could you paste the error you get so I can try to figure out exactly what breaks the build? Thanks!

The issue is that you're not allowed to put #defines in struct definitions on VC++, need to #define the whole struct

PR #7204, opened 2019-12-16
[MSC] Add a file in the virtual Taranis drive with radio information

fiam:
Add a new readonly file called RADIO.TXT to the root directory.
This file contains the unique ID of the CPU as well as a DFU
serial using the same algorithm as ST's DFU bootloader.

This way we can keep track of an specific radio when it's connected
in MSC mode.

### DIFF
--- a/radio/src/targets/common/arm/stm32/bootloader/CMakeLists.txt
+++ b/radio/src/targets/common/arm/stm32/bootloader/CMakeLists.txt
@@ -77,13 +77,14 @@ set(BOOTLOADER_SRC
   ../../../../../targets/${TARGET_DIR}/board.cpp
   ../../../../../targets/${TARGET_DIR}/pwr_driver.cpp
   ../../../../../targets/${TARGET_DIR}/bootloader/boot_menu.cpp
-  ../usbd_usr.cpp
-  ../usbd_storage_msd.cpp
+  ../cpu_id.cpp
   ../delays.cpp
-  ../usbd_desc.c
+  ../flash_driver.cpp
   ../usb_bsp.c
   ../usb_driver.cpp
-  ../flash_driver.cpp
+  ../usbd_desc.c
+  ../usbd_storage_msd.cpp
+  ../usbd_usr.cpp
   init.c
   boot.cpp
   bin_files.cpp
--- a/radio/src/targets/common/arm/stm32/cpu_id.cpp
+++ b/radio/src/targets/common/arm/stm32/cpu_id.cpp
@@ -34,3 +34,26 @@ void getCPUUniqueID(char * s)
   *tmp = ' ';
   strAppendUnsigned(tmp+1, cpu_uid[2], 8, 16);
 }
+
+void getCPUDFUSerial(char * s)
+{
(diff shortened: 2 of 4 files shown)


kilrah:
What is the purpose? This seems to be supportive of a feature it should be PRd together with

fiam:
It's for a custom app I'm working on, but I think it could be useful for companion too. The idea is to enable the USB host to reboot into ST's bootloader, flash everything, reboot into MSC and upgrade the SD card contents automatically.

fiam:
@kilrah In fact, it could entirely remove the need for the OpenTX bootloader, but now that it is already in place I don't see the harm on keeping it around even if users wouldn't have to interact with it anymore.

kilrah:
Nice, do you have that on a repo we can have a look at?
We definitely want to keep the bootloader since we want everything to be able to be done without the PC/USB connection...

fiam:
That's definitely a good reason to keep the bootloader around.

Regarding the app, I don't have anything meaningful to show yet. At this point I'm prototyping the host side with some quick and dirty python scripts.

raphaelcoeffic:
@fiam the main issue I see here is that ST's bootloader is incredibly slow. This might not be a concern for normal users, but it is for me as a dev. However, as long as the current bootloader is still functional, I'm fine.

kilrah:
> Regarding the app, I don't have anything meaningful to show yet. At this point I'm prototyping the host side with some quick and dirty python scripts.

OK, then we'll probably wait until that's in a shareable state to include this PR, please resubmit at that time.

fiam:
@raphaelcoeffic I wouldn't expect developers to flash using anything but gdb via stlink. Anything else is slower and if you're developing something for a radio you probably have stink plugged in anyway.

@kilrah Why close this? It's a useful feature on its own, certainly more useful than the already existing `FIRMWARE.TXT`, since this uses a well defined `<key>: <value>` format that's more robust to parse. This could be used by companion or by any 3rd party tool. I would expect some discussion regarding what should be exposed in `RADIO.TXT` or if we should redesign `FIRMWARE.TXT` instead, but closing this and asking me to resubmit once I have a complete implementation for the host side seems uncalled for.

raphaelcoeffic:
> I wouldn't expect developers to flash using anything but gdb via stlink. Anything else is slower and if you're developing something for a radio you probably have stink plugged in anyway.

@fiam I use it very rarely, as it works way easier without having to solder to the board to get SWD access. In fact, I don't even have SWD access readily soldered on all radios, and most of time, it's absolutely unnecessary. I'd like to keep it that way.

fiam:
@raphaelcoeffic Understood. There are different workflows and everyone should choose whatever floats their boat. Rest assured, I have no plans to remove the bootloader.

kilrah:
> Why close this? It's a useful feature on its own

On its own? No, nothing in our tools cares about the identity of a particular radio at this point. If a feature requiring it was to be added in companion then for sure it would make sense adding this.

Our view is not to add code that has no application yet, and to include all the bits that go together at the same time.

PR #7206, opened 2019-12-16
[build] Fix building with semihosting enabled

fiam:
Remove -DSEMIHOSTING from the bootloader CFLAGS unconditionally.

-DSEMIHOSTING is passed to the bootloader by the parent project when
SEMIHOSTING=ON, however the bootloader wouldn't fit on flash with
semihosting enabled.

If we leave the -DSEMIHOSTING flag, we would get a compilation error
on non-horus radios (since -DDEBUG is removed) and a link time error
on horus (since the bootloader won't fit in its allocated flash
region).


bsongis:
I don't think anyone seriously used the SEMIHOSTING option already, I added it for X10 if I am not mistaken. But it's good to fix on small radios.

PR #7210, opened 2019-12-17
[SEMIHOSTING] Enable semihosting for any targets that support it

fiam:
- Move semihosting initialization to main()
- Provide _exit and _kill even when semihosting is enabled. librdimon
doesn't include them.


bsongis:
OK for 2.3.4

PR #7212, opened 2019-12-17
Rework pwr_driver

3djc:
Got inspired by another PR

This use a single pwr_driver for both Horus and Taranis


3djc:
To make it clearer, I was inspired by #7202

radio/src/targets/taranis/CMakeLists.txt
```diff
@@ -160,6 +160,12 @@ elseif(PCB STREQUAL X7)
     set(BLUETOOTH YES)
     add_definitions(-DBLUETOOTH_PROBE)
     set(ROTARY_ENCODER YES)
+    if(DEFINED PCBREV_HARDCODED)
```
bsongis, line 163: Missing comment from the original PR:
```
# This *can* be used to prevent messing with PA14 (SWCLK), so stlink can be used
# without interruptions.
```

PR #7215, opened 2019-12-18
Fix x9lite-s having only 5 switches on simu

### LINKED ISSUE #7207: X9 Lite S companion missing switch SG and SF
I think this is missing config from #6820. If you can point me to where this config lives I can probably make a PR. SG and SF are 2-position switches. I see these two switches in firmware here: <https://github.com/opentx/opentx/pull/6820/files#diff-bb5dbdf59947abf74a8735ff61cf0aa4R137>

**To Reproduce**
1. openTX nightly 2019-12-16
1. open model simulator or move model from X9d+ or other radio
1. see that SG and SF are not options, create invalid models.

**Expected behavior**
SG and SF should be present and available

**Screenshots**
image

image
image

**General information**
 - OpenTX Version ?! what shipped on radio full SHA wasn't visible.
 - Hardware X9LITES
 - OpenTX Companion Version 2.3.3N71, Dec 16 2019 on macOS X Mojave

3djc:
Fix #7207

### DIFF
--- a/companion/src/firmwares/boards.cpp
+++ b/companion/src/firmwares/boards.cpp
@@ -330,7 +330,7 @@ int Boards::getCapability(Board::Type board, Board::Capability capability)
       if (IS_TARANIS_X9E(board))
         return 18;
       else if (IS_TARANIS_X9LITE(board))
-        return 5;
+        return 7;
       else if (IS_TARANIS_X7(board))
         return 8;
       else if (IS_JUMPER_T12(board))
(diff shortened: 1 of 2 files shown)


johnelliott:
Thanks @3djc and @bsongis I just got around to testing 2.3.4 on this hardware and the radio works. 
Follow-on work might be to make sure new documents using the x9lites default to the switches that ship with the radio.

PR #7223, opened 2019-12-20
Quick Access for Crossfire Config Menu

DukeyAir:
This is my suggestion for a quick access to Crossfire config

Especially for CRSF Micro TX without display fast access to the config menu is a big QOL feature.

(might be interesting to implement further, not only for crossfire, but radio tools in general).


3djc:
I believe this proposal might be a bit unflexible, and decides for users what he/she wants (what if you need BF script more than crsf setup ?). I think #7312 would give you an equaly fast access, but much more customizible

DukeyAir:
> I believe this proposal might be a bit unflexible, and decides for users what he/she wants (what if you need BF script more than crsf setup ?). I think #7312 would give you an equaly fast access, but much more customizible

I absolutely agree, just didnt implement it further because i wanted to hear general opinion first. But i like the proposed PR #7312 even more so +1 from me and this PR is obsolte now

Issue #5636, opened 2018-01-18
X9E SBus Trainer Mode not working


Fi156:
Hello,

I am trying to connect two X9E radios via SBUS Trainer mode.
Both radios and both receivers (X8R) are EU-Versions with FW 2.2.1.
(I never flashed the X8R's...)

Configuration file for both receivers:
[X9E_SBusTrainer_Minimal.zip](https://github.com/opentx/opentx/files/1642706/X9E_SBusTrainer_Minimal.zip)

Definition:
Master = Radio bound to the Model-X8R, this radio has the Student bound X8R connected to the middle 3 pins in the external module compartment. (X9D= Signal at the second pin, X9E=Signal at pin 4)
Student = bound to X8R in external module compartment of the master. NOT set to SBUS/Trainer mode.

When I am using an **X9D as master transmitter, everything works fine.** (So my wiring and configuration should be okay.)
When using one of two X9E transmitter as Master, the student receiver will be recognized by the student radio, but no stick input can be seen on screen of the master radio.

What can I do to narrow down the source of the problem?

Fi156:
Does the X9E lack the implementation of SBUS Trainer mode at the moment?

Can someone describe the existing problem and give some hints how to solve that?

kilrah:
No, but we've seen another report like yours so something is likely broken.

pointu:
Grrrr 3djc (on 15apr.),
Same Issue on my x9e, same testing with a x9d+ as master all fine, but with the x9e nothing on the trainers channels, and no signaling trainer on/off.

dohdohpub:
Hello,
so far I have always been able to fly by cable in trainer mode without problems, with X9D + Master and 9XR Pro as Slave. I think, the FW version has been 2.1.9 or older. Then I upgraded the slave to 2.2.3 and from there the trainer mode did not work anymore.
I thought it was up to the software update and so I also updated the Taranis, but it still did not work. Then I connected a FrSky XM receiver to connect the transmitter via SBUS in trainer mode, but that does not work either.
Is this a issue that can be fixed or is it a software bug in the firmware?

Gerberus:
Same problem here,
trainer with X9E and X12s not working (tested both as slave and master, otx 2.2.3), no signal in OpenTX Trainer menu.
I am aware of the bug regarding sbus input on the X12s, but at the current state i am unable to use trainer, regardless of which radio is master or slave.
Verified receiver (connected to master radio via module bay) SBUS output with DSO, looks good.
[...]

PR #7227, opened 2019-12-21
Bsongis/trainer sbus x9e

bsongis:


bsongis:
Fixes #5636

radio/src/gui/common/stdlcd/radio_hardware.cpp
```diff
@@ -558,6 +558,12 @@ void menuRadioHardware(event_t event)
 #if defined(BLUETOOTH)
       case ITEM_RADIO_HARDWARE_BLUETOOTH_MODE:
         lcdDrawTextAlignedLeft(y, STR_BLUETOOTH);
+#if defined(PCBX9E)
+        if (g_model.trainerData.mode == TRAINER_MODE_MASTER_SBUS_EXTERNAL_MODULE) {
```
bsongis: It remains an issue if you have BT enabled and you load a model with SBUS TRAINER enabled. In this case you will have BT enabled and impossible to remove it.

PR #7229, opened 2019-12-21
LUA sport/crossfireTelemetryPush protocol detection

klutvott123:
Make sportTelemetryPush/crossfireTelemetryPush return nil if they detect the wrong telemetry protocol.
This allows us to differentiate between wrong protocol and telemetry buffer not available. For lua scripts supporting both crossfire and smartport this will let us reliably detect the telemetry protocol using these functions.

### DIFF
--- a/radio/src/lua/api_general.cpp
+++ b/radio/src/lua/api_general.cpp
@@ -440,13 +440,15 @@ When called without parameters, it will only return the status of the output buf
 
 @retval boolean  data queued in output buffer or not.
 
-@status current Introduced in 2.2.0
+@retval nil      incorrect telemetry protocol.
+
+@status current Introduced in 2.2.0, retval nil added in 2.3.4
 */
 
 static int luaSportTelemetryPush(lua_State * L)
 {
   if (!IS_FRSKY_SPORT_PROTOCOL()) {
-    lua_pushboolean(L, false);
+    lua_pushnil(L);
     return 1;
   }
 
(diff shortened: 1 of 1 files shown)


radio/src/lua/api_general.cpp
```diff
@@ -438,15 +438,15 @@ When called without parameters, it will only return the status of the output buf
 
 @param value     value
 
-@retval boolean  data queued in output buffer or not.
+@retval boolean  data queued in output buffer or not. nil if incorrect telemetry protocol.
 
 @status current Introduced in 2.2.0
```
3djc: You need to mention the update here (2.3.4)

radio/src/lua/api_general.cpp
```diff
@@ -635,14 +635,14 @@ When called without parameters, it will only return the status of the output buf
 
 @param data table of data bytes
 
-@retval boolean  data queued in output buffer or not.
+@retval boolean  data queued in output buffer or not. nil if incorrect telemetry protocol
 
 @status current Introduced in 2.2.0
```
3djc: Same comment as above

radio/src/lua/api_general.cpp
klutvott123: Done!
Updated  ``@status`` and moved the initial comment to a new line with ``@retval nil`` to make things a little clearer.

bsongis:
Thanks, this idea is good

klutvott123:
@bsongis 👍 Thanks for accepting this

PR #7237, opened 2019-12-23
Update companion multi parts to bring in line with radio

3djc:


MRC3742:
This is probably information you already know? If so disregard.
https://github.com/pascallanger/DIY-Multiprotocol-TX-Module/blob/master/Multiprotocol/Multi.txt
There are three newer Protocols
59 - Bayang Rx
60 - Pelikan
61 - Tiger

3djc:
yes MRC3742, but the aim of this was just to be consistent between radio and companion. We will have to sync with multi again when we are about to release 2.3.5

PR #7243, opened 2019-12-28
Buttons for trims T5 and T6 were not working in simu (fix #7167)

### LINKED ISSUE #7167: Companion : T5 T6 not behaving correctly
Using companion, T5 and t6 are not behaving trully as trims (no beeps, no trims centered) when clicking on left / down arrow, while moving the cursor does work

kilrah:


radio/src/dataconstants.h
```diff
@@ -581,7 +581,7 @@ enum SwitchSources {
   #define SWSRC_TRAINER SWSRC_LAST_SWITCH,
 #endif
 
-#define SWSRC_LAST_TRIM                 (SWSRC_FIRST_TRIM + 2*NUM_TRIMS - 1)
+#define SWSRC_LAST_TRIM                 (SWSRC_FIRST_TRIM + NUM_TRIMS_KEYS - 1)
```
bsongis: Good on XLite(S)?

kilrah: Ah possibly not

PR #7267, opened 2020-01-04
2.4

qba667:
Fixed usage of USB mode menu - displayed just once.
Bertrand, can you please enable for me access to 2.4 - it is in protected state - so I can not push to it directly.

### DIFF
--- a/radio/src/main.cpp
+++ b/radio/src/main.cpp
@@ -31,7 +31,12 @@ uint8_t mainRequestFlags = 0;
 #if defined(LIBOPENUI)
 void openUsbMenu()
 {
-  auto menu = new Menu(&mainWindow);
+  static Menu* menu = nullptr;
+  if(menu) return;
+  menu = new Menu(&mainWindow);
+  menu->setCloseHandler([=]() {
+    menu = nullptr;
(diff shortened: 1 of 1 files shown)


3djc:
If you want to fix usage, there’s a settings for it, I often alternate between sd to make script changes and debug in the same session

qba667:
@3djc the menu was displayed multiple times - because it was check on every call to perMain.

3djc:
Right’ you reset it when done, overlooked that, great

radio/src/main.cpp
```diff
@@ -31,7 +31,12 @@ uint8_t mainRequestFlags = 0;
 #if defined(LIBOPENUI)
 void openUsbMenu()
 {
-  auto menu = new Menu(&mainWindow);
+  static Menu* menu = nullptr;
+  if(menu) return;
+  menu = new Menu(&mainWindow);
+  menu->setCloseHandler([=]() {
```
bsongis, line 37: I don't think it's needed to capture everything here

bsongis: I just pushed a change on top of your commit

PR #7278, opened 2020-01-06
Ident #ifdefs according to majority identing style

olliw42:
simple non-functional PR
just applies the code style found for most indents to all, to make it easier to read
also as exercise for me to contribute per PR to opentx ;)
thx, Olli


bsongis:
good :)

PR #7312, opened 2020-01-20
Move TOOLS menu as first radio screen

3djc:


bsongis:
What about putting the SD Manager as 2nd page?

klutvott123:
I think the icons for 480x272 needs reordering too

PR #7322, opened 2020-01-22
Crossfire LUA add missing support for PREC3 values

lichtl:
This is the PR for the crossfire device.lua to support 3 digit precission (eg. used on KISSFC) Credits to @fedorcomander


fedorcomander:
Thanks Carsten. There was not just 3 digits precision (btw, if u are on, please rename function round to something like formatFloat…

problem was, that min and max values were also neglected for float parameter type in general.

A.

3djc:
Thanks !

PR #7334, opened 2020-01-25
CRSF.LUA - fix nonworking retun to device overview on T16

lichtl:
This fixes the non working return to the device overview on the T16. In addition I also modified the inc/dec based on X9 file


3djc:
The more virtual the better ! Thx

lichtl:
Hi @3djc just noticed when testing nightly (to check crossfire fixes) nightly SD card content seem to be not updated.

3djc:
Yes, we try to avoid too many sd cards changes, even in nightly (seems to cause chaos), but I'll trigger an update, quite de few changes have gone in

PR #7337, opened 2020-01-26
Lua virtual replace

lichtl:
Based on previous PR I replaced button events with virtual ones (which also fixed wizards on T16). Also did some cleanup of some unused conditions


lichtl:
should be good now. If this should be squeezed/rebased into a single commit please let me know.

3djc:
Don't worry, I'll squash it. Thanks alot !

PR #7360, opened 2020-02-05
Move Module Tools behind Radio Script Tools

DukeyAir:
Move Module Tools after Radio Tools since they're less frequently used. Also it enables you to sort Radio Tools by names of the LUA-Scripts


bsongis:
Needs the same on all radios (or none)

DukeyAir:
Agreed, Updated Common LCD aswell.

PR #7374, opened 2020-02-10
Review sensor physID handling

3djc:
Fixes somes issues running  telemetry siumulator

Tested with ACCST, but not access


bsongis:
Thanks @3djc for the ACCESS as well!

PR #7386, opened 2020-02-14
Add color option to luaLcdDrawPoint()

olliw42:
in lua scripts only black points can be drawn currently with lcd.drawPoint() even on color displays. This PR corrects this, by adding an optional flag.

works for me on a Jumper T16

### DIFF
--- a/radio/src/lua/api_lcd.cpp
+++ b/radio/src/lua/api_lcd.cpp
@@ -71,6 +71,8 @@ Draw a single pixel at (x,y) position
 
 @param y (positive number) y position
 
+@param flags (unsigned number) drawing flags
+
 @notice Taranis has an LCD display width of 212 pixels and height of 64 pixels.
 Position (0,0) is at top left. Y axis is negative, top line is 0,
 bottom line is 63. Drawing on an existing black pixel produces white pixel (TODO check this!)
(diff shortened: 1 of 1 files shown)


bsongis:
Good change, thanks!

Would you please add an explanation for the parameter you introduced (so that the doc is automatically updated)?

olliw42:
ah, didn't consider this
I don't know which flags actually apply, so I just took the generic description I found in other places in the same file. It certainly would be useful to have a more complete description of which flags are possible, but that's beyond my sight.

3djc:
You probably also want to indicate the new flags introduction version, look just above in clear() for example with color  introduction

olliw42:
well, this actually raises more questions (and points to some inconsistencies in the docs)

e.g., since this param is optional I would have thought one would mark it as optional, and in fact in some places like in the clear() it is done while in most other places it is not done

in some places one finds a more complete description which flags apply or don't, while in many other one don't

I would also think that the comment in the @notice as regards the Taranis is somewhat odd, it looks like being written long ago. 

I would also think that some comment on the situation of color vs b/w screens would be appropriate.

I just can look at things and note, but I'm not in a position to resolve any of the inconsistencies. It looks to me that someone with knowledge would have to go through all of the description and make them consistent and up to date. 

So, suggestion, instead of this process of small iterations of getting it right for at least drawPoint() it probably would be shorter if anyone of you could just write down explicitly the text which I should do. :)

PR #7391, opened 2020-02-16
RAS update

3djc:
I can't remeber a reason why X7 and XLITE should not have RAS

I'm not sure about HORUS, @bsongis please advise


3djc:
x7/xlite do'nt have rad on their module

PR #7392, opened 2020-02-16
Bring back RAS to Horus

### LINKED ISSUE #6775: [2.3.0 RC3 - Horus X12S] - RAS always remains at 0
On my X12S the visualization of the RAS in the statistics is always of value 0, even changing antenna (INT or INT+EXT).
I'm sorry but this is a particular situation, that is I installed the IXJT+ RF module of the X10S (triple antenna), so I can't say it's a bug because maybe the firmware doesn't correctly interpret what comes from the module because it expects to find an IXJT and not a IXJT+.
It would still be welcome if he could intercept it anyway.
In this condition with OpenTX V2.2.4 the RAS is displayed only with INT+EXT antennas, it remains at 0 on the INT setting, but on the outside it reports it.
Surely with the original IXJT (X12S) the RAS was correctly reported on OpenTX V2.2.4 and in both situations (INT and INT+EXT antennas), I remember it well.

3djc:
Those were having it in 2.2.x, so I guess that should still be ok to have it. 

@bsongis please advise

This closes #6775

### DIFF
--- a/radio/src/telemetry/frsky.h
+++ b/radio/src/telemetry/frsky.h
@@ -282,15 +282,15 @@ inline bool isRasValueValid()
 {
   return true;
 }
-#elif defined(PCBX10)
-inline bool isRasValueValid()
-{
-  return false;
-}
 #elif defined(PCBX9DP) || defined(PCBX9E)
 inline bool isRasValueValid()
 {
   return telemetryData.xjtVersion != 0x0000 && telemetryData.xjtVersion != 0x00FF;
+}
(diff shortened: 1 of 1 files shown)


bsongis:
@kilrah do you remember? 

It seems we removed RAS on X10 on purpose, and I can't remember why

kilrah:
No idea either

3djc:
Checked 2.2.4, x10 seem to have RAS displayed

PR #7394, opened 2020-02-16
Allow range check for LR12 with 212 GUI

### LINKED ISSUE #7364: "Range Test" is not selectable in X9D X9D+ in LR12 mode.
The cursor will not enable the range test menu item to open on the X9D and X9D+ transmitter that I have. 

It works fine on my X-Lite, X10S and X12S. 

All running some version of OTx 2.3.

Even though LR12 is a non-telemetry mode and the Tx RSSI report will be 0 the ability to reduce power is useful for range testing when a FC outputting RSSI on an OSD is being utilised. This is the most common use case for LR12 mode.

If the reason for the X9D behaviour is that power is not actually being reduced, then it would be preferable for the "range test" menu item on the Tx to not be shown at all.

3djc:
This fixes #7364

### DIFF
--- a/radio/src/gui/212x64/model_setup.cpp
+++ b/radio/src/gui/212x64/model_setup.cpp
@@ -353,7 +353,7 @@ void menuModelSetup(event_t event)
     LABEL(InternalModule),
       INTERNAL_MODULE_TYPE_ROWS,
       MODULE_CHANNELS_ROWS(INTERNAL_MODULE),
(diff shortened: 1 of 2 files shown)


radio/src/gui/212x64/model_setup.cpp
```diff
@@ -353,7 +353,7 @@ void menuModelSetup(event_t event)
     LABEL(InternalModule),
       INTERNAL_MODULE_TYPE_ROWS,
       MODULE_CHANNELS_ROWS(INTERNAL_MODULE),
-      IF_NOT_ACCESS_MODULE_RF(INTERNAL_MODULE, IF_INTERNAL_MODULE_ON(isModuleFailsafeAvailable(INTERNAL_MODULE) ? (uint8_t)2 : (uint8_t)1)),
```
bsongis, line 356: isModuleRangeCheckAvailable(INTERNAL_MODULE)

PR #7396, opened 2020-02-17
Add copy min/max to outputs

### LINKED ISSUE #7378: Default servo travel setting
*Before suggesting a new feature here, please search the open issue if there is already a request for the feature*

**Is your feature request related to a problem? Please describe.**
The default servo travel is +-100%. Many of my models require a different travel, e.g. +-77.8%, +-82%(1100-1900) or +- 97.7% (1000-2000).

**Describe the solution you'd like**
Add a "default servo travel" setting as a per-model setting so that the servo travel is set to this value when "resetting" the servo travel value via the popup menu. Add a new menu item to the servo screen "reset all servos", which applies the default travel value, e.g. +-82%

**Describe alternatives you've considered**
Manually setting required channels to +-82%.....

**Additional context**

3djc:
image

This closes #7378 (please read issue for details)

### DIFF
--- a/radio/src/gui/128x64/model_outputs.cpp
+++ b/radio/src/gui/128x64/model_outputs.cpp
@@ -215,6 +215,9 @@ void onLimitsMenu(const char *result)
   else if (result == STR_EDIT) {
     pushMenu(menuModelLimitsOne);
   }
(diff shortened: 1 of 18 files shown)


loetefix:
I think it would be better to copy the PPM center too.

PR #7397, opened 2020-02-18
Bring mixer monitor to 128 screen

3djc:
image

image


bsongis:
@mhotar help! I do like the idea, but not the result on screen!

kilrah:
I'd rather see a fullscreen channel monitor like on 212...

PR #7402, opened 2020-02-20
128 chan monitor

3djc:
image

image

Can be access as mainview, and from curve edit, mix edit and input edit using long menu on X7 and long shift on xlite.

When running PPM US option
image


3djc:
Another one bites the dust

PR #7403, opened 2020-02-20
Add optional alternative button layout for Xlite

netmute:
This adds an optional compile-time flag that changes the button layout of the Xlite transmitter.
New layout looks like this: https://downloads.simonernst.com/xlite/after.jpg

There is a long story behind this layout, that I will spare you 😝
I have been maintaining a fork with these modifications for about a year. My hope is that this is useful enough to be in the main codebase, so I can stop maintaining that fork 😉


bsongis:
How do you want we maintain this? I undestand your wish, but I am afraid it is too difficult for us to accept this kind of changes!

@kilrah @3djc what do you think of it?

kilrah:
I am not for it. Seems to me the "long story" is "I broke the joystick, but since I only fly quads I don't care for the trims".
Can't really consider that pretty special case and workaround as mainstream...

3djc:
I share Kilrah's comment. Forks should be the way to handle that 'I like your software, but I want to it a bit different just for me'

netmute:
Fair enough. Thanks for looking at it though 👍

PR #7406, opened 2020-02-21
Update de.h.txt

loetefix:
More / new translations

### DIFF
--- a/radio/src/translations/de.h.txt
+++ b/radio/src/translations/de.h.txt
@@ -961,15 +961,15 @@
 #define TR_PRESET                      "Preset"
 #define TR_MIRROR                      "Spiegeln"
 #define TR_CLEAR                       "Löschen"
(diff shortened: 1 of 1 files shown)


kilrah:
Did you check them on all UI sizes? The reason they were kind of abbreviated instead of proper German is likely so the strings are short enough to fit on the screen.

loetefix:
i only changed a few things for the 480x272 . 
for example :
 #define TR_COPY_TRIMS_TO_OFS TR("Copy Trim->Subtrim",  "Kopiere Trimmposition auf Subtrim")

the second translation ( Kopiere Trimmposition auf Subtrim ) is only for 480x272 right? or am i wrong?

loetefix:
Ok, i think i was wrong, tomorrow i have to check if it fits.Sorry.

Issue #6581, opened 2019-07-20
X9Lite otx 2.30N49  Modell generating with wizzard  Error at Input Name Ias IGas


hrenz:
*Before submitting,  search the open issue if there is already a report of the bug*

Also often minor bugs are already fixed in nightly builds, often it is
a good idea to test (in a controlled environment like a bench setup)
with a nightly firmware to see if the bug still exists.

**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**General information**
 - OpenTX version (e.g. 2.1.23 or N123)
 - Hardware (e.g. Frsky X9D+ or 9X Pro)
 - OpenTX Companion (e.g. 2.1.23 on Ubuntu 16.04)

hrenz:
Testing X9Lite with different Languages German, France, Englich, Czech,.....
otx 2.30 N49
Modell generating with wizzard

English is ok, Input Name is correct
but at France, and German the Input Name is one Character too left 
so its  Eas  insteed of EGas    or    Ias insteed  IGas

![2019-07-20_151845](https://user-images.githubusercontent.com/6217896/61579572-1cee2280-ab07-11e9-8000-9d83a8830956.jpg)

kilrah:
SD card up to date with 2.3 contents?

hrenz:
its a SD-Card V2.3.0 V0019 Version, original Tx from Frsky /Engel
All test were made with same SD-card, so its the same wizzard for the languages
en, ge, fr, cz

other users have same issues with german,  Input  one character too left
I am surpriced too because this was ok since  V2.10 
so it must be at openTx V2.3.0 German ?  
![2019-07-20_165306](https://user-images.githubusercontent.com/6217896/61580209-7c503080-ab0f-11e9-8f07-efa83b24dcbd.jpg)

3djc:
I cannot seem to reproduce
![image](https://user-images.githubusercontent.com/5167938/61589459-562b9e80-abaa-11e9-8a1e-0e0713e5fb74.png)

3djc:
This is linked to the wizard process, models created with Otx default are fine

I cannot reproduce for french firmware, but german definitively has this issue

hrenz:
Ah, ok  at the German version 
And there is an error with spezialchar  äöü too  
allways H8h insteed of Höh  for  Elevator

kilrah:
[...]

PR #7411, opened 2020-02-25
Fix the special character problem for ö,ü,ä in german

loetefix:
There is still a problem with ä , ö , ü when crate a new model. these changes fix the problem,

tested on T16, Horus and XLITE

### DIFF
--- a/radio/src/strhelpers.cpp
+++ b/radio/src/strhelpers.cpp
@@ -55,6 +55,9 @@ int8_t char2zchar(char c)
 #if LEN_SPECIAL_CHARS > 0
   if ((int8_t)c < 0 && c+128 <= LEN_SPECIAL_CHARS) return 41 + (c+128);
 #endif
+#if defined(TRANSLATIONS_DE)
+  if (c >  'z') return 'a' - c - 10;
+#endif
   if (c >= 'a') return 'a' - c - 1;
   if (c >= 'A') return c - 'A' + 1;
   if (c >= '0') return c - '0' + 27;


bsongis:
What is the issue you want to fix? Logs on SD? On which radio?

loetefix:
When you create a new model in german the stick Höh (ELE) is shown as H8h. In all radios.

loetefix:
The change in strhelpers will fix it.

3djc:
This is 2.3.5

image

loetefix:
Created with wizzard or without?

3djc:
without, and the wizard in english only, so would not exhibit that issue anyway

bsongis:
@loetefix you had the issue with the wizard? I remember of a fix I did some months ago on this topic, are you sure you have the very last wizard lua scripts?

3djc:
Created with wizard

image

loetefix:
No. Not with wizard. I am confused. A friend habe same problem

kilrah:
@loetefix Did you transfer that model into companion and back to the radio?

loetefix:
No. When i create a new model in the transmitter, the höh appears allways as H8h.  This evening i will try it again.I have no wizzards on sd card, because i dont use it

kilrah:
Then can't reproduce it here with 2.3.5. Transfer to companion and back will break special chars in names though, that's known.

loetefix:
Hi, today i loaded the 2.3.5 on a taranis x7. I created a new model with wizard on the taranis and then there ist also H8h instead Höh. (Not created on companion)

3djc:
What wizard did you use ? The one on our pack ?

loetefix:
Yes, from the sd card content. The same problem is when create a model without wizard

3djc:
Well just above you said with wizard. Is it with or without ?

kilrah:
Do you have anything configured in Radio/Hardware for the custom names of the sticks?

loetefix:
No i have not configured custom names and it is with and without wizard. Tomorrow i can make a youtube video.

loetefix:
https://fpv-community.de/threads/jumper-t16.84369/page-75#post-1047064

There are some people they have the same issue.

kilrah:
Ref. #6581

Xillius178:
On an color screen aka Horus, Jumper T16 or Radiomaster TX16S i have no Problem with this german Vowels...


The problem is more likely that the old displays from Taranis etc.

These cannot be shown internally in the display controller.
Since the character sets are not available in this.

PR #7425, opened 2020-03-07
Reset ISRM upgrade moduel counter reset when mode is changed

### LINKED ISSUE #7422: sporadic message: Module upgrade required
Hello everyone,
I converted my Horus X10S to the ISRM module, installed OpenTx 2.3.5 with the options internalaccess, EU and LUA and then flasheded the ISRM firmware update 1.1.5. Everything works as far as it should. I have already bound some D16 receivers but also three Access receivers, all without problems.
I occasionally get a message window: Module upgrade repuired
While this message is on the screen, the model can continue to be controlled.
An external module is not available and is also not activated in the model preset.
The message also comes with different models presets.
Does any of you have any idea where it comes from or whether I have to worry?

image

If I then press RTN, everything continues as normal.

Thanks for your tips
Kind regards Ralf

3djc:
This fixes #7422

### DIFF
--- a/radio/src/gui/480x272/radio_power_meter.cpp
+++ b/radio/src/gui/480x272/radio_power_meter.cpp
@@ -52,9 +52,7 @@ bool menuRadioPowerMeter(event_t event)
     moduleState[g_moduleIdx].readModuleInformation(&reusableBuffer.moduleSetup.pxx2.moduleInformation, PXX2_HW_INFO_TX_ID, PXX2_HW_INFO_TX_ID);
 #if defined(ACCESS_LIB)
     // the module will reset on mode switch, we need to reset the authentication counter
-    if (isModuleISRM(INTERNAL_MODULE)) {
-      globalData.authenticationCount = 0;
-    }
+    globalData.authenticationCount = 0;
 #endif
     /* wait 1s to resume normal operation before leaving */
     watchdogSuspend(500 /*5s*/);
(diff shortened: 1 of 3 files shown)


radio/src/gui/480x272/radio_power_meter.cpp
```diff
@@ -62,6 +67,11 @@ bool menuRadioPowerMeter(event_t event)
     reusableBuffer.powerMeter.attn = 4;
     reusableBuffer.powerMeter.dirty = true;
     moduleState[g_moduleIdx].mode = MODULE_MODE_POWER_METER;
+#if defined(ACCESS_LIB)
```
bsongis: I am said this is not needed (no authentication in Power Meter mode)

radio/src/gui/480x272/radio_spectrum_analyser.cpp
```diff
@@ -56,8 +56,12 @@ bool menuRadioSpectrumAnalyser(event_t event)
   if (menuEvent) {
     lcdDrawCenteredText(LCD_H / 2, STR_STOPPING);
     lcdRefresh();
-    if (isModulePXX2(g_moduleIdx))
+    if (isModulePXX2(g_moduleIdx)) {
       moduleState[g_moduleIdx].readModuleInformation(&reusableBuffer.moduleSetup.pxx2.moduleInformation, PXX2_HW_INFO_TX_ID, PXX2_HW_INFO_TX_ID);
+#if defined(ACCESS_LIB)
```
bsongis: I am said this is not needed (no authentication in Spektrum Analyser mode)

bsongis:
Thanks @3djc

ParkerEde:
Hi Bertrand,
now I've updated to 2.3.7 but I still got the message, if I change from model preset with ACCST to model preset with ACCESS and back to model preset with ACCST.

Kind regards
Ralf

PR #7426, opened 2020-03-07
add lcd.backlightOn()

olliw42:
this PR adds a function lcd.backlightOn() to lua scripts. 

It can be very useful if a script can call backlightOn(). Note that popular scripts such as e.g. the Yappu telemetry script try to achieve exactly this through kind of strange workarounds, which should be taken as further evidence for the need of such a function.

Works great for me on my JumperT16.


radio/src/lua/api_lcd.cpp
```diff
@@ -62,6 +62,20 @@ static int luaLcdClear(lua_State *L)
   return 0;
 }
 
+/*luadoc
+@function lcd.backlightOn()
+
+Switches backlight on
+
+@status current Introduced in 2.2.6
```
bsongis: I am not sure it's right

bsongis:
@kilrah @3djc I am OK with this one once the version is fixed (2.3.6 instead of 2.2.6), before pressing the button I will need your opinion on it

kilrah:
Maybe an optional brightness argument that would adjust the ON backlight level? Could take care of a few other open issues...

radio/src/lua/api_lcd.cpp
olliw42: upppsss ... sorry, corrected to 2.3.6

olliw42:
hm ... I would think this would be a separate function

pl note that this here calls the backlightOn() function, which despite of it's name doesn't switch the lcd on but retriggers it to become bright for a certaan time. It's the function which is called when e.g. a switch is toggled or a stick is moved, and which makes the lcd to go from dim to bright upon an activity. 

This function doesn't have a brightness parameter itself, and I would argue that it would imply some twisting of the current concept if one would want to set it and fall back correctly. The brightnesses appear to be held in the global system setting, and I would thus think one better should a such a function, like it's done with the few other setting, if one thinks one needs it (I don't).

bsongis:
It's easy to change the C function name, but it's less easy to change the Lua function or we will break all existing scripts. So we need to be sure that we give it the right name now!

3djc:
I think I agree that brightness level should be handled a different way, possibly closer to what we do with sound volume (and I would think that in that case, both brightnessOn and brightnessOff levels should be handled).

On the function name, yes it turns LCD on, but it does more than that, it resets the going off timer too. I think it is actually it's way of working, maybe it should be something like resetBacklightTimer ?

bsongis:
We decided to change the function to `resetBacklightTimeout`

olliw42:
this makes lots of sense

many thanks!

PR #7432, opened 2020-03-09
Prevent center beep at boot

### LINKED ISSUE #7431: Center beep is played after boot even if input was already centered
**Describe the bug**

When **Center Beep** is enabled for an input, sound and haptic feedpack are played after boot has completed even if the said potentiometer was already centered.

**To Reproduce**

1. In the current model's setting turn on `Center Beep` for a potentiometer.
2. Ensure that `Pot positions` is set to have the chosen pot centered at boot.
3. Turn off the radio.
4. Center the pot.
5. Turn on the radio.
6. As soon as the splash screen goes away (and the model notes come up, if you have any), the radio briefly beeps and vibrates as the center beep does when crossing mid point.

**Expected behavior**

Center beep (and relative haptic feedback) should only play after the radio has booted and the mid point has been crossed once. Behavior should be similar to how the `!1x` frequency option works in Special Functions.

**General information**
 - OpenTX version: 2.3.5
 - Hardware: FrSky Horus X10S
 - ~~OpenTX Companion~~

nmaggioni:
Fixes #7431, tested on my Horus X10S.

### DIFF
--- a/radio/src/mixer.cpp
+++ b/radio/src/mixer.cpp
@@ -460,7 +460,7 @@ void evalInputs(uint8_t mode)
     if (mode == e_perout_mode_normal) {
       if (tmp==0 || (tmp==1 && (bpanaCenter & mask))) {
         anaCenter |= mask;
-        if ((g_model.beepANACenter & mask) && !(bpanaCenter & mask) && !menuCalibrationState) {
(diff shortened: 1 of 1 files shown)


bsongis:
Accepted - after a long phone call ;)

PR #7444, opened 2020-03-16
Radio tools related fixes

3djc:


radio/src/gui/128x64/view_main.cpp
```diff
@@ -364,7 +364,11 @@ void menuMainView(event_t event)
       break;
 
     case EVT_KEY_GENERAL_MENU:
+#if defined(RADIO_TOOLS)
       pushMenu(menuRadioTools);
```
bsongis: menuTabGeneral[0] ?

PR #7454, opened 2020-03-24
Nicholas/telemetry for rb30 40

Nicholas-Luoyi:
Added telemetry for rb30 40


radio/src/telemetry/frsky_sport.cpp
```diff
@@ -308,6 +317,22 @@ void sportProcessTelemetryPacketWithoutCrc(uint8_t origin, const uint8_t * packe
           }
 #endif
         }
+        else if (dataId >= RB3040_CH1_2_FIRST_ID && dataId <= RB3040_CH1_2_LAST_ID) {
```
bsongis: why 4 tests here, only one is needed, right?

radio/src/gui/common/stdlcd/draw_functions.cpp
```diff
@@ -407,6 +407,30 @@ void drawSensorCustomValue(coord_t x, coord_t y, uint8_t sensor, int32_t value,
           }
         }
       }
+      else if (telemetrySensor.id >= RB3040_OUTPUT_FIRST_ID && telemetrySensor.id <= RB3040_OUTPUT_LAST_ID) {
+        if (telemetrySensor.subId == 0) {
+          if(value == 0) {
```
bsongis: cosmetics: space missing between `if` and `(`

radio/src/gui/common/stdlcd/draw_functions.cpp
```diff
@@ -407,6 +407,30 @@ void drawSensorCustomValue(coord_t x, coord_t y, uint8_t sensor, int32_t value,
...
+            lcdDrawText(x, y, "OK", flags);     
+          }
+          else {
+            for (uint8_t i=0; i<9; i++) {
```
bsongis: In 99% cases value should be 0, right? If yes I would enter into this loop only if `value != 0`

PR #7472, opened 2020-03-29
Support for Multi FrSky 2.1.x

3djc:
Includes GUI review for 128 and 212 (was too packed)

image

image


martimbrepost:
@kilrah 
Do you think that the code from the multi protocol branch for Frsky ACCST v2.1x can also be added to the SPI code in Betaflight ? 
If yes, in what release do you think it could be implemented ?

Thanks a lot,
Martin

(I'm thinking Tinyhawk 2 with MatekF411RX and Xlite Pro ACCESS :-) )

kilrah:
I don't know, I have nothing to do with betaflight. But of course it's possible technically, the RX code is available in the multiprotocol project.

PR #7478, opened 2020-03-31
Bsongis/ras fix

### LINKED ISSUE #7471: RAS in OTX 2.3.7 on TaranisX9D now works only partially.
I simulate the RAS value by sliding the metal tube onto the antenna.(internal XJT, original antena)

On the hardware page (6/7) is **RAS --- / 33**, the value changes as the tube moves along the antenna,
RAS > 51 triggers an antenna warning.
screen-2020-03-29-110718

But the following doesn't work:

1) The analog submenu shows **RAS 0** all the time.
screen-2020-03-29-104749
2) A simple LUA script with **getRAS ()** doesn't work, return 0 (in OTX 2.2.4 without problems).

There are 2 RAS values:
telemetryData.swr**Internal**.value()
telemetryData.swr**External**.value()
 
I think the hardvare page displays **Internal / External**,
the antenna warning uses both values, but the analog submenu and getRas() only **Internal**.

It looks as if the internal and external have been reversed. (internal XJT has telemetryData.swrExternal.value () )

Of course I can be wrong.

bsongis:
Fixes #7471

### DIFF
--- a/radio/src/gui/212x64/radio_diaganas.cpp
+++ b/radio/src/gui/212x64/radio_diaganas.cpp
@@ -38,12 +38,4 @@ void menuRadioDiagAnalogs(event_t event)
     lcdDrawNumber(x+10*FW-1, y, (int16_t)calibratedAnalogs[CONVERT_MODE(i)]*25/256, RIGHT);
 #endif
   }
-
-  // RAS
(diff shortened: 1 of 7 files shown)


robustini:
@bsongis, could you explain to me the principle of value about this RAS?
In my Horus X10S Express I always have RAS 0/--- in any condition, external or internal antenna.

robustini:
@bsongis, could you explain to me the principle of value about this RAS?
In my Horus X10S Express I always have RAS 0/--- in any condition, external or internal antenna.
With X12S (ISRM) and internal antenna ---/---, with the external 0/---.

kilrah:
ISRM only measures RAS on external, IXJT did both

PR #7483, opened 2020-04-01
Display INT_MIN correctly

kilrah:


bsongis:
I propose a first fix to avoid a cast to int64_t by checking INT_MIN

PR #7484, opened 2020-04-01
Adds a page to the windows installer to select the location of the start menu entries

### LINKED ISSUE #7475: Create Windows start menu shortcuts for all users instead of "current" user
When installing the Companion under Windows, the installer requires admin privileges. During the installation process it creates start menu entries for the "current" user, which is the Administrator account, if your own user does not have admin privileges (which it never should have).

In consequence, that start menu entries are only visible for the Administrator but not for all other users of the system.

**Expected behavior**
Start menu entries are created for "all" users so that each user is able to start the Companion through the start menu.

**General information**
 - OpenTX Companion up to and including 2.3.7 on Windows

**Note**
I am able to create a pull request for the installer. The PR would modify the creation of the start menu entries and it would be backward compatible in the uninstall process, so that start menu entries in the old and new location will be removed.

pre-martin:
Fixes https://github.com/opentx/opentx/issues/7475

- New installer page for selecting the start menu location (either for current user or all users)
- Default is always for "current user"
- Installer retains radio button state during "Back" and "Next", but not during different invocations of the installer
- Page is only shown if "Do not create shortcuts" on the previous page was not selected
- English and French translations
- Uninstaller always tries to remove start menu entries at both locations

installer-page-en

installer-page-fr

### DIFF
--- a/companion/targets/windows/companion.nsi.in
+++ b/companion/targets/windows/companion.nsi.in
@@ -6,6 +6,7 @@
 ;Include Modern UI
 
   !include "MUI2.nsh"
+  !include "nsDialogs.nsh"
   !include "@CMAKE_CURRENT_SOURCE_DIR@\..\targets\windows\FileAssociation.nsh"
 
 ;--------------------------------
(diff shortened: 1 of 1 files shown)


3djc:
Thanks a lot

PR #7491, opened 2020-04-02
Adds the version into the Windows registry (and "Programs and Features")

pre-martin:
Displays the installed version in the Windows "Programs and Feature" application.

It always displays the full version.

Currently without version information:

without-version

After applying this PR with version number:

with-version

Note: The version is "2.x" and "2.x.7" because I am using modified values for the CMake variables "VERSION" and "VERSION_FAMILY". The final build will contain correct values, of course, like "2.3" and "2.3.8".


kilrah:
Nice idea, thanks!

Issue #7473, opened 2020-03-30
Make custom sensor use precision from set value, not definition


RealTadango:
When adding / discovering a custom sensor (not known ID by OpenTX firmware) setting the precision does not work. When the raw value is 9999 and you set the precision to 0.00 the value is shown as 9999,00, which makes no sense. The value should be shown as 99,99.

RealTadango:
Just my personal notes (trying to fix this :))

const FrSkySportSensor * sensor = getFrSkySportSensor(id, subId);
  uint8_t precision = 0;
  if (sensor) {
    precision = sensor->prec;
  }

So this part is the issue. precision is only set when a known FrSky SPort sensor can be matched to the id. The solution would be to get the configured sensor data and use the precision from there.

RealTadango:
struct TelemetrySensor contains prec, but newVal = sensor.getValue(newVal, unit, prec); uses prec as a parameter, which isn't set for custom sensors

RealTadango:
Fixed

PR #7508, opened 2020-04-09
Using the right precision for unknown (custom) sensors

RealTadango:
Precision is set to out of range value 255 unless there is a known sensor found. If out of bound spec is found when setting the value then the configured spec is used.


RealTadango:
#7473

Issue #6589, opened 2019-07-28
Add Lua serialRead() function for Serial Port


Ryanf55:
Hello,

This is for a continuation of support of Lua access to the serial port. While serial port access has been implemented in #6008 , that only included serialWrite. This feature request is for the reverse,serialRead(). Because communication with other hardware is now possible through Lua, it would be good to be able to also receive data from outside hardware. 

**Describe the solution you'd like**
Please implement a serialRead(numBytes) function, perhaps serialReadLine('\n'), where the line ending char is passed in. From the last discussion, it was mentioned that serialRead required was computationally expensive and difficult to implement. As a consolation, digitalRead() of the RX line would also be ok to at least get true/false responses from 3rd party hardware. 

**Describe alternatives you've considered**
I worked with S.Port a bit and trying to convert incoming serial data into S.port packets but it was very slow and complicated. 

**Additional context**
None required.

Ryanf55:
![image](https://user-images.githubusercontent.com/25047695/73575440-6cf25e00-4435-11ea-9bc4-7f3d3f7b0ef8.png)
Here is the use case: I am able to configure my ClearView receiver, but I have no way of getting confirmation the settings took from the device. I would love to set up audio alerts when the ClearView's battery is low. These all require serialRead().

I am happy to provide testing of any implementations for this.

PR #7520, opened 2020-04-14
lua serialRead

tomvand:
This PR adds a lua serialRead function. The function can be used to read up to a maximum number of characters from the serial input fifo, or up to a newline character (if present).

The serialRead function can be used when opentx is compiled with -DCLI=NO. Otherwise, it is not clear whether serial characters should go to lua or to the cli. Characters are sent to lua using the same mechanism as the cli and with an equally sized fifo, so memory consumption and runtime should be about the same.

Tested on Taranis X9D+

This is my first contribution to opentx, so let me know if I missed something obvious ;)

Related: #6589

### DIFF
--- a/radio/src/lua/api_general.cpp
+++ b/radio/src/lua/api_general.cpp
@@ -1597,6 +1597,23 @@ static int luaSerialWrite(lua_State * L)
   return 0;
 }
 
+/*luadoc
+@function serialRead???
+@param Maximal number of bytes to read
+// TODO or if 0 read to newline if present? Need peek to check that first...
+
+@retval number of bytes read
+@retval string
+
+Writes a string to the serial port. The string is allowed to contain any character, including 0.
+
+@status current Introduced in TODO
+*/
+static int luaSerialRead(lua_State * L)
+{
+  return 0;
+}
+
 const luaL_Reg opentxLib[] = {
   { "getTime", luaGetTime },
   { "getDateTime", luaGetDateTime },
@@ -1644,6 +1661,7 @@ const luaL_Reg opentxLib[] = {
   { "multiBuffer", luaMultiBuffer },
 #endif
   { "serialWrite", luaSerialWrite },
+  { "serialRead", luaSerialRead },
   { nullptr, nullptr }  /* sentinel */
 };
 
--- a/radio/src/lua/api_general.cpp
+++ b/radio/src/lua/api_general.cpp
@@ -1615,7 +1615,8 @@ static int luaSerialRead(lua_State * L)
   if (num > LUA_FIFO_SIZE) num = LUA_FIFO_SIZE;
   uint8_t str[LUA_FIFO_SIZE];
   uint8_t *p = str;
-  while (luaRxFifo.pop(*p++)) {
+  while (luaRxFifo.pop(*p)) {
+    p++;  // increment only when pop was successful
     if (num == 0) {
       if (*(p - 1) == '\n' || *(p - 1) == '\r') break;
     } else {
--- a/radio/src/lua/api_general.cpp
+++ b/radio/src/lua/api_general.cpp
@@ -66,6 +66,10 @@
   { "EVT_"#xxx"_LONG", EVT_KEY_LONG(yyy) }, \
   { "EVT_"#xxx"_REPT", EVT_KEY_REPT(yyy) }
 
+#if defined(LUA) && !defined(CLI)
+Fifo<uint8_t, 256> luaRxFifo;
+#endif
+
 /*luadoc
 @function getVersion()
 
--- a/radio/src/lua/api_general.cpp
+++ b/radio/src/lua/api_general.cpp
@@ -1598,12 +1598,15 @@ static int luaSerialWrite(lua_State * L)
 }
 
 /*luadoc
-@function serialRead???
-@param num (optional) maximum number of bytes to read. If left out, will read up to and including newline character or end of buffer.
+@function serialRead([num])
+@param num (optional): maximum number of bytes to read.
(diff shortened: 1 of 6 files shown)


radio/src/lua/api_general.cpp
```diff
@@ -1593,6 +1597,43 @@ static int luaSerialWrite(lua_State * L)
...
+  int num = luaL_optunsigned(L, 1, 0);
+
+#if defined(LUA) && !defined(CLI)
+  if (num > LUA_FIFO_SIZE) num = LUA_FIFO_SIZE;
+  uint8_t str[LUA_FIFO_SIZE];
+  uint8_t *p = str;
+  while (luaRxFifo.pop(*p)) {
+    p++;  // increment only when pop was successful
+    if (num == 0) {
+      if (*(p - 1) == '\n' || *(p - 1) == '\r') break;
```
bsongis: there is a buffer overflow here

bsongis: in case num == 0

radio/src/lua/lua_api.h
```diff
@@ -41,6 +41,11 @@ extern "C" {
   #endif
 #endif
 
+#if !defined(CLI)
+#define LUA_FIFO_SIZE 256
+extern Fifo<uint8_t, LUA_FIFO_SIZE> luaRxFifo;
```
bsongis: 256 bytes RAM
we take the risk that some scripts which are OK in 2.3.5 are KO in 2.3.6

bsongis: this buffer could be dynamically created on first `serialRead`

radio/src/targets/common/arm/stm32/usbd_cdc.cpp
```diff
@@ -223,6 +223,13 @@ static uint16_t VCP_DataRx (uint8_t* Buf, uint32_t Len)
     cliRxFifo.push(Buf[i]);
   }
 #endif
+#if defined(LUA) && !defined(CLI)
+  // copy data to the LUA FIFO
+  for (uint32_t i = 0; i < Len; i++)
```
bsongis: code style

radio/src/lua/api_general.cpp
tomvand: Oops, you're right! Fixed.

radio/src/lua/lua_api.h
tomvand: Agreed. Buffer is now dynamically created. This should also help with the ram compatibility, right? (Since older scripts do not call serialRead and therefore do not instantiate the fifo?)

radio/src/targets/common/arm/stm32/usbd_cdc.cpp
tomvand: Fixed the brackets, was that what you meant?

radio/src/lua/lua_api.h
bsongis: exact

radio/src/lua/api_general.cpp
```diff
@@ -1593,6 +1597,50 @@ static int luaSerialWrite(lua_State * L)
...
+  }
+  uint8_t str[LUA_FIFO_SIZE];
+  uint8_t *p = str;
+  while (luaRxFifo->pop(*p)) {
+    p++;  // increment only when pop was successful
+    if (num == 0 && (*(p - 1) == '\n' || *(p - 1) == '\r')) {
+      break;  // found newline
+    }
+    if (p - str >= num || p - str >= LUA_FIFO_SIZE) {
+      break;  // num reached or buffer full
```
bsongis: Still not good, it won't work with num == 0 (you want a full line)

tomvand: Wow I'm not sure how I missed that :S

I've pushed a new fix, hope it is correct this time.

3djc:
Maybe this would be better ? (unsure)

```
if (num == 0) {
      if (*(p - 1) == '\n' || *(p - 1) == '\r') {
        break;  // found newline
      }
 }
 else if (p - str >= num || p - str >= LUA_FIFO_SIZE) {
      break;  // num reached or buffer full
    }
}
```

tomvand:
That is very close to what I had in my first version, but in this case the buffer size is not checked when num == 0, which could cause an overflow if no newline character is found.

3djc:
true, it is just that you test num twice in a row so maybe 

 ```
while (luaRxFifo->pop(*p)) {
    p++;  // increment only when pop was successful
    if (p - str >= LUA_FIFO_SIZE) {
      break;  // Buffer full
    }
    if (num == 0) {
	if (*(p - 1) == '\n' || *(p - 1) == '\r')) {
	  break;  // Found newline
	 }
    }
    else if (p - str >= num) {
      break;  // Requested number of characters reached
    }
  }
```

Btw, I'm loving it, I use serialWrite to debug my scripts, really handy !

tomvand:
I see what you mean. I wrote it this way to be as explicit as possible, assuming that the compiler would probably optimize this anyway. If you want me to change it let me know ;).

Happy to hear it! I am tbh still a bit unsure about the lua api of this function (returning strings before a newline is found if that character is not in the buffer, returning empty strings when the buffer is empty or the fifo could not be created, not being able to change the newline character). If you have any thoughts on that (or better: some experience testing this function in lua) I would be very interested to hear what you think. 
So far I have used the serialRead command myself to set up a bidirectional serial link over FrskyX/S.Port to the paparazzi autopilot ([here](https://github.com/tomvand/paparazzi/blob/trashcan_spi_cc2500_pr/sw/tools/opentx_lua/sp2ser.lua)), with success.

PR #7531, opened 2020-04-18
Check multi channel order

3djc:
Prevent flashing a multi firmware that does not match radio setting


3djc:
Not such a good idea apprently

PR #7542, opened 2020-04-21
Add es and pt special characters

dgatf:
Hi there,

Yet another locale related PR. This PR adds the missing special characters for spanish language for stdlcd (fonts std and sqt5) and colorlcd (font Roboto)

Adds the following characters: ÁáÉéÍíÓóÚú

Also updates the es translation including the new characters

Cheers,
Daniel

x7-std
x7-sqt5
x10-Roboto


dgatf:
Ok, there were two issues. One related to the ttf horus fonts causing some tests to fail and other related to char mapping. Both fixed

First issue: The generation of the ttf fonts for the horus with `make ttf_horus_fonts` (using python pillow) was producing different font png images from current ones with few characters being slightly wider so it will fail any test with text drawn. Therefore reference images needs to be updated. I used the 'failed' screenshots of the tests to update the reference images

Also I've noted that at the end of the ttf fonts images _pt_ characters were missing and _fi_ were duplicated. This is fixed also

Second issue: I didn't remap the special characters after _es_ so it was breaking _it_ and _pl_ special characters for the color display (in lcd.cpp)

PR summary:

- Add _es_ special characters for all radios
- Add missing _pt_ special characters for the color display
- Update the reference images generated with python pillow with ttf fonts
- Update es translation adding special characters

Side note for myself: remember to run all tests before commiting a PR

bsongis:
good job, thanks!

PR #7551, opened 2020-04-25
Horus batt widget, reduce usage + fix issues 7539

### LINKED ISSUE #7539: Possible error in calculating the% in the battery script
**Horus X10S Express V2.3.7 EU** 

Hi, I'm not sure it's a problem, but it seems to me.
In the last two or three versions I have seen that improvements have been made in the battery script, however the value of the battery in % that is presented every time is smaller and moving away from the battery value that my CallMeter 8 presents. , which I use as a pattern.
I have verified that the voltage values of the cells and the battery are the same that the FLVSS presents, however the calculation in % of battery capacity has been moving away (decreasing) version to version of the one indicated by my meter. I am aware that my meter can do it wrong, but it certainly does as well or as bad today as it does three versions.

Could you check the calculation of the % of the battery?

offer-shmuely:
reduce usage from 80-90% to 12-25%
Fixes #7539 - calculate % by average value (instead of min-cell-voltage)


kilrah:
You need to rebase your branch on 2.3

offer-shmuely:
Hi
Sorry, but I did not understand.
Can you **_please_** explain what I need to do? 
I tought merging the opentx/opentx/2.3 to my clone repo, is good enough.

image

3djc:
I wonder if this has not become too compilcated for its own good. Whats the point of computing per cell % ? keep per cell lowest % history.

All we really need is a live total % (tot volt / cell cnt -> %), lowest cell values (not %). Back to basics !

offer-shmuely:
we can "back to basic", however I believe that in lipo you are referring to each of the cell individually, so you need to see it as such
please note the modern field-tester (I consider the ISDT as the best one) do show a bar on each cell.
and since we did like to be similar to a field tester (from the other thread), 
I went as close as I can to a good one (in contrast to older field-tester with simple black-and-white, and minimal graphics that show mostly text values)



isdt-battgo-field-tester

3djc:
I have (finaly, sorry) found time to have a look. We have discussed and do not want it based on lowest cell. Also:

 1/4 is quite broken
image

Also on a wider view, we can see that right allignement of cell table is not the same as channel view
image

I also noticed that we compute cellDataHistoryLowestPercent, which really doesn't make sense, and is only used in 1/4th view. I would just remove that

Also, you compute cellDataLivePercent, I'm not sure what the value is to compute cell by cell %, especialy when we know it is a cpu intensive task

3djc:
I have put a slimed down version here, https://github.com/opentx/opentx/blob/3djc/battcheck/radio/sdcard/horus/WIDGETS/BattCheck/main.lua

Tell me what you think

bsongis:
Done in another commit it seems? Feel free to reopen if needed

PR #7559, opened 2020-04-28
RM TX16S support

kilrah:


Eldenroot:
Great! Mine is on the way :)

Anyway OpenTX 2.4 is quite far away (touch panel control) - can we expect in late 2020 - at least beta version please or could you make some video how it will work?

PR #7565, opened 2020-05-01
Option to prompt for radio profile on companion start

### LINKED ISSUE #7514: Companion - optional selection for Radio Profile on startup
Add a "pop up" selection box on startup that prompts user to select a saved radio profile to be active.

For those of use with multiple radio types (X9D+, 10S, etc., etc.)  it is common to start companion and forget to choose the correct saved profile, resulting in starting over when the error found, possible wrong radio information (calibrations, etc.) saved to the radio, or errors due to wrong type.

Whether this feature to be active can be enabled/disabled via a checkbox under "Application Settings".

elecpower:
Fixes #7514
The default setting is 'off' to maintain current functionality.
If the setting is 'on' and more than 1 profile is defined, the user will prompted to select a profile.
The default profile is the one used in the previous session.

Application Settings

Startup Profile Chooser

### DIFF
--- a/companion/src/CMakeLists.txt
+++ b/companion/src/CMakeLists.txt
@@ -250,6 +250,7 @@ set(companion_SRCS
   wizarddialog.cpp
   styleeditdialog.cpp
   dialogs/filesyncdialog.cpp         # TODO move to own lib with other dialogs
+  profilechooser.cpp
   )
 
 set(companion_MOC_HDRS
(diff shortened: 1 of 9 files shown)


3djc:
omg, i love this feature !

bsongis:
nice idea

sande005:
THANK YOU for taking up 7514 and implementing this!

Miami-Mike:
Many of us help others out by loading and working with their otx files, often without knowing in advance what the radio type is. When we find that we have the wrong type selected we then have to decline the offer to convert, go back to square one, change radio profiles, and start over. This is the problem that I had hoped was being addressed.

Upon an otx file load with the wrong radio type selected, this option should then show what type of radio the file is meant for and then offer to either switch to the correct profile and continue, or keep the currently selected profile and convert the otx file accordingly.

PR #7568, opened 2020-05-02
Improve TX16s UARTS

3djc:
More logicaly, left UART is AUX, right is AUX2 (side benefit, allows usage of AUX even when using BT).
Also use 115200 for telem mirror for TX16s


secretman28:
is it possible to get this uart and make it do f.port for flashing receiver directly from there like the x-lite? its the only miss on the tx16s

3djc:
You can flash receiver from external module bay on tx16s, and no, sport requirtes hardware inverter and single wire rx/tx, so cannot be done on a uart without additional hardware

secretman28:
I know on the tx bay but was wondering ok tanks four taking the time

Issue #7569, opened 2020-05-02
Multi chan namer script won't run on X9D+ (Script panic - not enough memory)


MRC3742:
MultiChannelsUpdater.lua will not run on the X9D+ (2014) Transmitter.
Although this script runs fine on the X9D+ Simulator or the Jumper T16 Transmitter.
I have also tried the updated MultiChannelsUpdater.lua from the current 2.3 branch.
I have deleted all but one model to help free up memory but it still will not open.
It immediately throws the Script panic - not enough memory error.

Commenting out the last 30 Protocols in the "local Protocols =" listing allows it to open and run.
Is there a way to load the protocols full information only when it is called or detected?

Also I feel there needs to be a confirmation screen before running the script.
such as "Update Channel Names?" No->[EXIT]   Yes->[ENTER]
This would be especially useful when using a scroll wheel TX as sometimes it moves selections just as it is pressed to execute a tool script.

**General information**
 - OpenTX version is Compiled 2.3.8 from current 2.3 branch
 - Hardware Taranis X9D+ (2014)
 - External iRangeX plus with MPM version 1.3.0.91

MRC3742:
I have downloaded updated "MultiChannelsUpdater.lua" and new file "MultiChan.txt" to the V0027 SD_Card  TOOLS folder on the Taranis X9D+ (2014) which is updated in above 7577

I happy to report it is now working as expected with no memory errors.
It now also requires an "Enter" to execute the script with a chance to "Exit" before it runs.

Closing this issue as it has been fixed pending #7577 merge - Rich

PR #7577, opened 2020-05-03
Update Multi Channels Namer LUA script

pascallanger:
Solve issue: https://github.com/opentx/opentx/issues/7569
Additional MultiChan.txt file which is parsed 20 lines at a time for every loop of the LUA script.
Changes to output names are only applied if you press enter.


MRC3742:
I have downloaded updated "MultiChannelsUpdater.lua" and new file "MultiChan.txt" to the V0027 SD_Card  TOOLS folder on the Taranis X9D+ (2014)

I happy to report it is now working as expected with no memory errors.
It now also requires an "Enter" to execute the script with a chance to "Exit" before it runs.

Thanks - Rich

PR #7578, opened 2020-05-04
Update es translation of Companion

dgatf:
Hi,

This is a complete update of the spanish translation of Companion. I've also added myself to the translators list in credits.txt, if this is not ok let me know and I'll remove it

Cheers,
Daniel


CREDITS.txt
```diff
@@ -11,6 +11,7 @@ Arne Schwabe (schwabe)
 
 [Translators]
 [fr] Sebastien Charpentier (LapinFou)
+[es] Daniel Gorbea (dgatf)
```
bsongis, line 14: good :)

dgatf: Thanks

PR #7581, opened 2020-05-04
Preliminary T18 support

3djc:


Eldenroot:
Do you have a link to T18? I can see only an advertisement that more info will be released on 4th of May but still nothing...thx. anyway good job!

kilrah:
They released a video earlier today that doesn't show much.

danrough:
I have a T18 and a computer which runs OS X. If I can be of assistance in testing (I won't be of any use in developing the changes, but I have a track record of being able to break things in weird and wonderful ways!) in anyway, let me know. Thanks for doing what you've done already!

PR #7601, opened 2020-05-12
Fix 7598 - Wrong Flight Mode Index when exiting from Gvar Menu (128x64 display)

### LINKED ISSUE #7598: Wrong Flight Mode Index when exiting from Gvar Menu (128x64 display)
Exiting from Gvar Menu opens a wrong flight mode and not the one from the gvar menu is opened. Instead it opens the flight mode of the gvar minus one. This menu is only used in 128x64 displays. Other displays are not affected

Tested with opentx 2.3.8-X7

dgatf:
This PR fixes #7598

Adds a variable for gvar index to keep the flight mode index


radio/src/gui/128x64/gui.h
```diff
@@ -216,6 +216,7 @@ extern uint8_t s_copyMode;
 extern int8_t s_copySrcRow;
 extern int8_t s_copyTgtOfs;
 extern uint8_t s_currIdx;
+extern uint8_t s_currIdxGvar;
```
bsongis: can be merged with s_curveChan

dgatf: Yes, updated

bsongis:
Thanks!

PR #7632, opened 2020-05-25
TX16S internal gps option

3djc:
This option allows using an ublox GPS connected to BT soldering pad

image

GPS is then manage internaly by OpenTX, and is used to set time and pilot position, or LUA getTxGPS()

Status of GPS can be viewed in statistics screen 2
image


StefanKellerAC:
would this be also possible with X10 and internalACCESS (where internal SBUS is not available anymore in module bay) ?

kilrah:
It's completely unrelated?

StefanKellerAC:
In a x10 classic you can put an openXSensor GPS inside, connecting it to the S.Port Pin in the module bay and get the TX Position.
with an ACCESS update module this is not possible anymore,
or is there another possibility for an internal GPS?

ItsMeAubey:
Excuse me if this is the wrong place to ask, but is an internal bluetooth module for tx16s going to be released or supported by opentx?

3djc:
You can use the FrSky one for Horus, or use an external one on the UART, but that need external configuration

kolaa:
I tried this mod on 2.3.9 -TX16S freese on startup, with black screen. Please, share the GPS module settings.

ItsMeAubey:
@kolaa It is likely that you have damaged your radio or soldered incorrectly. There is no reason why any of these changes should stop your radio from booting.

Tygrys-1:
Can this be external GPS on AUX2 port?

pierrotm777:
Does exist a solution for have a BT and GPS options ?

PR #7635, opened 2020-05-26
Additional MPM 1.3.1.1 support updates

MRC3742:
Companion and Radio GUI Updates for Protocols and Sub Protocols to latest MPM version 1.3.1.1

Cross referenced from Multi_Names.ino for accuracy.
Reordered a few Protocols in lists to better follow the Protocol numerical listings.


3djc:
Thanks a lot

MRC3742:
@3djc You're welcome, Happy to contribute!

One item that I did not address is in Companion multiprotocols.cpp
DSM Protocol should have an "Enable max throw" checkbox available in Companion.
It has been added to the Radio GUI but is not listed in the current Companion.
I left the STR_MULTI option as nullptr as I believe there are other changes needed to make it work.
These changes are beyond my minimal code knowledge.

If time permits - Rich

PR #7643, opened 2020-05-27
Fix  6853

### LINKED ISSUE #6853: Curve points in one curve changing when editing another curve
2.2.2 and 2.2.4 same issue
X10S
Opentx Companion

When I edit a curve it corrupts the other curves.
First discovered this when the battery died whilst I was programming my tx when I recharged and restarted it the curve points had changed. So I tried to update firmware to 2.2.4

Now curves points are changing when I edit another curve. I can corrupt 1 or many other curves.

Any ideas I see this as critical, what should I do.

dgatf:
Fix #6853. This issue is affecting radios with rotary encoders

When increasing the number of points of the curve, it has to be done in increments of one. If the rotary goes faster then the new points are also copied to adjacent curves. So lets force low rotary speed when changing the number points of the curves


bsongis:
Excellent, you found an issue which was raised on curves but I had no clue it was when the rotary encoder was moving fast

PR #7665, opened 2020-06-03
Fix tts for polish language, need new sounds files.

ectogens:
Fix for tts.pl to correct spell polish language. To work properly need a new sounds files.
Where I can share a new files?


3djc:
You probaly need to create a polish sound creating routine (exemple here opentx/radio/util like cz for exemple)

PR #7672, opened 2020-06-04
fix: reverse t12 pot direction. up -> high, down -> low

knoopx:
did this for the t12 pro, which i don't know why its setup like that because it feels completely unnatural. i don't know if this affects the original t12.

### DIFF
--- a/radio/src/targets/common/arm/stm32/adc_driver.cpp
+++ b/radio/src/targets/common/arm/stm32/adc_driver.cpp
@@ -26,6 +26,8 @@
   const int8_t adcDirection[NUM_ANALOGS] = {1,-1,1,-1,  1,1,1,  -1,1,  -1,1};
 #elif defined(RADIO_TX16S)
   const int8_t adcDirection[NUM_ANALOGS] = {1,-1,1,-1,  1,1,1,  -1,1,  -1,1};
+#elif defined(RADIO_T12)
+  const int8_t adcDirection[NUM_ANALOGS] = {-1,1,-1,1,  -1,-1,  1,1};
 #elif defined(PCBX10)
   const int8_t adcDirection[NUM_ANALOGS] = {1,-1,1,-1,  -1,1,-1,   1,1,    1, -1};
 #elif defined(PCBX9E)


kilrah:
Yes it does, so can't be changed.

knoopx:
@kilrah could you elaborate more? how does the t12 behave, the same way as the pro? are you willing to accept this behaviour change if I properly implement this 
 (by differentiating pro vs original maybe by introducing a new `RADIO_T12_PRO`)?

3djc:
There are users out there with programmed models. Because of that, it cannot be changed anymore

kilrah:
No, they are the same radio, and there is no point in either breaking people's models (people have been getting T12Pros for 2 months) nor having it different between both.

knoopx:
> There are users out there with programmed models. Because of that, it cannot be changed anymore

well, not even on 2.4? did you manage to reach 2.3 without ever breaking people's configuration? do you agree at least the current direction does not feel natural?

> No, they are the same radio, and there is no point in either breaking people's models (people have been getting T12Pros for 2 months) nor having it different between both.

so t12 pot direction is the same as t12 pro?

3djc:
yes we agree, and yes t12 and t12 pro directions are the same

kilrah:
We have in a few occasions broken configuration but not on essential things that directly control the model, and that was in order to introduce major new functionality. This is just cosmetic.

If we were involved like we are with other manufacturers we would have changed that before the initial release, but in the case of the T12 Jumper did things their way back in the day by just taking the FrSky X7's firmware and hacking so the buttons worked without any other considerations.

knoopx:
alright, thanks for the explanation. I have no idea how manufacturers/opentx collaborate when developing new products but I guess I will need to mantain my own fork unless this issue gets some traction.

3djc:
Traction or not this will not change, it is too late for that, but as long as user are getting radios before we can (again with t18...) this kind of things will continue to happen. Things are improving tho

lshems:
I have suggested in the past to dive this by calibration. Allow all ADC values to be set as top, middle and low value, and the problems are gone for all inputs. So instead of asking for center, acknowledge and asking for extremes, acknowledge, ask for center, acknowledge, max, acknowledge, min, acknowledge.

Simpel Universal solution, I would think.

knoopx:
@lshems indeed that would be even better but will break backwards compatibility too... or not?

lshems:
Why?

PR #7677, opened 2020-06-05
Throttle trim - Fixes 5635

### LINKED ISSUE #5635: OpenTx 2.2 - Trim Idle Only still doesn't follow the assigned throttle trim when you cross the trims
I'm a fan of crossed trims and would really like to see this problem solved. For electric models the crossed trims work great since the Trim Idle Only function isn't needed. But for gas/glow engines where the throttle trim is only effective at idle, selecting the Trim Idle Only function doesn't follow the selected trim input for the throttle channel. Instead the Trim Idle Only function always affects the THR trim regardless of which trim input device is assigned to the throttle channel. The result is that when cross trims are selected the Trim Idle Only function ends up affecting the elevator channel trim instead of the throttle channel trim.

This has been an open issue since way back on November 29, 2013
https://github.com/opentx/opentx/issues/170

It's been over four years and OpenTx is now on 2.2.1.  Is anybody working on solving this bug?

dgatf:
This PR fixes #5635. Though it is not noted in the linked issue, a similar problem occurs with throttle inversed + non default trims

The PR adds a model option, _throttle trim switch_, that allows to redefine the trim switch linked to the throttle stick. The option defaults to the throttle trim switch, so there is no impact on current models

When using non defaults trims (e.g. cross trims), the new trim linked to the throlle stick can be redefined with this option and trims with different behauviour (inversed and idle) are applied to the desired stick


radio/src/mixer.cpp
```diff
@@ -527,13 +527,19 @@ int getStickTrimValue(int stick, int stickValue)
...
-      trim = ((g_model.throttleReversed ? (trim+trimMin) : (trim-trimMin)) * (RESX-stickValue)) >> (RESX_SHIFT+1);
-    }
-    if (g_model.throttleReversed) {
+  uint8_t thrTrimSw;
+  if (g_model.thrTrimSw == 0) //this is needed for compatibility with existing models
+      thrTrimSw = MIXSRC_TrimThr - MIXSRC_FIRST_TRIM;
+    else if (g_model.thrTrimSw == MIXSRC_TrimThr - MIXSRC_FIRST_TRIM)
+      thrTrimSw = 0;
+    else
+      thrTrimSw = g_model.thrTrimSw;
```
dgatf: This is needed for compatibility with current models. May be removed with next eeprom update

sande005:
Does this play nicely with defining trim behavior in Flight Modes, and using Trims>Subtrims in Outputs?  So very pleased to see this issue moving forward!

dgatf:
Yes, same as now for flight modes and subtrims. Now you can use other trims than defaults and _throttle idle trim_ or _inverse throttle_ will be applied through the selected trim switch defined in the new option _Throttle trim switch_ (eg. if you define cross trims for throttle, select _Ele_ in _Throttle trim switch_)

mcddd:
THANK YOU dgatf for fixing this long standing issue!
I'm looking forward to testing it out.

dgatf:
Added two non regression tests, _throttleTrimWithCrossTrims_ and _invertedThrottlePlusThrottleTrimWithCrossTrims_ to check trims applied to ele and thr

dgatf:
Added two non regression tests, _throttleTrimWithCrossTrims_ and _invertedThrottlePlusThrottleTrimWithCrossTrims_ to check trims applied to ele and thr

radio/src/mixer.cpp
```diff
@@ -527,13 +527,19 @@ int getStickTrimValue(int stick, int stickValue)
...
-  if (stick == THR_STICK) {
-    if (g_model.thrTrim) {
-      int trimMin = g_model.extendedTrims ? 2*TRIM_EXTENDED_MIN : 2*TRIM_MIN;
-      trim = ((g_model.throttleReversed ? (trim+trimMin) : (trim-trimMin)) * (RESX-stickValue)) >> (RESX_SHIFT+1);
-    }
-    if (g_model.throttleReversed) {
+  uint8_t thrTrimSw;
+  if (g_model.thrTrimSw == 0) //this is needed for compatibility with existing models
+      thrTrimSw = MIXSRC_TrimThr - MIXSRC_FIRST_TRIM;
+    else if (g_model.thrTrimSw == MIXSRC_TrimThr - MIXSRC_FIRST_TRIM)
```
bsongis: I don't get why this else if is needed

dgatf: I've made the assumption that non used eeprom values are 0 (this assumption has been checked with several models on a X7 radio). So  _g_model.thrTrimSw_  will be 0 in existing models when updating opentx from previous releases. And the default switches order is RETA, so T has to go in first place (0) (this is safe as on previous releases throttle trim could only be used correctly on the default throttle switch). This is achieved by interchanging R (2) and T (0). And this is reason of the _else if_

This _else if_ is clearer as:

_else if (g_model.thrTrimSw == 2)_

bsongis: wow, it was impossible to understand!

Eldenroot:
Good nob, thank you for yoir contribution, I hope that we will see more valuable PRs from you :)

companion/src/modeledit/setup.cpp
```diff
@@ -1348,11 +1356,29 @@ void SetupPanel::populateThrottleSourceCB()
...
 
+void SetupPanel::populateThrottleTrimSwitchCB()
+{
+  Board::Type board = firmware->getBoard();
+  lock = true;
+  ui->throttleTrimSwitch->clear();
+  int idx=0;
+  for (int i=0; i<getBoardCapability(board, Board::NumTrims); i++, idx++) {
+    QString trim = RawSource(SOURCE_TYPE_TRIM, i).toString(model, &generalSettings);
+    trim = (trim == "TrmR") ? "TrmT" : (trim == "TmrT") ? "TmrR" : trim;
```
kilrah, line 1367: @dgatf can you explain this? It doesn't work with translations, and doesn't for English either (TrmT is there twice, TrmR is nowhere). Also don't see why you'd fool the names?

dgatf: Yes, is easy to explain... it is a typo. This is the line 1367 fixed:

`trim = (trim == "TrmR") ? "TrmT" : (trim == "TrmT") ? "TrmR" : trim;`

Please read the entire post to understand why R and T are interchanged...

_I've made the assumption that non used eeprom values are 0 (this assumption has been checked with several models on a X7 radio). So g_model.thrTrimSw will be 0 in existing models when updating opentx from previous releases. And the default switches order is RETA, so T has to go in first place (0) (this is safe as on previous releases throttle trim could only be used correctly on the default throttle switch). This is achieved by interchanging R (2) and T (0). And this is reason of the else if_

Fixed with #8141

kilrah: But `trim == "TmrT"` still won't work when companion is set say to French and you've got "TrmG" instead?

dgatf: You' re right. Fixed in last commit 4be12eb

Issue #6163, opened 2018-09-09
POT for set HORUS(s) backlight, Please...


CcLaFleche:
With S2 Like as Volume with S1

Amenotepia:
This would be great for Nirvana as well. :)

Luznatural:
And also for the rest of the Horus community....
Is there any way to move this issue a little bit forward? This is the single issue for 2.4 already proposed.
It shouldn't be so difficult as already Frsky has shown us how to implement it in FrOs.
Thanks in advance

CcLaFleche:
It's going well ?

Luznatural:
Thanks for moving this issue up in priority.

markran:
I would also like this.

bsongis:
Still valid @bot

PR #7680, opened 2020-06-05
Brightness adjust from backlight function

### LINKED ISSUE #4071: 2.2N359 Assign Brightness to any available pot or switch
Not really an issue, but I'd like the ability to adjust screen color using any assignable knob or switch.  I guess this can be considered an enhancement.

Thank You

dgatf:
This PR closes #4071 and #6163

Adds a source to backlight function to control the brightness:

- Source: if assigned, controls the brightness. If not assigned brightness in radio configuration is used
- Switch: if assigned to ON, auto off as defined in radio config. Else light on when switch on (as current behaviour)

If assigned to switch ON the function is always active. Due to this it is better to have enable option but this would require an eeprom change. Enable option can be added in 2.4


3djc:
Colorlcd radios have varying acceptable brighness range per model, and limits that should not be crossed. Not sure this is taken into account

bsongis:
The HAL should define a MIN and MAX which are acceptable. And we should use limit(min, value, max) to be sure we stay inside the allowed brightness range

3djc:
or percentage of MIN/MAX so that assigning a pot gives a sentation of full range instead of having a dead area ?

bsongis:
Good idea

dgatf:
Yes, much better for color lcd. Updated

3djc:
I absolutely love it. One observation is that their is no enable checkbox like there is one for the very similar "volume" function. Not sure it should be changed, just adding this comment to discusss with @bsongis

3djc:
Just an update. We had a discussion. The functionnality is really extremely comparable to (audio) volume, so they really need to behave the same, yet they don't because of the enable tickbox. Current thinking is to remove "enable" from volume function (since it needs a switch anyway)

dgatf:
Good. I agree volume function doesn't need the checkbox. Just take into account that I've added the _ON switch_ behavior, which is slightly different by only overriding the default brightness but keeping the auto off option. In this case the checkbox would be good to deactivate the brightness function. Maybe this is too convoluted tough

dgatf:
Well, also volume function may have the ON switch assigned. Shouldn't have enable checkbox all functions that may be active all the time with the ON switch? (So they can be deactivated without modifying or deleting the function)

dgatf:
In order to behave the same as the volume function maybe it shouldn't get the auto off feature with the ON switch
What do you think?

3djc:
What do you call auto off ?

dgatf:
Backlight duration

3djc:
Hum i personally don’t believe so. It is ideal to be able to edit able to set brightness through a function, it allows quickly to adjust to indoor/outdoor situations (like volume really). But you might really want to raise brightness while looking at something at the field, yet benefit from battery saving by auto turning screen off while flying

dgatf:
Ok, for me is fine either way

3djc:
We had a look:
- we need the enable box for consistencies
- when set on a switch (not ON), the auto dim does not work, we think it should remain

PR #7683, opened 2020-06-06
Filter instant trim only default trims

### LINKED ISSUE #1619: Instant trims adjusts the trims it expects to be attached to sticks, instead of trims actually in use.
I always have my trims crossed, so I can fly with my right hand on the right hand stick, and trim with the left, on the left hand side of the radio.

Instant trim doesn't check for this and if I were to use it, applies the trim to the wrong controls.

dgatf:
This PR fixes #1619

Adds a filter to instant trim only the inputs with default trims


radio/src/opentx.cpp
```diff
@@ -1660,21 +1660,31 @@ void instantTrim()
...
-      int16_t delta = 0;
-      for (int e=0; e<MAX_EXPOS; e++) {
-        ExpoData * ed = expoAddress(e);
-        if (!EXPO_VALID(ed)) break; // end of list
-        if (ed->srcRaw-MIXSRC_Rud == stick) {
-          delta = anas[ed->chn] - anas_0[ed->chn];
+
+    if (stick!=THR_STICK) { // don't instant trim the throttle stick
+      bool addTrim = true;
+      for (uint8_t i=0; i<MAX_INPUTS; i++) {
```
bsongis: why MAX_INPUTS and not MAX_EXPOS (agree that the name is not very nice)?

bsongis:
I have pushed a different implementation, would you please double-check?

radio/src/opentx.cpp
dgatf: Yes, better MAX_EXPOS

dgatf (commented):
Yes, way better. More efficient code. Good for me

PR #7690, opened 2020-06-06
Add dinamic bit shift for flight modes fading - Fixes #7623

### LINKED ISSUE #7623: Flight mode transition bug
Another RCG user posted a problem he was having.
I looked at his OTX, and it seems like a bug to me in how flight mode transitions are done when a mixer channel is over +/-200.  The value changes sign and is offset by the difference from 200 during the transition.

Simple OTX attached, only one mixer line and one flight mode.  Channel value changes sign during flight mode transition, even though flight mode change should not affect its value at all.

dgatf:
Fixes #7623 

Add dinamic bit shift to flight modes fading

Advantages:
- Chans value never overflows the mixer when fading flight modes
- Increases accuracy for low chans values


bsongis:
Perfect, thanks!

PR #7706, opened 2020-06-08
Another attempt to fix #6063

### LINKED ISSUE #6063: Modelselection does not remember the actual model / modelcategory
Hi,
since 2.2.2:
while open modelselectionscreen, the Horus X10 marks always the first model in the first modelcategory.
This is not good.
In 2.2.1 the modelslectionscreen is coming up with the last selected model in the last modelcategory. This is much more better than the behavior in 2.2.2. 
Can you please fix it?

Kind regards
Ralf

bsongis:


bsongis:
@raphaelcoeffic you just forgot to call the `modelslist.setCurrentCategory(...)`

PR #7707, opened 2020-06-08
Add cursor to custom curves - Closes #6399

### LINKED ISSUE #6399: Display Stick and uSec position on Curve graph
Currently, there seems to be no way to overlay the stick and micro second position on the curve graph.

Even my old Spektrum and JR radio's can do this.  

DX6 display

Its very helpful when setting up a throttle curve, if you can see exactly were about's the stick is positioned with the Input and Output uSec values. So the curve can be plotted more precisely and quickly.

The Inputs page has a similar feature. but it doesn't relate to the Curves page.

It makes it very tedious when trying to fine tune any curve. If we have to keep jumping back and forth between the Curve and Channel monitor screens.

dgatf:
Closes #6399

Adds a cursor and in/out values to custom curves (as in edit input menu) when editing from edit input menu or edit mix menu


radio/src/gui/128x64/model_input_edit.cpp
```diff
@@ -114,6 +92,8 @@ void menuModelExpoOne(event_t event)
         lcdDrawTextAlignedLeft(y, STR_SOURCE);
         drawSource(EXPO_ONE_2ND_COLUMN, y, ed->srcRaw, RIGHT|STREXPANDED|attr);
         if (attr) ed->srcRaw = checkIncDec(event, ed->srcRaw, INPUTSRC_FIRST, INPUTSRC_LAST, EE_MODEL|INCDEC_SOURCE|NO_INCDEC_MARKS, isSourceAvailableInInputs);
+        s_currSrcRaw = ed->srcRaw;
```
bsongis: why is it done here?

radio/src/gui/212x64/model_mix_edit.cpp
```diff
@@ -112,6 +112,8 @@ void menuModelMixOne(event_t event)
         lcdDrawTextAlignedLeft(y, STR_SOURCE);
         drawSource(MIXES_2ND_COLUMN, y, md2->srcRaw, STREXPANDED|attr);
         if (attr) CHECK_INCDEC_MODELSOURCE(event, md2->srcRaw, 1, MIXSRC_LAST);
+        s_currSrcRaw = md2->srcRaw;
```
bsongis: same question here, why not outside the display loop?

bsongis:
@dgatf would you please have a look to my "cosmetics" commit, I think the variables were not set on the right place

bsongis:
I need it for the next PR merge, please don't forget to look at my changes! Thanks

kilrah:
Cursor no longer follows any other preset curve, only custom ones

screenshot_tx16s_20-10-15_15-12-27

kilrah:
Actually even if you change the selected custom curve it still follows the previous one as well

screenshot_tx16s_20-10-15_15-16-18

PD5DJ:
Possible to remove this function for the time being? until fixed in nightlies?
Better no Cursor then a Cursor that does not function properly.

PR #7713, opened 2020-06-10
Remove Avr boards

dgatf:
Removes remaining code for avr boards

Also the following files/folder from companion may be removed: ~~_burnconfigdialog.*_~~ _fusesdialog.*, er9x_

Should I delete them?

PR #7716, opened 2020-06-10
Companion enhanced telemetry sensors context menu and other refinements across context menus

### LINKED ISSUE #6365: add copy/paste functions to telemetry page under companion
Problem:
Due to hardware changes (new ESC, GPS of another manufacturer etc..) certain telemetry sensors are deleted and read in again via automatic recognition.
As a result, the sorting sequence has changed.
If logical switches have been created, they are now functionless or based on wrong telemetry sensors, since the assignment is made via the sorting number of the telemetry sensors.

current workaround
- if not known: read "new hardware sensors" in a dummy memory to recognize ID, instances etc.
- in the target memory, enter the ID's / instances  manually to the adequate sensors so that the sorting sequence is not lost.

Desired solution (companion):
Similar to logical switches and special functions, copy & paste functionality on the telemetry page would also be an advantage within Companion.
You could create reference memories that contains hardware-specific telemetry sensors and easily migrate them by drag & drop to a desired memory in a certain order.
Side effect would be an easy way to resort the sensors and group them in a more logical way than they are automatically recognized

thanks & regards
Udo

elecpower:
Continues implementation of enhanced context menus for telemetry sensors including updating references.
Common update source references needed to be modified to handle blocks of 3 sources per sensor index.
Note: I have only been able to perform limited testing as I only have a few sensors.

Also included standardisation of prompts across all context menus.

Enhanced flight mode edit of global variables to cut copy and paste all flight mode values for a global variable when on flight mode zero tab
Fixes #6365


elecpower:
Sorry I just realised this clashes with #7713. Please advise how you want to handle the merge of this PR as not sure of the timing of both.

elecpower:
Cascading refs need fixing

elecpower:
I'll manually merge the latest 2.3 with 7713 included as there are merge conflicts.

bsongis:
Seems all good, thanks!

PR #7719, opened 2020-06-11
Check model is off before changing model - Fixes #7715

### LINKED ISSUE #7715: Check Rx is off before selecting a new model
There is a check to see if the Rx is off before powering the Tx off.
If you select a different model while the current model is still powered on, and they share the same receiver number, then you would end with TWO live models being controlled at the same time.

I suggest checking the current model's receiver is off (no telemetry RSSI) before swapping to the new model (I've just done this in erskyTx, just needed to call the same routine as I call at power off).

This was prompted by a report that this actually happened, someone had a live model in the pits operating as well as the intended model.

Mike

dgatf:
Fixes #7715. Check there is no telemetry when changing/creating model

May be linked to the option _disableRssiPoweroffAlarm_ and update var name and translations accordingly?

### DIFF
--- a/radio/src/gui/480x272/model_select.cpp
+++ b/radio/src/gui/480x272/model_select.cpp
@@ -260,7 +260,7 @@ bool isModelOff() {
 void onModelSelectMenu(const char * result)
 {
   if (result == STR_SELECT_MODEL) {
-    if (isModelOff()) {
+    if (isModelOff() && !g_eeGeneral.disableRssiPoweroffAlarm) {
       // we store the latest changes if any
       storageFlushCurrentModel();
       storageCheck(true);
@@ -279,7 +279,7 @@ void onModelSelectMenu(const char * result)
     deleteMode = MODE_DELETE_MODEL;
   }
   else if (result == STR_CREATE_MODEL) {
-    if (isModelOff()) {
+    if (isModelOff() && !g_eeGeneral.disableRssiPoweroffAlarm) {
       storageCheck(true);
       modelslist.addModel(currentCategory, createModel());
       selectMode = MODE_SELECT_MODEL;
--- a/radio/src/gui/480x272/model_select.cpp
+++ b/radio/src/gui/480x272/model_select.cpp
@@ -239,24 +239,6 @@ void onDeleteModelConfirm(const char * result)
   }
 }
 
-bool confirmModelChange() {
-  if (TELEMETRY_STREAMING()) {
-    RAISE_ALERT(STR_MODEL, STR_MODEL_STILL_POWERED, STR_PRESS_ENTER_TO_CONFIRM, AU_MODEL_STILL_POWERED);
-    while (TELEMETRY_STREAMING()) {
-      RTOS_WAIT_MS(20);
-      if (readKeys() == (1 << KEY_ENTER)) {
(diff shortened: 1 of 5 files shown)


3djc:
Shouldn't g_eeGeneral.disableRssiPoweroffAlarm be taken into consideration ?

3djc:
Sorry, didn't see your initial text. Yes, I think both should probably obey to disableRssiPoweroffAlarm

dgatf:
Ok. Will do

dgatf:
Linked to the rssi alarm. I did't find a good new string to fit in the length, but I believe current one still cover the new functionality

radio/src/gui/480x272/model_select.cpp
```diff
@@ -239,36 +239,58 @@ void onDeleteModelConfirm(const char * result)
...
-    storageFlushCurrentModel();
-    storageCheck(true);
-    memcpy(g_eeGeneral.currModelFilename, currentModel->modelFilename, LEN_MODEL_FILENAME);
-    modelslist.setCurrentModel(currentModel);
-    modelslist.setCurrentCategory(currentCategory);
-    loadModel(g_eeGeneral.currModelFilename, true);
-    storageDirty(EE_GENERAL);
-    storageCheck(true);
-    chainMenu(menuMainView);
+    if (isModelOff() && !g_eeGeneral.disableRssiPoweroffAlarm) {
```
3djc: Shouln't those checks be swapped so that alert is not raised if not desired ?

radio/src/gui/480x272/model_select.cpp
```diff
@@ -239,36 +239,58 @@ void onDeleteModelConfirm(const char * result)
...
   }
   else if (result == STR_CREATE_MODEL) {
-    storageCheck(true);
-    modelslist.addModel(currentCategory, createModel());
-    selectMode = MODE_SELECT_MODEL;
-    setCurrentModel(currentCategory->size() - 1);
-    modelslist.setCurrentModel(currentModel);
-    modelslist.setCurrentCategory(currentCategory);
-    modelslist.onNewModelCreated(currentModel, &g_model);
+    if (isModelOff() && !g_eeGeneral.disableRssiPoweroffAlarm) {
```
3djc: Same as previous  (Shouln't those checks be swapped so that alert is not raised if not desired ?)

radio/src/gui/480x272/model_select.cpp
dgatf: Indeed the this wasn't working when rssi check was disabled...Thanks. Fixed

radio/src/gui/480x272/model_select.cpp
dgatf: Fixed

radio/src/gui/480x272/model_select.cpp
```diff
@@ -239,9 +239,30 @@ void onDeleteModelConfirm(const char * result)
   }
 }
 
+bool confirmModelChange() {
```
bsongis: Is it possible that this function is shared for all LCD dimensions?

dgatf: Sure. Updated

PR #7724, opened 2020-06-12
Add 6-position S3 to Taranis x9d+ 2019 SE

10man:
I am trying to add S3 (6 position) to the Taranis X9D+ 2019 SE transmitter. I know people say you have to replace one of the other pots but I think it can be done. I thought the easiest way to go about figuring out what to setup in the firmware, was to just add some code and put it up for PR. I have this marked as draft because it's not working properly, the TX is doing something strange...

When I go to the debug screen for analogs and turn pot S2 the field for S3 follows in the opposite direction... So if I turn the pot CW the value of S2 goes up and the value of S3 goes down. I verified that this voltage is not present on the input line. I verified that the proper voltage is present on pin 58 (PD11) and that it follows the switch. Turning S3 does nothing in the radio display.

Clearly I have missed something and I am at a loss as to how to go about debugging. I was hoping I could get some guidance, thanks. I'm new to this codebase.

### DIFF
--- a/radio/src/targets/taranis/hal.h
+++ b/radio/src/targets/taranis/hal.h
@@ -781,7 +781,7 @@
     #define ADC_GPIO_PIN_POT3           GPIO_Pin_11  // PD.11
     #define ADC_GPIOB_PINS              (ADC_GPIO_PIN_POT2)
     #define ADC_GPIOD_PINS              (ADC_GPIO_PIN_POT3)
-    #define ADC_CHANNEL_POT3            ADC_Channel_9
+    #define ADC_CHANNEL_POT3            ADC_Channel_4
     #define ADC_VREF_PREC2              300
   #endif
   #define ADC_GPIO_PIN_SLIDER1          GPIO_Pin_4  // PC.04
--- a/radio/src/targets/taranis/hal.h
+++ b/radio/src/targets/taranis/hal.h
@@ -759,7 +759,7 @@
   #define HARDWARE_POT1
   #define HARDWARE_POT2
   #define HARDWARE_POT3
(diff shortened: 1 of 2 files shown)


radio/src/targets/taranis/hal.h
```diff
@@ -778,8 +778,10 @@
     #define ADC_CHANNEL_POT3            ADC_Channel_9
     #define ADC_VREF_PREC2              330
   #else
+    #define ADC_GPIO_PIN_POT3           GPIO_Pin_11  // PD.11
     #define ADC_GPIOB_PINS              (ADC_GPIO_PIN_POT2)
-    #define ADC_CHANNEL_POT3            0
+    #define ADC_GPIOD_PINS              (ADC_GPIO_PIN_POT3)
+    #define ADC_CHANNEL_POT3            ADC_Channel_9
```
10man: If I reset the `ADC_CHANNEL_POT3` to `0` then the 7th analog (on the analog debug screen) follows the reverse of the elevator stick. I also tried setting the analog channel to 7 and S3 again followed !S2.

radio/src/targets/taranis/hal.h
```diff
@@ -781,7 +781,7 @@
     #define ADC_GPIO_PIN_POT3           GPIO_Pin_11  // PD.11
     #define ADC_GPIOB_PINS              (ADC_GPIO_PIN_POT2)
     #define ADC_GPIOD_PINS              (ADC_GPIO_PIN_POT3)
-    #define ADC_CHANNEL_POT3            ADC_Channel_9
+    #define ADC_CHANNEL_POT3            ADC_Channel_4
```
10man, line 784: This has decoupled the S3 and the 7th analog from the other sticks and pots but the analog value does not follow the switch...

10man:
Duhh... Cannot just use any pin for ADC input...

PR #7734, opened 2020-06-14
Fix negative bit shift

### LINKED ISSUE #7733: [Companion 2.3.8 Simu] Mixer values collapse to zero during FM fade in/out
This issue is new in Companion 2.3.8 (no problem in 2.3.7), unfortunately it completely breaks the sim for many of my templates, setting off wild oscillations in multiple channels. So far, I've not been able to reproduce this issue on the radio, but I wanted to report it before doing further testing.

**The problem**
When activating a flight mode with non-zero fade in/out, mixer values collapse to zero during the transition.

**To Reproduce**
1. Download [ this file (minimalist example)](https://www.dropbox.com/s/rlygv36hl5dbyxz/FMslow_bug_238_T16.otx?dl=0)
2. Start Companion 2.3.8 with Jumper T16 profile and open the file
3. Pull SH to switch flight modes while monitoring CH1

**Expected behavior**
CH1 mixer value should remain constant, instead it collapses to zero during the fade in/out (set to 0.3 secs in FM1).

**General information**
 - Companion 2.3.8 (release version)
 - Jumper T16 profile

dgatf:
Fixes #7733

### DIFF
--- a/radio/src/mixer.cpp
+++ b/radio/src/mixer.cpp
@@ -947,7 +947,7 @@ void evalMixes(uint8_t tick10ms)
         mixerCurrentFlightMode = p;
         evalFlightModeMixes(p==fm ? e_perout_mode_normal : e_perout_mode_inactive_flight_mode, p==fm ? tick10ms : 0);
         for (uint8_t i=0; i<MAX_OUTPUT_CHANNELS; i++)
-          sum_chans512[i] += (chans[i] >> (17 - __builtin_clz(abs(chans[i])))) * fp_act[p];
+          sum_chans512[i] += (chans[i] >> (__builtin_clz(abs(chans[i])) >= 17 ? 0 : 17 - __builtin_clz(abs(chans[i])))) * fp_act[p];
         weight += fp_act[p];
       }
     }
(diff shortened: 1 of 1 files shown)


radio/src/mixer.cpp
```diff
@@ -947,7 +947,7 @@ void evalMixes(uint8_t tick10ms)
         mixerCurrentFlightMode = p;
         evalFlightModeMixes(p==fm ? e_perout_mode_normal : e_perout_mode_inactive_flight_mode, p==fm ? tick10ms : 0);
         for (uint8_t i=0; i<MAX_OUTPUT_CHANNELS; i++)
-          sum_chans512[i] += (chans[i] >> (17 - __builtin_clz(abs(chans[i])))) * fp_act[p];
+          sum_chans512[i] += (chans[i] >> (__builtin_clz(abs(chans[i])) >= 17 ? 0 : 17 - __builtin_clz(abs(chans[i])))) * fp_act[p];
```
bsongis, line 950: is it possible you create a macro to hide this? it has become difficult to read

bsongis: may I ask you to add a new non regression test for this bug?

dgatf: Sure

kilrah:
Someone reports that it doesn't work... are you using the same compiler version as we do for release when you test?

dgatf:
I'm using arm-none-eabi-g++ v4.7.4

Also tested with arm-none-eabi-g++ v6.3.1

Using Ubuntu 18.04.4 LTS

All is working fine

3djc:
[slow_issue.zip](https://github.com/opentx/opentx/files/4776712/slow_issue.zip)

Run the only model, watch chan monitor, CH6, play with SD switch

bsongis:
OK I revert the commit which introduced this regression. It will be worked again in 2.3.10

TomasJot:
I can confirm - it doesn't work. (fix, i mean  - after reverting commit 991deec it's ok.)

PR #7747, opened 2020-06-17
Fix mixer overflow

### LINKED ISSUE #7623: Flight mode transition bug

dgatf:
Fixes mixer overflow #7623 by limiting the chans values when fading flightmodes

Added two non regresion tests for mixer modes transitions:

- _flightModeOverflow_ evaluates the initial overflow error (where current mixer fails) 
- _flightModeTransition_ evaluates the transition between two values (where dynamic bit shift fails)

### DIFF
--- a/radio/src/tests/gtests.h
+++ b/radio/src/tests/gtests.h
@@ -65,6 +65,7 @@ inline void MODEL_RESET()
   memset(&anaInValues, 0, sizeof(anaInValues));
   extern uint8_t s_mixer_first_run_done;
   s_mixer_first_run_done = false;
+  evalMixes(1);  // this is needed to reset fp_act
   lastFlightMode = 255;
 }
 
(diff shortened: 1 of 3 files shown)


radio/src/tests/gtests.h
```diff
@@ -65,6 +65,7 @@ inline void MODEL_RESET()
   memset(&anaInValues, 0, sizeof(anaInValues));
   extern uint8_t s_mixer_first_run_done;
   s_mixer_first_run_done = false;
+  evalMixes(1);  // this is needed to reset fp_act
```
dgatf, line 68: needed to reset static variable fp_act

bsongis (approved):

PR #7753, opened 2020-06-18
Provide a way to add pot S3 to Taranis X9D+ 2019

10man:
FrSky removed the pads for S3 and rearranged the GPIO usage to make it a bit difficult to add the extra pot (or 6 position switch) to the radio. This change set provides a `cmake` switch to swap a couple GPIO's around to make the required analog input available. I switched the right horizontal trim right from `PC.01` to `PD.11`. Then the newly available analog input `ADC_Channel_11` which is connected to `PC.01` gets used for the S3 input.

__NOTE: This requires a modification to the hardware to work !__

I have tested this on target and verified that both the trim and analog input function properly.

### DIFF
--- a/radio/src/targets/common/arm/CMakeLists.txt
+++ b/radio/src/targets/common/arm/CMakeLists.txt
@@ -2,6 +2,7 @@ set(TIMERS 3 CACHE STRING "Timers count (2 or 3)")
 set_property(CACHE TIMERS PROPERTY STRINGS 2 3)
 option(CLI "Command Line Interface" OFF)
 option(DEBUG "Debug mode" OFF)
+option(S3_HACK "Setup S3 Hack" OFF)
 option(LOG_TELEMETRY "Telemetry Logs on SD card" OFF)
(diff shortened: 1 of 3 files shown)


HippoDan:
What is the required hardware modification?  Where are the new locations to solder S3 on the 2019 SE?  Thanks.

10man:
Looking at the main board from inside the radio, the connector on the upper right holds the signal from the trim. The trim is the 4th pin from the right. The signal from the trim is moved to the 7th pin from the right on the next connector which is also the top side of R15. That slot is open so one could get the right pin an plug into that connector real clean, I just soldered onto the top of the resistor. The you put the signal from the pot where the trim was. To get power for the pot I used C90 (top side is +), this provides a nice return path with twisted wire.

In summary:
- trim signal to topside of R15
- pot signal to where the trim was

bsongis:
The problem is we don't want to take care of support of all possible hacks, it already takes so much time not to break anything on the currently supported radios!

PR #7787, opened 2020-07-03
Tx16s sbus in

3djc:
Side note, it requires NON inverted TTL input. Works fine with a TBS receiver in non inverted sbus output

image

### DIFF
--- a/radio/src/gui/480x272/radio_hardware.cpp
+++ b/radio/src/gui/480x272/radio_hardware.cpp
@@ -372,7 +372,7 @@ bool menuRadioHardware(event_t event)
 #if defined(RADIO_TX16S)
         lcdDrawText(lcdNextPos, y, " (TTL)");
 #endif
-        g_eeGeneral.auxSerialMode = editChoice(HW_SETTINGS_COLUMN2, y, STR_AUX_SERIAL_MODES, g_eeGeneral.auxSerialMode, 0, UART_MODE_MAX, attr, event);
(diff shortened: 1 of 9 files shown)


wimalopaan:
Is it possible to use a lower baudrate for the sbus input? Does this break something in the internal queing? 

This would break sbus compatibility but since the physical signal is TTL, full compatibility is already not there.

kilrah:
No, makes no sense...

wimalopaan:
What do you mean?
Would it technically be not possible, to simply change the baudrate of the uart? 
Or would this make no sense because it breaks sbus compatibility (I readily would accept that)

wimalopaan:
Well, I tries to test that. I used a µC to generate a sbus signal (ttl, 100kB, 8e2) sending all 14ms a sbus packet (25 Byte, startbyte = 0x0F, endbyte = 0x00). 
I set the model to trainer/Serial, the aux2 to Sbus-input. Looks like the signal is recognized, because if I remove the cable, I get a lost / recovered pupil signal  message.
But if I setup a mixer with input e.g. TR1, I can't see the values, always 0.
Something I missed? Any hints to debug this?

3djc:
Because while it needs non inverted SBUS, it works out of the box with device providing that, like CRSF receivers, or buy manually getting non inverted from existing receivers (https://oscarliang.com/uninverted-sbus-smart-port-frsky-receivers/) and likely others we have not tested

wimalopaan:
Well, it works for me. I generate the sbus with an external µC.

My question is: can I reduce the baudrate (changing the source code) to say 9600 Bd? Or does this break things?

kilrah:
You'll have to try and see but in theory it should work, just with reduced framerate/increased latency, an sbus frame at 9600 is about 30ms...

PR #7808, opened 2020-07-09
Companion do not prompt for radio profile when file name supplied at start up

### LINKED ISSUE #7806: Companion opens OTX file before radio profile is selected, causing conversion prompt
With radio profile section enabled, when launching Companion by opening an OTX file, the file is loaded before the 'Select Profile' prompt is shown.  This means that if the file does not match the radio from the last-used profile there is a prompt to convert it (which is most likely unnecessary).

If the conversion prompt is ignored for a few seconds, both dialogs appear at the same time, with the select profile prompt on top of the conversion prompt.

dialogs

The prompt to select the radio profile should come before the file is opened so that the correct radio profile could be loaded before the file is parsed.  The conversion prompt would only appear if a radio profile is selected which does not match the file.

An advanced maneuver might be automatically selecting the radio profile based on the OTX file, but I don't know if that's feasible.  E.g. if the OTX file is for a TX16S, and there's only one TX16S profile, select it.

I'm happy to have a swing at this myself, but am guessing that someone who knows the code would resolve it much faster :-)

elecpower:
Fixes #7806

### DIFF
--- a/companion/src/mainwindow.cpp
+++ b/companion/src/mainwindow.cpp
@@ -116,33 +116,6 @@ MainWindow::MainWindow():
   connect(windowsListActions, &QActionGroup::triggered, this, &MainWindow::onChangeWindowAction);
   connect(&g, &AppData::currentProfileChanged, this, &MainWindow::onCurrentProfileChanged);
 
-  // give time to the splash to disappear and main window to open before starting updates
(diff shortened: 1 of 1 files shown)


elecpower:
Closed in favour of 7822

PR #7813, opened 2020-07-12
Update companion_de.ts -Some words incomprehensible

Xillius178:
The reason for the change:
Incomprehensible!

In my opinion the wrong German words were chosen.

Timer = "Takt" - english - "clock"
Edge = "Puls"  - english  - "pulse"
Sticky = "SRFF" (absolutely incomprehensible)

The correct translation would normally be:

Timer = Zeitgeber, Zeitmesser
Edge = Kante, Schwellwert
Sticky = Kurzzeitig Ein

Background, I wanted to select a momentary switch for arming, and I just couldn't find the parameter "Sticky" in the logic switch ...
I found out by switching from German to English, I found the words.

Very cumbersome and not understandable what the translator was thinking ...

It is also easier to understand in German if you leave these words as they are.

Greetings

### DIFF
--- a/companion/src/translations/companion_de.ts
+++ b/companion/src/translations/companion_de.ts
@@ -5660,17 +5660,17 @@ Sind Sie sicher?</translation>
     <message>
         <location filename="../firmwares/logicalswitchdata.cpp" line="102"/>
         <source>Timer</source>
-        <translation>Takt</translation>
+        <translation>Timer</translation>
     </message>
     <message>
         <location filename="../firmwares/logicalswitchdata.cpp" line="104"/>
         <source>Sticky</source>
-        <translation>SRFF</translation>
+        <translation>Sticky</translation>
     </message>
     <message>
         <location filename="../firmwares/logicalswitchdata.cpp" line="106"/>
         <source>Edge</source>
-        <translation>Puls</translation>
+        <translation>Edge</translation>
     </message>
     <message>
         <location filename="../firmwares/logicalswitchdata.cpp" line="108"/>
@@ -7997,17 +7997,17 @@ If blank then the mix is considered to be &quot;ON&quot; all the time.</source>
     <message>
         <location filename="../modelprinter.cpp" line="548"/>
         <source>Edge</source>
-        <translation>Puls</translation>
+        <translation>Edge</translation>
     </message>
     <message>
         <location filename="../modelprinter.cpp" line="551"/>
         <source>Sticky</source>
-        <translation>SRFF</translation>
+        <translation>Sticky</translation>
     </message>
     <message>
         <location filename="../modelprinter.cpp" line="554"/>
         <source>Timer</source>
-        <translation>Takt</translation>
+        <translation>Timer</translation>
     </message>
     <message>
         <location filename="../modelprinter.cpp" line="581"/>


kilrah:
This would no longer match what's on the radio, nor what's in the German manual the same person wrote. Makes no sense to change it.

Maybe use OpenTx in English instead?

Xillius178:
Where is it in the German manual? Can't Find it...
There is no official German manual at all...

Why are not expected in the remote control?
Then there are two translations, one for the Campanion and one for the radio firmware, right?
If so, where is this translation file?

This makes perfect sense.
Since I'm not the only one who didn't find the "Sticky" Word command.

To offer an abbreviation like "SRFF" as a translation. I beg you
And what sense does a translation make if you can handle it even worse than the original English?

Everyone should understand the word timer, but the word "clock" in English is out of place in this context.

Why should I switch to English? Just because a few words were translated poorly, so honestly!

Please think about it first ...

kilrah:
See this thread: https://openrcforums.com/forum/viewtopic.php?f=92&t=3563&start=30

Xillius178:
@ kilrah
Great help!
Brings nothing.
And doesn't solve the problem.
Not answered any of my questions.
That is not a help, and I can do without that.
And this are the unofficial instructions, right?!

And if I am on an airfield where there is no WiFi, no cell phone network and what I want to change, what do I do?

To summarize it:
I have to download the instructions now, before I have to register in this forum. 
Then to look for the word somewhere. Just so that I can find out the Word Sticky as "SRFF"?

It can't be, absolutely not.
A translation should be clear and understandable, right ?!
Otherwise you don't need it, my opinion.

kilrah:
These translations have been in place since March of 2014. People have been able to do with them and we didn't get complaints about them not understanding. If you change it you'll invalidate the documentation and confuse people who have been using this for 6 years and know what is what.

Is that really worth it? In my opinion it isn't.

hrenz:
For Xillius178
Man muss das vom einfachen Englisch unterscheiden
und muss es in ein passendes technisches Deutsch übersetzen.
Das ist oft nicht leicht und man kann das oft nicht 1:1 übersetzen 
Die techn. Funktion muss erfasst werden und dann möglichst genau den passenden dt. techn. Fachbegriffen treffen.

In open9x , openTx, hat man nur 4-6 Zeichen zur Verfügung

Beispiel: Timer
Timer, als echte Uhr, wie bei Timer 1, Timer2, Timer3 passt für diese Funktion gar nicht,
denn es ist keine Uhr sondern ein Clockgenerator also ein Taktgenerator
deshalb
Timer = Takt = Taktgenerator mit einstellbaren EIN/AUS-Zeiten

Edge = Puls = Ein Impuls, einstellbar, flankengetriggert, deshalb das edge für Flanke

Sticky = SRFF von der Funktion ein Set/Reset-Flipflop mit Reset dominant, daher ein SRFF und kein RS-FF

Das aktuelle dt. Handbuch, Stand Mai 2020 gibt es bei Engel Modellbau im FrSky Forum
und wurde über die Jahre in div Foren zu zigtausenden geladen.
Dort sind auch über 200 Beispiel enthalten, ca 990Seiten als PDF
Auch das was du suchst in zig Varianten

-----------------------------
Diese dt. Bezeichnungen gibt es schon sehr lange
schon zu Th9x und Companion9x Zeiten, denke so ca 8-10 Jahre

Jeder Elektrotechniker weiß mit diesen deutschen Bezeichnungen sofort was gemeint ist. 
Mit Sticky kann kein Elektrotechniker was anfangen, aber mit Flipflop bzw RSFF, SRFF sofort

Vielleicht hilft es zum Verständnis
Du wärest der erste der damit nicht klar kommt

HR

PR #7840, opened 2020-07-24
Change volts by Volts

activa73:


kilrah:
Units should be lowecase when not abbreviated, so the current text is correct.
https://english.stackexchange.com/questions/173906/should-units-of-measure-be-capitalized

activa73:
Yes you are right. Sorry for the mistake.
BR
Cyprien

Issue #7549, opened 2020-04-24
macOS 10.15.4  Compiling OpenTx Error info = fatal error: 'Availability.h' file not found


1371030:
macOS:10.15.4
Xcode:11.4.1

gcc -v
Configured with: --prefix=/Library/Developer/CommandLineTools/usr --with-gxx-include-dir=/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/c++/4.2.1
Apple clang version 11.0.3 (clang-1103.0.32.59)
Target: x86_64-apple-darwin19.4.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin

cd build
cmake -DPCB=X10 -DPCBREV=T16 -DHELI=YES -DDEBUG=NO -DPYTHON_EXECUTABLE=/usr/local/cellar/python/3.7.7/bin/python3 -DCMAKE_BUILD_TYPE=Release ..
make firmware

-- OpenTX 2.3.7 @8b080683
-- Python found, version: 3.7.7
-- Qt Version: 5.14.1
-- Foxlib found at /usr/local/lib/libFOX-1.6.dylib
-- Found SDL: /usr/local/lib/libSDL.dylib;-framework Cocoa (found version "1.2.15") 
-- SDL Lib: /usr/local/lib/libSDL.dylib;-framework Cocoa; Headers: /usr/local/include/SDL; Version: 1.2.15
TARGET simu/libsimulator: cpp compiler /Library/Developer/CommandLineTools/usr/bin/c++ v11.0.3
-- Added optional gtests target
-- The ASM compiler identification is Clang
-- Found assembler: /Library/Developer/CommandLineTools/usr/bin/cc
TARGET bootloader: cpp compiler arm-none-eabi-g++ v4.9.3
TARGET firmware: cpp compiler arm-none-eabi-g++ v4.9.3
TARGET companion: cpp compiler /Library/Developer/CommandLineTools/usr/bin/c++ v11.0.3
-- Added optional gtests-companion target

Compiling Error info:
  /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/cstddef:44:15: note: in file included from /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/cstddef:44:
[...]

PR #7861, opened 2020-08-05
Hotfix build mac os

cvetaevvitaliy:
fix build OpenTX for MacOS
I got the same error, the problem is in the python script
https://github.com/opentx/opentx/issues/7549


raphaelcoeffic:
@cvetaevvitaliy your "fix" is not really a hotfix, rather a rewrite of functions into a worse way... You need to identify the exact thing that makes the change work, then we can talk about it.

In my case, it compiles when I remove the GCC includes (`${arm_gcc_include_path}`):
```diff
diff --git a/radio/src/targets/horus/CMakeLists.txt b/radio/src/targets/horus/CMakeLists.txt
index 525c8a4ad..1e036e929 100644
--- a/radio/src/targets/horus/CMakeLists.txt
+++ b/radio/src/targets/horus/CMakeLists.txt
@@ -329,7 +329,7 @@ if(PYTHONINTERP_FOUND)
   add_custom_command(
     OUTPUT ${PROJECT_BINARY_DIR}/radio/src/datacopy.cpp
     WORKING_DIRECTORY ${RADIO_DIRECTORY}/src
-    COMMAND ${PYTHON_EXECUTABLE} ${RADIO_DIRECTORY}/util/generate_datacopy.py datastructs.h -DPCBHORUS -DPCBX10 -DCPUARM -DSTM32F4 -DSTM32F429_439xx -DCOLORLCD -DBACKUP -DSIMU -I. -Itargets/horus -Itargets/common/arm/stm32 -I${THIRDPARTY_DIR} -I${THIRDPARTY_DIR}/STM32F4xx_DSP_StdPeriph_Lib_V1.4.0/Libraries/CMSIS/Device/ST/STM32F4xx/Include ${arm_gcc_include_path} > ${PROJECT_BINARY_DIR}/radio/src/datacopy.cpp
+    COMMAND ${PYTHON_EXECUTABLE} ${RADIO_DIRECTORY}/util/generate_datacopy.py datastructs.h -DPCBHORUS -DPCBX10 -DCPUARM -DSTM32F4 -DSTM32F429_439xx -DCOLORLCD -DBACKUP -DSIMU -I. -Itargets/horus -Itargets/common/arm/stm32 -I${THIRDPARTY_DIR} -I${THIRDPARTY_DIR}/STM32F4xx_DSP_StdPeriph_Lib_V1.4.0/Libraries/CMSIS/Device/ST/STM32F4xx/Include > ${PROJECT_BINARY_DIR}/radio/src/datacopy.cpp
     DEPENDS ${RADIO_DIRECTORY}/src/datastructs.h ${RADIO_DIRECTORY}/util/generate_datacopy.py
     )
   add_custom_target(datacopy
```

cvetaevvitaliy:
Yes that's right, I rolled back to the previous version generate_datacopy.py

I find problem in CMake 
execute_process(COMMAND "${THE_PROGRAM}" argument OUTPUT_QUIET)

radio/src/targets/horus/CMakeLists.txt
```diff
@@ -294,7 +294,7 @@ MACRO(GET_GCC_INCLUDE_PATH gcc_include_path)
   if (WIN32)
     execute_process(COMMAND arm-none-eabi-gcc -v -x c -E NUL ERROR_VARIABLE _gcc_output OUTPUT_QUIET)
   else()
-    execute_process(COMMAND arm-none-eabi-gcc -v -x c -E - INPUT_FILE /dev/null ERROR_VARIABLE _gcc_output OUTPUT_QUIET)
+    execute_process(COMMAND arm-none-eabi-gcc argument -v -x c -E - INPUT_FILE /dev/null ERROR_VARIABLE _gcc_output OUTPUT_QUIET)
```
bsongis, line 297: I don't see what is this "argument"?

bsongis: if it was something for cmake, it should be ARGUMENT, but I don't see this in the documentation of execute_process:
https://cmake.org/cmake/help/v3.18/command/execute_process.html

PR #7881, opened 2020-08-15
Add rx num handling to CRSF

3djc:


elecpower:
@3djc should Companion ModuleData::supportRxNum() also be updated?

3djc:
> @3djc should Companion ModuleData::supportRxNum() also be updated?

Good catch !

PR #7886, opened 2020-08-18
Companion fix radio setup console runtime qt warnings

elecpower:


companion/src/generaledit/generalsetup.cpp
```diff
@@ -505,13 +505,13 @@ void GeneralSetupPanel::on_splashScreenDuration_currentIndexChanged(int index)
   emit modified();
 }
 
-void GeneralSetupPanel::on_pwrOnDelay_valueChanged()
+void GeneralSetupPanel::on_pwrOnDelay_valueChanged(int i)
```
bsongis, line 508: would you change with (int) ?

bsongis: ok I have done it, no need to change :)

PR #7887, opened 2020-08-18
CZ Month name correction based on rules

valfa14:
http://steam.g6.cz/index.php?title=Pravopis_a_nejasnosti


valfa14:
Simple months shortcut correction based on grammar rules http://steam.g6.cz/index.php?title=Pravopis_a_nejasnosti

PR #7895, opened 2020-08-23
Companion improve switch conversions

elecpower:
Move from more and more complex index shifts due to the number of supported radios to switch name mapping


bsongis:
Thanks!

PR #7904, opened 2020-08-27
[sd-card] X7 delta setup wizard fix scroll changing page rather than field

tgreer:
The X7 script (and thus what's used on the Tango2) had a bug where when you scrolled down it changed page rather than changing between fields.
I compared with the multi script (most recently updated) and have tested that the changes work in the delta setup wizard.

### DIFF
--- a/radio/sdcard/taranis-x7/SCRIPTS/WIZARD/delta.lua
+++ b/radio/sdcard/taranis-x7/SCRIPTS/WIZARD/delta.lua
@@ -93,11 +93,11 @@ local function navigate(event, fieldMax, prevPage, nextPage)
       dirty = blinkChanged()
     end
   else
-    if event == EVT_PAGE_BREAK or event == EVT_RIGHT_FIRST then
+    if event == EVT_VIRTUAL_NEXT_PAGE then
       page = nextPage
       field = 0
       dirty = true
-    elseif event == EVT_PAGE_LONG or event == EVT_LEFT_FIRST then
+    elseif event == EVT_VIRTUAL_PREV_PAGE then
       page = prevPage
       field = 0
       killEvents(event);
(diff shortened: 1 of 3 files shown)


radio/sdcard/taranis-x7/SCRIPTS/WIZARD/delta.lua
```diff
@@ -93,11 +93,11 @@ local function navigate(event, fieldMax, prevPage, nextPage)
       dirty = blinkChanged()
     end
   else
-    if event == EVT_VIRTUAL_NEXT then
+    if event == EVT_PAGE_BREAK or event == EVT_RIGHT_FIRST then
```
3djc: EVT_VIRTUAL_NEXT_PAGE ?

radio/sdcard/taranis-x7/SCRIPTS/WIZARD/delta.lua
```diff
@@ -93,11 +93,11 @@ local function navigate(event, fieldMax, prevPage, nextPage)
...
       page = nextPage
       field = 0
       dirty = true
-    elseif event == EVT_VIRTUAL_PREV then
+    elseif event == EVT_PAGE_LONG or event == EVT_LEFT_FIRST then
```
3djc: EVT_VIRTUAL_PREV_PAGE ?

3djc:
`lcd.drawText(48, LCD_H-8, "[Enter Long] to confirm", 0);` needs to be changed to` lcd.drawText(0, LCD_H-8, "[Enter Long] to confirm", 0);` too

tgreer:
Updated to use the correct calls, and tested on Tango2. Also fixed the alignment of the text as per the comments above.

3djc:
Thanks a lot ! 👍

PR #7905, opened 2020-08-28
Horus - Trainer driver uses the same interrupt / flags for trainer / trainee. It seems wrong to me. Untested.

bsongis:


3djc:
Working fine :)

PR #7907, opened 2020-08-30
Companion add external R9M 2019 to X7 ACCESS

### LINKED ISSUE #7790: 2.3.9 - Companion incorrectly shows R9M 2019 module set to "OFF"
**Conditions**
OpenTx 2.3.9
QX7 ACCESS FCC radio running 2.1.0 firmware
R9M 2019 module running ACCESS 1.3.0 firmware
R9MX OTA receiver
Companion running on Mac OS Catalina 10.15.5

**To reproduce**
1. Install R9 module in radio.
2. Set internal module to OFF.
3. Set external module protocol to R9M ACCESS.
4. Register and bind an R9 receiver. I used an R9MX OTA.
5. In module options, set Power to 20dBm. I don't think this affects the bug outcome.
6. On a Mac computer, read models and settings from the radio.

**Outcome**
In Companion, in the model settings tab, the external module is shown as set to "OFF".

**Expected outcome**
In Companion, the external module would be shown correctly i.e. as set to "R9 ACCESS" for the conditions above.

**Related issues**
1. In Companion, in the model settings tab, the drop menu for the external module does not include all valid options e.g. for the conditions above, "R9 ACCESS" does not appear.

Screen Shot 2020-07-03 at 10 30 55 AM
Screen Shot 2020-07-03 at 10 47 48 AM

elecpower:
Fixes #7790
Tested using Companion profile only as do not have a radio to test reading and writing settings.

### DIFF
--- a/companion/src/firmwares/opentx/opentxinterface.cpp
+++ b/companion/src/firmwares/opentx/opentxinterface.cpp
@@ -763,7 +763,7 @@ bool OpenTxFirmware::isAvailable(PulsesProtocol proto, int port)
           case PULSES_GHOST:
             return true;
           case PULSES_ACCESS_R9M:
-            return IS_TARANIS_XLITE(board) || IS_TARANIS_X7_ACCESS(board) || IS_TARANIS_X9LITE(board) || board == BOARD_TARANIS_X9DP_2019 || board == BOARD_X10_EXPRESS || (IS_FAMILY_HORUS_OR_T16(board) && id.contains("internalaccess"));
+            return IS_ACCESS_RADIO(board, id);
           case PULSES_PXX_R9M_LITE:
           case PULSES_ACCESS_R9M_LITE:
           case PULSES_ACCESS_R9M_LITE_PRO:
--- a/companion/src/firmwares/opentx/opentxinterface.cpp
+++ b/companion/src/firmwares/opentx/opentxinterface.cpp
@@ -763,14 +763,14 @@ bool OpenTxFirmware::isAvailable(PulsesProtocol proto, int port)
           case PULSES_GHOST:
             return true;
           case PULSES_ACCESS_R9M:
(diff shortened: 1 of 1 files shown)


companion/src/firmwares/opentx/opentxinterface.cpp
```diff
@@ -763,14 +763,14 @@ bool OpenTxFirmware::isAvailable(PulsesProtocol proto, int port)
           case PULSES_GHOST:
             return true;
           case PULSES_ACCESS_R9M:
-            return IS_TARANIS_XLITE(board) || IS_TARANIS_X9LITE(board) || board == BOARD_TARANIS_X9DP_2019 || board == BOARD_X10_EXPRESS || (IS_FAMILY_HORUS_OR_T16(board) && id.contains("internalaccess"));
+            return IS_TARANIS_XLITE(board) || IS_TARANIS_X7_ACCESS(board) || IS_TARANIS_X9LITE(board) || board == BOARD_TARANIS_X9DP_2019 || board == BOARD_X10_EXPRESS || (IS_FAMILY_HORUS_OR_T16(board) && id.contains("internalaccess"));
```
3djc: maybe return` IS_TARANIS_XLITE(board) || IS_ACCESS_RADIO(board, id) `?

3djc (changes requested):

companion/src/firmwares/opentx/opentxinterface.cpp
elecpower: I didn't think all ACCESS radios have an external module and thus why I chose the specific radio here

elecpower: ...or maybe if there are a few exceptions return (IS_ACCESS_RADIO(board, id) && !IS_radiox && !IS_radioy)?

3djc: I checked and all seem to accept it on the radio, even things like xlites, adapters do exist and soem prefer the larger modules

elecpower: Thanks for the info so we'll go with your recommendation

3djc:
sweet, thanks

PR #7911, opened 2020-08-31
Ensure mixes match channel order after edits in Companion

### LINKED ISSUE #7908: Mixers getting invisible in HorusX12S after moving channels in companion
It is possible to move channels via "Move Up" and Move Down" 
This is great because all corresponding mixer lines are automatically moved with the channel.

grafik

Before moving Channels everything works fine...
grafik

When I move channel 2 with "Move up" to Channel 1 nearly all other Mixers get invisible.
grafik

The mixers work as they should, but they are invisible an can not be edited.
grafik

You can reproduce it in a real radio and in the simulator.

When you download the Model from the radio, the mixers are visible in companion.

Here my test template
[ChannelFinder.zip](https://github.com/opentx/opentx/files/5146447/ChannelFinder.zip)

Regards Dirk

elecpower:
Fixes #7908
Tested using simulators

### DIFF
--- a/companion/src/firmwares/modeldata.cpp
+++ b/companion/src/firmwares/modeldata.cpp
@@ -656,6 +656,8 @@ int ModelData::updateReference()
       }
     }
   }
+  if (updRefInfo.type == REF_UPD_TYPE_CHANNEL)
+    sortMixes();
   //s1.report("Mixes");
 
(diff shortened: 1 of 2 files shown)


3djc:
👍

PR #7918, opened 2020-09-04
LUA: add ability to read rotenc speed

3djc:
tested by @pascallanger (thx to him)


pascallanger:
Works really well. I'm now using it in the new upcoming DSM forward programming LUA script.
It gives the same feeling when changing over a large value (-150%..+150%) compared to the normal GUI.

PR #7930, opened 2020-09-09
Update indentation in radio/src/targets

anton:
By using the provided script clang-format.sh, indentation for all the .c and .h-files under radio/src/targets are updated.


anton:
Oh. Interesting. Six out of seven builds are failing with travis-ci.

jtaylor2:
Why do you want to change the coding format from what has been used for years, and is documented in the OpenTX Code Style Guide https://github.com/opentx/opentx/wiki/OpenTX-Code-Style-Guide ?

anton:
Thanks for the comment. The specification in clang-format.sh is now updated to match the OpenTX-Code-Style-Guide.

bsongis:
We cannot do this now, until we will cry for 2.3 => 2.4 merge!
Is it possible you open another PR with only the script?
Also it seems Travis is not happy ;)

anton:
Thank you for the comment. It seems reasonable to not update the indentation of all of the files if there is a pending merge between 2.3 and 2.4.

One suggestion might be to run the automatic indentation script on both 2.3 and 2.4 before the merge. One of the reasons to unify and standardize the indentation with a tool is to simplify e.g. merging.

A new pull request with only the script is created: https://github.com/opentx/opentx/pull/7940

PR #7651, opened 2020-05-30
Rotary Encoder Switches / Sources removed

bsongis:

PR #7939, opened 2020-09-17
Companion refactor of ui datamodels

elecpower:
Moves from per tab to per model ui datamodels to improve performance, reduce memory and simplify refreshing across tabs for sources, switches and channel curves.
There a many other places where ui datamodels can be implemented for performance and memory reduction. They will be incorporated as part of other housekeeping.

### DIFF
--- a/companion/src/datamodels/rawitemdatamodels.cpp
+++ b/companion/src/datamodels/rawitemdatamodels.cpp
@@ -30,7 +30,7 @@ RawSourceItemModel::RawSourceItemModel(const GeneralSettings * const generalSett
   Firmware * fw = getCurrentFirmware();
 
   addItems(SOURCE_TYPE_NONE,           RawSource::NoneGroup,     1);
-  for (int i=0; i < fw->getCapability(LuaScripts); i++)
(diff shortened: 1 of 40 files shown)


elecpower:
After refactoring the app it continues to crash under certain circumstances. Sometimes you get a crash report which indicates a memory corruption from malloc in a QT string class. The section of code is reused many times and works as expected except in the specific test cases. The strange thing is that outputting qDebug messages to trace the fault stops the crashes. Time for memory monitoring deep dive...

elecpower:
Ready for review. App seems to be stable. I used QT IDE with GDB debugger and valgrind to monitor memory and nothing stood out as making a code change changed the object the crashes were detect to occur in. I suspect it is the sequence QT releases memory in the parent child trees and/or garbage collection and could even change depending on OS.

elecpower:
Warning: this PR will have multiple merge conflicts with PR #7651. Just let me know what I can do to help when you get to merging the two into 2.3.

bsongis:
Excellent PR, thanks!

PR #7953, opened 2020-09-21
Prevent overflow on text sensors

3djc:


radio/src/telemetry/telemetry_sensors.cpp
```diff
@@ -198,8 +198,10 @@ void TelemetryItem::setValue(const TelemetrySensor & sensor, int32_t val, uint32
     }
   }
   else if (unit == UNIT_TEXT) {
-    *((uint32_t*)&text[prec]) = newVal;
-    setFresh();
+    if (prec < TELEMETRY_SENSOR_TEXT_LENGTH) {
```
bsongis, line 201: it's still possible to get an overflow here, in case we write 4 bytes on TELEMETRY_SENSOR_TEXT_LENGTH - 1

3djc:
Superseeded

PR #7964, opened 2020-09-24
Wismy/ota

bsongis:


radio/src/gui/480x272/radio_sdmanager.cpp
```diff
@@ -394,6 +394,10 @@ bool menuRadioSdManager(event_t _event)
                 if (isReceiverOTAEnabledFromModule(EXTERNAL_MODULE, information.productId))
                   POPUP_MENU_ADD_ITEM(STR_FLASH_RECEIVER_OTA_BY_EXTERNAL);
               }
+              if (information.productFamily == FIRMWARE_FAMILY_FLIGHT_CONTROLLER) {
+                POPUP_MENU_ADD_ITEM(STR_FLASH_RECEIVER_OTA_BY_INTERNAL);
```
bsongis: what is the product you want to flash? I think the original string asks to the user to flash the receiver, and here it seems you want to flash the flight controller

PR #7992, opened 2020-10-05
Add 4+2 modular layout

### LINKED ISSUE #7757: New telemetry screen layout for 480x272
Hello, I would like to propose a new layout for telemetry screen. My idea is 4+2 or 2+4 layout. It is difficult to add all wanted values to one screen especially when some telemetry sensors cant be used with small cell (GPS for example). 2 pages is not always way to go.

ngng

3djc:
This add a configurable 4+2 widget, that adapts to use the available space based on its option.

This closes #7757

Default option
image

Mirror
image

Without trims
image

Without trims and sliders
image

Lets go crazy !
image


offer-shmuely:
great!! 
love it

Networker-001:
great!
I've written a widget, that's capturing sensor values.
if I inserting it twice, the second one does not work.
If I use the '2 widget' it works

PR #8001, opened 2020-10-07
Update Dockerfile for 2.3

nmaggioni:
Dockerfile was removed in 810cdc3 but the wiki [still references it](https://github.com/opentx/opentx/wiki/OpenTX-2.3-Docker-Build-Instructions) as a build method for 2.3. I've updated the latest known version of it to support Python3 and the needed more recent libs.

Tested by compiling 2.3.9 and 2.3.10 and flashing them on an X10.

### DIFF
--- a/radio/util/Dockerfile
+++ b/radio/util/Dockerfile
@@ -1,0 +1,23 @@
+# A Debian image for compiling OpenTX 2.3
+
+FROM debian:buster
+
+RUN DEBIAN_FRONTEND=noninteractive apt-get -y update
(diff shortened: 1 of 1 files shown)


bsongis:
Not in this repo!

nmaggioni:
@bsongis Care to elaborate on where this should go, then? The wiki still references it, and I didn't find any other (official) repo that has it.

gretel:
@bsongis think you should mind the motivational aspects of behaving like a superior.

PR #8005, opened 2020-10-09
Allow backlight function to turn off backlight entirely

3djc:
This change allow you to turn off screen when backlight function is feed with -MAX value
If user want to have the function to fully take control of backlight, set backlight mode to ON
If user want to adjust level, but keep auto dimming feature, then set backlight mode to keys/controls or both


jfrickmann:
I cannot select ON for backlight in global functions. I can select MAX, and it does not work; the backlight turns off after a few sec. If I leave it empty "---" however, then it does work. Using e.g. RS will adjust the strength, but not to off (yes, I did calibrate inputs ;-) ).

3djc:
Hi Jesper, not fully sure what you are trying to achieve, but this should give you some guidance : https://github.com/opentx/opentx/pull/8018

PR #8019, opened 2020-10-11
Fixed 115200 bps soft serial on F2.

qba667:
While checking implementation of non inverted software serial I spotted few problems, that are addressed in this commit.
First of all handling of shared EXTI was not correct, in case when shared IRQ was used we should avoid disabling it just disable source. Other problem was using to low priority when used with rotary encoder, so the priority must be set when soft serial is started. For higher speeds also order of handling inside handler method was import, handling soft serial as first.

### DIFF
--- a/radio/src/targets/taranis/telemetry_driver.cpp
+++ b/radio/src/targets/taranis/telemetry_driver.cpp
@@ -94,17 +94,17 @@ static uint16_t probeTimeFromStartBit;
 uint32_t oldPriority = 0xFFFF;
 void telemetryPortInvertedInit(uint32_t baudrate)
 {
-  //TODO:
+  // TODO
   // - handle conflict with HEARTBEAT disabled for trainer input...
   // - probably need to stop trainer input/output and restore after this is closed
   // - There is no need to stop trainer - we need to just enable EXT IRQ if not enabled by HEARTBEAT handling
-  //   see code below, but it is always nescesary to configure EXTI_Line for software uart.
+  //   see code below, but it is always necessary to configure EXTI_Line for software uart.
 
   if (baudrate == 0) {
     if (oldPriority != 0xFFFF) {
       NVIC_SetPriority(TELEMETRY_EXTI_IRQn, oldPriority);
     }
-    //Never disable EXTI_IRQ just remove TELEMETRY_EXTI_LINE from configuration as EXTI can be reused
+
     NVIC_DisableIRQ(TELEMETRY_TIMER_IRQn);
 
     EXTI_InitTypeDef EXTI_InitStructure;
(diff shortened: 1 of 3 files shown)


bsongis:
On which radio(s) did you test this fix?

qba667:
@bsongis affected one QX7. 
It is very easy to simulate it.
1. Add print out just before line telemetryFifo.push(rxByte) in the file telemetry_driver.cpp.
2. Connect USB to serial adapter to QX7 using putty or other client establish the connection  using speed 115200 bps 8 data bit parity none, stop bit 1.
3. Try to send some over putty
4. Observe traces on debug port of the radio

The problem in case of X7 was to low priority of the rotary encoder, plus the logic that was not ensuring external IRQ is enabled at all.

bsongis:
OK so the fix is only for QX7?

qba667:
As far I can tell yes (also ACCESS variant), but I tries to keep it as universal as possible - removing if def for special flags makes it more portable and easier to test.

qba667:
@bsongis were you able to take a look at this one?

radio/src/targets/taranis/telemetry_driver.cpp
```diff
@@ -162,17 +165,17 @@ void telemetryPortInvertedInit(uint32_t baudrate)
   EXTI_InitStructure.EXTI_Trigger = TELEMETRY_EXTI_TRIGGER;
   EXTI_InitStructure.EXTI_LineCmd = ENABLE;
   EXTI_Init(&EXTI_InitStructure);
-
-  //TODO:
-  // - handle conflict with HEARTBEAT disabled for trainer input...
-  // - probably need to stop trainer input/output and restore after this is closed
-#if !defined(TELEMETRY_EXTI_REUSE_INTERRUPT_ROTARY_ENCODER) && !defined(TELEMETRY_EXTI_REUSE_INTERRUPT_INTMODULE_HEARTBEAT)
+  oldPriority = NVIC_GetPriority(TELEMETRY_EXTI_IRQn);
```
bsongis, line 170: is it needed?

radio/src/targets/taranis/telemetry_driver.cpp
```diff
@@ -91,17 +91,20 @@ static uint8_t rxByte;
...
+  //   see code below, but it is always nescesary to configure EXTI_Line for software uart.
 
-    //TODO:
-    // - handle conflict with HEARTBEAT disabled for trainer input...
-    // - probably need to stop trainer input/output and restore after this is closed
-#if !defined(TELEMETRY_EXTI_REUSE_INTERRUPT_ROTARY_ENCODER) && !defined(TELEMETRY_EXTI_REUSE_INTERRUPT_INTMODULE_HEARTBEAT)
-    NVIC_DisableIRQ(TELEMETRY_EXTI_IRQn);
-#endif
+  if (baudrate == 0) {
+    if (oldPriority != 0xFFFF) {
```
bsongis, line 104: needed?

qba667: @bsongis Comments improved, typos fixed - up to you if you want to keep it.

bsongis: I meant: is it needed to keep the oldPriority in memory to set it again when we don't need the high priority anymore? Because we should know which is the value of "oldPriority", it should be a constant, do I miss anything?

PR #8024, opened 2020-10-12
I am a member of Radiomaster, and I did a little Chinese translation,…

zyren:
… because I am using it for the first time and I am not sure whether the submission is successful, so this is the first test.


3djc:
謝謝 !

DusKing1:
@3djc No offense, but the proper string should be 谢谢 . They are different between zh_CN and zh_TW.

PR #8033, opened 2020-10-14
add racing mode when ACCESS + 8CH

WismyYao:
when use ACCESS with 8ch output,  user can choose racing mode .  In this mode , no telemetry passes back


raphaelcoeffic:
@WismyYao why is the custom function function necessary for something which is basically a module / protocol specific option? This does not really make sense to me.

radio/src/gui/gui_common.h
```diff
@@ -184,6 +184,19 @@ inline uint8_t MODULE_CHANNELS_ROWS(int moduleIdx)
   }
 }
 
+inline uint8_t IF_ALLOW_RACING_MODE(int moduleIdx)
+{
+  if (!IS_MODULE_ENABLED(moduleIdx)) {
+    return HIDDEN_ROW;
+  }
+  else if (isModulePXX2(moduleIdx)) {
+    if (0 == g_model.moduleData[moduleIdx].channelsCount)
```
bsongis, line 201: it is right, but difficult to read, I can't remember if we have a macro for this

bsongis:
Yes, I do agree with @raphaelcoeffic on this one, I don't understand the need of the function. 
Let's imagine and try to go a little bit further: would it mean that all module options should have their own function?

raphaelcoeffic:
Normally, the way to set module options “on-the-fly” would be to use a LUA function and tie it to some knob. This much more flexible and useful than crafting a special function for the same purpose.

kilrah:
But it doesn't need to be on the fly, so should be a checkbox in the module settings instead.

bsongis:
I just asked to Wismy this morning. Here is the way she have thought about it:
1) The user starts its flight, with telemetry enabled (no racing mode)
2) Then he wants to switch to racing mode (improved delays, but no telemetry) => he enables the special function
3) At any moment he can recover the telemetry by switching off the racing mode

Useful? Not useful? The checkbox in the module settings is far easier for her to implement, but she thought it would add some more possibilities to do it this way

3djc:
Beside creating useless functions for tons of users, it also creates lots of confusions for relevant users, if you change receiver for example, since not all seem to support it, you don't have any way to know your function is now usless. I can see only confusion resulting of that

kilrah:
That's actually a pretty interesting way to go about it, if they get the switch to work right on the RF side it's a good idea.
Actual racers don't give a damn about telemetry, but maybe some people will...

The function should only be shown if the protocol selection is appropriate and hidden otherwise.

radio/src/gui/gui_common.h
```diff
@@ -184,6 +184,28 @@ inline uint8_t MODULE_CHANNELS_ROWS(int moduleIdx)
...
+  static uint8_t unenableCount = 0;
+  if (flag && moduleIdx==INTERNAL_MODULE) {
+    if (!IS_MODULE_ENABLED(moduleIdx)) {
+      return HIDDEN_ROW;
+    }
+    else if (isModulePXX2(moduleIdx)) {
+      if (0 == g_model.moduleData[moduleIdx].channelsCount)
+        return 1;
+    }
+    g_model.moduleData[moduleIdx].pxx2.enableRacingMode = 0;
```
bsongis: This changes something in your model, depending if you are inside the menu or not!

raphaelcoeffic:
> I just asked to Wismy this morning. Here is the way she have thought about it:
> 
> 1. The user starts its flight, with telemetry enabled (no racing mode)
> 2. Then he wants to switch to racing mode (improved delays, but no telemetry) => he enables the special function
> 3. At any moment he can recover the telemetry by switching off the racing mode
> 
> Useful? Not useful? The checkbox in the module settings is far easier for her to implement, but she thought it would add some more possibilities to do it this way

Such corner cases are really best tackled with LUA, IMHO. For the main users, the receiver settings way is really simple and probably intuitive as well. That's where I would look for it.

radio/src/gui/212x64/model_setup.cpp
```diff
@@ -1098,6 +1100,15 @@ void menuModelSetup(event_t event)
         break;
       }
 
+      case ITEM_MODEL_SETUP_INTERNAL_MODULE_RACING_MODE:
+      {
+        uint8_t moduleIdx = CURRENT_MODULE_EDITED(k);
+        ModuleData & moduleData = g_model.moduleData[moduleIdx];
+        lcdDrawTextAlignedLeft(y, STR_RACING_MODE);
+        moduleState[moduleIdx].racingMode = editCheckBox(moduleState[moduleIdx].racingMode, MODEL_SETUP_2ND_COLUMN, y, "", attr, event);
```
bsongis, line 1108: I don't understand here, it will change moduleState which is volatile, then on next startup are you sure the state will be restored?

sp00ls:
@WismyYao 
Hi, has there been any progress on getting this worked out? It looks like FrSky dropped the new racing mode firmware for the Archer RS today, but there is no way to use it aside from compiling this branch manually.

Edit: I did compile this branch myself to try out race mode. X9D+2019 build is attached for anyone who finds this pull request while trying to figure out how to enable race mode and just wants to try it. 
[opentx-x9d+2019-2.3.12-noheli-ppmus-lua-flexr9m-racing_mode.bin.zip](https://github.com/opentx/opentx/files/5884579/opentx-x9d%2B2019-2.3.12-noheli-ppmus-lua-flexr9m-racing_mode.bin.zip)

bsongis:
I guess I can close this one, the feature seems reimplemented in #8267

PR #8044, opened 2020-10-17
Wrap call to auxSerial function with define check

olliw42:
self explanatory

see https://github.com/opentx/opentx/blob/2.3/radio/src/trainer.cpp#L118-L122 for comparison

may not have bitten you so far since the combination of TRAINER_BATTERY_COMPARTMENT and not AUX_SERIAL didn't yet occur
(was found when trying to compile for the TX16S without AUX_SERIAL enabled)


3djc:
Not needed anymore

PR #8048, opened 2020-10-17
Fix print telemetry sensor 

### LINKED ISSUE #8035: Companion: "Print model" of analog sensor prints the range wrong if precision is other than 1 digit
I want to use the analog sensor for RSSI. After discovery of the A2 I change its range to 100 and the unit to dB.
image
The print model then gives me this.

TELE6:2RSS Custom Id: F103 Inst: 27 Unit: dB Prec: 1 Ratio: **100** Offset: 0 A/Offset: N Filter: Y Positive: N Log: Y 

This looks fine, but when I change the precision to 0 decimals the print model reports a range of 1000 even though I selected 100

image

TELE6:2RSS Custom Id: F103 Inst: 27 Unit: dB Prec: 0 Ratio: **1000** Offset: 0 A/Offset: N Filter: Y Positive: N Log: Y

elecpower:
Fixes #8035
Precision and offset to match Companion
Code housekeeping

### DIFF
--- a/companion/src/modelprinter.cpp
+++ b/companion/src/modelprinter.cpp
@@ -23,7 +23,6 @@
 #include "multiprotocols.h"
 #include "boards.h"
 #include "helpers_html.h"
-#include "multiprotocols.h"
 #include "appdata.h"
 
 #include <QApplication>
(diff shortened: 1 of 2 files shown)


3djc:
@elecpower could you rebase your PR so that we have Travis working ?

PR #8052, opened 2020-10-19
Extra widget layouts added

PD5DJ:
Added 4 new layouts

1x2, 1x3, 1x4 and a 5 way layout

All for wide/long horizontal/vertical widget creation


bsongis:
Which widget do you put in these layouts? 
I wanted to have Lua layouts, but I didn't implement it (as done for Themes and Widgets), perhaps it would be suitable for you?

PD5DJ:
Widgets that prefer full width of screen.

Creating layouts with lua would be even more great ofcourse.

Examples of these layout usage.

https://github.com/Hobby4life/OTX-Widget-480x272-Widget-Layouts/raw/main/layout-example.png

https://github.com/Hobby4life/OTX-Widget-480x272-Widget-Layouts

raphaelcoeffic:
@Hobby4life I think this is something that should be looked at in 2.4. 2.3 should not have these kind of changes anymore. It would be much better to participate in the 2.4 effort, so that we can finally move on.

raphaelcoeffic:
@Hobby4life hi! Would you like to move this PR to EdgeTX? We have some great advances on 2.4, and it would be nice to have this PR.

Eldenroot:
> @Hobby4life hi! Would you like to move this PR to EdgeTX? We have some great advances on 2.4, and it would be nice to have this PR.

maybe this can be done right now by edgetx team (if has not been added already)?

pfeerick:
All but the 5 way layout were added to ETX in 2.5.

PR #8066, opened 2020-10-29
Fix insertMix

### LINKED ISSUE #8021: model.insertMix(..) fail OTX 2.3.10
OpenTX 2.3.10
X12s Horus.
i have tried this:
local Data = model.getMix(0,0)
model.insertMix(0+15,0,Data)
This cmd crashs my function script. it stops after the cmd.
if you edit the new Mixer(ch16), you can see that the source is incorrect transfered.
I have no problems with the model.insertMix(...) cmd in 2.3.9

3djc:
This fixes #8021
This also fixes I_SOAR scripts issues among others


lshems:
Perhaps reformulate because there are no script issues in isoar scripts:

This also allows to use isoar scripts again, as well as other scripts impacted by the bug created by the last opentx release in the in insrtmix lua API function.

PR #8072, opened 2020-11-01
Fix issue 8060 - Taranis fails to update avr multiprotocol module

### LINKED ISSUE #8060: Updating AVR multiprotocol module firmware fails from Taranis X9D at openTX 2.3.10
openTX 2.3.10
Taranis X9D
AVR Multiprotocol Module

Updating AVR multiprotocol module firmware from Taranis X9D fails with error "Firmware update error - NoSync" at openTX 2.3.10.  This worked with openTX 2.3.9.

This was reported and discussed on RCG in the DIY Multiprotol Module thread in the following posts by myself and others:
https://www.rcgroups.com/forums/showpost.php?p=45700669&postcount=20192
https://www.rcgroups.com/forums/showpost.php?p=45704105&postcount=20196
https://www.rcgroups.com/forums/showpost.php?p=45692937&postcount=20188

As reported in the above links, flashing a stm module from X9D works and flash a avr module from TX16s works.  Flashing a avr module from X9D fails.  I believe the error is coming from multi_firmware_update.cpp line 247.

jtaylor2:
firmware by changing probeTimeFromStartBit back to original value of 48

### DIFF
--- a/radio/src/targets/taranis/telemetry_driver.cpp
+++ b/radio/src/targets/taranis/telemetry_driver.cpp
@@ -123,7 +123,7 @@ void telemetryPortInvertedInit(uint32_t baudrate)
       break;
     case 57600:
       bitLength = 35; //34 was used before - I prefer to use use 35 because of lower error
-      probeTimeFromStartBit = 52; //round down - 48 used in original implementation
+      probeTimeFromStartBit = 48; //48 used in original implementation
       break;
     default:
       bitLength = 2000000/baudrate; //because of 0,5 us  tick


3djc:
I don't think we got a clear response for Jakub on why it was changed, or did I miss it ?

qba667:
@3djc to fit better into center of the bit, I was debugging software serial and analysed when data is probed. It should be around 1.5x bit cycle time +-  some correction because of processing time. Need to look at my communication dumps.

qba667:
1 bit length at 57600 bps = 34,72 us. Because we want to probe the bus in the middle of the pulse, we should start at 1.5 * bit length in us 34,72 * 1.5 = 52,08. At the end it means we have error margin of 17us, because set bit time 35 is not exactly equals 34,72 (difference 0,28) we can successfully probe (17 / (35-34,72)) = 60 bytes. We can keep 48 it will give us more time for recognizing valid signals.

jtaylor2:
Bit length of 34 with probeTime of 52 also works, but as the failure with avr update is reading the sync byte it seems that probeTime is a better fix.   I had tried 50 with bit length of 35 to split the difference and that didn't work either.

I see you edited your comment to say we can keep 48 so I will leave the pull request as is, but can change it if desired.

PR #8076, opened 2020-11-03
Always display output channel name on mix screen

3djc:
Currently, output name is shown on B&W LCD in mixer screen ONLY if a mixer lines exists.

This display the name also when one doesn't


pascallanger:
Great, thanks!!!

PR #8117, opened 2020-11-22
Improve TW display

zyren:
Authored by Jacky Wu from Taiwan.


3djc:
Thx Zyren , will get updated firm to you

PR #8124, opened 2020-11-25
Try to fix the functions list in Logic Switch

zyren:
After "Δ}x\0", the display is disordered. I suspect that the problem may come from the length of "Δ"


3djc:
Zyren, closing this. The delta character is a strange beast that needs to be done with STR_CHAR_DELTA, but there is another issue on our side with '}" not producing the expected result (and altering count). I'll update the translation guide when this gets fixed. Please stop progressing on  this until this gets fixed.

Issue #7562, opened 2020-04-30
Lua function to reset telemetry sensor


jfrickmann:
There is a request (#6415) for reset telemetry/flight but not for an individual sensor.

When I publish model templates for people, sensor integration is hard to do, because they have not yet "discovered" their sensors when I make the template. One way is to do everything in OTX with calculated sensors, and then require users to add the real sensors as sources to the calc sensors. Works great until someone hits "delete all sensors". Alternative is to do it in Lua. But here I cannot reset e.g. the altimeter. I could hack a solution where Lua modifies a SF to do it, but that would break if FUNC_RESET_PARAM_FIRST_TELEM is changed in a future version.

Therefore, it would be great to have a lua function model.resetSensor(sensor) doing exactly the same as the Special Custom Function Reset TELEx:Alt

In the meantime, I do not reset the sensor. I correct internally in my lua programs, knowing that the log file may be garbage if the altimeter drifts up and down during a flying session. But at least, things look good in my own graphing scripts because it zeros the start of every flight for Alt.

lshems:
You can fix that by writing your own alt value to a virtual sensor, and have them discover that.
You can than reset whenever you want.

Since you posted this, I came to the conclusion that a lot of people suffer from the big drift on the alti on the new receivers where it is integrated. In my scripts I also offset the show alti with the value just at release of the launch switch, so it behaves good enough to evaluate launch height, which is determined a few seconds after start. The offset was introduced to get repeatable launch height measurements. 

But now the auto-offset at launch has become a precondition to have a reasonable graph shown on the radio. But the logs are still a mess, without the offset.

I will certainly add a virtual sensor and write the altitude I show on the radio graph during flight also to the log file :).

jfrickmann:
I have made a pull request for code adding the desired function - please see #8137.

PR #8137, opened 2020-12-04
Added model.resetSensor(sensor) function

jfrickmann:
Issue #7562


3djc:
I wonder if this function should not return nil when sensor is not found and maybe sensor number when found ?

jfrickmann:
I assume that you have the current discussion on RCGroups "LUA scripting Technical discussion" thread in mind? Since the purpose of this function is not to read any value back, I do not think that it matters here, but I am not against it either. Please let me know your preference!

PR #8146, opened 2020-12-08
Dgatf/increase popup size smalllcd

dgatf:
This PR increases the size of the popups for small lcds as is easy to lose part of the popup message, in particular for lua errors

Before:

captura_x7_20-12-08_10-14-11

After:

captura_x7_20-12-08_11-21-04


kilrah:
You don't need to close a PR and recreate one when you make changes, it updates as you continue pushing to the branch.

dgatf:
Yes, I know. But I included an unwanted commit from another PR. Then when I removed this PR commits with a rebase, it got closed automaticly so I needed to open a new one... Sorry for the mess

PR #8147, opened 2020-12-08
Improve script error output for small lcds

dgatf:
This PR improves the output for script errors:

- Adds a TR for small lcds to translatables
- Removes the script path
- Adds a second error line only if the error len is bigger than 20. No changes on the popup menu size

Before:

screenshot_x7_20-12-22_08-33-07

After:

captura_x7_20-12-22_08-18-57
captura_x7_20-12-22_08-24-32

### DIFF
--- a/radio/src/gui/128x64/popups.h
+++ b/radio/src/gui/128x64/popups.h
@@ -23,16 +23,16 @@
 
 #include <keys.h>
 
-#define MESSAGEBOX_X                   1
+#define MESSAGEBOX_X                   10
 #define MESSAGEBOX_Y                   16
-#define MESSAGEBOX_W                   (LCD_W - 2)
+#define MESSAGEBOX_W                   (LCD_W - 19)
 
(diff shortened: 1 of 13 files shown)


bsongis:
The menu is less beautiful :(

bsongis:
Perhaps remove the directory (coming from Lua): TOOLS

bsongis:
And at the end add an "ellipsis" function which would write "No SBEC sensor..."

dgatf:
> The menu is less beautiful :(

Yep, I thought that after putting all the screenshots together

Or maybe this bigger popup but only for lua errors

3djc:
We can also leave only the script name and remove the path elements

dgatf:
New attempt. Edited PR description

bsongis:
It seems very better!

bsongis:
I keep it for 2.3.12 :)

PR #8156, opened 2020-12-10
Remove module choice from special function set failsafe

### LINKED ISSUE #7737: Companion : issue in SF set failsafe
Companion does offer "set failsafe internal" and "set failsafe external" when only one should be proposed

image

Exists in 2.3.9 and has probably be present for some time

elecpower:
Fixes #7737


bsongis:
Thanks for your 3 PRs, I see many small interesting fixes!

PR #8169, opened 2020-12-20
Dynamic ui data models

### LINKED ISSUE #8187: Insert Logical switch in Companion causes incorrect SF
When inserting a logical switch often the switch condition of the Special Functions is affected. Insert function is available in Companion only.

**To Reproduce**
1. Add at least 2 LS
2. Add 2 SF with switch condition depending on the LS

3a, Insert new LS line between LS1 and LS2. So that LS2 is now LS3.
-> Switch condition in SF2 will be moved to FM0
Expected behavior: SF2 switch condition should be LS3

3b. Insert new LS line before LS1. So that LS1 is now LS2 and LS2 will be LS3.
-> Switch condition in SF1 will be moved to LS2 correctly. SF2 stuck on LS2
Expected behavior: SF2 switch condition should be LS3 also

Edit: Nearly same behavior when deleting lines.

- OpenTX Companion 2.3.10 with Radio Profile X9D+ 2019

elecpower:
Define more dynamic ui data model classes and helper functions to data classes.
These new classes will be progressively incorporated into the model and general edit uis.
To reduce conflicts and the code review effort, each edit ui will be a separate PR.
Fixes #8187


elecpower:
Whilst working on the non-updating data  models I have come to the conclusion that it would be beneficial to refactor the base data and filter classes as part of this PR to support both model types.
This will result in the PR touching lots of places but in most cases it will be straight renames with no logic changes. But better to do now whilst the of models implemented is relatively small.

elecpower:
Support added for:
- non-updating item models
- unfiltered and filtered item factories to simplify parameter passing and ensure memory released
- factories can be shared eg per radio model and local eg model edit tab

Model edit, telemetry, inputs, mixes and curve reference demonstrate these uses.
To minimise memory and maximise execution performance, unfiltered item models should be singletons and referenced by one or more filtered item models.

bsongis:
We trust you ;)

PR #8172, opened 2020-12-20
Remove retry timer for set trim on simulator

### LINKED ISSUE #7722: [Companion 2.3 Simu] Trim value spontaneously resets when switching between flight modes
**Problem**
Simple setup with two flight modes:
- FM0 has TrmT as 'disabled'
- FM1 has TrmT as 'own trim', value=+100. 

When switching between the two flight modes, the 'own trim' value spontaneously resets to zero, sometimes immediately, other times after a couple of cycles.

So far I can only reproduce this in Companion.

**To Reproduce**
1. Start Companion 2.3.7, with X9D profile
2. Open [this .OTX file](https://www.dropbox.com/s/vy02biivyakev9o/TrimBug_For_OTX_237_Simu.otx?dl=0)
3. Open the simulator
4. Keep an eye on the TrmT graphic (or CH1, which uses TrmT as source)
5. Switch rapidly between between flight modes using SH.

**Expected**: TrmT should alternate between +100 and zero
**Actual**: after a couple of cycles, TrmT shows zero for both FMs.

**General information**
 - Companion 2.3.7 (but also occurs in 2.2.x)
 - Frsky X9D profile
 - Windows 10 X64

dgatf:
Fixes #7722

The issue only affects the simulator, not the firmware

The delay timer applied to the trim is too long, so it may be applied to the wrong flight mode when switching fast between flight modes in the simulator

BTW... why is needed a delay for the trim on the simulator?

### DIFF
--- a/radio/src/targets/simu/opentxsimulator.cpp
+++ b/radio/src/targets/simu/opentxsimulator.cpp
@@ -213,17 +213,7 @@ void OpenTxSimulator::setTrim(unsigned int idx, int value)
   if (i < 4)  // swap axes
     i = modn12x3[4 * getStickMode() + idx];
   uint8_t phase = getTrimFlightMode(getFlightMode(), i);
-
-  if (!setTrimValue(phase, i, value)) {
-    QTimer *timer = new QTimer(this);
-    timer->setSingleShot(true);
-    connect(timer, &QTimer::timeout, [=]() {
-      emit trimValueChange(idx, 0);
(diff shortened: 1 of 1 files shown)


bsongis:
Why do we emit an event when setTrimValue fails, why this delay?

bsongis:
No answer :)
I propose we remove it, if later we understand why it is needed we won't forget to add a comment ;)

dgatf:
I agree. Timer removed

PR #8189, opened 2020-12-30
Turn off module on model change

3djc:
This turns off module on commited model change to allow module to start properly on new model settings


3djc:
Put in draft state, Multi team will be having a look at the multi bootloader first

CapnBry:
Just to help clarify why this PR exists at all, it was created in response to this issue I was seeing when switching between models.
[Youtube](https://youtu.be/N_HzjatSFbw)

I realize this isn't the Issue tracker so it isn't for discussion, but I thought it would be helpful as a reminder for what this was addressing based on discussion on RC Groups.

3djc:
It is is draft mode because speaking to the multi guys, they feel it should be fixed in multi bootloader, or at least they would like to try

CapnBry:
Oh yeah no worries at all, I just wanted to provide a link for the video (before I forgot about it) in case in a couple months someone wonders why this PR even exists and by that time it would be buried in the RC Groups thread. :)

bsongis:
Why the module is not shutdown in this function:

```cpp
void preModelLoad()
{
  watchdogSuspend(500/*5s*/);

#if defined(SDCARD)
  logsClose();
#endif

  if (pulsesStarted()) {
    pausePulses();
  }
  pauseMixerCalculations();

  stopTrainer();
}
```

CapnBry:
76e12ef Confirmed still works perfectly switching CRSF to MULTI, CRSF to CRSF, MULTI to MULTI, and MULTI to CRSF. I'm sure you knew that but I tested just to be sure.

PR #8197, opened 2021-01-04
Companion - MPM Updates to 1.3.1.92

MRC3742:
@3djc as it appears 2.3.11 is near to release and maybe these small changes, additions could be added to bring Companion and Radio GUI up to the current MPM version 1.3.1.92

Thanks as always - Rich


3djc:
Thanks a lot @MRC3742 , this is very helpfull

PR #8218, opened 2021-01-14
Properly space swich/pots check text

### LINKED ISSUE #8217: (Minor) In Preflight Checks EX1 display label is shortened to EX (EX2 on the other hand is shown correctly as "EX2")
**Bug description**

When optional inputs EX1 and EX2 are enabled with setting "Pot with detent" and they are included in model Preflight checks, their state is checked on power-up.
If EX1 is not at the center, a warning is given, but the display text says "EX" without a number. Correct would be "EX1".

(On the other hand EX2 is handled correctly -> the display shows "EX2")

**To Reproduce**
Tested with OpenTX v2.3.11 on RadioMaster TX16S with two additional pots connected to inputs SI & SJ.
1) Enable EX1 & EX2 as "Pot with detent"
2) In model setup, enable under Preflight checks EX1 & EX2
3) power off
4) make sure EX1 and EX2 are not centered
5) power up
6) observe the wrong warning label for EX1

**Expected behavior**
Preflight check dialog should say "EX1" if EX1 is not centered and not just "EX"

**General information**
 - OpenTX version v2.3.11
 - RadioMaster TX16S with additional two pots connected to SI & SJ

3djc:
This fixes #8217 

image

### DIFF
--- a/radio/src/switches.cpp
+++ b/radio/src/switches.cpp
@@ -663,6 +663,9 @@ void checkSwitches()
       }
       int x = SWITCH_WARNING_LIST_X;
       int y = SWITCH_WARNING_LIST_Y;
+#if defined(COLORLCD)
+      lcdNextPos = SWITCH_WARNING_LIST_X;
+#endif
       int numWarnings = 0;
       for (int i=0; i<NUM_SWITCHES; ++i) {
 #if defined(COLORLCD)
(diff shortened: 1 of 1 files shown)


elecpower:
@3djc just a cosmetic idea but would it not be more consistent to have the pot symbol after the name as for switches?

3djc:
Actually, the switch symbol is not displayed to save space, the arrow is the position, and when displayed, symbol is always to the right (look in hardware screen for example)

elecpower:
@3djc apologies for the poor description on my part. As you say the switch position is to the right of the switch name. So for pots place the 'dial' symbol to the right of the pot name as it implies variable position.

3djc:
no, mixed up my words, sorry, symbol is left, always

image

On the right is the position, which is only available for switches

PR #8221, opened 2021-01-16
Colorlcd Multirotor wizard

3djc:
Add a multirotor wizard like the one available on b&w radio.
Graphics courtesy of Radiomaster


Eldenroot:
Can we expect heli wizard as well?

3djc:
> Can we expect heli wizard as well?

Probably not, at least not until the community defines standards in heli setup

lshems:
> > Can we expect heli wizard as well?
> 
> Probably not, at least not until the community defines standards in heli setup

Not fair to point to the community. You need flight modes to set up a Heli in a sensible way.

Flight modes switches cannot be set up in Lua.

But never mind.

PR #8225, opened 2021-01-16
Fix deprecated functions

sfjuocekr:
Fixed some deprecated functions and an obsolete stackpointer clobber.


radio/src/targets/horus/CMakeLists.txt
```diff
@@ -85,7 +85,19 @@ if (PCB STREQUAL X10)
     add_definitions(-DRADIO_FAMILY_T16)
     option(INTERNAL_MODULE_MULTI "Support for MULTI internal module" ON)
     option(BLUETOOTH "Support for bluetooth module" OFF)
+    set(HARDWARE_TOUCH YES)
+    set(USB_CHARGER YES)
     add_definitions(-DMANUFACTURER_JUMPER)
+  elseif (PCBREV STREQUAL TX18S)
```
kilrah, line 91: Please keep that for a separate PR, knowing we don't add support for things when we don't have the hardware to test and direct contact with the manufacturer is preferred.

radio/src/targets/horus/CMakeLists.txt
```diff
@@ -85,7 +85,19 @@ if (PCB STREQUAL X10)
     add_definitions(-DRADIO_FAMILY_T16)
     option(INTERNAL_MODULE_MULTI "Support for MULTI internal module" ON)
     option(BLUETOOTH "Support for bluetooth module" OFF)
+    set(HARDWARE_TOUCH YES)
```
kilrah, line 88: T18 doesn't have working touch, and again please separate refactorings from functional changes

sfjuocekr: Shit sorry about the goof up there, I was working on adding TX18S support and finally figured out what was wrong with my builds.

I also "fixed" the fromSet deprecation, but apparently the QT libs on travis still don't have those changes so that needs to be reverted as well.

Is there a means of leaving a commit out of a PR? Or else Ill just rebase on the latest master.

kilrah: Don't think your can exclude a commit, but you can just revert them. Or if you want to make it clean rename your branch, recreate one with the same name and cherry pick the commits you want, as long as the branch name is the same it'll still link to the PR

sfjuocekr:
I'll close this and make a new PR without the TX18S references, this needs to be a new branch for now anyway.

As Im fairly new to OpenTX development, could someone give me a few pointers to debugging? I would like to see at least the log output like the simulator does.

kilrah:
If you build with DDEBUG=YES you should have the debug output on the serial port, and also on USB if it is set to SERIAL.

sfjuocekr:
Doh, that makes sense!

PR #8226, opened 2021-01-16
Fixed some function deprecations, a bogus stackpointer clobber and a …

sfjuocekr:
…redundant __packed reference the compiler complained about it not doing anything.

### DIFF
--- a/radio/src/targets/horus/lcd_driver.cpp
+++ b/radio/src/targets/horus/lcd_driver.cpp
@@ -142,6 +142,7 @@ static void LCD_NRSTConfig(void)
 }
 
 // TODO delay function
+#if !defined(RADIO_T18)
 static void delay3(uint32_t nCount)
 {
   uint32_t index = 0;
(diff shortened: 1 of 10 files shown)


radio/src/targets/horus/lcd_driver.cpp
```diff
@@ -142,6 +142,7 @@ static void LCD_NRSTConfig(void)
 }
 
 // TODO delay function
+#if !defined(RADIO_T18)
```
3djc, line 145: What's that for ?

sfjuocekr: Because that function is only used if that radio is not defined if you look down at line 158

bsongis: it would be nice to use delay_us anyhow!

sfjuocekr: I left the original code mostly as-is and just fixed some compiler complaints.

Looking at the code that would be a huge number to put into delay_us, the person who put it there probably had a reason for it and sadly it is not commented either.

I'm guessing the datasheet of the display/controller would have the reason documented.

PR #8228, opened 2021-01-18
Allow setting serial baurate for LUA

3djc:
Also:
- default back to 115k on TX16S
- defines MAX source for LUA
- some code housekeeping


stronnag:
Tested with 2.3.12N420 / UART1/ Bluetooth adapter and a LUA script. Data received as expected.

PR #8229, opened 2021-01-18
Fix SBUS Trainer Module Input for X7 Access

bug400:
Trainer SBUS-Input on the external module port does not work on the TARANIX Q X7
ACCESS. If I connect a bound Frsky XM reveiver to the Heartbeat, Batt and
Gnd pins of the module port and activate Trainer/SBUS then I see no
input in the Trainer Menu.

I found that the file radio/src/targets/taranis/hal.h has an illegal
configuration for the INTMODULE_HEARTBEAT_XXX definitions.
According to the documentation for the STM32F407 there does not exist an
alternate function mapping to an USART for pin 7 of Port A.

With the configuration in this patch I got the SBUS-Input for the external
module port working.


radio/src/targets/taranis/hal.h
```diff
@@ -1500,13 +1500,14 @@
 #elif defined(RADIO_X7ACCESS)
   #define INTMODULE_HEARTBEAT
   #define INTMODULE_HEARTBEAT_REUSE_INTERRUPT_ROTARY_ENCODER
-  #define INTMODULE_HEARTBEAT_RCC_AHB1Periph      RCC_AHB1Periph_GPIOA
-  #define INTMODULE_HEARTBEAT_GPIO                GPIOA
-  #define INTMODULE_HEARTBEAT_GPIO_PIN            GPIO_Pin_7  // PA.07
-  #define INTMODULE_HEARTBEAT_EXTI_PortSource     EXTI_PortSourceGPIOA
+  #define INTMODULE_HEARTBEAT_RCC_AHB1Periph      RCC_AHB1Periph_GPIOC
```
bsongis, line 1644: I think you will fix the Trainer input on external module, but in the same time you break the internal module heartbeat which is used to improve the latency.

I need the schematics to fix it properly, will get back to this soon!

raphaelcoeffic: @bsongis did you check whether or not on that radio the heartbeat input is share with the rotary encoder? If this is the case, aren't the defs here bogus anyway?

mariusvds: Is this going to be fixed?
The problem is that I want to connect a ESP32 to the input, and I can find many libraries to generate a SBUS signal, but no for CPPM. (Does this bug also affect the input of CPPM-signals?)

bug400:
Should be removed.

PR #8243, opened 2021-01-24
#if #endif clauses look screwed up

olliw42:
the #if #endif clauses seem to be intertwined and not properly located. I tried to decorrelate according to what I think the intention would be. 

I tried to add code at the end of the file and it gave compile errors with AUX2_SERIAL disabled, that's how I found that.


3djc:
Since travis is broken, I pushed a change myself that I could check localy. Thx for spotting that

olliw42:
oh wau, this was really fast
I was concerned that it failed, but it seems it wasn't me, good to know
THX

PR #8255, opened 2021-01-26
Introduce model telemetry streaming

3djc:
This change allow to differentiate between module and model telemetry.

When model telemetry is lost ("rssi = 0"), the telemetry lost is heard as usual. The non streaming sensor go 'dark', but still actively transmitted sensors, like module transmit power for exemple, are still been updated


olliw42:
great PR. I was initially enthusiastic since I though it could help solving a rssi issue we have with the Mavlink for OpenTx stuff. It however doesn't do what I naively would have expected, so pl allow me to present what my thoughts are on this.

I do see two sources for RSSI, namely 
1. what one can set via setTelemetryValue(), which is the RSSI as listed in the TELEMETRY page under senors, and as it also can be displayed in widgets.
2. what one can set via telemetryData.rssi.set(), which is what triggers the alarms, and also determines the rssi bars in the right of the top bar, and is what TELEMETRY_RSSI() refers to

Now, from the fact that (2) triggers the alarms and sets the rssi bars, I would have considered it the "built-in" RSSI and thus would have thought that it should NOT be related to whether there is a RSSI sensor discovered or not. However it is, since in radio/src/telemetry/telemetry.cpp the voice e.g. is tested against MODEL_TELEMETRY_STREAMING().

I would have arranged things such as to disentangle (1) and (2). I would have made (2) to relate to the old TELEMETRY_STREAMING(), and would have made (1) to relate to the new MODEL_TELEMETRY_STREAMING(). And I would have not tied MODEL_TELEMETRY_STREAMING() to implicitly assume TELEMETRY_STREAMING() (as it is except for ghost), since sensor values might be valid even though there is no TELEMETRY_STREAMING().

just my 2 cents

PR #8259, opened 2021-01-29
Companion sensors refactoring

elecpower:
Move data driven code from ui, edit and print to data class


companion/src/firmwares/helpersdata.h
```diff
@@ -0,0 +1,53 @@
...
+ * it under the terms of the GNU General Public License version 2 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ */
+
+#ifndef _HELPERS_DATA_H_
```
3djc: Just a cosmetics here, we are moving away from this to favor `#pragma once` instead, so 21 can be replaced by the pragma, 22 and 53 can be deleted (we have enough #ifdef already :D. We are replacing those gradually as we make changes to those .h files

3djc:
Could you rebase on 2.3 ? That should make our friend Travis happier ;)

elecpower:
Rebased to kick Travis again

PR #8261, opened 2021-01-29
Fix multi per channel failsafe

3djc:
As per discussion with Pascal


pascallanger:
This is correcting 2 issues:
 - No pulse=None setting per channel was evaluated as Hold
 - It was possible to have a failsafe value being evaluated as Hold

PR #8267, opened 2021-02-01
Wismy/racing mode

WismyYao:


radio/src/gui/128x64/model_setup.cpp
```diff
@@ -1230,6 +1232,16 @@ void menuModelSetup(event_t event)
         break;
       }
 
+#if defined(HARDWARE_INTERNAL_MODULE)
+      case ITEM_MODEL_SETUP_INTERNAL_MODULE_RACING_MODE:
+      {
+        ModuleData & moduleData = g_model.moduleData[moduleIdx];
+        lcdDrawTextAlignedLeft(y, STR_RACING_MODE);
+        moduleState[moduleIdx].racingMode = editCheckBox(moduleState[moduleIdx].racingMode, MODEL_SETUP_2ND_COLUMN, y, "", attr, event);
```
bsongis: it won't be saved in the model storage, it means that on watchdog reset we will be in an unknown state

radio/src/gui/212x64/model_setup.cpp
```diff
@@ -1098,6 +1100,15 @@ void menuModelSetup(event_t event)
         break;
       }
 
+      case ITEM_MODEL_SETUP_INTERNAL_MODULE_RACING_MODE:
+      {
+        uint8_t moduleIdx = CURRENT_MODULE_EDITED(k);
+        ModuleData & moduleData = g_model.moduleData[moduleIdx];
+        lcdDrawTextAlignedLeft(y, STR_RACING_MODE);
+        moduleState[moduleIdx].racingMode = editCheckBox(moduleState[moduleIdx].racingMode, MODEL_SETUP_2ND_COLUMN, y, "", attr, event);
```
bsongis: same comment here, the variable should be persistent

radio/src/gui/gui_common.h
```diff
@@ -184,6 +184,26 @@ inline uint8_t MODULE_CHANNELS_ROWS(int moduleIdx)
   }
 }
 
+inline bool isRacingModeAllowed()
+{
+  if (isModulePXX2(INTERNAL_MODULE) && 0 == g_model.moduleData[INTERNAL_MODULE].channelsCount && moduleState[INTERNAL_MODULE].racingMode)
```
bsongis: expression can be simplified

radio/src/gui/gui_common.h
```diff
@@ -184,6 +184,26 @@ inline uint8_t MODULE_CHANNELS_ROWS(int moduleIdx)
...
+  else
+    return false;
+}
+
+inline uint8_t IF_ALLOW_RACING_MODE(int moduleIdx)
+{
+  if (!IS_MODULE_ENABLED(moduleIdx)) {
+    return HIDDEN_ROW;
+  }
+  else if (isModulePXX2(moduleIdx)) {
```
bsongis: `else if (isModulePXX2(moduleIdx) && 0 == g_model.moduleData[moduleIdx].channelsCount)`

radio/src/gui/gui_common.h
```diff
@@ -184,6 +184,26 @@ inline uint8_t MODULE_CHANNELS_ROWS(int moduleIdx)
...
+    if (0 == g_model.moduleData[moduleIdx].channelsCount)
+      return 1;
```
bsongis: return 1 means 2 fields, do we have 2 fields or only one? in this case it should be 0

radio/src/pulses/pulses.cpp
```diff
@@ -526,10 +526,11 @@ void setCustomFailsafe(uint8_t moduleIndex)
         g_model.failsafeChannels[ch] = channelOutputs[ch];
       }
     }
+    storageDirty(EE_MODEL);
```
bsongis, line 640: good catch

radio/src/pulses/pxx.h
```diff
@@ -29,7 +29,7 @@
 
 #define PXX2_LOWSPEED_BAUDRATE             230400
 #define PXX2_HIGHSPEED_BAUDRATE            450000
-#define PXX2_PERIOD                        4 // 4ms
+#define PXX2_PERIOD                        50 // 4ms
```
bsongis: why 50?

radio/src/pulses/pxx2.cpp
```diff
@@ -47,14 +47,20 @@ void Pxx2Pulses::addFlag1(uint8_t module)
   else {
     subType = g_model.moduleData[module].subType;
   }
-  Pxx2Transport::addByte(subType << 4);
+
+  uint8_t flag1 = subType << 4u;
+  if (isFunctionActive(FUNCTION_RACING_MODE)) {
```
bsongis: + test for "allowed"!

radio/src/translations/pt.h.txt
```diff
@@ -1244,7 +1245,7 @@
 #define TR_JITTER_FILTER       "ADC Filter"
 #define TR_RTC_CHECK           TR("Check RTC", "Check RTC voltage")
 #define TR_AUTH_FAILURE        "Auth-failure"
-
+#define TR_RACING_MODE         INDENT "Racing mode"
```
bsongis: don't introduce INDENT there, we try to remove them slowly

radio/src/gui/480x272/model_setup.cpp
```diff
@@ -1247,6 +1249,17 @@ bool menuModelSetup(event_t event)
         }
         break;
       }
+
+#if defined(HARDWARE_INTERNAL_MODULE)
+      case ITEM_MODEL_SETUP_INTERNAL_MODULE_RACING_MODE:
+      {
+        ModuleData & moduleData = g_model.moduleData[moduleIdx];
```
bsongis, line 1256: moduleData is not used anywhere!

bsongis:
The Companion part is missing

radio/src/pulses/pxx.h
WismyYao: You suggested me change it to 50ms . Racing mode is 4.6ms , we should set it more then 4.6ms

3djc: 50 doesn't look right at all to me. It is also used on external module btw

bsongis: right, I have pushed a proposal, it needs testing (internal module + heartbeat)

PR #8318, opened 2021-02-21
new LUA function disableTelemetryPopop

bug400:
On Taranis the KEY LONG ENTER event calls the telemetry popup menu which is not  
wanted in some use cases. Therefore, the LUA function "disableTelemetryPopup()"
was added that disables the call of the telemetry popup for the            
current execution of the telemetry script run function.

Thus, the KEY LONG ENTER event can be finally handled in the LUA script.

My use case is a telemetry script which realizes a lot of additional "software
switches" along with a mixer script. This is necessary for RC ship models with 
many special remote controlled features where the number of Taranis hardware
switches is insufficient.

The LUA function enables the implementation of pus buttons and prevent the call 
of the telemetry popup for switches if the ENTER key was pressed too long      
by mistake.

Generally it would be desirable to have full control in LUA key event handlers 
whether an event is finally handled by OpenTx or not.

### DIFF
--- a/radio/src/lua/api_general.cpp
+++ b/radio/src/lua/api_general.cpp
@@ -1916,6 +1916,24 @@ static int luaSerialRead(lua_State * L)
 
   return 1;
 }
+#if !defined(COLORLCD)
+/*luadoc
+@function disableTelemetryPopup()
+ 
+This function disables the Telemetry Popup Menu for the current key event.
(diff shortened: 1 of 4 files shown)


lshems:
Really nice to deactivate an indeed undesirable function for lots of people using Lua scripts.

lshems:
Why is merging blocked??

PR #8339, opened 2021-03-05
Companion curves text fix for MacOS

elecpower:
Attempt to fix issue 8053
~~Included commentary in diff to support change~~
Tested on Ubuntu 20.04
Need some help with Win and Mac build and test


elecpower:
Tested on Ubuntu 20.04 only as my Windows build is currently broken.
Also needs Mac tester once in Nightlies.

Issue #8347, opened 2021-03-08
Companion will crash and exit while editing Models Inputs when pressing Ctrl+A if no line is selected


wrdouglass:
**To Reproduce**
1. Select properties for any Model
2. Click on 'Inputs' tab
3. Select any row
4. Press Ctrl+A
5. Edit dialog will be displayed
6. Enter data as desired
7. Click OK
8. Press Ctrl+A (again)
9. Companion will hang briefly then exit, no error details are displayed, all changes are lost

**Expected behavior**
Step 8 should not cause Companion to crash and exit. It should ignore the Ctrl+A as there is nothing it can do in that scenario.

Note: Pressing Cancel in step 7 will also cause the crash in step 8

**General information**
 - Found in OpenTx Companion 2.3.11

wrdouglass:
I have code fixes for this and a couple similar issues in InputsPanel & MixesPanel but appears i don't have permission to push a branch for a pull-request.  Let me know how i can proceed here, i've looked around but i can't seem to locate any documentation on the steps for this project. Thanks.

3djc:
You can submit PR out of fork branches

kilrah:
https://github.com/opentx/opentx/wiki/OpenTX-Git-Development-Workflow

wrdouglass:
@kilrah thank you.

elecpower:
Closed by merging PR

PR #8349, opened 2021-03-08
Resolve issue 8347 by checking for nullptr.  

wrdouglass:
Additionally resolve a couple other similar issues for keyboard accelerators in inputs & mixers

### DIFF
--- a/companion/src/modeledit/inputs.cpp
+++ b/companion/src/modeledit/inputs.cpp
@@ -351,7 +351,8 @@ void InputsPanel::expoOpen(QListWidgetItem *item)
   if (!item)
     item = ExposlistWidget->currentItem();
 
-  if (item == nullptr) return;
+  if (item == nullptr) 
+    return;
 
   int idx = item->data(Qt::UserRole).toByteArray().at(0);
   if (idx < 0) {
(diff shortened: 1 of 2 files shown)


elecpower:
Fixes #8347 (link to Issue)

companion/src/modeledit/inputs.cpp
```diff
@@ -351,6 +351,8 @@ void InputsPanel::expoOpen(QListWidgetItem *item)
   if (!item)
     item = ExposlistWidget->currentItem();
 
+  if (item == nullptr) return;
```
elecpower: @wrdouglass thanks for the PR we appreciate the assistance. Could you please change this and similar changes to use the project's current standards of line 351 for nullptr test and if statement(s) on following lines to test.

wrdouglass: i'm not quite sure what you are asking

do you mean

if (item == nullptr)
  return;
  
 i wondered about that but i could see that there were varying standards in that file itself and went with what i was accustomed to, however i can certainly change that.

3djc:
This might come handy : https://github.com/opentx/opentx/wiki/OpenTX-Code-Style-Guide

And yes, code reformatting is still in progress :)

PR #8358, opened 2021-03-11
Resolve two sets of repeated warnings when built with VS2019

wrdouglass:
These warnings are clearly a minor nuisance but given i don't see any other warnings during the build i presume that the goal is to have zero warnings.  

The first in opentx.h(323,16): warning C5208 appears to be "new" to VS2019 and i resolved it by simply naming the struct based on the resulting typedef name (see: https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/c5208?view=msvc-160)

The second in layout4+2.cpp(68,32): warning C4805 is somewhat more interesting and its unclear to me why it doesn't like that comparison but allows coercion from bool to uint32_t (and vice-versa) in other places without any complaints - i saw there was a macro to cast the uint32_t value to bool so i wrapped all the offending compare operands with that macro. (see: https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-1-c4805?view=msvc-160)

### DIFF
--- a/radio/src/opentx.h
+++ b/radio/src/opentx.h
@@ -320,7 +320,7 @@ void memswap(void * a, void * b, uint8_t size);
 #define MASK_CFN_TYPE  uint64_t  // current max = 64 function switches
 #define MASK_FUNC_TYPE uint32_t  // current max = 32 functions
 
-typedef struct {
+typedef struct _CustomFunctionsContext {
   MASK_FUNC_TYPE activeFunctions;
   MASK_CFN_TYPE  activeSwitches;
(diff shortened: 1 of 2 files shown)


radio/src/opentx.h
```diff
@@ -320,7 +320,7 @@ void memswap(void * a, void * b, uint8_t size);
 #define MASK_CFN_TYPE  uint64_t  // current max = 64 function switches
 #define MASK_FUNC_TYPE uint32_t  // current max = 32 functions
 
-typedef struct {
+typedef struct _CustomFunctionsContext {
```
bsongis, line 323: why not:
```
struct CustomFunctionContext 
{
...
}
```

PR #8363, opened 2021-03-13
telemetry dir pin init missed definining OType

rotorman:
Added missing OType definition to telemetry direction pin.


raphaelcoeffic:
@rotorman is this linked to an issue you have experienced?

rotorman:
Lucikly no, this is just an imperfection fix (by reading the code and trying to understand how OpenTX works under the hood, I noticed that other GPIO pins have it explicitly defined, for this pin the output type was not listed).

PR #8368, opened 2021-03-14
Update of startup_stm32f42_43xx.s

rotorman:
STM32F429/439 are Cortex M4 processors (OpenTX version listed m3 and in general, the file seemed to stem from stm32f40_41 and not from stm32f42_43 as used in horus radios).

By comparing the file in OpenTX GitHub with the startup_stm32f429bitx.s file generated by the current STM32CubeIDE for STM32F429BI (as in TX16S, T16 & T18), I can see that the present OpenTX version misses an initial stack pointer setting.

Both issues fixed in this PR.

I attach for reference/comparison the startup_stm32f429bitx.s from ST's original STM32CubeIDE (in order to not have to ZIP-attach it I added an extension .txt to the file):
[startup_stm32f429bitx.s.txt](https://github.com/opentx/opentx/files/6136677/startup_stm32f429bitx.s.txt)

### DIFF
--- a/radio/src/targets/horus/startup_stm32f42_43xxx.s
+++ b/radio/src/targets/horus/startup_stm32f42_43xxx.s
@@ -1,16 +1,16 @@
 /**
   ******************************************************************************
-  * @file      startup_stm32f40_41xxx.s
+  * @file      startup_stm32f42_43xx.s
   * @author    MCD Application Team
   * @version   V1.3.0
   * @date      08-November-2013
-  * @brief     STM32F40xxx/41xxx Devices vector table for RIDE7 toolchain.          
+  * @brief     STM32F42xxx/43xxx Devices vector table for RIDE7 toolchain.          
   *            This module performs:
   *                - Set the initial SP
   *                - Set the initial PC == Reset_Handler,
   *                - Set the vector table entries with the exceptions ISR address
   *                - Configure the clock system and the external SRAM mounted on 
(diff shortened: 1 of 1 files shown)


rotorman:
Travis CI check fails not due to induced code changes in this PR, but due to apt-get failing on Travis. According to Travis CI log:

```
The following packages have unmet dependencies:

 lib32stdc++6 : Depends: lib32gcc-s1 (>= 9.3.0-21ubuntu1~16.04) but it is not installable

E: Unable to correct problems, you have held broken packages.
apt-get.diagnostics

apt-get install failed

The command "sudo -E apt-get -yq --no-install-suggests --no-install-recommends $(travis_apt_get_options) install curl libmpfr4 libmpc3 libfox-1.6-dev libgtest-dev lib32stdc++6 libclang-common-6.0-dev clang-6.0 python3-pip" failed and exited with 100 during .


```

raphaelcoeffic:
Interesting finding. As far I know, SP is set in the bootloader, but you're right, it is probably better to have things clean here as well. @bsongis @kilrah this is probably something that should be checked.

radio/src/targets/horus/startup_stm32f42_43xxx.s
```diff
@@ -1,16 +1,16 @@
 /**
   ******************************************************************************
-  * @file      startup_stm32f40_41xxx.s
+  * @file      startup_stm32f42_43xx.s
```
bsongis, line 3: xxx :)

Issue #8241, opened 2021-01-23
Anomaly in the FRSKY telemetry output stream on TX16S


b14ckyy:
There is some kind of anomaly in the Telemetry output of the Radiomaster TX16S that makes at least one known device as a receiver unusable. 

Setup: 

- Radiomaster TX16S MAX
- Upgraded with Horus X10S Bluetooth LE Module
- Internal R9ACCESSMOD installed
- Genuine DSD Tech HM-10 BTLE Module paired with the radio BT
- OpenTX 2.3.11 with Bluetooth and INTERNAL R9 ACCESS Mod enabled (also tried 2.3.12NB421 with same issue)
- Antenna Tracker with U360GTS Universal Telemetry Controller

Before the TX16S I used a Taranis X9D with an TTL Inverter and a Pair of HC-05 BT modules on 57k baud to mirror telemetry to the tracker and that worked fine with 2.4GHz and R9 System. 

Now after switching to the TX16S the same chain was built just with HM-10 and Horus BT module. But now the following issue appears: The U360GTS Protocol detection, recognizes a FRSKY_X Telemetry stream but does not show any data. All packages are disregarded as corrupt. I have tested many configurations and it was impossibly to get the tracker reading the telemetry from the radio. The Android FrSky Telemetry Viewer app on the other hand works fine in multiple chains (see test setups below). But the telemetry transmission from a Crossfire Module over BT works just fine and correctly. 

1. Take any TX16S (maybe BT mod is not even needed, Telemetry mirror to UART works too)
2. Bind a plane with FrSky 2.4GHz D16 receiver or use R9 System
3. Enable telemetry mirror on UART or BT
4. Flash U360GTS firmwareTo a NAZE32 or SPRF3 Board https://github.com/raul-ortega/u360gts-configurator
5. Connect OLED screen over SPI and enable Protocol auto detect and provider INAV (optional, add buttons to use the OLED menu)
6. Bind HM-10 to the Horus module and connect it to the tracker
7. Power all up .... 
Now the Tracker will detect FRSKY_X Protocol but all data stay 000 

I have tested all the following link chains to rule out as much as possible: 

##################
_I always used a plane with either R-XSR receiver (Multi Module) or R9mm receiver. Both always show same result.
X9D - Serial Out - HC-05 baud 57k ..... HC-05 57k - tracker               | fine
TX16S - internal BT on Mirror mode ...... HM 10 115k/57k - tracker     | problem
[...]

Issue #7798, opened 2020-07-06
Different output on telemetry Mirror with f.port addition: AND S.PORT configuration after 2.2.3 


brainbubblersbest:
Hello, 
I dont know if there is a Bug or already documented. 
But I dont found Information about this. 
Even in the official f. Port Documentation. 
If I use f.Port, and Trace the data stream over a bt connection to a device that need gps Position and altitude over S.port, it cant work with this data. 
Data is received and inverted. This is all checked with a two channel scope. And counterchecked with normal s.port receiver firmware. 

Is there any Information available that f.port affect the data stream on bt mirror? 
And in which way it does? 
I thought Till today. F.port affect only the communication between RX and FC. 
Hardware:
X12s
Hm-10
S.port inverter to uninvert the signal
Mfd crossbow Mini AAT

kilrah:
We are not aware of any difference on our side.

brainbubblersbest:
I will try to make a Video of the two setups.
The Baudrate of the data came out of the HM-10  is not different in any way?
I use 119200 for HM-10 to internal BT
And 57600 for a connection between to hc-05 modules. 
But here i use the sport connector on r9m. 
This is btw only the configured Speed in the BT Modules .
But is that baudrate also correct for the data that goes trough this BT Connection? "used at BT Mirror"?

brainbubblersbest:
Now im totally confused. 😅
I need uninverted s.port on the tracker. 
First picture Shows uninverted s.port at top after uninverter, and s.port came out of the hm-10. Nice and symetrical.

Second picture Shows uninverted s.port at top after uninverter (with other time setting) and s.port measured at the back of the r9m.
Maybe not symetrical in relation to the transmition lag?
Or related to wrong baudrate setting? 

But the s.port at back of r9m has the same Signal direction? Wtf? Any Suggestions?

This is same behavior with s.port and f.port setup. 🤔
![IMG_20200707_190306](https://user-images.githubusercontent.com/48190425/86820141-f105dd00-c088-11ea-9a53-966f30199afe.jpg)
![IMG_20200707_191920](https://user-images.githubusercontent.com/48190425/86820148-f2370a00-c088-11ea-83f4-7f1d2056d29b.jpg)

brainbubblersbest:
![IMG_20200707_190306](https://user-images.githubusercontent.com/48190425/86820001-b8fe9a00-c088-11ea-8451-ded0a523c34c.jpg)
[...]

PR #8386, opened 2021-03-19
STM32 WordLength 9b -> 8b data + parity

rotorman:
Fixed comment. STM32 WordLength 9b is here necessary due to parity bit (8b data + 1b parity) and not due to 2 stop bits.

### DIFF
--- a/radio/src/targets/common/arm/stm32/aux_serial_driver.cpp
+++ b/radio/src/targets/common/arm/stm32/aux_serial_driver.cpp
@@ -124,7 +124,7 @@ void auxSerialInit(unsigned int mode, unsigned int protocol)
       break;
 
     case UART_MODE_SBUS_TRAINER:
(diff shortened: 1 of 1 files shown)


raphaelcoeffic:
@rotorman I think we should concentrate our efforts on 2.4 now, and you would be very much welcome :-)

rotorman:
I understand that you prefer the Issues & PRs in 2.4 and not so much more in 2.3. Possibly, this PR/comment could nevertheless be merged, as large parts of code in v2.4 stem from v2.3 (at least that is what my diff says).

But, I took your suggestion to heart. My first baby steps with getting to know v2.4 resulted in two issues, both I listed in #8417.

brainbubblersbest:
@rotorman just for clarification, has this issue anything to to with #8241 or #7798?

Or with the fix that needs to merged into the u-360-gts project?.https://github.com/raul-ortega/u360gts/issues/91#issue-638041921

Because All these issues then, can be closed officially. 😊

rotorman:
@brainbubblersbest This PR here fixes just the comment and stems from a STM32 and StdPeriphLib peculiarity (parity bit counted to data length).
I do not believe it has anything to do with the other issues you linked.

PR #8391, opened 2021-03-24
Fix building with TRACE_SD_CARD CMake option enabled

rotorman:


3djc:
We do prefer #if !defined(BOOT) but I'll fix it myself

rotorman:
Thanks! I'll keep your preference in mind for future! Cheers!

PR #8405, opened 2021-04-01
Add space to SD edit name

### LINKED ISSUE #8402: Rename issue SD CARD menu on X9D+ unable to shorten names of files and folders
I have discovered that in the SD CARD menu it is not possible to rename files and folders to a shorter name than the one it already has. The space character does not appear when browsing through the characters so it is not possible to go from name123 to name12. It is possible to make it longer, but not shorter.
It would be very nice if this would work for those situations that you don't have a pc at hand while out flying or for those who do everything from the transmitter itself.

I enter through the character positions with ENT and use + and - keys to go through the available characters, but space (' ') is missing.

 - OpenTX version 2.3.11 (official release version)
 - Frsky X9D+

It was confirmed by Miami Mike on rcgroups. Risto confirms it does work on his TX16S, so it might just be a B&W radio issue.

Thank you.

dgatf:
Fixes #8402. Only affects B&W radios

This PR allows up to the space character in SD rename file for B&W radios. This allows to shorten the file name.

### DIFF
--- a/radio/src/gui/common/stdlcd/draw_functions.cpp
+++ b/radio/src/gui/common/stdlcd/draw_functions.cpp
@@ -145,7 +145,7 @@ void editName(coord_t x, coord_t y, char * name, uint8_t size, event_t event, ui
           if (c <= 0) v = -v;
         }
         else {
-          v = checkIncDec(event, abs(v), '0', 'z', 0);
+          v = checkIncDec(event, abs(v), ' ', 'z', 0);
         }
       }
 


alm-nl:
I have tested this fix on my X9D+ and it works perfectly! I used the sourcecode of 2.3.11 on January 6th, 2021 (few days before the release) and applied the fix to the mentioned file, compiled it (using https://github.com/vitas/opentx-docker-build) and it does exactly what it needs to do. 😎👍👍

PR #8415, opened 2021-04-04
2.4 Pot-pourri

raphaelcoeffic:
This allows no to actually delete a model (with confirmation dialog).

@3djc in case this one is not very pretty, please correct as you like ;-)


raphaelcoeffic:
Added basic screen setup.

PR #8421, opened 2021-04-05
Add synced pulses to mainstream OTX (yeah baby!)

3djc:


radio/src/mixer_scheduler.h
```diff
@@ -0,0 +1,84 @@
...
+ * GNU General Public License for more details.
+ */
+
+#ifndef _MIXER_SCHEDULER_H_
+#define _MIXER_SCHEDULER_H_
+
+#define MIXER_SCHEDULER_DEFAULT_PERIOD_US 4000u // 4ms
+
+#define MIN_REFRESH_RATE      1750 /* us */
+#define MAX_REFRESH_RATE     25000 /* us */
```
AlessandroAU, line 27: Can this be increased to say 70,000? to allow 15hz GHST mode and 25hz ELRS mode.

3djc: Discussed it with @raphaelcoeffic, and we are both in agreement to keep the value as is and suggest you ask for double your period, in order to keep frequent enough mixer run

PR #8427, opened 2021-04-05
Bluetooth trainer byte parsing: catch XOR state errors

rotorman:
In XOR state only 0x7E ^ 0x20 = 0x5E or 0x7D ^ 0x20 = 0x5D are expected.
This PR handles the other cases appropriately.


bsongis:
I will check carefully this PR when I will add the feature to Ethos

bsongis:
Yes this one is good, it should be merged into 2.3

PR #8438, opened 2021-04-09
Multi module config support

3djc:


raphaelcoeffic:
@3djc @pascallanger We need this in 2.4 as well.

PR #8444, opened 2021-04-12
Add basic support for Github Actions

raphaelcoeffic:
This is based on the [opentx-commit-tests docker image](https://github.com/raphaelcoeffic/build-opentx).

Please that this is only the start, it is still missing some stuff:
- adding other targets
- use matrix strategy instead of multiple steps (see [here](https://docs.github.com/en/actions/learn-github-actions/managing-complex-workflows#using-a-build-matrix))

Further useful informations on Travis conversion can be read [here](https://docs.github.com/en/actions/learn-github-actions/migrating-from-travis-ci-to-github-actions).

**Please note:** it seems that GH Actions, the way it's configured right now, does the tests on the **merged version of the MR**, and **not on the MR's HEAD commit** (see [this github issue](https://github.com/actions/checkout/issues/15#issuecomment-524093065) for more details).

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -1,0 +1,23 @@
+name: OpenTX Commit Tests
+on:
+  push:
+    branches: [ 2.4 ]
+  pull_request:
+    branches: [ 2.4 ]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    container:
+      image: ghcr.io/raphaelcoeffic/opentx-commit-tests
+      volumes:
+        - ${{ github.workspace }}:/src
+    steps:
+      -
+        name: Check out the repo
+        uses: actions/checkout@v2
+      -
+        name: Test COLORLCD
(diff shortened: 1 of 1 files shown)


raphaelcoeffic:
For example builds, see https://github.com/raphaelcoeffic/opentx/actions

raphaelcoeffic:
Another nice example of why we should move to Github Actions: https://travis-ci.org/github/opentx/opentx/jobs/766986668

bsongis:
> Another nice example of why we should move to Github Actions: https://travis-ci.org/github/opentx/opentx/jobs/766986668

I am not sure you have experience with Github Actions, it's not perfect as well, and timeouts do occur from time to time

raphaelcoeffic:
> I am not sure you have experience with Github Actions, it's not perfect as well, and timeouts do occur from time to time

Thx! We will see! For now, both are run in //, and the Actions always won while I was trying.

PR #8447, opened 2021-04-14
2.4: Use the same font all the way for a specific language

raphaelcoeffic:
This is more complicated as it seems. So that we get consistent sizes across fonts and languages, we will have to have a look at this: https://stackoverflow.com/questions/43060479/how-to-get-the-font-pixel-height-using-pils-imagefont-class

And try to compute what we need to give PIL.ImageFont the proper pixel size.

TODO:
- [ ] find a way to have a consistent way to generate font bitmaps with consistent size across fonts & sizes.

### DIFF
--- a/cmake/Bitmaps.cmake
+++ b/cmake/Bitmaps.cmake
@@ -34,14 +34,12 @@ macro(add_truetype_font_target name size subset effect)
   set(target font_${name})
   if(${effect} STREQUAL "bold")
     set(font ${FONT_BOLD})
-    set(cjk_font ${CJK_FONT_BOLD})
   else()
     set(font ${FONT})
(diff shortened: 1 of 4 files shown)


3djc:
2.4 is ok
image

This (broken)
image

raphaelcoeffic:
@3djc thanks for the test, I'll fix this ASAP.

raphaelcoeffic:
Better now?

Screenshot 2021-04-14 at 09 15 47

Screenshot 2021-04-14 at 09 17 13

The main issue here is that we apparantly do not adapt to the size of the controls to the size of the fonts, which is bad, in fact.

raphaelcoeffic:
This has been merged into EdgeTX already.

PR #8459, opened 2021-04-19
Update ModuleSyncStatus::update signature with proper type

CapnBry:
The function prototype for `ModuleSyncStatus::update()` takes the newInputLag parameter as unsigned, when all Multiprotocol, CRSF, and Ghost pass a signed value. The [member variable](https://github.com/opentx/opentx/blob/2.3/radio/src/telemetry/telemetry.h#L298) update() puts this value in is also signed, as it should be because the offset can be negative.

This is pedantic cleanup because I not believe it is causing any issues. I've not seen any sync delayed by (uint16_t)(-1) us if our value goes negative (I work on ExpressLRS). I do think it is important enough to match the signedness through the entire call chain though.

It also updates getRefreshString to use the proper signed types, and display correctly scaled values, since the refreshRate is in us and the existing code appears to think it is in ns?

### DIFF
--- a/radio/src/telemetry/telemetry.cpp
+++ b/radio/src/telemetry/telemetry.cpp
@@ -398,7 +398,7 @@ ModuleSyncStatus::ModuleSyncStatus()
   memset(this, 0, sizeof(ModuleSyncStatus));
 }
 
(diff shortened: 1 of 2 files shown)


raphaelcoeffic (approved):
Thanks! It looks ok to me.

PR #8469, opened 2021-04-24
Stop Companion crash when conversion switch index out of range

### LINKED ISSUE #8403: Companion conversion GUI Close
Companion force closes the GUI when opening an .otx model file and converting, this worked on a previous nightly (e.g N460 and earlier) in this case from a Jumper T16 to a T-Lite

Hardware Jumper T-Lite
latest nightly 2.3.12
Windows 10

### LINKED ISSUE #8466: Unexpected Termination of Companion on Converting an OTX File
I am using Companion 2.3.11 with several Radio profiles, i.e. Taranis X9D+ 2019, Taranis X7S ACCESS and Horus X10S Express. As I have a number of models that I fly with different Radios, I do regular conversions of the otx files for comparison, harmonisation, etc. In this context, I noticed the following erroneous behaviour.

If I work in the profile for the Taranis X7S ACCESS, and then open a Taranis X9D+ 2019 otx file and confirm the file conversion, Companion terminates withour any pre-warning and without saving anything. This problem does not occur if I open a Horus X10S Express otx file in the X7S profile.
For testing I have generated a Taranis X9D+ 2019 otx file with only one simple model in it, and the problem still occurs. So it does not seem to be related to any specific model setup.

elecpower:
Fixes #8466 
Fixes #8403
Conversion unable to identifying the correct sub board type from FourCC that have differing hardware capabilities. eg X9D+ 2019 has one more switch than other X9D boards.

### DIFF
--- a/companion/src/firmwares/rawswitch.cpp
+++ b/companion/src/firmwares/rawswitch.cpp
@@ -191,7 +191,10 @@ RawSwitch RawSwitch::convert(RadioDataConversionState & cstate)
     QStringList fromSwitchList(getSwitchList(cstate.fromBoard));
     QStringList toSwitchList(getSwitchList(cstate.toBoard));
     // set to -1 if no match found
-    newIdx = toSwitchList.indexOf(fromSwitchList.at(swtch.quot));
+    if (swtch.quot < fromSwitchList.count())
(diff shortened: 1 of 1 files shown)


3djc:
Great, Thanks !

PR #8476, opened 2021-04-28
Make SD card sync direction combo box consistent with folder labels

### LINKED ISSUE #8319: Improve SD-card synchronisation texts.
Hello, 

Thanks for the great work on Opentx! 

I have noticed some of my fellow RC-modelers getting confused by the dialog that comes up when synchronizing the SD and like to propose an improvement in Companion.

Its says 'Local folder' and 'Radio folder' in the folder selections. This is fine. Maybe 'Companion' instead of 'Local' would be even better, but that's just a matter of opinion.
But then in the 'Sync direction' drop down it talks about: 'Source folder' and 'Destination folder'. This is confusing, it is not obvious what device is the destination and what is the source. The terms in this drop-down should correspond to names of the selection boxes directly above it.  So something like 'Both directions, Local to Radio folder first' and "Only from Local folder to Radio folder.

Keep up the good work!

OS: Linux, Windows.
Companion version 2.3.11.

arunh:
Fixes opentx/opentx#8319

### DIFF
--- a/companion/src/dialogs/filesyncdialog.cpp
+++ b/companion/src/dialogs/filesyncdialog.cpp
@@ -143,10 +143,10 @@ void FileSyncDialog::setupUi()
   QWidget * wdgDst = folderSelectorWidget(leDst, this);
 
   AutoComboBox * syncDir = new AutoComboBox(this);
-  syncDir->addItem(tr("%1%2 Both directions, to destination folder first").arg(dstArw, srcArw), SyncProcess::SYNC_A2B_B2A);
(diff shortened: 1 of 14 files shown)


elecpower:
@arunh I agree the current wording is not ideal and thank you for going to the effort to submit this PR.
However FileSyncDialog is implemented as a generic folder syncing class so I cannot support fixing the way you have proposed.
That is why in this use case the source and destination field labels are set in the MainWindow::sdsync().
If you have no objection I'll work on another variation over the next week all going well that maintains the generic nature.

arunh:
@elecpower no problem

PR #8480, opened 2021-05-01
Move adcRead() to mixer scheduler IRQ to reduce sample jitter

AlessandroAU:
Currently the callback stack for reading the ADCs looks like this. doMixerCalculations() -> getADC() -> adcRead(). 

The function doMixerCalculations() triggered by releasing a mixerMutex lock. While the lock itself is released precisely by the hardware timer IRQ there is a variable jitter in when the adcRead() task gets around to actually sampling the sticks depending on what the RTOS is doing. 

The adcRead() can be moved into the mixer scheduler IRQ ensuring that the samples take more precisely and stored to the global array, afterwards the variable jitter in when the doMixerCalculations() actually runs is not relevant as long as the data is processed within the frame window. 

Minimising sampling jitter is very important for Betaflight and particularly high refresh rate links, some initial testing on the bench has shown that this can significantly the smoothness of feedforward traces. 

image

(left old, right new)

### DIFF
--- a/radio/src/opentx.cpp
+++ b/radio/src/opentx.cpp
@@ -1101,10 +1101,6 @@ void getADC()
   }
 #endif
 
-  DEBUG_TIMER_START(debugTimerAdcRead);
-  adcRead();
-  DEBUG_TIMER_STOP(debugTimerAdcRead);
-
   for (uint8_t x=0; x<NUM_ANALOGS; x++) {
     uint16_t v;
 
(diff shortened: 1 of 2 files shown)


raphaelcoeffic (approved):

ctzsnooze:
That's awesome!  Every little bit of jitter reduction pays huge dividends down the track!

PR #8482, opened 2021-05-02
Introduce YAML to Companion 2.4

elecpower:
Add a YAML format stub to Companion.
The YAML format needs to be able to read and write to the radio sd card plus import and export computer yml files. These computer files could be incomplete eg just a single model without general settings.
May need to develop a new interface class that is radio agnostic for constructing and reading the byte arrays. At the same time the interface needs to know the source and destination radio firmware type and version for conversions and upgrades.
Probably best to wait until the radio side is developed as some of these challenges are common.
Also Companion manipulates its generic data structures and byte array data with after import and before export functions.


elecpower:
Moved to EdgeTX

Issue #7221, opened 2019-12-20
[OpenTX V2.3.8] - Wrong signature warning on X12S when try to flash MPM


robustini:
Could you check if this is indeed the case?
On X10S Express and Jumper T16 no problem, when I try on X12S I have this error.

3djc:
Could you share the file you are trying to flash ? Fairly strange as the code should be the same

robustini:
For the multimodule or the tx?
If you mean the MPM firmware I flash the same file in the others tx and work fine.

robustini:
[multi-stm-1.3.0.52_EXT.zip](https://github.com/opentx/opentx/files/3989969/multi-stm-1.3.0.52_EXT.zip)

3djc:
For the multi, it will make it easier to replicate. Also is your x12 using internalaccess ?

robustini:
Yes, internalaccess enabled.

3djc:
Ok i will work on it tomorrow if I can replicate on simu. If not, I’ll need to find someone with x12

3djc:
OK, I cannot reproduce, but the wrong signature message seems to indicate the radio does not like the info comming back from the multi module bootloader. Does flashing this very module on another radio work ?

robustini:
Yes, with other radio work, with the same module.
I removed that pin from the ISRM module for Multi-Trainer but I don't think that's the problem.
![isrm_pin](https://user-images.githubusercontent.com/3864712/71305960-0c561700-23db-11ea-8054-d1f93425ac5b.jpg)

3djc:
@raphaelcoeffic your thoughs ?

robustini:
If the X12S uses this pin to flash the MPM then I've cut out the possibility of flashing MPM with OpenTX, but on the X10S Express I did the same thing and the flashing works.
But it seems to me somewhat strange that this is so.

robustini:
https://youtu.be/6mlW7paMfAA

3djc:
Yes confirm it is a issue while talking to the module in BL mode

robustini:
Hardware or software issue?

3djc:
Don’t know yet, someone with the right hardware is going to have to check that

gvt70:
Just updated to 2.3.4 and tried updating MPM with  
[...]

PR #8487, opened 2021-05-03
Do not use DMA on x12s for multi firmware update

### LINKED ISSUE #8477: "Wrong signature" when flashing ELRS from X12S
When trying to flash ExpressLRS on a R9M module from a Horus X12S, I get a "wrong signature" error.

The .frk bootloader flashes fine.
The error happens during the next step, flashing the .elrs firmware using the "Flash External ELRS" action.

Maybe linked to #7221 ?

 - OpenTX version 2.3.12N487
 - Frsky X12S (with inverter mod and internal ISRM)
 - R9M ACCST (with resistor mod)
 - ELRS 1.0.0 RC2

trigger7:
Fixes #8477 (and possibly #7221)

Not enabling DMA seems to allow the x12s to flash the external module correctly.

Verified with R9M ACCST, flashed with ExpressLRS.

I patched both classes (MultiExternalUpdateDriver and MultiExtSportUpdateDriver) as the issue happens with ELRS and MPM, but I haven't verified flashing MPM as I do not have such module.

### DIFF
--- a/radio/src/io/multi_firmware_update.cpp
+++ b/radio/src/io/multi_firmware_update.cpp
@@ -118,7 +118,11 @@ class MultiExternalUpdateDriver: public MultiFirmwareUpdateDriver
       if (inverted)
         telemetryPortInvertedInit(57600);
       else
+#if defined(PCBX12S)
+        telemetryPortInit(57600, TELEMETRY_SERIAL_WITHOUT_DMA);
+#else
         telemetryPortInit(57600, TELEMETRY_SERIAL_DEFAULT);
+#endif
     }
 
     bool getByte(uint8_t & byte) const override
@@ -164,7 +168,11 @@ class MultiExtSportUpdateDriver: public MultiFirmwareUpdateDriver
 
     void init(bool inverted) const override
     {
+#if defined(PCBX12S)
+      telemetryPortInit(57600, TELEMETRY_SERIAL_WITHOUT_DMA);
+#else
       telemetryPortInit(57600, TELEMETRY_SERIAL_DEFAULT);
+#endif
     }
 
     bool getByte(uint8_t & byte) const override


raphaelcoeffic:
@trigger7 ~~do you have MPM at hand to test the update on X12S?~~ Ok, I'll test then this weekend.

3djc:
I think we should understand why DMA isn't working on x12s rather than bindly disabling it. Maybe it is the right fix, maybe not

robustini:
Yes, it fix for sure #7221, tested now, awesome, thanks!

robustini:
Ehm, i'm sorry, it corrects the problem at 50%, since after the flashing the module does not communicate with the tx, it must be disabled and re-enabled, so we might as well disable it before flashing.

raphaelcoeffic:
> I think we should understand why DMA isn't working on x12s rather than bindly disabling it. Maybe it is the right fix, maybe not

Right, but in the meantime? I’ll try to have a look this weekend when I get hold of my friend’s X12S. Or maybe you have the hardware to research it?

trigger7:
> I think we should understand why DMA isn't working on x12s rather than bindly disabling it. Maybe it is the right fix, maybe not

True, but I'm afraid I reached my limits there.

I first tried to see why the x12s was reading a wrong signature, and it seems it sometimes reads unexpected "extra" bytes on the fifo (STK_OK and STK_INSYNC from what I saw). Couldn't find where these are coming from, nor why they aren't cleared by the call to telemetryClearFifo.

Then, basically what I did is hunt the differences between x12s and other radios, saw that DMA thing, tried disabling it during multi firmware update and it worked...

> Ehm, i'm sorry, it corrects the problem at 50%, since after the flashing the module does not communicate with the tx, it must be disabled and re-enabled, so we might as well disable it before flashing.

Indeed,I did flash my R9M with the module disabled, and manually enabled it afterward.

robustini:
> Indeed,I did flash my R9M with the module disabled, and manually enabled it afterward.

So it doesn't solve my issue, but thanks anyway!

3djc:
You did great trigger7, one of us needs to have a look and understand why in this particuliar radio, DMA is behaving this way. There might be a very valid reason for that, or an error somewhere else and this is only the symptom

3djc:
Thanks for the investigation and proposal @trigger7 , I have implemented it in a simpler way (no defined needed)

robustini:
> I have implemented it in a simpler way (no defined needed)

I don't see it available yet in the code, do you still have to pull it?

robustini:
> 
> 
> Thanks for the investigation and proposal @trigger7 , I have implemented it in a simpler way (no defined needed)

I tried and partially solves it, in the sense that you can write MPM but then the module must be turned off and on again or its status varies from displaying the firmware version to no serial input.
Can't you put an automatic reset of the MPM after the flash?

robustini:
No, perhaps the problem lies elsewhere.
After flashing the module and with the latest firmware compiled by the master I have this result on the X12S.

https://youtu.be/I5TbUxCmeos

robustini:
Same problem with X10S Express, damn.

robustini:
@3djc I tried with an OpenTX firmware compiled at the end of March and on both X12S and X10S Express the problem is gone, always with the latest MPM firmware.
Is there something broken on the master?
Here now MPM is unusable with ISRM enabled, whereas before it worked without problems.
Do I open another issue for this thing?
However, I confirm that your [commit](https://github.com/opentx/opentx/commit/af17efaddaaa75bedddf62113ef2ef2c18955407) solves this issue of flashing on X12S.

robustini:
@3djc i found the damn commit: https://github.com/opentx/opentx/commit/81c7338ab32e37f9c607e1df24f7132812d44218
What the hell is this damn racing mode? ;-)
Is there any way to get rid of it when compiling or in any case disable it via the menu?
@bsongis please if possible solve the problem, thank you!

PR #8494, opened 2021-05-10
new LUA function getAvailableMemory

Lucas1188:
added LUA function to get radio available memory remaining for display in telemetry/widgets scripts or for LUA staging to save memory

### DIFF
--- a/radio/src/lua/api_general.cpp
+++ b/radio/src/lua/api_general.cpp
@@ -1535,6 +1535,7 @@ static int luaGetUsage(lua_State * L)
   lua_pushinteger(L, instructionsPercent);
   return 1;
 }
+
 /*luadoc
 @function getAvailableMemory()
 
@@ -1549,6 +1550,7 @@ static int luaGetAvailableMemory(lua_State * L)
   lua_pushunsigned(L, availableMemory());
   return 1;
 }
+
 /*luadoc
 @function resetGlobalTimer([type])
 
--- a/radio/src/lua/api_general.cpp
+++ b/radio/src/lua/api_general.cpp
@@ -1535,7 +1535,20 @@ static int luaGetUsage(lua_State * L)
   lua_pushinteger(L, instructionsPercent);
   return 1;
 }
+/*luadoc
+@function getMemory()
 
+Get available memory remaining in the Heap for Lua.
+
+@retval usage (number) a value returned in b
+
+@status current Introduced in 2.4
+*/
+static int luaGetMemory(lua_State * L)
+{
+  lua_pushinteger(L, availableMemory());
+  return 1;
+}
 /*luadoc
 @function resetGlobalTimer([type])
 
@@ -1763,6 +1776,7 @@ const luaL_Reg opentxLib[] = {
   { "chdir", luaChdir },
   { "loadScript", luaLoadScript },
   { "getUsage", luaGetUsage },
+  { "getMemory", luaGetMemory },
   { "resetGlobalTimer", luaResetGlobalTimer },
 #if LCD_DEPTH > 1 && !defined(COLORLCD)
   { "GREY", luaGrey },
--- a/radio/src/lua/api_general.cpp
+++ b/radio/src/lua/api_general.cpp
@@ -1546,7 +1546,7 @@ Get available memory remaining in the Heap for Lua.
 */
 static int luaGetMemory(lua_State * L)
 {
-  lua_pushinteger(L, availableMemory());
+  lua_pushunsigned(L, availableMemory());
   return 1;
 }
 /*luadoc
--- a/radio/src/lua/api_general.cpp
+++ b/radio/src/lua/api_general.cpp
@@ -1536,7 +1536,7 @@ static int luaGetUsage(lua_State * L)
   return 1;
 }
 /*luadoc
-@function getMemory()
+@function getAvailableMemory()
 
 Get available memory remaining in the Heap for Lua.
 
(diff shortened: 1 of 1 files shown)


lshems:
collectGarbage is doing that already???

Lucas1188:
from my knowledge, the collectgarbage command could only return the mem used by lua in the lua env but not memory remaining in the heap on the radio (Lua + OTX), unless im missing something in the implementation of lua in opentx itself.

radio/src/lua/api_general.cpp
```diff
@@ -1535,7 +1535,20 @@ static int luaGetUsage(lua_State * L)
   lua_pushinteger(L, instructionsPercent);
   return 1;
 }
+/*luadoc
+@function getMemory()
```
3djc: I would name it getAvailableMemory()

radio/src/lua/api_general.cpp
```diff
@@ -1535,7 +1535,20 @@ static int luaGetUsage(lua_State * L)
   lua_pushinteger(L, instructionsPercent);
   return 1;
 }
+/*luadoc
```
3djc, line 1539: CR before

radio/src/lua/api_general.cpp
```diff
@@ -1535,7 +1535,20 @@ static int luaGetUsage(lua_State * L)
...
 
+@retval usage (number) a value returned in b
+
+@status current Introduced in 2.4
+*/
+static int luaGetMemory(lua_State * L)
+{
+  lua_pushunsigned(L, availableMemory());
+  return 1;
+}
```
3djc, line 1552: CR after

lshems:
Could you explain its intended use?just confused.

lshems:
> from my knowledge, the collectgarbage command could only return the mem used by lua in the lua env but not memory remaining in the heap on the radio (Lua + OTX), unless im missing something in the implementation of lua in opentx itself.

But what use is it to knwow that? Isn't the available memory in the LUA environment the only thing that counts??

How are they interconnected?

I am just interested to know what you could do with that information. I'm always struggling with memory, so perhaps I missed something in my comprehension of things??

radio/src/lua/api_general.cpp
Lucas1188: > Could you explain its intended use?just confused.
Its just a hookup for LUA to the same function call by OTX that shows the bytes remaining on the debug statistics page, that way, any LUA scripts created by users can take into account mem remaining on the radio (heap) to prevent itself from being killed (by staging scripts) or for debugging without needed to exit the telemetry page (If running custom Telemetry Screen).

Lucas1188:
Could you explain its intended use?just confused.
Its just a hookup for LUA to the same function call by OTX that shows the bytes remaining on the debug statistics page, that way, any LUA scripts created by users can take into account mem remaining on the radio (heap) to prevent itself from being killed (by staging scripts i.e unload and loadscript) or for debugging without needed to exit the telemetry page (If running custom Telemetry Screen).

radio/src/lua/api_general.cpp
Lucas1188: do I take away the luadoc portion until it is approved?

3djc: No no, it's fine. Since it is purely optional and informative (ie not affecting existing scripts), you could do a 2.3.x version if this is usefull to you

lshems:
I was always told that there was no way that a big OTX model would impact the amount of memory available for LUA. But then how exactly does the LUA environment claim or release memory from the heap?

I am still really confused. I have no clue what actions I could take using this information during development, or even during execution of code.

No misunderstanding here, I am not questioning the pull request.
I am just curious how I could use this information in a sensible way.

Lucas1188:
Lua in otx from looking at the src defines it's mem allocation from OTX under the field LUA_MEM_AVAIL. On certain radios, it's defined as a fixed amount and on others (i.e Taranis) it's defined as 0 (unlimited) meaning it gets what it gets. Because of this fluid nature, it is not possible to have accurate feedback of mem remaining on the radio inside a Lua script by using collectgarbage unless you already know the size of memory allocated (i.e 60000b -collectgarbage("count"). OTX knows it's remaining heap through _heap_end and _heap fields defined by the drivers (this drives the debug statistic page mem remaining): this includes overheads transparent to collectgarbage (luac radio compilation etc...).

So this pull just exposes that same number from OTX to LUA so that the LUA scripts can now do its own mem allocation of sorts to prevent an overflow 

for example: if your script is storing a buffer of data from telemetry/switch states/inputs (for speed) etc into a Lua table and the mem available reaches a defined threshold, it starts to write it to the SD card with io and clear the Lua table by nil and collectgarbage // you are going to load a script as a chunk with a know file size but there's not enough mem available, you could throw an err or unload something else first.

This is just my understanding from the limitations I have observed from the mem limited stm32f2x on my old Taranis x9D, I would be happy if someone would correct my misunderstanding in anyway, I'm still new to the otx environment.

lshems:
Thanks a million.

So that means than on a Taranis, the lesser memory used by a smaller model (just the default 4 mixes for instance) will allow for more memory to be allocated to the LUA environment than when a large model is loaded?

Because in that case I understand the use of it. But I was always told that the model size did not have an impact on the available LUA size.

?!

Lucas1188:
Model size from what I've seen doesn't affect Lua avail mem (seems to be allocated by otx on init during runtime: anyone more experienced in this please correct me if I'm wrong) but loading text scripts for the first time before the radio compiles to binary has one huge impact in terms of file size and compiler overhead, as well as user scripted Lua runtime allocation of memory (both of which at least I hope will now be trackable and more manageable from the user Lua script side).

Edit: at least the logic is Lua will "overflow" at 0b available I don't know if there's a buffer put in place by otx.

radio/src/lua/api_general.cpp
Lucas1188: Should I rebase this to 2.3 branch?

Lucas1188:
moved to #8496 to rebase to 2.3.x

PR #8498, opened 2021-05-11
Improved sync

3djc:
We like this fix better , @raphaelcoeffic please check it is ok

### DIFF
--- a/radio/src/mixer_scheduler.cpp
+++ b/radio/src/mixer_scheduler.cpp
@@ -68,9 +68,13 @@ void mixerSchedulerSetPeriod(uint8_t moduleIdx, uint16_t periodUs)
   mixerSchedules[moduleIdx].period = periodUs;
 }
 
-bool mixerSchedulerWaitForTrigger(uint8_t timeoutMs)
+void mixerSchedulerClearTrigger()
 {
   RTOS_CLEAR_FLAG(mixerFlag);
+}
+
+bool mixerSchedulerWaitForTrigger(uint8_t timeoutMs)
+{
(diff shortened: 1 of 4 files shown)


raphaelcoeffic:
> We like this fix better , @raphaelcoeffic please check it is ok

Most of it is ok, it's just another way of putting it. In these kinds of critical things, I prefer to have an explicit semantic rather than an explicit one, as it allows the reader to better understand what is happening, without having to look at other places.

However, there is something that is IMHO not a good choice:
```
    for (int timeout = 0; timeout < MIXER_MAX_PERIOD; timeout += MIXER_FREQUENT_ACTIONS_PERIOD) {
      bool interruptedByTimeout = mixerSchedulerWaitForTrigger(MIXER_FREQUENT_ACTIONS_PERIOD);
      execMixerFrequentActions();
      if (!interruptedByTimeout) {
        break;
      }
    }
```

The issue here is that `execMixerFrequentActions()` is called on a time critical code path. It is important that the time lapse between the trigger and sending the packets to the module is as constant as possible (so that the module can better adjust the frequency and phase shift).

We know it is not possible to be 100% constant, because of the way the mixer works, but it's worth trying to. So I would move `execMixerFrequentActions()` up, before `mixerSchedulerWaitForTrigger()`.

CapnBry:
I'm just a random developer but for sure the code needs to check the not-timeout flag before doing the `execMixerFrequentActions` and delaying the mixer from running.

Whether the `mixerSchedulerEnableTrigger` should also clear it instead of an explicit call in the main function, that's a small style choice and personally I'd rather see the explicit call to know what the behavior is without looking into the mixerSchedulerEnableTrigger function, but that's splitting hairs.

PR #8507, opened 2021-05-14
Model heli config

elecpower:
Auto widgets and item models housekeeping


Eldenroot:
Omg am I dreaming or is it happening?! :) thx!

elecpower:
@Eldenroot sorry dreaming it is just some existing Companion ui housekeeping. As a fixed wing pilot I'm not even sure what exists is usable? I do not believe any of the devs are heli pilots and repeated requests for dev help in this area has received no practical support.

PR #8533, opened 2021-05-28
Fix stack alignment

rotorman:
29+1 is not dividable with 4.


raphaelcoeffic:
By the way, this enables compiling with more recent compilers without crashing the radio ;-)

PR #8545, opened 2021-06-07
Fixes trainer setup in Companion

### LINKED ISSUE #8541: Windows Companion can't be used without corrupting the Trainer Mode: field
You will see that I have reported this problem months ago as a part of issue #7873 which was I think mis-interpreted and re-written by others in authority here.  
I frankly do NOT care about the Trainer Mode Simulation issue with the Trainer in TRAINER ON status. That's never worked. That problem is still there in Companion too, but is I feel very low pri.. 

This issue is serious since it means Companion users using this wireless training mode can not edit ANY model in companion and Write it back to the Radio without forcing him/her to need to check and edit the Trainer Mode parameter and change it back to: "MULTI/Master" any time you edit ANY other model for any reason.     

The Windows Companion is a wonderful tool for editing and backing up models however for those of us that like to do Wireless Training using the Trainer Mode type : MULTI/MASTER  find this mode field is corrupted whenever models programmed to support it are read in from the Radio, and also in the other direction, when a user Writes Models and Settings and has pre-set that Trainer Mode:  to "Multi/Master" the Model Setup Page inside the Radio will show that this field was corrupted during the write and that the Model Setup Page displays the following Trainer Mode: "eck RSSI on Shutdo"  instead.       
**To Reproduce**
1. In Companion, Go to any Model with the Wireless Training mode set to "Multi/Master" and then Write Models and Settings to the radio.  Select Models then scroll up to the Trainer Mode field.  You will see "eck RSSI on shutdo" instead and must change it back to Master /Multi to re-enable the basic Wireless Trainer function to work.
and also...
2. In Companion connect any radio with Wireless Training mode Master Multi enabled. Select Read Models and Settings then go to that model's Radio Setup Page and look at the new value in this field..  It will read Master/Bluetooth instead of what was last set inside the Radio,  and must be edited back to "Multi/Master" to fixit.

These behaviors 1. and 2. above do NOT produce these errors as stated 100% of the time only approximately 98% of the time..why it seems to work some times is not clear.  

**Expected behavior**
One would expect that when he/she makes a change on either the Radio side or the Companion side and re-Writes, or re-Reads  it back across the USB interface that the parameter should not change to some other value as it is doing now.    

If applicable, add screenshots to help explain your problem... Ok I'll try to get a couple 
 - OpenTX version: 2.3.11
 - Hardware: Radiomaster TX16S
 - OpenTX Companion:  2.3.11 on Windows10 64bit 20H01)

elecpower:
Fixes #8541
Exposes missing ppm fields for Slave/Jack to match radio
Adds item model and toString for trainer mode
Remove static trainer modes from ui definition and use filtered item model
Print trainer mode use toString function.
Note: only tested using Companion Simulator for X9D+, TX16S, X12S, X7 and a couple of others with and without bluetooth trainer


jims1234:
I just looked at the "details" of what the Travis CI compiler did and My god this stuff is complicated..  Thank you so very much .. I'm heading over to the Donate page now because I know just keeping the tool chain that can manage such a huge and complex project as Companion and/or OpenTx firmware for all these radios is an incredibly complex challenge.. Thank God you're doing it, not dummies like me.. :)

PR #8547, opened 2021-06-08
Add drawHudRectangle() function to lua lcd

olliw42:
I just have seen @3djc 's post at rcgroups that 2.3.12 is getting close. Maybe somewhat late but I'd like with this PR suggest to add to 2.3.12 a HUD drawing function to lua scripts.

I'm using this since, well, I guess, a year or so, and found it to work well. I've added it to my branch since it massively speeds up lua scripts which draw a HUD compared to doing it in lua. I've observed values between 1.5 and 2 of speed up. It recently was also PR-ed to edgetx and found to speed up yappu in a similar range. 

The function should be more generally useful, since scripts with a HUD are among the most widely used scripts (yappu and alike).

Comment: For said purpose it would be also very useful to have line drawing function which clips to a user-specified rectangle. I can provide such a function quickly using CohenSutherland clipping. This might be however not what you want, hence I've not included it here but kind of ask if I should also PR this.

### DIFF
--- a/radio/src/gui/480x272/lcd.cpp
+++ b/radio/src/gui/480x272/lcd.cpp
@@ -586,9 +586,9 @@ void lcdDrawHudRectangle(float pitch, float roll, coord_t xmin, coord_t xmax, co
     if (roll > 0.0f) {
       for (int32_t s = 0; s < ywidth; s++) {
         int32_t yy = ymin + s;
-        int32_t xx = ox + ((float)yy - oy) / angle; // + 0.5f; rounding not needed
+        int32_t xx = ox + ((float)yy - oy) / angle;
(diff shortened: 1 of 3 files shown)


olliw42:
it seems some targets do use different lcdDrawSolidHorizontalLine functions... so I moved it to COLORLCD to see if checks are passed

olliw42:
I think I got it now, sorry for that. Let's see what test says... no still fails ... 
... it seems one needs to define it explictely for each screen...
... so moved it back to COLORLCD
again sorry for this

olliw42:
so, passes, but is now only for COLORLCD
hope it nevertheless has good chances to go into 2.3.12

PR #8549, opened 2021-06-09
MPM Updates to 1.3.2.84 for Companion and Radio GUI

MRC3742:
@3djc as it appears 2.3.12 is near to release.
These small changes, additions could be added to bring Companion and Radio GUI up to the current MPM version 1.3.2.84

Thanks as always - Rich


3djc:
Thanks a lot Rich

PR #8550, opened 2021-06-11
Confirmation prompt before writing models and settings to radio

### LINKED ISSUE #8542: Add warning in Companion prior to write to SD card
Many users have reported accidentally deleting all models on their transmitter. The issue usually arises when they are trying to add one more model supplied on an .otx file from another user. They either just write that one model to their radio thinking that it will be added to what's there already or have the wrong window active when doing the write command.

It would be helpful if whenever the "write models and settings to radio" command was enacted, for Companion to give a popup warning stating that this command will overwrite all existing models on the radio with whatever is showing in the active screen, do you want to proceed yes - no.

elecpower:
Fixes #8542
The prompt can be turned off

### DIFF
--- a/companion/src/mdichild.cpp
+++ b/companion/src/mdichild.cpp
@@ -1532,7 +1532,7 @@ void MdiChild::writeEeprom()  // write to Tx
 {
   if (g.confirmWriteModelsAndSettings()) {
     QMessageBox msgbox;
-    msgbox.setText(tr("You are about to overwrite ALL models on the Radio."));
+    msgbox.setText(tr("You are about to overwrite ALL models and settings on the Radio."));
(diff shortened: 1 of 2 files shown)


Oldgibber:
Something I asked for literally years ago, and was rejected, was a pop-up box to say "All models and settings written to radio", with perhaps a confirmatory click. 

At the moment a small message appears in the bottom left corner of the screen which disappears after a second or two and is easily missed.

3djc:
We signal errors and issues, not what works ok

Oldgibber:
> 

Surely it is an issue when there is no clear indication that the files have been written. Do you not want suggestions for improvements?

PR #8564, opened 2021-06-21
Allow both version of rotary encoder boards

3djc:
This should allow use of current rotary encoder boards who include a design error, and also fixed ones


rotorman:
For the interested, I posted some info about the hw-mod of the fixed roller board schematic and board layout for T16 and TX16S in EdgeTX PR: https://github.com/EdgeTX/edgetx/pull/261
You can find Eagle CAD files attached to the PR there, that can be sent directly to PCB manufacturers.

PR #8568, opened 2021-06-22
Fix ISRM pulses

### LINKED ISSUE #8565: Horus X10S Express and OTX 2.3.12 causes on some receivers huge delay in signals and causes malfunction on S-Bus and slow down in PWM Outputs
I updated my FRSky Horus X10S Express form OTX 2.3.11 to 2.3.12.

After the update some models working with FrSky Archer R8 Pro (FW2.1.7)and FrSky RSX-R (FW2.1.0) (both on ACCESS protocol) doesn't work with S-Bus to a Spirit FBL unit any more. 
Looking deeper I saw as well that there is a huge delay on signal processing (0,5 >2s) on the Archer R8 Pro to the PWM output with OTX 2.3.12. The PWM's of the Archer R8 Pro have nothing to do with the spirit unit. So it seems to be an issue between Horus and receiver.
This happend not to an Archer RS (FW2.1.8). This unit works fine together with the Spirit 

Roll back and forth between 2.3.11 and 2.3.12 let me reproduce the errors.

I expected that the Archer R8 Pro and the RSX-R work after the update to the new OTX version as before.

IMG_1870
IMG_1871

3djc:
Awaiting further tests

### DIFF
--- a/radio/src/pulses/pulses.cpp
+++ b/radio/src/pulses/pulses.cpp
@@ -268,12 +268,12 @@ void enablePulsesExternalModule(uint8_t protocol)
 #if defined(PXX2) && defined(EXTMODULE_USART)
     case PROTOCOL_CHANNELS_PXX2_HIGHSPEED:
       extmoduleInvertedSerialStart(PXX2_HIGHSPEED_BAUDRATE);
-      mixerSchedulerSetPeriod(EXTERNAL_MODULE, PXX2_PERIOD);
+      mixerSchedulerSetPeriod(EXTERNAL_MODULE, PXX2_NO_HEARTBEAT_PERIOD);
       break;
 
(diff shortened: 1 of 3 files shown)


radio/src/pulses/pulses.cpp
```diff
@@ -347,7 +347,7 @@ bool setupPulsesExternalModule(uint8_t protocol)
     case PROTOCOL_CHANNELS_PXX2_LOWSPEED:
       extmodulePulsesData.pxx2.setupFrame(EXTERNAL_MODULE);
 #if defined(PCBSKY9X)
```
raphaelcoeffic, line 349: This `define` does not look right. Shouldn't it be `!defined(PCBSKY9X)`?

PR #8580, opened 2021-06-27
Fixes 8562 missing negative sign for telemetry sources

### LINKED ISSUE #8562: Companion 2.3.12: When choosing a sensor to include in a "Calculated" sensor, the minus signs are missing.
In previous versions of Companion, when selecting a sensor to include in a "Calculated" sensor, each available sensor's name appeared in the top half of the list preceded by a minus (-) sign and again in the bottom half of the list without a preceding sign. Now the sensor's name still appears in both halves of the list but the minus sign is missing, making it unclear whether the value or its negative will be used in the calculation.

This is the issue that [sandet01 mentioned on RCGroups](https://www.rcgroups.com/forums/showthread.php?p=47330411&postcount=5536).

elecpower:
Fixes #8562

### DIFF
--- a/companion/src/firmwares/sensordata.cpp
+++ b/companion/src/firmwares/sensordata.cpp
@@ -322,10 +322,10 @@ QString SensorData::precToString(const int value)
 }
 
 //  static
-QString SensorData::sourceToString(const ModelData * model, const int index, const bool sign)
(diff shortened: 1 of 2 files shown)


3djc:
thanks !

Issue #8598, opened 2021-07-06
Your opinion: Lua scripts with touch screen?


olliw42:
hey
I'd like to ask whether this would be something you possibly would consider adding in a nearby version:

I've noted that it literally would need just few code lines to enable touch screen support and make touch available to lua widget scripts. To be clear, I am not talking about adding touch to OpenTx in general, only to read it and make it available to lua widgets.

I have this running here and I really just love it ... I bite my ass that I haven't noted this already a year ago how easy it was ...

this would add a full new layer of functionality to lua widgets, and I can't really see any cost of it

hence this question

3djc:
Not opposed to the idea personnaly, but can you tell more about what you uhave in mind ?

Reading coordinates is trivial, and would probably be enough in a full screen lua scenario, but how about widgets ? How would a widget know this swipe is for it ?

lshems:
> Not opposed to the idea personnaly, but can you tell more about what you uhave in mind ?
> 
> Reading coordinates is trivial, and would probably be enough in a full screen lua scenario, but how about widgets ? How would a widget know this swipe is for it ?

This 'focus' thing has been discussed in length and is a non issue. The issue is not what you question above, but the position opentx has taken on allowing Lua to know about button events to act on them. Position is that it will not happen.

So better be honest, instead of asking for info you'll never use.

lshems:
> hey
> I'd like to ask whether this would be something you possibly would consider adding in a nearby version:
> 
> I've noted that it literally would need just few code lines to enable touch screen support and make touch available to lua widget scripts. To be clear, I am not talking about adding touch to OpenTx in general, only to read it and make it available to lua widgets.
> 
> I have this running here and I really just love it ... I bite my ass that I haven't noted this already a year ago how easy it was ...
> 
> this would add a full new layer of functionality to lua widgets, and I can't really see any cost of it
> 
> hence this question

Did you already manage to get this running on edgeTX? I'm waiting desperately to get key and touch events in scripts and widgets.

olliw42:
>  can you tell more about what you uhave in mind ?

this is on EdgeTx, but I have exactly (= EXACTLY) the same also working on OpenTx (just the first 1:30 minutes are of relevance)
https://youtu.be/W0uaKVUnGps

I find it supercool to have buttons and selectors (and it's amazingly simple to do in the lua)

>Reading coordinates is trivial, and would probably be enough in a full screen lua scenario

yes
[...]

PR #8599, opened 2021-07-06
Touch for lua widgest

olliw42:
for further discussing issue https://github.com/opentx/opentx/issues/8598

some notes:
- I've only tested this to compile for a tx16s, but not by actually running a lua script. This is just the concept. If it gets more realistic I'll be happy to provide a test lua script.
- the method of how to forward the touch event to the lua widget can be debated, several approaches are possible, like e.g. forwarding as function argument to the refresh() function. I here took the simplest approach of simply defining a lua function.
- the extEvent value could possibly also be merged with the event value by a bit shift, like its done for the lcd flags
- the checkTouchTmo() function is probably superfluous and could be done at the top of touchPanelRead() (if it is guaranteed to always be regularly called)
- the wipe detection could need a bit of improvement, it works but needs a bit of "training" by the user

<strike>to a dev: I think PR's can be somehow marked as experimental or draft or alike, maybe this could be done with this one here</strike> LOL, google helps ...


olliw42:
FYI
I've actually "decided" for my own opentx flavor to forward the touch event as parameter to the refresh() function, and not a lua function (i.e. I dropped getTouchEvent() and call now refresh() as refresh(widget,events)). Function-wise it's pure cosmetics, i.e., NFC, but it may be more like what one would expect to see. I actually provide it as a table in a table, as I've concluded that's the only reasonable way to have it extensible, i.e., future proof. 
I'm not sure I should update this PR here for as long as it's not yet clear if it crosses red lines or not. I'll do as you advice.

olliw42:
btw, any statement possible already if it crosses red lines or not? :D

3djc:
Sorry quite busy currently, I need to check the event side with what we do on libopenui and 2.4, don't want two definitions of similar things

olliw42:
sorry if I came across pushy, I'm just curious, that's all.

concerning libopenui & 2.4:

That's why I've added this extEvent field thing, which holds what is more useful for a lua script. That is, the event field which libopenui is using is not touched or affected at all, so that there are no side effects at all, and linopenui/2.4 can do whatever they need to do or want to do. From this event the extEvent is then generated and it is which is forwarded to (and cleared by) the lua script. I think this decoupling of libopenui and lua has many benefits, like no side effects on libopenui and future-proofness, i.e. libopenui can change it's state machine anytime.

It's true however that in libopenui/2.4 one also can think of much more advance lua concepts, like a lua script which can call native libopenui windows like checkboxes, buttons, sliders, and so on, and I think/guess that this is what is on every ones mind. Any concept introduced now may have to be overriden later on - but as much as I can see this advanced lua concepts are more of a goal or idea currently and not in a state to materialize as a mature result in any near future.

So, it's true, this here is an idea for an intermediate approach to have touch lua scripts, which in some bright future should die out.

radio/src/lua/api_general.cpp
```diff
@@ -1992,6 +2012,15 @@ const luaR_value_entry opentxConstants[] = {
 
 #if defined(KEYS_GPIO_REG_DOWN) && defined(NAVIGATION_HORUS)
   { "EVT_RTN_FIRST", EVT_KEY_BREAK(KEY_EXIT) },
+  { "TEVT_DOWN", TE_DOWN },
```
3djc: Could you rename those EVT_TOUCH_xxxx ?

olliw42:
@3djc 
made the suggested change

also slightly improved the touch behavior

here you find a script to test it, copy it to WIDGETS folder and install as full screen on e.g. main view 2
[TTest.zip](https://github.com/opentx/opentx/files/6796369/TTest.zip)

olliw42:
from my side the question which would need to be decided if the touch events should be passed to the lua script 

a) via a function like the `getTouchState()` as it is now 
b) as parameter to the refresh() function, i.e., like `local function refresh(widget, events) ... end`

both work equally well, but have different pro&cons, mainly concerning potential forward migration

radio/src/options.h
```diff
@@ -100,6 +100,9 @@ static const char * const options[] = {
 #endif
 #if defined(BIND_KEY)
   "bindkey",
+#endif
+#if defined(HARDWARE_TOUCH)
```
3djc: Would prefer not to display this, as this would not be a firmware option

radio/src/main.cpp
```diff
@@ -500,6 +500,10 @@ void perMain()
 
   event_t evt = getEvent(false);
 
+#if defined(HARDWARE_TOUCH)
```
3djc, line 503: Maybe accademical, but I guess it should be #if defined(HARDWARE_TOUCH) && defined(LUA) ? A couple of other places too

3djc:
Apparently the EdgeTX folks are working on some issues around touch handling (unrelated to this PR), we probably will wait those to be fixed before merging this (no point in releasing something we know does have issues)

radio/src/options.h
olliw42: ok

olliw42: not sure though why it wouldn't be an option ... I mean, you would compile it into any boards?

radio/src/main.cpp
olliw42: interesting point. So far HARDWARE_TOUCH was independent on LUA. I would suspect that this would be so also on the long run. Also, the changes embraced in these defines basically "just" took out the touch init & read from the radio_diaganas and moved it to more generic places, and did allow some touch testing. Again totally independent on the lua additions. My feeling is that not linking this to lua is probably a good thing.

olliw42:
yeah, sadly, the touch can look up ... I do have that only very very rarely, but it seems that others can get that quite frequently ...

it's somewhat unfortunate that opentx doesn't have a means for experimental features

besides the possible look ups, you maybe nevertheless want to try the script ... it's fun :D

radio/src/lua/api_general.cpp
olliw42: ups, it now comes through to me too LOL
done already

radio/src/main.cpp
3djc: Well in a 2.3 codeline, we are not using touch, unless for this specific LUA scenario, so why read touch if you don't have LUA ? The touch was previously read only in diaganas for the same reason

radio/src/options.h
3djc: I mean we would include that by default in all radio that support touch, so this would not be an option (option are the tick boxes that you have in companion)

3djc:
Well we have that, it is called a fork, or a branch ! But code in master is expected to work.   Yes, we are only humans and there is, and will be, issues in the code we put in master, but there is a difference between that and knowingly put code you know doesn't quite work as it should.  :)

radio/src/main.cpp
olliw42: I think the point which wasn't clear to me was what you were explaining in the other comment, that it would be included by default and would not be an option

yet, I'm not fully understanding: touch in diaganas should be now also available only if also lua is available?

IMHO it would not make much sense to keep the old code in diaganas in case one would not use lua, I rather would consider it more clean to do it in diaganas as it is right now for both with and without lua - in which case we would be back to the defines as they are now...

radio/src/options.h
olliw42: got it

olliw42:
I would have some comments to your last response, but they probably would steer us into a too OT direction, so I better shut up LOL

I rather like to ask this:
EdgeTx has made some improvement to touch, in my understanding (= I might be mistaken) it is not totally perfect yet in the sense that it never locks up anymore, but lockups are mitigated to a point that they consider it now ripe for release. So, the question would be, if I would bring in EdgeTx's changes as a PR to OpenTx, would this then release the stop on this PR here?

radio/src/main.cpp
3djc: In diaganas, the touch panel is activated on entry in the screen, and desactiavted on exit. The only purpose of this existing was to allow buyers of the tx16s to reassure themselved confirm they actually have touch hardware, and it is working. We do not intend to have touch hardwrae running and polled if not usable, which is why TP is deinit on screen exit. In case of this PR, if LUA isn't specified, then touch becomes back not usable, and therefore should not be active, imho

olliw42: ok
so, if TOUCH and not LUA, then one should get the old behavior for diaganas (i.e. diaganas ints and reads touch)? Or not, becasue these users instead just could use the compile with LUA?
Just want to get it right :)

3djc: I'm personnaly ok if after this PR the diaganas part is disabled if LUA is disabled, I think both ways are ok. LUA option is not an end user option in Companion (unlike LUAC), but there are professional users that have LUA disabled in their builds, it is more is this sceneario that I believe we should not be pooling the TP

olliw42: ok, even I understand it now :)
I'll do this evening

olliw42: I did it now not by adding && defined(LUA) to all these instances, but rathe by not setting HARDWARE_TOUCH in the CMakeFile if LUA is defined. This achieves exactly the same, but is more flexible and IMHO just nicer. Hope that's acceptable. For the moment, I consider this resolved.

olliw42:
made HARDWARE_TOUCH dependent on LUA, by handling it in the CMakeFile
changed some uint to bool, as this is what they are and also to resolve some intermixed 0/1 and false/true
also took the liberty to somewhat clean uop the radio_diaganas
tested for HARDWARE_TOUCH = YES/NO and LUA=YES/NO on Tx16s and T16

olliw42:
> So, the question would be, if I would bring in EdgeTx's changes as a PR to OpenTx, would this then release the stop on this PR here?

comments to this?

olliw42:
@3djc 

now that EdgeTx2.4.0 has been officially released, and also OpenTx2.3.14 is out of the way, I'd like to ask if you could respond to this:

If I would bring in EdgeTx's touch screen handling changes as a PR to OpenTx, would this then release the stop on this PR here? That is, would EdgeTx'es changes be considered "to work" so that you would consider that one could go on with this PR here?

3djc:
Had a chat with Risto a couple of days ago about the changes that have been done in edge, that are more a workaround than a true fix. I think they would be acceptable if the debug code is removed and some way to prevent chain restart of gt911 is implemented

rotorman:
The I2C code in EdgeTX stems almost fully from STM32F4 HAL and is (in my opinion) by no means a non-production code. The biggest issue being that it is currently non interrupt, non DMA based, using while loops for waiting. This works, as reported by testers at EdgeTX Discord, without issues and also is no different than previously touch I2C was handled in OpenTX.

If you follow the original STM32F4 HAL I2C flow behind the scenes you can see that it has very awkward handling of triggering I2C STOP and ACK/NACK, differing if there are 3, 2 or a single data byte to read in. I wrote up the execution flow in nicely readable form in: https://github.com/EdgeTX/edgetx/pull/457#issuecomment-881677919

I have been testing DMA based I2C interaction with GT911 and have it already beautifully running on a dedicated, non-OpenTX/EdgeTX based test code on RM TX16S mainboard (a dedicated STM32CubeIDE project):
https://github.com/rotorman/GT911test/blob/master/TX16S_FreeRTOS_I2C_DMA/Core/Src/main.c
It is using again STM32F4 HAL, but this time the DMA I2C functions and also FreeRTOS. As soon as CooCox CoOS gets swapped in EdgeTX against FreeRTOS, I would integrate the I2C DMA based readout code.

In short, the new low-level I2C DMA based interaction using FreeRTOS can be described with these brief lines:
```
#define I2C_TIMEOUT pdMS_TO_TICKS(20) /* 20 ms */

bool I2C_GT911_ReadRegister(uint16_t reg, uint8_t * buf, uint8_t len)
{
    if (HAL_I2C_Mem_Read_DMA(&hi2c1, GT911_I2C_ADDR << 1, reg, I2C_MEMADD_SIZE_16BIT, buf, len) != HAL_OK)
        return false;
    if (xSemaphoreTake(hBinSemI2C, I2C_TIMEOUT) != pdPASS) // Wait for semaphore from HAL_I2C_MemRxCpltCallback()
        return false;
    return true;
}

bool I2C_GT911_WriteRegister(uint16_t reg, uint8_t * buf, uint8_t len)
{
    if (HAL_I2C_Mem_Write_DMA(&hi2c1, GT911_I2C_ADDR << 1, reg, I2C_MEMADD_SIZE_16BIT, buf, len) != HAL_OK)
        return false;
    if (xSemaphoreTake(hBinSemI2C, I2C_TIMEOUT) != pdPASS) // Wait for semaphore from HAL_I2C_MemTxCpltCallback()
        return false;
    return true;
}

void HAL_I2C_MemTxCpltCallback(I2C_HandleTypeDef *hi2c)
{
    BaseType_t xHigherPriorityTaskWoken = false;
    xSemaphoreGiveFromISR(hBinSemI2C, &xHigherPriorityTaskWoken);
    portYIELD_FROM_ISR(xHigherPriorityTaskWoken);
}

void HAL_I2C_MemRxCpltCallback(I2C_HandleTypeDef *hi2c)
{
    BaseType_t xHigherPriorityTaskWoken = false;
    xSemaphoreGiveFromISR(hBinSemI2C, &xHigherPriorityTaskWoken);
    portYIELD_FROM_ISR(xHigherPriorityTaskWoken);
}
```

3djc:
FreeRTOS is on my TODO list, I can see some synergy here :)

olliw42:
@rotorman 
hey Risto, thx for chiming in

I read your words to mean that the current touch code in EdgeTx is not a workaround, but indeed a fix. This was my understanding of the situation too. Is that the correct, you guys consider it fixed?

It is maybe not what one really wants to have since, I think, everyone agrees that polling code isn't the nicest, but I think just the fact that it is polling doesn't make it a workaround.

3djc:
The workaround part is that it still crahses and GT911 is reset and restarted

olliw42:
> Had a chat with Risto a couple of days ago about the changes that have been done in edge, that are more a workaround than a true fix. I think they would be acceptable if the debug code is removed and some way to prevent chain restart of gt911 is implemented

many thx for the response. This is a somewhat hesitant response, isn't it? I would prefer clear statements as these tend to avoid misunderstandings. So, with the changes you suggest you say: YES!

Removing the debug code makes sense to me. I'm not so convinced with the chain restart thing (if you refer here to what I think you refer too, the restart in case of issues). I2C is dodgy. So it is has just become good common practice to have rescue handlers. Thus, IMHO, and I think most would agree, the fact that there are rescue handlers does not mean it's bad code, but to the opposite.

olliw42:
> The workaround part is that it still crahses and GT911 is reset and restarted

@rotorman 

is this so? The hiccup counter, is it going up just very very sporadically, or kind of many times for many users?

(with the new code I myself just had not yet have seen it triggering)

rotorman:
I do not see the existing I2C code in EdgeTX as a workaround - the touch was extensively tested and deemed working with the code as is. Nevertheless, the following applies to it:
* polling I2C and waiting in a while() loop are not ideal solutions, which is why for the next iteration plans are afoot to change it against DMA based code
(the I2C DMA based GT911 readout already works, if you care to test my STM32CubeIDE project at https://github.com/rotorman/GT911test/tree/master/TX16S_FreeRTOS_I2C_DMA - it outputs touch data via AUX1 with 400000 baud 8N1 on TX16S. No other functionality is active, thus do not be alarmed that the TX16S display remains dark.)
* using the STM32F4 HAL blocking `HAL_I2C_Master_Transmit()` and `HAL_I2C_Master_Receive()` we have reports from two users saying that they saw each a single "I2C/GT911 hiccup" during a long testing session. The hiccup stems from GT911 pulling either SDA or SCL constantly low and not releasing it until it is hw-resetted. The current EdgeTX I2C/GT911 code detects this state and issues a hardware reset for the GT911 and also re-initializes the internal I2C handling. This has been shown in tests to work effectively and quickly (takes less than 200 ms from detection, over resetting, back to running touch state). No user has mentioned that they noticed this during usage (other than the hiccup/TSI2CEvent counter being displayed in analogs diagnose screen as non-zero).

Future tests will tell, if changing to DMA based I2C handling will fully eliminate the GT911 lockup - it may very well be an issue with GT911 or it's current config used. Different STM32 internal I2C handling (polling to DMA) might not remedy this.

olliw42:
@rotorman many THX, sir

so, the facts are on the table now, and it's OTX's decision now

I would fell that if it's good enough for EdgeTx, which depends soooo much more on touch than OpenTx, it should be good enough for OpenTx too, but as said, it's OTX's decision now

let's see what they decide :)

olliw42:
wipe handling significantly improved
also have taken over the i2c touch code form EdgeTx 2.4.0

works great for me on a Tx16s

I would not see a reason why now this could not be considered seriously, would be great to hear the current thoughts

3djc:
Need to touch base with Risto. We first need to merge FreeRTOS, and Risto was working on dma based drivers for it, haven’t been able to check the status of that

rotorman:
I do not have a working version yet together with StdPeriph (in a separate purely STM32F4 HAL based STM32CubeIDE Pj. works fine, but this does not help here).

3djc:
No rush, we haven’t merged FreeRTOS just yet !

rotorman:
The work in progress is here in case you or @olliw42 (or anyone else) wants to have a look/help out:
https://github.com/EdgeTX/edgetx/tree/i2c_dma
Do not know yet why DMA gives an error after first transaction (testing with RM TX16S).

olliw42:
> Need to touch base with Risto. We first need to merge FreeRTOS, and Risto was working on dma based drivers for it, haven’t been able to check the status of that

ah, so FRTOS is planed for v2.3.15? very cool

the lua touch support added here should be really very independent on whatever RTOS & I2C & touch-sensor handling is been used

maybe it was a mistake to add the EdgeTx thing to this PR
but it makes it testable, so that you guys can convince yourself that it accomplishes a thing (I'm in the stages of preparing a video demonstrating what's possible with this type of lua widget touch support - I would be surprised if it wouldn't blow you away LOL)

> The work in progress is here in case you or @olliw42 (or anyone else) wants to have a look/help out:
> https://github.com/EdgeTX/edgetx/tree/i2c_dma

many thx. I couldn't hold a promise I made before, so I won't make any promises anymore anytime LOL

> Do not know yet why DMA gives an error after first transaction (testing with RM TX16S).

yeah, STM32 DMA can be tricky and not so understandable sometimes
I don't see so much value in using DMA here, you have tried a ISR only solution and found to not work?

EDIT: just briefly looked into it: really, an extra task just for reading the touch screen?? This doesn't feel right at all, I would say.

rotorman:
> an extra task just for reading the touch screen?? This doesn't feel right at all, I would say.

For testing this provides IMHO relatively independent test bed for touch and I2C. When code is running, the extra task can be scrapped and merged with menus task. Most of the processing is done in IRQs anyhow.
Also the use of current FreeRTOS semaphores is more costly than task notifications, but for first they provide better encapsulation.

olliw42:
ah, so it's currently there for test&dev, makes sense for sure, sorry I was picking on it :)

olliw42:
here we go
part1 of the video
https://youtu.be/_56VRxEAWkM
part2
https://youtu.be/FOu4gnsiU8Y

olliw42:
@3djc 
is this PR still considered of potential relevance or you kind of put it off the list for e.g. 2.3.15 ?
(i.e. would I waste my time updating it or not?)

btw, since I have added the EdgeTx touch drivers, I had not a single case where the touch screen would have become unresponsive ... certainly just an isolated experience, yet ...

3djc:
Still on the list, less concerned about this PR than with the RTOS one it depends on

olliw42:
thx for the response

I hear you saying that this depends on the RTOS. I'm not fully understanding this, but I probably don't have to understand it :)
(EdgeTx is AFAIK on FREE_RTOS since more than a month, so concerning this PR it only can be a matter of taking this or that drivers :))

I was just asking since it hasn't a 2.3.15 label, while now more and more get that label ... I just wanted to ensure that you don't think that I left, I would still be interested to help getting this into a next version. :)

PR #8600, opened 2021-07-06
Add OS name to lua luaGetVersion()

olliw42:
adds OS name string to the lua luaGetVersion() function
this allows to discriminate between OpenTx and EdgeTx, which is potentially highly useful for scripts
this mirrors the change done by EdgeTx (https://github.com/EdgeTX/edgetx/blob/2.4/radio/src/lua/api_general.cpp#L136-L145)

no idea if it needs a change in simulator/companion, I have no knowledge here

tested on tx16s

### DIFF
--- a/radio/src/lua/api_general.cpp
+++ b/radio/src/lua/api_general.cpp
@@ -87,7 +87,7 @@ Return OpenTX version
 
 @retval string OpenTX version (ie "2.1.5")
 
-@retval multiple (available since 2.1.7) returns 5 values:
+@retval multiple (available since 2.1.7) returns 6 values:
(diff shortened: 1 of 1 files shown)


pfeerick:
You may also want to consider also adding the documentation changes (or something similar) for the return values that as was done in the EdgeTX PR this is replicating... i.e. lines 96, 103-104 of the PR. https://github.com/EdgeTX/edgetx/pull/323/files

Since line 90 is now inconsistently stating that there are five return values while documenting six. 😉

olliw42:
indeed, thx

3djc:
thanks

PR #8601, opened 2021-07-07
More pulse sync fixes

### LINKED ISSUE #8575: Horus X12 - ISRM Open-Tx 2.3.13 - Very choppy servo movement on ACCST D16 receivers
After updating my X12s to 2.3.13 the servo movement became very 'choppy' in ACCST D16 mode. A clubmate with the same Tx config (X12s with ISRM upgrade) has the same issue. On ACCESS all is fine.

I switched the D16 models I had on the field today to the MPM and there control was fine. 

I went from 2.3.11 to 2.3.13 so I don't know if 2.3.12 had the same issue.

The way it is now is unflyable so this is a serious bug!

### LINKED ISSUE #8585: LOST CONNECTION OF HORUS X10S EXPRESS AND RECEIVER
I have a Horus X10S Express and both in versions 3.3.12 and 3.3.13 (much less) I have had problems of loss of connection with the plane, the Horus X10S said "loss of telemetry" and the servos stopped during one or two seconds, then the telemetry would recover and the servos would work again.
To produce the fault, he moved the right stick continuously, commanding ailerons and depth.

Both the Horus X10S Express transmitter with the receivers (Archer R6 and Rx6R) are updated to the latest versions of Frsky, namely 2.1.6 and 2.1.7. respectively.

I am back to version 3.3.11 and everything seems to work fine.

3djc:
Fix Horus iXJT where HB and expired timers where both trigering

Fix ISRM ACCST broken by 2.3.13. Duration set to 10ms which is the longuest one (9ms for ISRM/ACCST) + 1ms backup

This fixes #8585 
This fixes #8575

### DIFF
--- a/radio/src/pulses/pulses.cpp
+++ b/radio/src/pulses/pulses.cpp
@@ -518,11 +518,6 @@ bool setupPulsesInternalModule(uint8_t protocol)
 #if defined(HARDWARE_INTERNAL_MODULE) && defined(PXX1) && !defined(INTMODULE_USART)
     case PROTOCOL_CHANNELS_PXX1_PULSES:
       intmodulePulsesData.pxx.setupFrame(INTERNAL_MODULE);
-#if defined(INTMODULE_HEARTBEAT)
-      mixerSchedulerSetPeriod(INTERNAL_MODULE, INTMODULE_PXX1_SERIAL_PERIOD + 1000 /* backup */);
-#else
-      mixerSchedulerSetPeriod(INTERNAL_MODULE, INTMODULE_PXX1_SERIAL_PERIOD);
-#endif
       return true;
 #endif
 
--- a/radio/src/pulses/pulses.cpp
+++ b/radio/src/pulses/pulses.cpp
@@ -457,12 +457,10 @@ static void enablePulsesInternalModule(uint8_t protocol)
     case PROTOCOL_CHANNELS_PXX1_PULSES:
       intmodulePxx1PulsesStart();
 #if defined(INTMODULE_HEARTBEAT)
-      // use backup trigger (1 ms later)
       init_intmodule_heartbeat();
-      mixerSchedulerSetPeriod(INTERNAL_MODULE, INTMODULE_PXX1_SERIAL_PERIOD + 1000/*us*/);
-#else
-      mixerSchedulerSetPeriod(INTERNAL_MODULE, INTMODULE_PXX1_SERIAL_PERIOD);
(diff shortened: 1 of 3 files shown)


radio/src/targets/common/arm/stm32/heartbeat_driver.cpp
```diff
@@ -79,6 +79,8 @@ void check_intmodule_heartbeat()
 #endif
     EXTI_ClearITPendingBit(INTMODULE_HEARTBEAT_EXTI_LINE);
 
+    mixerSchedulerResetTimer();
```
raphaelcoeffic, line 82: @3djc I believe that reseting the timer here is good, thus allowing to do it earlier in the game, and removing the special casing for heartbeat in `setupPulsesInternalModule()`. It seems to me as well the timer reset was missing in the same function for PXX1 (`setupPulsesInternalModule()`):
```
#if defined(PXX1) && defined(INTMODULE_USART)
    case PROTOCOL_CHANNELS_PXX1_SERIAL:
      intmodulePulsesData.pxx_uart.setupFrame(INTERNAL_MODULE);
      return true;
#endif
```

raphaelcoeffic: By the way, that would explain why the issue with iXJT showed up on X10/X12S, and not on Taranis-like targets, as only these targets use PXX1 and INTMODULE_USART.

radio/src/pulses/pulses.cpp
```diff
@@ -523,7 +519,6 @@ bool setupPulsesInternalModule(uint8_t protocol)
     case PROTOCOL_CHANNELS_PXX1_PULSES:
       intmodulePulsesData.pxx.setupFrame(INTERNAL_MODULE);
 #if defined(INTMODULE_HEARTBEAT)
-      mixerSchedulerResetTimer();
       mixerSchedulerSetPeriod(INTERNAL_MODULE, INTMODULE_PXX1_SERIAL_PERIOD + 1000 /* backup */);
```
raphaelcoeffic: @3djc I think that modules / protos using fixed period (with or without heartbeat) do not need to set the period every time, as it does not change, and even if that would change, the new period is used only if the timer hits (set only in `mixerSchedulerStart()` and `MIXER_SCHEDULER_TIMER_IRQHandler()`). For PXX2, this is a different story, as I'm not sure the module is reset when going from TOOLS_PERIOD to the "normal" period.

PR #8607, opened 2021-07-08
Heliwizard & Lua Swash access

jrwieland:
Heli wizard (for horus "color" based radio's) 
works for both FBL and flybar helicopters.

Added lua script access swash settings in to Model Edit for the fbl settings to be applied to the heli menu from the script


3djc:
Very interesting ! Has there been community test and discussion about this ? If yes, a link would be appreciated

jrwieland:
> 
> 
> Very interesting ! Has there been community test and discussion about this ? If yes, a link would be appreciated

@3djc  Here is a link for the files, there is a compiled companion (windows) as well as the C++ modeledit file and the lua script.
If you have another OS feel free to add the C++ file to it and compile a copy for yourself.
Poke around and see what needs changed... having more eyes on it than mine is always a plus !!

[Heli Test Files](https://drive.google.com/file/d/1Y0NOxN2OI29ajCTlWgCys6wGeZq7SU90/view?usp=sharing)

radio/src/lua/api_model.cpp
```diff
@@ -1586,11 +1586,69 @@ static int luaModelResetSensor(lua_State *L)
...
+
   lua_pushnil(L);
   return 1;
 }
+/*luadoc
+@function model.swashRingData(index, line,{parameters})
+
+Set heli swash parameters
+
+index = 0
```
3djc: Not sure what this means ? What does index param correspond to  ?

radio/sdcard/horus/SCRIPTS/WIZARD/heli/wizard.lua
```diff
@@ -0,0 +1,559 @@
...
+  if TypeFields[1][5]==0 then
+    drawNextTextLine("Type","FBL")
+  else
+    drawNextTextLine("Type","FB")
+    if TypeFields[2][5]==0 then
+      lcd.drawText(290,10,"Swash 120")
+    elseif TypeFields[2][5]==1 then
+      lcd.drawText(290,10,"Swash 120X")
+    elseif TypeFields[2][5]==2 then
+      lcd.drawText(290,10,"Swash 140") else
```
3djc: your "else" position and indentation needs to be changed (many instances of that)

radio/src/lua/api_model.cpp
```diff
@@ -1586,11 +1586,69 @@ static int luaModelResetSensor(lua_State *L)
...
+* @notice If a parameter is missing, then that parameter remains unchanged.
 
+@status current Introduced in 2.3.x
+*/
+static int luaModelSwashRingData(lua_State *L)
+{
+unsigned int idx = luaL_checkunsigned(L, 1);
+if (idx < MIXSRC_CYC3){
+  luaL_checktype(L, -1, LUA_TTABLE);
+for (lua_pushnil(L); lua_next(L, -2); lua_pop(L, 1)) {
```
3djc: wrong indent

radio/src/lua/api_model.cpp
```diff
@@ -1586,11 +1586,69 @@ static int luaModelResetSensor(lua_State *L)
...
+* 'elevatorWeight' (value) -100 to 100
+
+* @notice If a parameter is missing, then that parameter remains unchanged.
 
+@status current Introduced in 2.3.x
+*/
+static int luaModelSwashRingData(lua_State *L)
+{
+unsigned int idx = luaL_checkunsigned(L, 1);
+if (idx < MIXSRC_CYC3){
```
3djc: You seemt to be checking idx and not using it ?

radio/src/lua/api_model.cpp
jrwieland: Parameter renamed and explained.

Since this is basically an internal mix it follows the same format as a mix
index refereed to "which" mix (swash) and Line refereed to what line.

Since this is the only mix line for this item I simply described it so the user no option other than to use "0" for these 2 items.

I have tried changing the values to 1 & 2, from 0 they have no effect, however to eliminate any possible conflicts I suggest they be set to "0" in the descripton.

doc changes

radio/sdcard/horus/SCRIPTS/WIZARD/heli/wizard.lua
jrwieland: else position and indentation changed

radio/src/lua/api_model.cpp
jrwieland: indentation repaired

radio/src/lua/api_model.cpp
jrwieland: unused idx removed,
it was added to reflect the other lua set commands format.
I didn't think this would work without an initial "if" statement ... I stand corrected. 

That's what I get for not coding C in over 35 years... second guessing myself and adding unnecessary items

radio/sdcard/horus/SCRIPTS/WIZARD/heli/wizard.lua
```diff
@@ -0,0 +1,592 @@
...
+  --Set Swash Parameters
+  if TypeFields[1][5]==1 and TypeFields[2][5]==0 then
+    model.swashRingData(0,0,{type="1",collectiveSource=89,aileronSource=90,elevatorSource=88,collectiveWeight=60,aileronWeight=60,elevatorWeight=60})
+  elseif TypeFields[2][5]==1 then
+    model.swashRingData(0,0,{type="2",collectiveSource=89,aileronSource=90,elevatorSource=88,collectiveWeight=60,aileronWeight=60,elevatorWeight=60})
+  elseif TypeFields[2][5]==2 then
+    model.swashRingData(0,0,{type="3",collectiveSource=89,aileronSource=90,elevatorSource=88,collectiveWeight=40,aileronWeight=40,elevatorWeight=60})
+  elseif TypeFields[2][5]==2 then
+    model.swashRingData(0,0,{type="4",collectiveSource=89,aileronSource=90,elevatorSource=88,collectiveWeight=35,aileronWeight=35,elevatorWeight=60})
+  else
```
3djc: Something is missing or too much

radio/src/lua/api_model.cpp
```diff
@@ -1586,11 +1586,68 @@ static int luaModelResetSensor(lua_State *L)
...
+
   lua_pushnil(L);
   return 1;
 }
+/*luadoc
+@function model.swashRingData(swash, line, value)
+
+Set heli swash parameters
+
+@param swash (unsigned number) number (use 0)
```
3djc: If it needs to be 0, does it make sense to ask inputing it ?

jrwieland: No, but it has to be in the command otherwise it will throw an error when executing it, a similar situation with the line number, so there has to be an explanation.

Would something like
@param swash  must be zero "0"

be acceptable?

radio/sdcard/horus/SCRIPTS/WIZARD/heli/wizard.lua
jrwieland: Without the else end (do nothing) , it always adds info to the heli screen.  This information is not needed for FBL helicopters, this screen pertains to flybar helicopters.

instead of a "do nothing" at the end I can move it to the beginning

if TypeFields[1][5]==0 then 
end
else
if TypeFields[2][5]==0 then 
...
else
model.swashRingData(0,0,{type="4" ....
end

It changes the location of the "do nothing" so that the command ends with a flybar choice.

Which method is preferred?

jrwieland: Changed the location of the do nothing  to the front of the command - unwanted results.
With an "else statement" at the end, the swash parameters are always filled in regardless of helicopter type - flybarless or flybar.  (Since it didn't meet any of the other requirements it chose the last one).

Item reverted back to original setting with do nothing at the end.

radio/src/lua/api_model.cpp
jrwieland: language changed to required
Capture

3djc:
@jrwieland there was a number of issues, making it difficult to comment line by line. I have created a branch here : https://github.com/opentx/opentx/tree/3djc/lua-swash

Could you try it and let me know what you think ?

jrwieland:
files updated

jrwieland:
@3djc Just wondering if you had a chance to recheck the heli Lua so it can make the 2.3.15 update when it comes out. 
I see that the edgeTX team  is looking to incorporate the code and tweak the wizard to work with the changes to their code

bsongis:
I merge it, but I propose those changes:
- rename the new function to `setSwashRing` 
- add a function `getSwashRing`

jrwieland:
Thank you for merging it.  Any changes you believe would be beneficial are welcomed

bsongis:
`swashRingData` renamed to `setSwashRing` in origin/2.3

3djc:
Could you do a new PR to add a getSwashRing function (so we have the get/set pair). The other piece has been done already

jrwieland:
Can do.. it'll be a little later today..that work thing is interfering with my fun time

3djc:
No rush ! Thanks for your help actually !

PR #8609, opened 2021-07-09
[radio/src/dump.cpp] Improvement 

JulioCesarMatias:
Improvements only for this extension

### DIFF
--- a/radio/src/dump.cpp
+++ b/radio/src/dump.cpp
@@ -26,31 +26,31 @@
 #define dumpPrintf(...) serialPrintf(__VA_ARGS__)
 #endif
 
+#if defined(DEBUG) || defined(CLI)
 uint8_t dumpPosition;
 
-void dumpStart(unsigned int size)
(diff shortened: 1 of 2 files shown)


radio/src/dump.cpp
```diff
@@ -26,31 +26,31 @@
...
 }
 
-void dumpBody(const void *data, unsigned int size)
+static void dumpBody(const void *data, unsigned int size)
 {
   for (unsigned int i=0; i<size; i++) {
     dumpPrintf("%.2X ", ((uint8_t *)data)[i]);
     dumpPosition++;
-    if ((dumpPosition & (32-1)) == 0) {
+    if ((dumpPosition & 31) == 0) {
```
3djc: The compiler will do the change itself. When we write things like that, it is meant to help understand that our target value is 32, and that a -1 needs to be applied, so given our lastest style, it should be
` if ((dumpPosition & (32 - 1)) == 0) {`

PR #8614, opened 2021-07-10
RadioKing TX18S 

shaman7036:
add RadioKing TX18S support )


raphaelcoeffic:
@shaman7036 apart from the changes to make it different, what is the actual difference to the T18? I see absolutely none.

PR #8636, opened 2021-07-25
WIP: FreeRTOS

### LINKED ISSUE #8588: Flat spots in BF BB Logs caused by late OpenTx pulses delivery
This is going to be a fun one... 

For some time now we have been noticing 'flat spots' in betaflight Black Box logs, which appeared to be due to duplicate packets on the Rx side. 

Instead, what appears to be happening is that synced protocols (GHST for example, but quite probably all of them) occasionally run the mixer late (1ms typically), and the transmitter module has no 'fresh' data to deliver OTA to the receiver. 

To see it happening, setup a logic analyzer/scope to monitor the serial data stream from the radio, and look for packets which are late by approx. 1ms. 

These occur at random times, with a period of 'a few seconds', and no particular pattern. 

Seems to happen on all radios, confirmed on the TX16s, and a XLite Pro (so not Color display related). 

The below scope screenshot shows the problem occurring. The cyan trace being the mixer scheduler timer interrupt, which correctly sets the mixerFlag. The magenta trace being the mixer actually running (mixerSchedulerWaitForTrigger completing without a timeout). 

Myself and 3djc have spent many hours on this so far, with no real progress, and a suspicion that the issue lies in the RTOS itself. Almost seems that the RTOS detects the flag being set, but doesn't correctly task switch. 

image

Some testing that has been performed so far: 
- Change CFG_SYSTICK_FREQ from 1ms to 2ms, mixer runs 2ms late when the problem occurs, as kindof expected
- Switch from using flags to semaphores to signal a mixer run, no change in behavior
- Shut down the audio task, no change

3djc:
Very largely based on https://github.com/EdgeTX/edgetx/pull/499

With much appreciated help from @rotorman 

Fixes #8588

ImmersionRC have been provided with test builds, waiting on their confirmation the issues is fixed by this (internal test suggest it is)

And much more future proof (based on FreeRTOS 2020-12-01 LTS)


radio/src/FreeRTOSConfig.h
```diff
@@ -20,15 +20,20 @@
...
+#endif
 
 #define configUSE_PREEMPTION            1
 #define configUSE_IDLE_HOOK             0
 #define configUSE_TICK_HOOK             0
 #define configCPU_CLOCK_HZ              ( SystemCoreClock )
 #define configTICK_RATE_HZ              ( ( TickType_t ) 1000 )
 #define configMAX_PRIORITIES            ( 5 )
-#define configMINIMAL_STACK_SIZE        ( ( unsigned short ) 130 )
+#define configMINIMAL_STACK_SIZE        ( ( unsigned short ) 30 )
```
raphaelcoeffic, line 36: Is that really sufficient for the idle task? That would be cool! Saves another 100x4 bytes RAM!

3djc: Tested on x7 and tx16s so far, can't see an issue, I GAIN 8 bytes on x7 compared to Coocox with this

radio/src/thirdparty/FreeRTOS/include/FreeRTOS.h
```diff
@@ -175,7 +175,7 @@
 #endif
 
 #ifndef INCLUDE_uxTaskGetStackHighWaterMark
-    #define INCLUDE_uxTaskGetStackHighWaterMark    0
+    #define INCLUDE_uxTaskGetStackHighWaterMark    1
```
rotorman: By FreeRTOS convention
`#define INCLUDE_uxTaskGetStackHighWaterMark  1`
should be defined in https://github.com/opentx/opentx/blob/3djc/FreeRTOS/radio/src/FreeRTOSConfig.h
and here under thirdparty\FreeRTOS the setting should be left unchanged.

PR #8649, opened 2021-08-01
Handle volume correctly also for HELLO

olliw42:
this one really annoyed me forever and made me not use that feature as it made it pointless:

While with a special function the speaker volume can be made adjustable, e.g., with a poti, this did NOT affect also the HELLO startup sound, which was played at full volume (kicking out your ears)

This is because the special functions mixer has not yet run at the time the HELLO was played. Hence, this PR makes the audio thread wait for the mixer to have run at least once if the HELLO should be played. Some additional delay was needed to get a smooth sound, which was determined a bit experimentally. 

Works great for me on Tx16S and T16.

### DIFF
--- a/radio/src/audio.cpp
+++ b/radio/src/audio.cpp
@@ -531,10 +531,9 @@ void audioTask(void * pdata)
 
   if (!globalData.unexpectedShutdown) {
     while (!s_mixer_first_run_done) { // wait until the first mixer run was completed
-      RTOS_WAIT_MS(25);
+      RTOS_WAIT_MS(1);
     }
-    RTOS_WAIT_MS(25);
-    AUDIO_HELLO();
+    RTOS_WAIT_MS(100); // give it some time to update currentSpeakerVolume
   }
 
   while (1) {
--- a/radio/src/audio.cpp
+++ b/radio/src/audio.cpp
@@ -530,6 +530,10 @@ void audioTask(void * pdata)
 #endif
 
   if (!globalData.unexpectedShutdown) {
+    while (!s_mixer_first_run_done) { // wait until the first mixer run was completed
+      RTOS_WAIT_MS(25);
+    }
+    RTOS_WAIT_MS(25);
     AUDIO_HELLO();
   }
 
--- a/radio/src/opentx.cpp
+++ b/radio/src/opentx.cpp
@@ -2028,6 +2028,9 @@ void opentxInit()
 #endif
 
   referenceSystemAudioFiles();
+  if (!globalData.unexpectedShutdown) {
+    AUDIO_HELLO();
+  }
   audioQueue.start();
   BACKLIGHT_ENABLE();
 


raphaelcoeffic:
Funky fix Olli! Much easier than doing it "the correct way", which would require pulling the volume control via pot out of the mixer (same for backlight I guess?).

lshems:
Did you consider checklist and other preconditions? Or are these after the mixer runs once?

raphaelcoeffic:
> Did you consider checklist and other preconditions? Or are these after the mixer runs once?

I believe these are indeed run before the module is/are turned on, but that does not mean the mixer does not run. It is started very early at the same time the UI task is started as well. And it runs at a much smaller frequency when no module is turned on, but still around every 50ms.

lshems:
Just checking. Don't have anything to check myself as I'm abroad.

olliw42:
> Funky fix

yeah ... 

only ... it did give me only few days of happiness ... as it actually does not work satisfactorily

two issues
- lua scripts can play very early after they got started, and as a result the lua sound can come before the HELLO sound ... I've modified the PR, and it does circumvent this now on my Tx16S and T16 ... so, may look ok, one could think, but...
- switch warnings ... they really screw up the idea ... that is, the corresponding sound is ONLY emitted after the main screen started ... really not the right moment LOL
 
> doing it "the correct way", which would require pulling the volume control via pot out of the mixer

I was also first thinking about this, but I think this is going to be a headache, and actually not the "correct way" ... the mixer thing I assume will have soo many dependencies to all get right that it should be called just a flawed approach

I actually came to the conclusion that the basic concept of doing it via SF is just the wrong way ... I understand that this has looked very compelling and elegant, but...

I think what one one should do is to add corresponding options to the radio setup and handle them just right away ... this IMHO also makes much more sense user-wise, since they would be in the radio setup and not in the model setup like now

I don't know what to do with this PR here, unless is gives inspiration, it probably should be closed.

lshems:
I remember a discussion on the opentx GitHub about this.

I suggested back then, that storing the last volume level so it can be used on startup would be perfect.

Does that hint in a direction we could go?

raphaelcoeffic:
> I think what one one should do is to add corresponding options to the radio setup and handle them just right away ... this IMHO also makes much more sense user-wise, since they would be in the radio setup and not in the model setup like now

Exactly, this is what I meant with "correct way". Just check in the UI task the state of the pot, and set the volume. The GF needs to go away then.

lshems:
Nop. That is not what I mean.

If you shut down the radio, you cannot change the volume setting until it is started again.

So by storing the VOLUME LEVEL (sorry for caps) on shutdown, and read it at startup to use as starting volume, no need to do anything related to HOW the volume arrived at a certain level.

3djc:
Clearly not something I would want. Use the radio at a noisy location, auto save, bring it back home, turn on and it screams. No thanks

lshems:
Sorry, but EVERY device you have works like that.

But hey, everything I say is not true as well.

lshems:
Sorry, couldn't hear you, my phone was put on mute after I restarted it. 

Really, you are not serious.

But ok, I'll stop posting on opentx.

raphaelcoeffic:
> Clearly not something I would want. Use the radio at a noisy location, auto save, bring it back home, turn on and it screams. No thanks

Me neither, as this would be reproducing a similar behaviour we have right now: having a volume that does not correspond to what the pot says (even if only for a sec or 2).

lshems:
Not at all. The pot determines what the volume is on shutdown. Not on startup. 

Most of times you have the radio on the couch, need to restart, and ten the volume is not blockable. That's where the annoyance is.

But ok. My suggestion is clearly not an improvement to ALWAYS having max volume on start up in your opinion.

3djc:
> But ok, I'll stop posting on opentx.

I should have known not to keep my hopes high ...

3djc:
> But ok. My suggestion is clearly not an improvement to ALWAYS having max volume on start up in your opinion.

 Obviously you know that startup volume is already for you to choose, doesn't have to be 100% if you prefer say 30%

PR #8650, opened 2021-08-02
USB telemetry mirror by vimalb

3djc:
Allow mirroring telemetry (in native format) on USB.

Given only 2 bits are available for mode in EEprom, you have either serial when DEBUG is set, or telem when non DEBUG builds


olliw42:
great

can I make a suggestion:

I had a similar issue in my MAVLink for OpenTx project, when I wanted to add MAVLink as an selectable option (here it was a 4 bits issue in the model struct). I solved it by placing further extra bits at the end of the structure, and introducing methods to get and set the value (see https://github.com/olliw42/opentx/blob/mavtelem2314-v30.1/radio/src/datastructs.h#L403-L514). Please note that this is not a revolutionary new approach for OpenTx; composing helper methods also had been taken before for the rfProtocol field (see same link).

I think the same could - should(?) - be done here too. The limitation with respect with debug is not so nice, and the issue may come up again and again in future (e.g., why can't GPS be via usb, etc. pp).

If you want to, I can make a respective change and PR it, on top of which this here could be placed.

olliw42:
@3djc 
question: the merge of this, does it mean that the idea to extend the bitrange by helper functions along the lines of what is done for rfProtocol is not desired, or would a PR which does do it be appreciated and considered for merge?

3djc:
Not necessarily closed to the idea, but do we have spare bits in radio data section? (Did not check)

olliw42:
I can't exactly answer this from top of my head and am not sure in how far it would be relevant, all I can say is that I was joyfully adding all sorts of bits and bytes to the structures and had not yet any issue with backwards compatibility. Might have been pure luck, but given the number of bist&bytes I#ve added/changed it would have been quite a lucky punsh :)

Adding such helpers would be quite a substantial amount of work (at least it was when I did it for the case I mentioned in the above), hence I would appreciate knwoing in before the chances :)

What I think one (I) can quikcly do is to add the bits to the structure without doing the helpers and see if the addition has any sideeffect I can observe. Does this make sense?

3djc:
Well you cannot change the size of radio data for existing radio without facing consequences, like invalid eeprom if opened with an older version

olliw42:
one needs to adapt the size checkers to the new size for sure, but with doing that I could not yet observe any negative sideeffects
e.g. for MAVLink for OpenTx I'm doing this: https://github.com/olliw42/opentx/blob/mavtelem2314-v30.1/radio/src/datastructs.h#L1017-L1027

3djc:
Have you opened eeprom with an older firmware or version of companion?

olliw42:
I do not use companion (except for flashing), i.e. am totally incompetent on the matter

btw, by looking (again) at what I did I note that I actually didn't do the handler function magic for the USBMode field, here I simply marked the original ones as spare and added new bits below (in contrast to what I did - I presume had to do - in the ModelData struct), 
https://github.com/olliw42/opentx/blob/mavtelem2314-v30.1/radio/src/datastructs.h#L844-L847
https://github.com/olliw42/opentx/blob/mavtelem2314-v30.1/radio/src/datastructs.h#L867-L875

there is btw a spare bit: https://github.com/olliw42/opentx/blob/mavtelem2314-v30.1/radio/src/datastructs.h#L802

I wonder however if it wouldn't be smart to add additional spare bits now ... I mean, the time will likely come when one has to do it anyhow ...

PR #8656, opened 2021-08-06
Aux2: DMA stream incorrect, was not RX but TX

olliw42:
title says it

### DIFF
--- a/radio/src/targets/horus/hal.h
+++ b/radio/src/targets/horus/hal.h
@@ -444,7 +444,7 @@
   #define AUX2_SERIAL_GPIO_PinSource_TX        GPIO_PinSource14
   #define AUX2_SERIAL_GPIO_PinSource_RX        GPIO_PinSource9
   #define AUX2_SERIAL_USART_IRQHandler         USART6_IRQHandler
-  #define AUX2_SERIAL_DMA_Stream_RX            DMA2_Stream6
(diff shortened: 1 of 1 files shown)


rotorman:
Good catch (USART6 RX can only be DMA2 Stream 1 or 2 and Channel 5) and crazy that this has not caused issues earlier:
grafik

olliw42:
it wasn't used with dma by anything so far, I think, only non-dma

3djc:
Yeap, maybe we should move to dma too, will check when I come back :)

olliw42:
I think for relatively slow and short transitions dma is a bit of overhead
as much as I could see there is no really traffic intense use of the aux2 currently

(I discovered the bug when I started using dma for my mavlink stuff some time back)

btw, as it is currently it also only works for rx, and not tx, if one uses dma, that's because of the nvic is not enabled. I haven't put this into this pr here since aux2 is only used for rx, as much as I can see, so correcting it to work also for tx would be technically a functional change

raphaelcoeffic:
By the way, I found this to be a very useful resource: https://stm32f4-discovery.net/2017/07/stm32-tutorial-efficiently-receive-uart-data-using-dma/

Please note the `DMAFifo` already works like that, however without `HC` or `TC`. Just plain streaming.

3djc:
This fixes #8677

3djc:
Thanks

PR #8665, opened 2021-08-16
Add TBS Tango2/Mambo support

ghost:
Pull request to merge FreedomTX into OpenTX. Code tested on all major hardware platforms. Companion support is currently not tested, but will be worked on in the next step.


3djc:
The Companion part is not working yet with this. Since it is already a fairly large chunk, we asked things to be splited to ease the review process

wbertrand:
Hi,
Just to let you know that since the last TBS Tango2 update the playing of time units (min,sec) is not working anymore (at least in french)
Bertrand W

PR #8679, opened 2021-08-23
Improve module protocol conversion in Companion

elecpower:
Only allow supported protocols for each module on conversion


3djc:
Great idea

Issue #8685, opened 2021-08-28
Changing the communication speed of ports AUX1, AUX2


dobosatti:
Is it possible to change the baud rate of ports AUX1, AUX2?
My antenna tracker can only communicate on 57600.

3djc:
This place is not for questions or end user support, but to track OpenTX issues. Please go to community sites like rcgroups or openrx forums for community support

PR #8686, opened 2021-08-28
TX16S : mirror FrSky telem at 57k

3djc:
Looking for feedbacks, for 57k seems more logical for AUX FrSky telemetry mirror

### DIFF
--- a/radio/src/telemetry/frsky.h
+++ b/radio/src/telemetry/frsky.h
@@ -34,11 +34,7 @@ enum FrSkyDataState {
 };
 
 #define FRSKY_SPORT_BAUDRATE          57600
-#if defined(RADIO_TX16S)
-#define FRSKY_TELEM_MIRROR_BAUDRATE   115200
-#else
 #define FRSKY_TELEM_MIRROR_BAUDRATE   FRSKY_SPORT_BAUDRATE
-#endif
 
 #define FRSKY_D_BAUDRATE          9600
 


brainbubblersbest:
Whooops. A lot of users in the AAT Community already use that connection with 115200baud.

Because a long time OTX dont supported AUX baudrate changes, the Tracker Projects changed their used Baud speeds instead. 

This commit would cause i compatibility again and straight away. 

A lot better would be a Baudrate setting by Users Choice.

A Big Enhancement would be to  change the Baudrate on both Aux Serials in General for enhanced Pheriphal compatibility. 57600/115200/450000
#8685 

Please @3djc !

This would be Nice for Users of that Connection
And of Course for the Mav Link Branch in combination with different Telemetry Radio Types.

3djc:
Choice is unfortunatly not an option before 2.4 as we are out of eeprom space

brainbubblersbest:
Thank you for that acceptable argument.
As another option:
Maybe that commit can wait then till 2.4?
I Just expect alot of Users struggling again, 
with setup their Hardware again over AT Commands.
and Custom Compiled Tracker FWs and/or Applications after the next OpenTX Update.
I just wanna mention my Thoughts. 
After all its the Decission from you Devs.

3djc:
Changing it now would probably create more issues than it solves, so will be done when we can introduce speed choices

Issue #8710, opened 2021-09-21
widget timer2 does not show days


offer-shmuely:
**To Reproduce**
1. set persitent timer 
2. run the timer for more than 24 hours
3. display the timer with widget "timer2"

widget display:
39:41

**Expected behavior**
1d 15:41

![image](https://user-images.githubusercontent.com/7026911/134149376-77a0d97c-3078-48ee-ad6e-a2e0c1ba7b8a.png)

![image](https://user-images.githubusercontent.com/7026911/134149406-98127e2c-6d02-418a-8c17-7424c4226a61.png)

3djc:
We got a request to actually make it display hours... but since this is a LUA widget, you can do your own version that displays days as you want. Maybe you should use the timer widget instead ?

offer-shmuely:
now it does.
if the time is below 1 hours --> it will display mm:ss
if the time is above 1 hours but below 1 day --> it will display hh:mm:ss
if the time is above 1 day --> it will display dd hh:mm:ss

Note: I use the timer3 for recrding total fllight time of the model, so it reach high values

I do have my own version,  but I like to share it with others.
BTW, the "timer" widget have the samem problem, and it is too small for reading during flight

MlCHEL1234:
Unbelievable - another example of how OpenTX implements sensible user suggestions.

No other system will be able to keep up with this.

kilrah:
I prefer seeing it in hours and no doubt others will, so maybe make it a widget option to enable the "day" display.

offer-shmuely:
added option "use_days"

PR #8711, opened 2021-09-21
widget timer2: fix display above 24hours (issue: #8710)

offer-shmuely:
widget timer2: fix display above 24hours
widget timer2: font size relative to timer value


offer-shmuely:
moved to edgeTx

PR #8714, opened 2021-09-21
Fix model input trims

### LINKED ISSUE #8707: Companion not updating 'Include Trim' when adding Input rows
**Describe the bug**
When creating a model in Companion 2.3.14, sending it to the radio, and reading it back, that the Include Trim value was changing from Yes to No (Inputs tab list would say NoTrim), and then realised that Companion is not showing that, but either is setting it correctly or the transmitter is correcting it.

**To Reproduce**
1. In an existing or new model
2. Goto Inputs tab, and an input, setting the source to 'MAX'
3. Notice that the 'Include Trims' is still saying Yes (at this point, on the radio, it would say 'No')
4. Write the model to the radio, read it back, and see that the Inputs page now says 'NoTrim' for that input (when it didn't say it before)

**Expected behavior**
That Companion would change between Include Trims On and Off as required, similar to the actual radio behaviour. 

**Screenshots**
[Video] Creating the input line in Companion: https://user-images.githubusercontent.com/5500713/133980525-bf177c5e-52ce-4490-a3dd-497b4006f923.mp4

[Video] Creating the input line in Simulator:
https://user-images.githubusercontent.com/5500713/133980569-f2423b1d-04f3-4340-ab92-d728a88e9124.mp4

After reading back from RM TX16S:
companion notrim bug - read from transmitter

**General information**
 - OpenTX 2.3.14
 - RM TX16S
 - OpenTX Companion 2.3.14

elecpower:
Fixes #8707 
Enable/disable scale fields rather than show/hide for a cleaner interface

### DIFF
--- a/companion/src/firmwares/input_data.cpp
+++ b/companion/src/firmwares/input_data.cpp
@@ -20,6 +20,8 @@
 
 #include "input_data.h"
 #include "radiodataconversionstate.h"
+#include "eeprominterface.h"
+#include "compounditemmodels.h"
 
 void ExpoData::convert(RadioDataConversionState & cstate)
 {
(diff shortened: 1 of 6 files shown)


3djc:
Thx :)

PR #8718, opened 2021-09-24
Implement top bar widget layout for Timer2

pfeerick:
Timer2 never had the display code present for top bar, so would display nothing if used as a top bar widget.

image


3djc:
Thanks

PR #8738, opened 2021-10-08
Remove tx battery calibration offset from Companion radio settings

### LINKED ISSUE #8676: Incorrect Battery Offset Displayed in Companion
I have calibrated the display of the battery voltage in my Horus X10S Express (OpenTx 2.3.14) with a precise multimeter. To match the voltage display of the Radio with the value indicated on the multimeter, I had to increase the voltage value by 40mV. When I then imported the OTX file into Companion, a battery offset of 0,4V was displayed on the hardware page. Not a big issue but I thought it's worth mentioning.

elecpower:
Fixes #8676

### DIFF
--- a/companion/src/firmwares/generalsettings.cpp
+++ b/companion/src/firmwares/generalsettings.cpp
@@ -494,20 +494,6 @@ FieldRange GeneralSettings::getPPM_MultiplierRange()
 }
 
 //  static
-FieldRange GeneralSettings::getTxVoltageCalibrationRange()
-{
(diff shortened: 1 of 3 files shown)


elecpower:
After more testing realised the field holds the voltage offset as the number of voltage steps +/-127. Each step represents a value specific to the radio. As the field is a hardware calibrated value and its accuracy when displayed in Companion even with a per radio step will have an error factor then IMHO best to not display at all. So I will modify this PR.

PR #8756, opened 2021-10-26
Fix mix list highlighting

### LINKED ISSUE #8746: Companion 2.3.14: Selecting "Toggle highlight" on the Mixes page produces a weird result.
In Companion 2.1.9, right-clicking a mix line and then left-clicking _Toggle_ _highlight_ changes the selected channel number and its associated label to bold type, and the function only allows one channel at a time to be highlighted. If you then switch tabs away from and back to the Mixes page the highlighted channel remains highlighted, but the highlighting is lost when the editing window is closed. Apparently this was the intended behavior of the _Toggle_ _highlight_ option, but as of version 2.2 it behaves very differently.   

Now, in versions 2.2.4, 2.3.14, and presumably all Companion versions in between, right-clicking any mix line and left-clicking _Toggle_ _highlight_ causes all of the channel numbers for channels with entries, and all of their associated labels, to disappear. Nothing remains in the left column of the Mixes page except the channel numbers and labels of mix channels with no entries, and those appear in gray.

And if you switch to a different tab and back again you'll find that the page has returned to normal, and if the channel you intended to highlight has one or more entries, the highlighting will be there.

You might decide to simply remove the _Toggle_ _highlight_ function since its usefulness seems questionable, or you might decide to fix it, but I don't think you'd want to leave it as it is.

elecpower:
Fixes #8746

### DIFF
--- a/companion/src/modeledit/mixes.cpp
+++ b/companion/src/modeledit/mixes.cpp
@@ -355,7 +355,7 @@ void MixesPanel::mixersDuplicate()
 void MixesPanel::mixerOpen()
 {
   QListWidgetItem *item = mixersListWidget->currentItem();
-  if (item == nullptr) 
+  if (item == nullptr)
     return;
 
   int idx = item->data(Qt::UserRole).toByteArray().at(0);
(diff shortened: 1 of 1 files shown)


3djc:
Thanks !

PR #8760, opened 2021-10-29
Fix lua error message for small LCDs

klutvott123:
It's not guaranteed that the error string contains '/'. ```strrchr``` will return a null pointer if the character isn't found and the result is that we get only the error title and no message. 
The companion will crash if this happens in the sim.


3djc:
Thanks

Issue #8780, opened 2021-11-19
Jumper T-Lite switch locations


FelixSchreiber:
In the radios Main View the switch locations are not ordered according to the real hardware switch locations.

On the radio, C and A are left to the antenna,  B and D are right to it; C is the leftmost and D the rightmost switch. C and D are located more to the back whereas A and B are located more to the front of the radio.
In the main view, A and B are left to the left stick, C and D are right to the right stick.
![T-Lite_switches](https://user-images.githubusercontent.com/94545140/142624999-0d75ff0e-1205-415e-99ed-8ac4009ac8da.png)
A more intuitive order would e.g be :
Left to the left stick: First Switch A and below switch C
Right to the right stick: First Switch B and below switch D

In Companion the switches are ordered A, B, C, D (from left to right)
![SimulatorSwitches](https://user-images.githubusercontent.com/94545140/142625052-dbc3e83d-9512-4572-8c91-dd38a590c57f.png)
A more intuitive order would be according to the real switch locations: C, A, B, D (from left to right)

Furthermore it would be very helpful if the other buttons in the radio image could be named (as long as you are not fully used to this radio and know it by heart).
![SimulatorGraphics](https://user-images.githubusercontent.com/94545140/142625080-0e490265-1de4-42c7-8cc2-e6bef8760b34.png)

**General information**
 - OpenTX version 2.3.14
 - Hardware Jumper T-Lite
 - OpenTX Companion .3.14

3djc:
Switch are alphabeticaly ordered by design

FelixSchreiber:
sorry, but this answer is too simple and doesn't comply with reality (probably you are referring only to the Companion?).

In radio/src/gui/128x64/view_main.cpp there are for example many different arrangements for the switches defined.
Why not add something similar for this radio like

```
#elif defined(PCBXLITES)
        static const uint8_t x[NUM_SWITCHES] = {2*FW-2, 16*FW+1, 2*FW-2, 16*FW+1};
        static const uint8_t y[NUM_SWITCHES] = {4*FH+1, 4*FH+1, 6*FH+1, 6*FH+1};
        for (int i=0; i<NUM_SWITCHES; ++i) {
          if (SWITCH_EXISTS(i)) {
            getvalue_t val = getValue(MIXSRC_FIRST_SWITCH + i);
            getvalue_t sw = ((val < 0) ? 3 * i + 1 : ((val == 0) ? 3 * i + 2 : 3 * i + 3));
[...]

PR #8781, opened 2021-11-19
Display switches in main view according to their phhysical locations

FelixSchreiber:
solves point 1 in #8780 by rearranging Jumper T-Lite switches in main view according to their physical location

screenshot_tlite_21-11-19_17-59-06


3djc:
Wrong solution, will deal with it when I have a minute (and delete that wrong xlite code too)

Issue #8783, opened 2021-11-21
Quad Wizard uses wrong inputs in mixers


FelixSchreiber:
On a radio with default channel order AETR, when creating a new model using the Quad wizard, the inputs Rud and Ail are mixed to the wrong output channels.

1. Default channel order set up to AETR
2. Create new model, using Quad wizard
3. Accept all default settings (Throttle=Ch3, Roll=Ch1, Pitch=Ch2, Yaw=Ch4, ...)
4. Open Inputs page. Inputs are ordered Ail, Ele, Thr, Rud
4. Open Mixes page in model. Rud input is assigned to Ch1 (named Roll) and Ail input is assigned to Ch4 (Yaw)
![screenshot_tx16s_21-11-21_02-54-18](https://user-images.githubusercontent.com/94545140/142746536-d2882ed5-6a86-4f30-83d6-ad6d3c350d45.png)

**Expected behavior**
Rud input should be assigned to Ch4 and Ail input should be assigned to Ch1

**General information**
 - OpenTX version 2.3.14
 - Hardware Radiomaster Tx16S
 - OpenTX Companion 2.3.14

PR #8784, opened 2021-11-21
Horus SD - quad wizard bugfix

FelixSchreiber:
solves #8783 and also lets you select switch G for arm, beeper and modes (previously only up to switch F)


3djc:
Thanks

PR #8832, opened 2022-01-18
Function switch : fix toggle switch

3djc:
Fix T-Pro toggle behavior for function switches

Thanks @pfeerick for the report and tests


pfeerick:
To give credit where it is due, it was Joshua Bardwell who actually found this... I just confirmed and tested the fix ;)

To follow on with what he was commenting on re function switch behaviour, a third state would be nice... lets call it "Latched"... whereby you can exactly mimic the behaviour of the 6Pos, as when you press, it turns on, and if you press again, it stays on. If you press another "latch" button in the group, it turns on, all the others turn off. i.e. one must always be on, and is mutually exclusive. If you think this is worth pursuing, I can open an issue for this?

3djc:
You can, I have a test version for this ready :D

PR #8843, opened 2022-01-23
Companion function switches add data item models and include in model print

elecpower:
Sync with EdgeTX


elecpower:
Note: this needs to be merged before T Pro fixes WIP

elecpower:
These changes were successfully merged as part of #8847

PR #8851, opened 2022-01-30
Companion recover memory on model edit dialog close

### LINKED ISSUE #8627: Companion 2.3.13 and Windows 10: Companion gradually slows down with extended usage.
I've found that when using Companion 2.3.13 over long periods to edit and simulate model setups, the program gradually slows down, with ever-growng lags between mouse clicks and program response, until it becomes virtually unusable. At that point I can save my working otx file, close Companion, open it again, reload the otx file, and everything will be normal until the lag slowly builds up again.

Using Task Manager, I've found that right after starting Companion and loading an otx file, memory usage is about 25 MB and power usage is "Low".

Fast

As the lag between mouse clicks and program response grows during use, I've found that memory usage also grows, and power usage progresses from "Low" at program start to "Very high" when lag is extreme.

Slow
  
I've seen memory usage climb to over 500 MB and it would presumably have climbed higher if I had let Companion keep running.

I don't know much about how Windows 10 works, but my impression is that some fault in Companion is causing it to consume an ever-increasing amount of computer resources. I don't recall ever encountering another program that behaved like this.

elecpower:
Partially fixes #8627
Forces memory recovery on dialog close.

### DIFF
--- a/companion/src/modeledit/modeledit.cpp
+++ b/companion/src/modeledit/modeledit.cpp
@@ -48,6 +48,7 @@ ModelEdit::ModelEdit(QWidget * parent, RadioData & radioData, int modelId, Firmw
 
   ui->setupUi(this);
   setWindowIcon(CompanionIcon("edit.png"));
+  setAttribute(Qt::WA_DeleteOnClose);
   restoreGeometry(g.modelEditGeo());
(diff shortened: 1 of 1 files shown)


3djc:
Thanks

PR #8856, opened 2022-02-03
Show default names in Companion if switch/source names only spaces

elecpower:
Port of https://github.com/EdgeTX/edgetx/pull/1524


3djc:
Thanks @elecpower

PR #8857, opened 2022-02-03
Some modules improvements

3djc:
Improvement for proper ELRS support at full speed, including updated GUI to see module link performance
Support for high speed serial nano modules on RM Zorro (r9m lite in ACCESS mode for example)


CapnBry:
Is the Zorro an F407 clocked at 168MHz? If so, shouldn't it be able to do 5250000 baud at 16x oversampling still? I'll admit I don't know where the RCC setup occurs in OpenTX :(

3djc:
I have a bit more bit errors than I would like to at 5.35, so down a notch at 3.75, far less errors

CapnBry:
> I have a bit more bit errors than I would like to at 5.35, so down a notch at 3.75, far less errors

On the internal module? I get zero bit errors at 5.25, only overflow errors. There do seem to be less overflow errors at lower bauds, but there's a sharp uptick going from 400k to 921k and then just moderately more all the way up to 5.25M. I mentioned this in my more robust resync PR on EdgeTX but I am not sure where the RX interrupt is getting starved.
https://github.com/EdgeTX/edgetx/pull/1533

PR #8876, opened 2022-03-12
MPM Updates to 1.3.3.11 for Companion and Radio GUI

MRC3742:
@3djc as it appears 2.3.15 is near to release.
These small changes, additions could be added to bring Companion and Radio GUI up to the current MPM master branch 1.3.3.11

Thanks as always - Rich


3djc:
Just in time ! Thanks Rich !

PR #8878, opened 2022-03-12
MPM_Update_MissingStrings

MRC3742:
@3djc This should fix the compile errors on Companion and Simulator

Thanks - Rich


3djc:
Thanks Rich, seems much better  :)

PR #8892, opened 2022-03-27
about screen cosmetics for 128x64

TheIsotopes:
this fix the about screen headline for 128x64 displays e.g. Zorro


3djc:
Any specific reason for closing it ?

TheIsotopes:
If I deleted it while cleaning, feel free to copy it.

PR #8895, opened 2022-04-01
Show MPM version for internal and external modules

pfeerick:
Rather than just show internal MPM on radios with HARDWARE_EXTERNAL_ACCESS_MOD, 
show for all radios where present. And also show for external MPM if present.

It would also be possible to show the CRSF module info here, but IMO it's not worthwhile yet
as we've not figured out the cleanest way way to pull the version info from the module yet.


3djc:
Really unsure why it was done this way !  Must have been smoking somethign realy strong that day :D

pfeerick:
Domino effect and feature creep? :grin:

TheIsotopes:
@pfeerick Do not see anything on the version screen of my tx16s.

pfeerick:
That would be because the module info entry is only shown for PXX2... doh! https://github.com/opentx/opentx/pull/8903

PR #8903, opened 2022-04-13
Followup #8895 - actually show module info entry

pfeerick:
Enable the `Modules / RX version` entry if MULTIMODULE

I didn't have a OTX build environment set up when doing #8895 so only focused on the actual `menuRadioModulesVersion` code... and neglected to notice it was only active for `PXX2`. Have since tested this on TX16S hardware and the menu entry is now visible, and shows the module information.


3djc:
Thanks, and sorry for no catching up this one

pfeerick:
*This never happened* ... the commit log is a lie ;)

PR #8905, opened 2022-04-15
Add iFlight Commando8 ELRS

XING-IF:
Add custom code for iFlight Commando 8 ELRS

Specs and software adaptations:
- Chipset STM32F407VET6
- Internal ELRS + Multi protocol transmitter on dual-band or single band antenna
- Only internal or external module can be active; not both!

Product link: https://shop.iflight-rc.com/Commando-8-Radio-Transmitter-Pro1696?search=commando
image


radio/src/gui/128x64/model_setup.cpp
```diff
@@ -1074,15 +1074,34 @@ void menuModelSetup(event_t event)
...
-                  if (checkIncDec_Ret) {
+                  #if defined(RADIO_Commando8)
+                    if (reusableBuffer.moduleSetup.newType!=MODULE_TYPE_NONE)
+                    {
+                      reusableBuffer.moduleSetup.newType=MODULE_TYPE_NONE;
+                      g_model.moduleData[moduleIdx].type = reusableBuffer.moduleSetup.newType;
+                      reusableBuffer.moduleSetup.previousType = reusableBuffer.moduleSetup.newType;
+                      setModuleType(EXTERNAL_MODULE, MODULE_TYPE_NONE);
+                    }
+                  #endif
```
raphaelcoeffic, line 1088: I believe you should better implement this behaviour inside `isInternalModuleAvailable()`. Basically by returning `false` to everything, or even better, by not defining `HARDWARE_INTERNAL_MODULE` (probably the better solution).

radio/src/gui/128x64/model_setup.cpp
```diff
@@ -1074,15 +1074,34 @@ void menuModelSetup(event_t event)
...
                     setModuleType(moduleIdx, moduleType);
                   }
                 }
                 else
+                {
 #endif
-                  reusableBuffer.moduleSetup.newType = checkIncDec(event, reusableBuffer.moduleSetup.newType, MODULE_TYPE_NONE, MODULE_TYPE_MAX, 0, isExternalModuleAvailable);
+                    reusableBuffer.moduleSetup.newType = checkIncDec(event, reusableBuffer.moduleSetup.newType, MODULE_TYPE_NONE, MODULE_TYPE_MAX, EE_MODEL,isExternalModuleAvailable);
+                  #if defined(RADIO_Commando8)
+                    if (checkIncDec(event, g_model.moduleData[moduleIdx].type, MODULE_TYPE_NONE, MODULE_TYPE_MAX, EE_MODEL,isInternalModuleAvailable)!=MODULE_TYPE_NONE)
```
raphaelcoeffic, line 1099: What is the logic implemented here? I'm not sure I understand it properly. It seems however not really the proper solution to call `checkIncDec()` here again.

radio/src/gui/128x64/radio_setup.cpp
```diff
@@ -543,14 +543,18 @@ void menuRadioSetup(event_t event)
         lcdDrawTextAlignedLeft(y, STR_PWR_ON_DELAY);
         lcdDrawNumber(RADIO_SETUP_2ND_COLUMN, y, g_eeGeneral.getPwrOnSpeed(), attr|LEFT);
         lcdDrawChar(lcdLastRightPos, y, 's');
-        if (attr) CHECK_INCDEC_GENVAR(event, g_eeGeneral.pwrOnSpeed, -1, 2);
+        #if !defined(RADIO_Commando8)
```
raphaelcoeffic, line 546: If you wish to disable `pwrOnSpeed` and `pwrOffSpeed`, I believe you should use a `define` that is more explicit, and add it to the proper section in the target specific `CMakeLists.txt` (probably in `radio/src/taranis/CMakeLists.txt`).

radio/src/gui/128x64/radio_setup.cpp
```diff
@@ -543,14 +543,18 @@ void menuRadioSetup(event_t event)
...
+          if (attr) CHECK_INCDEC_GENVAR(event, g_eeGeneral.pwrOnSpeed, -1, 2);
+        #endif  
         break;
 
       case ITEM_RADIO_SETUP_PWR_OFF_SPEED:
         lcdDrawTextAlignedLeft(y, STR_PWR_OFF_DELAY);
         lcdDrawNumber(RADIO_SETUP_2ND_COLUMN, y, g_eeGeneral.getPwrOffSpeed(), attr|LEFT);
         lcdDrawChar(lcdLastRightPos, y, 's');
-        if (attr) CHECK_INCDEC_GENVAR(event, g_eeGeneral.pwrOffSpeed, -1, 2);
+        #if !defined(RADIO_Commando8)
```
raphaelcoeffic, line 555: See previous comment.

radio/src/opentx.cpp
```diff
@@ -301,7 +301,8 @@ void generalDefault()
 #if defined(LCD_CONTRAST_DEFAULT)
   g_eeGeneral.contrast = LCD_CONTRAST_DEFAULT;
 #endif
-
+//Modified T8 applies to commando8
+  g_eeGeneral.contrast = 25;
```
raphaelcoeffic, line 305: You should define `LCD_CONTRAST_DEFAULT` to the value suitable for the radio instead of this.

radio/src/opentx.h
```diff
@@ -33,6 +33,11 @@
 
 #include "board.h"
 
+//Modified T8 applies to commando8
+#if defined(RADIO_Commando8)
+#undef VARIO
```
raphaelcoeffic, line 38: Is there a particular reason for preventing the use of a variometer?

radio/src/pulses/modules_helpers.h
```diff
@@ -644,7 +644,13 @@ inline void resetMultiProtocolsOptions(uint8_t moduleIdx)
   else {
     g_model.moduleData[moduleIdx].multi.autoBindMode = 0;
   }
+
+#if defined (RADIO_Commando8)
+  g_model.moduleData[moduleIdx].multi.optionValue = -27;
```
raphaelcoeffic, line 649: This should not be necessary, please explain.

radio/src/targets/common/arm/stm32/audio_dac_driver.cpp
```diff
@@ -59,7 +59,12 @@ void dacInit()
   GPIO_InitStructure.GPIO_OType = GPIO_OType_PP;
   GPIO_InitStructure.GPIO_PuPd = GPIO_PuPd_UP;
   GPIO_Init(AUDIO_MUTE_GPIO, &GPIO_InitStructure);
+#if defined(RADIO_Commando8)
+  GPIO_ResetBits(AUDIO_MUTE_GPIO, AUDIO_MUTE_GPIO_PIN);
```
raphaelcoeffic: If an inverted logic is required here, it is probably better to define a more generic macro/function which has a specific implementation for this radio in its `board.h`/`board.cpp`.

raphaelcoeffic: All of the further specific definitions in this file would be avoided.

radio/src/targets/common/arm/stm32/bootloader/boot.cpp
```diff
@@ -268,7 +268,9 @@ int main()
 
   // init screen
   bootloaderInitScreen();
-
+#if defined(RADIO_Commando8)   //Modified T8 applies to commando8
+    lcdSetRefVolt(25);
```
raphaelcoeffic, line 272: This should be in `bootloaderInitScreen()`.

radio/src/targets/common/arm/stm32/bootloader/boot.cpp
```diff
@@ -509,16 +511,27 @@ int main()
...
 
 #if defined(RTC_BACKUP_RAM)
       rtcInit();
       RTC->BKP0R = SOFTRESET_REQUEST;
 #endif
 
+      #if defined(RADIO_Commando8)
+      lcdClear();
+      lcdDrawText(2, 22,"Press the power button.");
+      lcdDrawText(2, 33,"Exit the flashing mode.");
```
raphaelcoeffic, line 523: Why should this be specific to this radio? Either it is some new feature / improvement, which should be justified separately and applied to all similar radios, or it shall not be here at all. In general, UI code should not be included directly in the generic bootloader code, as it is common to all radios, with completely different screen sizes, orientation, etc.

radio/src/targets/common/arm/stm32/sticks_pwm_driver.cpp
```diff
@@ -117,10 +117,26 @@ extern "C" void PWM_IRQHandler(void)
...
+#define _CHANNEL3     3
+
+#else
+
+#define _CHANNEL0     0
+#define _CHANNEL1     1
+#define _CHANNEL2     3
+#define _CHANNEL3     2
+
+#endif
```
raphaelcoeffic, line 134: I'd need to check, but there is probably a better way to achieve channel mapping (especially with a more meanful `define` name: `CHANNEL_1023` is a bit cryptic, isn't it?).

radio/src/targets/taranis/CMakeLists.txt
```diff
@@ -219,6 +219,40 @@ elseif(PCB STREQUAL X7)
...
+    if(iFlight_RELEASE)
+      set(BIND_KEY NO)
+    endif()
+    add_definitions(-DRADIO_Commando8)
+    add_definitions(-DCommando8_Float_Show_OFF) #Commando8_Float_Show_OFF
+    add_definitions(-DCommando8_STICKS_PWM_EN) #Commando8_STICKS_PWM_EN
+    add_definitions(-DiFlight_RELEASE)
+    add_definitions(-DEEPROM_VARIANT=0x4008)
+    add_definitions(-DMANUFACTURER_iFlight)#add_definitions(-DMANUFACTURER_RADIOMASTER)
+    add_definitions(-DCHANNEL_1023)
```
raphaelcoeffic: @3djc can you please have a look at the defines above? That does not really look like the way it is done normally.

radio/src/targets/taranis/board.h
```diff
@@ -660,14 +665,15 @@ enum Analogs {
   #define NUM_TRIMS_KEYS                (NUM_TRIMS * 2)
 #endif
 
-#if defined(STICKS_PWM)
+#if defined(STICKS_PWM) || defined(Commando8_STICKS_PWM_EN)
```
raphaelcoeffic: Why not using just `STICKS_PWM` ?

raphaelcoeffic (changes requested):

companion/src/CMakeLists.txt
```diff
@@ -351,6 +351,8 @@ elseif(PCB STREQUAL X7 AND PCBREV STREQUAL TX12)
   set(FLAVOUR tx12)
 elseif(PCB STREQUAL X7 AND PCBREV STREQUAL ZORRO)
   set(FLAVOUR zorro)
+elseif(PCB STREQUAL X7 AND PCBREV STREQUAL Commando8)
```
3djc: Please follow casing rules, all those are upper case

companion/src/CMakeLists.txt
```diff
@@ -351,6 +351,8 @@ elseif(PCB STREQUAL X7 AND PCBREV STREQUAL TX12)
...
+  set(FLAVOUR Commando8)
```
3djc: all lower

companion/src/firmwares/boards.cpp
```diff
@@ -96,6 +96,8 @@ uint32_t Boards::getFourCC(Type board)
       return 0x4778746F;
     case BOARD_RADIOMASTER_T8:
       return 0x4378746F;
+    case BOARD_iFlight_Commando8:
```
3djc: all UPPER, I stop here, I think you got the idea

companion/src/firmwares/opentx/opentxinterface.cpp
```diff
@@ -1406,6 +1416,17 @@ void registerOpenTxFirmwares()
...
+  /* iFlight Commando8 board */
+  firmware = new OpenTxFirmware("opentx-Commando8", QCoreApplication::translate("Firmware", "iFlight Commando8"), BOARD_iFlight_Commando8);
+  addOpenTxCommonOptions(firmware);
+  firmware->addOption("noheli", Firmware::tr("Disable HELI menu and cyclic mix support"));
+  firmware->addOption("nogvars", Firmware::tr("Disable Global variables"));
+  firmware->addOption("lua", Firmware::tr("Enable Lua custom scripts screen"));
+  addOpenTxFontOptions(firmware);
+  registerOpenTxFirmware(firmware);
+  addOpenTxRfOptions(firmware, NONE);
+  //firmware->addOption("bindkey", Firmware::tr("Allow bind using bind key"));
```
3djc: remove line

radio/src/gui/128x64/radio_setup.cpp
3djc: and why disable it to start with ?

radio/src/gui/screenshot.cpp
```diff
@@ -62,9 +62,11 @@ const char * writeScreenshot()
     return error;
   }
 
+#if defined(RTCLOCK) //This is a bug. An error is reported when RTC is closed.
```
3djc: fix the bug in strAppendDate sounds like a better option

radio/src/main.cpp
```diff
@@ -81,7 +81,9 @@ void handleUsbConnection()
           POPUP_MENU_ADD_ITEM(STR_USB_SERIAL);
 #endif
 #if defined(USB_SERIAL)
+  #if !defined(RADIO_Commando8)
```
3djc, line 84: any reason why usb serial telem would not work ?

radio/src/opentx.cpp
```diff
@@ -387,6 +388,12 @@ void generalDefault()
   g_eeGeneral.pwrOnSpeed = 1; // 1 second
 #endif
 
+  //Modified T8 applies to commando8
+#if defined(RADIO_Commando8)
+  g_eeGeneral.splashMode=3;
+  g_eeGeneral.pwrOnSpeed=2;
```
3djc: blabla = value (spaces)

radio/src/stamp.cpp
```diff
@@ -62,6 +64,9 @@
   const char vers_stamp[]  = "FW" TAB ": opentx-" FLAVOUR    "\036VERS" TAB ": RM Factory (" GIT_STR ")" "\036BUILT BY : OpenTX" "\036DATE" TAB ": " DATE " " TIME "\036EEPR" TAB ": " EEPROM_STR;
 #elif defined(JUMPER_RELEASE)
   const char vers_stamp[]  = "FW" TAB ": opentx-" FLAVOUR    "\036VERS" TAB ": Factory (" GIT_STR ")" "\036BUILT BY : OpenTX" "\036DATE" TAB ": " DATE " " TIME "\036EEPR" TAB ": " EEPROM_STR;
+#elif defined(iFlight_RELEASE)
+  //#if defined(RADIO_T8)   //Modified T8 applies to commando8
+  const char vers_stamp[]  = "FW" TAB ": iFlight-Commando 8"  "\036VERS" TAB ": OpenTX-2.3.14" "\036BUILT BY : iFlight" "\036DATE" TAB ": " DATE " " TIME "\036EEPR" TAB ": " EEPROM_STR;
```
3djc: no hardcoding of version number !

radio/src/targets/taranis/CMakeLists.txt
3djc: yeah, need to work on a proper version, this is fairly broken too (check for iFlight_RELEASE and set it later)

radio/src/targets/taranis/hal.h
```diff
@@ -282,6 +303,25 @@
   #define TRIMS_GPIO_PIN_RVU            GPIO_Pin_2  // PC.02
   #define TRIMS_GPIO_REG_RHR            GPIOE->IDR
   #define TRIMS_GPIO_PIN_RHR            GPIO_Pin_4  // PE.04
+
+#elif defined(PCBX7) && defined(RADIO_Commando8)
+  #define TRIMS_GPIO_REG_LHL            GPIOD->IDR
+  #define TRIMS_GPIO_PIN_LHL            GPIO_Pin_15 // PD.15
+  #define TRIMS_GPIO_REG_LHR            GPIOC->IDR
+  #define TRIMS_GPIO_PIN_LHR            GPIO_Pin_6  // PC.06 //CHANGE
```
3djc: remove all // change

radio/src/translations/en.h.txt
```diff
@@ -313,6 +313,9 @@
 #elif defined(RADIO_T8)
   #define LEN_VKEYS                    "\005"
   #define TR_VKEYS                     "RTN\0 ""ENTER""PGUP\0""PGDN\0""SYS\0 ""MDL\0 ""UP\0  ""DOWN\0"
+#elif defined(RADIO_Commando8)
```
3djc, line 316: isns't that the same as RADIO_T8 ?

XING-IF:
thanks for the review, changes are in work and should be finished within the next couple of days.

radio/src/pulses/modules_helpers.h
XING-IF: Because our D8 module crystal frequency is 26.000050MHZ, and the R81 receiver crystal frequency is 25.999340MHZ, so there is a frequency deviation and the frequency needs to be compensated. This variable g_model.moduleData[moduleIdx].multi.optionValue cannot save the setting value after switching to another module, so it is set to -27 by default. Is there a better approach for this?

radio/src/gui/128x64/model_setup.cpp
XING-IF: The built-in and external RF module cannot be turned on at the same time, because it controls the RF electronic switch that is connected to a Dual-Band antenna with 2.400 and 900 Mhz. Both RF modules use the same antenna. We tried to use the isInternalModuleAvailable() function and finally failed. Is there a better approach for our case?

radio/src/targets/common/arm/stm32/bootloader/boot.cpp
XING-IF: Because our power switch is controlled by a STM32F031. After entering the BootLoader and exiting opentx through the UI interface, the STMF031 still thinks that opentx is in BootLoader mode. At this time, pressing the power button will cause the STM32F031 to shut down, so you cannot exit through the exit option of the BootLoader UI interface. We'd need to guide the user to press the power button to exit the BootLoader mode. But in order to not affect the general code, we added the COMMANDO8 definition to the BootLoader. Please advice, thanks!

companion/src/CMakeLists.txt
XING-IF: code cleaned up

companion/src/CMakeLists.txt
XING-IF: code cleaned up

companion/src/firmwares/boards.cpp
XING-IF: code cleaned up

companion/src/firmwares/opentx/opentxinterface.cpp
XING-IF: comments removed

radio/src/opentx.cpp
XING-IF: comments removed

radio/src/targets/taranis/hal.h
XING-IF: comments removed

XING-IF:
feedback reviewed and changes committed

radio/src/translations/en.h.txt
XING-IF: T8 base yes

radio/src/stamp.cpp
XING-IF: hardcoding removed in last commit

radio/src/targets/taranis/board.h
XING-IF: code fixed to STICKS_PWM

radio/src/gui/128x64/model_setup.cpp
XING-IF: code fixed, please review

radio/src/opentx.cpp
XING-IF: changed code

radio/src/gui/128x64/radio_setup.cpp
XING-IF: code has been removed

radio/src/gui/128x64/radio_setup.cpp
XING-IF: code fixed

Issue #8945, opened 2022-07-19
9XRPRO External RF menu crashes/missing.


jarno-r:
**Describe the bug**
On Turnigy 9XR Pro & OpenTX 2.3.15, new models with external multimodule cannot be created and old models cannot be modified. A model created in an older version of OpenTX works, but trying to change the protocol causes the menu to disappear. 

Occasionally, selecting another model causes a reboot without the model being changed. Not sure if this is related.

Everything worked ok on 2.3.7, but not on 2.3.15.

**To Reproduce**
Flash 2.3.15 on 9XRPRO and try to create a model with External RF.

**Expected behavior**
External RF menu exists and works.

**Screenshots**
How it looks with a model created in an older version of OpenTX:
![External RF](https://user-images.githubusercontent.com/19144726/179742535-e209be62-f6cc-44eb-be3d-49d6e8c20064.png)

How a newly created model looks like:
![No External RF](https://user-images.githubusercontent.com/19144726/179742547-71d50d42-a201-46a7-84ba-b8fa294d119f.png)

Also note that the "External RF" title is missing. 

**General information**
 - OpenTX version 2.3.15 (opentx-9xrpro-autosource-autoswitch-battgraph-en-2.3.15-otx)
 - Hardware Turnigy 9XR Pro

jarno-r:
Bisected the firmware. First bad commit is 2.3.14-697184a5 (the parent 2.3.14-dfd72140 is good).

SwissAirplane:
Hi!

This is my problem too!
I have a 9XR PRO remote and I can't set the module.. no signal goes out.
(FrSKY DJT)
can you send me a 2,3,13 or earlier FW?
Maybe there is a solution, what can I do to make it work?
Thanks

jarno-r:
@SwissAirplane I don't have a FW at hand, but I've setup a repo where I'm trying to fix all issues with 9XR PRO (and hopefully get the fixes back into OpenTX at some point). The repo is here: [https://github.com/jarno-r/opentx-9xrpro](https://github.com/jarno-r/opentx-9xrpro)
I'll probably have pre-built FWs to download there in the near future, but it's not hard to build it yourself.
FYI, OpenTX from and including 2.3.12 suffer from random crashes, which I've yet to figure out and fix. They're not harmful, but annoying as they kill the audio. So I'd recommend 2.3.11 or earlier, if you can get that somewhere.

PR #8946, opened 2022-07-19
Fix broken External RF module menu.

jarno-r:
Fix for [Issue 8945 - 9XRPRO External RF menu crashes/missing.](https://github.com/opentx/opentx/issues/8945).

Taranis specific menu options were partially included, which caused the External RF menu to be messed up.

### DIFF
--- a/radio/src/gui/128x64/model_setup.cpp
+++ b/radio/src/gui/128x64/model_setup.cpp
@@ -85,8 +85,10 @@ enum MenuModelSetupItems {
   ITEM_MODEL_SETUP_CHECKLIST_DISPLAY,
   ITEM_MODEL_SETUP_THROTTLE_WARNING,
   ITEM_MODEL_SETUP_SWITCHES_WARNING1,
+  #if defined(PCBTARANIS)
   ITEM_MODEL_SETUP_SWITCHES_WARNING2,
   ITEM_MODEL_SETUP_POTS_WARNING,
+  #endif
   ITEM_MODEL_SETUP_BEEP_CENTER,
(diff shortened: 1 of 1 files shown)


radio/src/gui/128x64/model_setup.cpp
```diff
@@ -85,8 +85,10 @@ enum MenuModelSetupItems {
   ITEM_MODEL_SETUP_CHECKLIST_DISPLAY,
   ITEM_MODEL_SETUP_THROTTLE_WARNING,
   ITEM_MODEL_SETUP_SWITCHES_WARNING1,
+  #if defined(PCBTARANIS)
   ITEM_MODEL_SETUP_SWITCHES_WARNING2,
   ITEM_MODEL_SETUP_POTS_WARNING,
```
pfeerick, line 90: Er... the 9XR-Pro has pots also... think you roped in one line too many there, and removed that from being shown ;)

jarno-r: Yes, it has pots, but for some reason, the pot warning was only ever working on PCBTARANIS. To get the pot warnings working on 9XR-Pro requires some changes in other parts of the code too. Since it seems the feature was originally intended only for PCBTARANIS, I figured it would be safer to just put it back the way it was.

pfeerick: Indeed... a non-broken menu is always a good start. I'll have to dig mine out and get it working again... it's not been used for a few years now due to the various gremlins that have been creeping in for the 9X and 9XRPRO targets... might be time to finally try and fix some of them! Great to see someone else giving it some love! :)

pfeerick:
Shush, bot!

PR #9024, opened 2023-09-05
Fix vario for last generation of ELRS receivers

3djc:


mha1:
Hi JC, please have a look at https://github.com/EdgeTX/edgetx/pull/4010. I went a bit further to properly implement the CRSF spec for packets 0x07 and 0x09.

